PR #745, opened 2013-10-19
Do not use toJSON for data serialization in views

AmIMeYet:
This is an updated version of #599 following @samccone's suggestions.

As per the Backbone documentation (see documentcloud/backbone#2134), do not use _model.toJSON_ or _collection.toJSON_ when rendering views. This allows developers to override _toJSON_ without affecting rendering. It was implemented by adding a _serializeModel_ method to View and a _serializeCollection_ method to ItemView mimicking Backbone's implementation of _toJSON_. Appropriate tests where added and the documentation was updated.

This may be a breaking change for those that have overridden _toJSON_ for data serialization.

### DIFF
--- a/docs/marionette.itemview.md
+++ b/docs/marionette.itemview.md
@@ -180,11 +180,11 @@ Backbone.Marionette.ItemView.extend({
 
 ## ItemView serializeData
 
-Item views will serialize a model or collection, by default, by
-calling `.toJSON` on either the model or collection. If both a model
-and collection are attached to an item view, the model will be used
-as the data source. The results of the data serialization will be passed to the template
-that is rendered. 
+Item views will serialize a model or collection by respectively
+returning the model's attributes or doing so for each model in a collection.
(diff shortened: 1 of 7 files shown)


rpastorelle:
So... Since `serializeData` no longer uses `model.toJSON` how do I place additional attributes on the model pre-render?

I know about `templateContext` but that is not ideal to me because it would require that I duplicate code in every view that renders this model. Had this PR not been applied I could have just override `toJSON` in the model and every view using that model would receive the additional attributes.

rafde:
If you need to keep using `model.toJSON` you can change `serializeData` to use it.

rpastorelle:
@rafde How? Wouldn't I need to override `serializeData` for every view?

rafde:
If you extended all the Marionette classes, i.e. `MyView = Marionette.View.extend({})`, you would only need to update your extend class that all other views extend from.

Issue #1034, opened 2014-03-21
Consider supporting multiple event handlers when using string declaration for events.


MeoMix:
I have the following:

```
var ForegroundView = Backbone.Marionette.Layout.extend({
    events: {
        'click': function (event) {
            this.tryResetContextMenu(event);
            this.onClickDeselectCollections(event);
        }
    }
    ...
```

This is consistent with Backbone's implementation of event handling, but feels inconsistent when taking Marionette into account.

Please consider supporting the following:

```
var ForegroundView = Backbone.Marionette.Layout.extend({
    events: {
        'click': 'tryResetContextMenu onClickDeselectCollections'
    }
    ...
```

PR #1288, opened 2014-05-04
DO NOT MERGE: proof of concept multiple event handlers

samccone:
Ok this is a proof of concept for #1034 

Ended up with this syntax for defining multiple event handlers 

``` js
      events: {
        'click @ui.bar': ['attack', 'yield']
      }
```

Which I think is much better than defining another space delimitated DSL.
This also has the added advantage of not being breaking!

---

Let me know what you guys think and I will extend this functionality to the other events: 
- `collectionEvents`
- `modelEvents`

^ I think that is all right?


hrishikeshs:
+1 for this, 3 years later :smile:

PR #1900, opened 2014-09-15
Provide Marionette objects as individual UMD modules

### LINKED ISSUE #1332: Provide Marionette components as commonJS modules
I'm using Backbone and Marionette in a CommonJS environment (browserify) and from that perspective, it would be great, if the CommonJS build did not include everything, but would let me cherry-pick the components that I want to load. That would reduce the amount of KB, which would translate to faster load times.

For example - I may not use Marionette.Application at all. Then why load it in? Perhaps I'd only like to take advantage of CompositeView and ItemView? It'd be great, if I could just do something like this:

var CompositeView = require('marionette/compositeview');
CompositeView would, of course, handle all of it's dependencies itself (requiring Marionette.View, etc).

Personally, I only want to use Marionette views and regions in my app, I do not need router, controller, app, vent or any of that.

jmeas:
#### [Play with it here](https://github.com/jmeas-test/test-umd-modules)

Resolves #1332 

This allows for users to pick and choose the Marionette classes they want to use. This updates our build system to transparently bundle each class as a UMD module.

```sh
# build everything, including the components
grunt build
```

Usage:

```js
// Pull in just what you want to use
var ItemView = require('marionette/lib/item-view');
var Region = require('region');

// And if you're a power user who wants to access Marionette
// directly, for, say, `triggerMethodOn`
var Mn = require('marionette/lib/core');
```

:tada:

=====

Caveats of the playground repo above:

– I don't test Behavior/Behaviors classes
– I *only* test initialization. So if a dependency is used in a particular function, that error won't be caught

Please take a look through all of the source files to see if I missed any dependencies. Do remember that the helper methods like TriggerMethod, and the Error class, are all loaded when you load in the core.


samccone:
Closing this for now.

Issue #1920, opened 2014-09-22
Region.empty is always called twice


rjmackay:
When `region.empty()` is called it triggers a 2nd call to itself.

Because `region.show()` binds to the view's `destroy` event and `region.empty()` destroy's the view. A 2nd call to `region.empty` is always triggered.

If you have an `onEmpty` callback in your region, it is always triggered twice when calling `empty()`

`region.empty()` needs to unbind from the views `destory` event before it destroys the view

Issue #1911, opened 2014-09-19
Unexpected behavior when showing a view in a region with preventDestroy: true


jmirmina:
While upgrading from Marionette 2.0.1 to 2.2.1, I've noticed that one of our layouts is not working as expected . I've tracked it down to [marionette.region.js line 165](https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.region.js#L165)

Essentially, when calling show on a region and setting preventDestroy: true, if that view is swapped out and then later manually destroyed, it will call empty() on the region, removing the new view that was added to the region.

I've created a [jsfiddle](http://jsfiddle.net/569wc2ch/) to demonstrate this behavior. I would expect that View 2 would still be displayed, but the region has been emptied.

Edit: I had proposed some solutions, but after looking into it some more, I don't think they would work. I'm wondering if _shouldDestroyView is false, maybe it needs to unbind the listener to the previous view's destroy method so empty isn't called on the region?

PR #1925, opened 2014-09-22
Add preventDestroy option to region.empty()

rjmackay:
This is more of a change of behaviour, not just a fix for #1920 or #1911 
This allows calling region.empty() without destroying the view. The regions node is emptied, and events are unbound, but the view is left in tact.

When changing views, we use region.empty() .. forwarding the preventDestroy option if it was passed. But region.empty() will still unbind the events for us.

This may not be an approach you want to take, feel free to disregard if so. But it was easier to code than explain :)


samccone:
moved to #2204

Issue #1796, opened 2014-09-01
v3.0.0 Plan


samccone:
Because v3 is a long ways away, and given that we're still not sure exactly what will land in it, the current discussion around that release will be had in [**a Google doc**](https://docs.google.com/document/d/1Sx1YE2SJg-NGSGsd8mELf3wpywI-UyOvCkufbm6klOw/edit#). This cleans up Github issues that are aren't actionable at this time.

Once the release nears and we begin to solidify our plans we'll migrate back to Github issues.

jasonLaster:
Here's a trello board for ongoing work:
https://trello.com/b/oN61mg80/board

PR #2073, opened 2014-11-14
Motion to deprecate getOption and this.options hash

thejameskyle:
**NOTICE OF MOTION TO DEPRECATE "GETOPTION" AND "OPTIONS" HASH**
**CASE NO. 2073**

On the date of April 9th, 2013, Brad Dunbar, core contributor of popular javascript library "Backbone.js" (hereinafter referred to as simply "Backbone") committed "Dispense with View#options merging." (Ref: https://github.com/jashkenas/backbone/commit/51eed189bf4d25877be4acdf51e0a4c6039583c5) which was later released as part of version 1.1.0 of Backbone.

This commit removed the `this.options` property from the Backbone View Class, opting in favor of merging options directly onto the view instance.

Over a year and a half later popular javascript framework "Marionette.js" (hereinafter referred to as simply "Marionette") has yet to remove `this.options` from it's codebase.

On the date of October 27th, 2012, Derick Bailey, original author of Marionette, committed "added Marionette.getOption method" (Ref: https://github.com/marionettejs/backbone.marionette/commit/03f7d69bdeaf0dd393e858a070695de50ac45233) which was later released as part of version 1.0.0 of Marionette.

`Marionette.getOption`'s primary use case was as an internal helper to select a value from either the `this.options` hash or from a class instance.

Several conversations in the past 8 months have taken place about how it is preferable to merge `options` directly onto the instance if any properties are needed later on (See EXHIBIT A), and using `getOption` for user code is considered a bad practice.

Removing `getOption` has already been added to the plans for Marionette 3.0.0 (Ref: https://github.com/marionettejs/backbone.marionette/issues/1796).

**EXHIBIT A.**
```js
initialize(options) {
  this.objection = options.objection;
}
```

**RESPECTFULLY SUBMITTED**
**PRINTED NAME** _"THE" JAMES KYLE_


samccone:
@thejameskyle did you want to close this and we can figure out how this fits in with the new v3 plans?

thejameskyle:
@samccone Sure

samccone:
:boom:

PR #2100, opened 2014-12-01
added grunt-jsdoccer v1.0.10

ChetHarrison:
here is your doccer sir! (at least the first stab at it). It should become a little more configurable. You can just bump the version in the package.json from here on out as I a make updates. Docs are here https://github.com/ChetHarrison/jsdoccer and https://github.com/ChetHarrison/grunt-jsdoccer


samccone:
kk going to close for now, obviously this will get in @ChetHarrison we will just pair on it for an hour and we should be good to go 

:clap:

PR #2114, opened 2014-12-06
Add options hash into TemplateCache

Molchanov:
In one project I had a situation, when I needed to have options object in TemplateCache.prototype.loadTemplate function for loading template from different folders.

And I think that this simple modification will be very useful for others.


samccone:
@thejameskyle @jmeas are you guys still ok with this PR?

thejameskyle:
If it gets some tests. I'm good with it.

samccone:
kk I will add them

samccone:
moved to #2205

PR #2138, opened 2014-12-13
Add `isRendered` property to views

duncanbeevers:
To aid in defensive coding against all phases of a view's lifecycle, an isRendered property is provided on each view instance.

The value is falsey if the view has not yet been rendered, or has been destroyed.
The value is true if the view is currently rendered.

CompositeView's isRendered state-tracking property is renamed _isRendering to indicate its internal-only nature, and to prevent collisions with the now-exposed isRendered property.

Sequencing of the onBeforeRender event, mutation of the view isRendered state, and onRender is consistent and spelled out explicitly in the tests.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -111,6 +111,7 @@ Marionette.CollectionView = Marionette.View.extend({
     this._ensureViewIsIntact();
     this.triggerMethod('before:render', this);
     this._renderChildren();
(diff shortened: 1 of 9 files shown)


jmeas:
This broke major. I'm dropping the commit so that I can do other work on major; we should open a new PR against major and fix it.

PR #2141, opened 2014-12-14
LayoutView should remove its element before destroying child regions

### LINKED ISSUE #2134: Update layoutView destroy mode
LayoutViews [destroy their regions before destroying themselves](https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.layoutview.js#L48-L49). This creates `n` DOM paints, one for each layer of views.

Imagine a three-level nested LayoutView tree. The tree is attached to the document. You call destroy on the parent view, which calls destroy on its RegionManager before it, itself, is destroyed. That destroy call will remove the children separately from the parent.

LayoutView needs a custom destroy for efficiency.

megawac:
Resolves #2134, the layout views element should be removed before destroying the child regions and views

### DIFF
--- a/api/layout-view.yaml
+++ b/api/layout-view.yaml
@@ -33,7 +33,14 @@ examples:
       });
       
       // Create a new LayoutView
-      var layoutView = new Marionette.LayoutView();
+      var layoutView = new Marionette.LayoutView({
+        // This option removes the layoutView from
+        // the DOM before destroying the children
+        // preventing repaints as each option is removed.
+        // However, it makes it difficult to do close animations
+        // for a child view (false by default)
+        destroyImmediate: true
+      });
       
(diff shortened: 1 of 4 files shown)


samccone:
@megawac mind rebasing off of major

jasonLaster:
That or a feature flag.

samccone:
so based on the view talks that we have had does this still have a place? Is this something we want to be messing with for the next release?

samccone:
yes it is

megawac:
:wolf:

samccone:
1 commit, and some docs and then :ship:

jasonLaster:
:+1: Lets get this in

samccone:
we just need to add docs to the layoutView md file

PR #2145, opened 2014-12-18
Add LayoutView shortcut methods

### LINKED ISSUE #2143: Possible new LayoutView syntax
I was chatting with @tbranyen today, and we came to realize that Marionette and LayoutManager handle nested views rather similarly. The biggest difference in that LayoutManager, the API causes Regions to be more an implementation detail than a type of object that users need to worry about.

Compare these two APIs:

```js
rootView.getRegion('main').show(new ChildView());
rootView.setView('main', new ChildView());
```

The top API is Marionette's implementation of nesting. The bottom one is LayoutManager's.

Also compare:

```js
rootView.getRegion('main').currentView;
rootView.getView('main');
```

I think that these 'shortcut' methods could be good for most folks who never even think about Regions a as a thing other than a container – I'm one of them. This is so much the case that in LayoutManager, what we would call a 'region name' is called a 'selector alias,' as it just provides a way to describe an element that would otherwise just be a selector. But never a new concept, like a 'Region.'

I think that introducing these two methods to LayoutView is a good idea, but I have some ideas for our view layer that might be considered pretty radical.

First, I would get rid of ItemView so that there is *just* LayoutView. This has been discussed a number of times, and I still think it's a solid idea. Separating the two provides no substantial boost in performance while introducing an unnecessary concept (1 view class is easier to learn than 2). Further, it makes the entire view layer of Marionette more consistent. Instead of "3 of our 4 views support nested views..." it becomes "All of our views support nesting. It's just a matter of whether you want control over the children views, or whether you want it to be bound to a collection."

None of the above is all that radical, I think, but my plans for regions might be. I think we should, like LayoutManager, make regions an implementation detail. I've put more thought into a transition region, and I think we could implement something as elegant as React's animation system, which was inspired by Angular's. The tl;dr is...you don't need to override regions to do it.

There are outstanding questions about the future behavior of regions. Should the element be a placeholder, as in [minionette](https://github.com/jridgewell/minionette/blob/master/docs/minionette.region.md)? Should they support multiple views, like the [Multiregion](https://github.com/joezimjs/MultiRegion)? I don't think that the ideas presented in this issue affect these future decisions. By adjusting the API made available to the view we could support either or both of these things, if we chose.

====

So what would I suggest now?

We add two new methods to the LayoutView to bypass dealing with Regions. We deprecate the old methods for v3 or v4. We ask around to see why people make custom regions, and make sure that we can still support all of their use cases if we make region an implementation detail.

Thoughtz?

jasonLaster:
fixes #2143 

Adds two shortcut methods inspired by @tbranyen's LayoutManager. 

[wip] If we all like, i'll add some tests and docs and things. 

Thanks @jmeas for being social and things...

### DIFF
--- a/api/layout-view.yaml
+++ b/api/layout-view.yaml
@@ -203,6 +203,48 @@ functions:
           layoutView.destroy();
           ```
 
+  showChildView:
+    description: |
+      Shows a view in a region.
+      
+      This is a very useful method for showing child views inside a layout's regions in `onBeforeAttach` or `onAttach`.
+      
+      @api public
+      @param {String} regionName - The name of the region you're trying to get
+      @param {View} view - The view instance that you\'re showing in the region
+
+    examples:
+      -
+        name: Showing a view in a region
+        
+        example: |
+          You'll often want to show a view inside of a layout's regions before the layout is attached to the document.
+          The easiest way to do this is inside the layout's `before:attach` handler.
+          
+          ```js
+          var LayoutClass = Mn.LayoutView.extend({
+            onAttach: function() {
+              this.showChildView('sidebar', new SidebarView());
+              this.showChildView('list', new ListView());
+            }
+          });
+          
+          var layoutView = new MyLayoutView();
+          layoutView.render();
+          ```
+
+  getChildView:
+    description: |
+      Retrieves a child view from the LayoutView's region.
+
+      This is a useful method for grabbing a child view from a layout's region.
+      
+      In general, layouts should not reference their child views, but in the cases where you find yourself wanting to do this `getChildView` is a useful method.
+
+      @api public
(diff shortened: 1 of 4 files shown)


samccone:
@jasonLaster any thoughts on the state of this?

jasonLaster:
Could someone run with this? I won't be in front of a computer until Tuesday.

jasonLaster:
@marionettejs/marionette-core just updated. should be good to go

samccone:
ok I am 100% :+1: this going to merge unless any objections...

jmeas:
:+1:

samccone:
:wave:

samccone:
in minor now! yay

PR #2171, opened 2014-12-30
Adds the childViews property to a behavior. When the behavior's "view" i...

keithstric:
1. I have been using behaviors to add buttons and various functionality that is needed throughout my app. I wanted to create one behavior say for a filter feature that is used in many different places. This 'filter' button of course would require a view and corresponding template be created. However, when the behavior's view was destroyed I was seeing zombie views surfacing for the views I had created in the behavior. This pull request is meant to address this issue.

2. I added a property to the behavior class called childViews. This is an array placeholder to place any views created by the behavior. Then in the destroy method, call destroy on each of the views in the childViews array.

3. When the behavior's parent view is destroyed, it calls the destroy method in the behavior. I added functionality that will call the destroy method on each of the views in the childViews array. Also, in the behavior.spec.js file, there was a test there for destroy that stated destroy should return the behavior. This was not being checked, so I also added a 'return this;' line to the destroy method and fixed the unit test to actually verify that the destroy method returns the behavior. I also added another unit test to ensure that destroy was being called on the views in the childViews property.

4. Since adding the 'return this;' line to the destroy method of a behavior, I couldn't determine if this will have any unwanted side effects. From my initial testing (in my app) it doesn't appear that it does but I of course could be wrong. Also, I'm sure there is a better way to keep track of the views that a behavior may/may not create, it would be great if this were a little more elegant but I'm not sure how I could accomplish that.


jfairbank:
Hi, @keithstric! Thanks for the PR. I'm not sure I'm in favor of adding an array of `childViews` to `Behavior` in core, though. This is not something I think most people would use, and I'm not sure I like the idea of encouraging coupling a behavior to multiple child views.

Is there something `CollectionView` or `CompositeView` aren't doing for you that made you take this approach? If so, can you share your use case in a separate issue?

As far as returning the behavior upon destruction, that may have been an unintended spec that was added, since it was a noop test. I don't immediately see the merit in returning the behavior after it is destroyed besides consistency with other Marionette methods that return `this`. @marionettejs/marionette-core?

keithstric:
Hi @jfairbank, Hopefully I can describe this use case, sometimes I get a little long typed so I apologize in advance. One of the ways I've been using behaviors:

### Use Case
One of the common patterns I've been using is a form representing an object and then a series of tables that show other things that initial form is related to. Each of these tables are within it's own panel and in the header of those panels I have one or more behaviors which have visual representations. 

For example one of the behaviors is a search field, another is a filter button with a drop down list of things to filter by and yet another is a drop down menu of actions. All of these different behaviors create their own little view structure and since they are behaviors are easily added to these related tables and I don't have to write the exact same code over and over. So to add a search field to one of these related tables I just include the searchBehavior in the table's composite view, set a couple of properties (i.e. where the behavior's view should be placed) and that's it. I then get a search field where I said it should go and since the behavior knows about it's containing view and collection, the search field will just work without any further effort.

### Why not a Composite/Collection/Item View
At first, the behaviors I explained above started as ItemViews, but I found myself having to define the same events over and over in the parent views, it was easier to just move those ItemViews into a behavior, define the event listener in the behavior then just add the behavior and "Bob's your uncle" everything works.

Hopefully that explains how I'm using these behaviors. I agree it's not something everyone would use, but if someone did need to have a behavior create a view(s) it would prevent them from creating zombie views unknowingly, which was one of the main reasons I started using Marionette.

jfairbank:
I'm still a little confused why your use case needs this change, but I understand your desire to reduce code duplication with behaviors. I'm more curious if there was a reason why including your behaviors in a collection or composite view wasn't working. Since collection and composite views manage child views already, I feel like this is an unnecessary addition to behaviors. We don't want to duplicate what's already been solved. If behaviors mixed into collection and composite views isn't cutting it for you, then I would like to know some more detail about that please. Maybe a JSFiddle of a small code sample might help.

As far as the other change of returning `this` from `destroy`, I'm good with that.

keithstric:
@jfairbank ok I've given this some consideration and moved some of these behaviors into their own (? extends View) views. I still get the same type of behavior so I concede your points. Maybe I was too new to Marionette when I started down this road a year ago.

jasonLaster:
Thanks @keithstric I think I have an idea of what you're ding and it's pretty cool. I don't think we're ready to expand the role of behaviors, but we are exploring expanded region APIs. Could you attach some code samples.

PR #2175, opened 2015-01-05
Fixing broken links in docs.

bzalasky:
These were previously referencing *.md files. However, they need
to point to their respective *.html files on the docs site.

### DIFF
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -66,7 +66,7 @@ add them after the app is started, they will run immediately.
 ## Application Event
 
 The `Application` object raises a few events during its lifecycle, using the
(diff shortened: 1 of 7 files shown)


paulfalgout:
These can actually stay .md and the new docs will accommodate.

samccone:
hi @bzalasky thanks for the work, we are working on a new version of the docs and actually will be launching soon!

http://dev.marionettejs.com/docs/current/


Thanks to this line there is no need for this change
https://github.com/marionettejs/marionettejs.com/blob/990ee340a81208d4f5ba1b3d3256f141fd1015eb/tasks/compile-docs.js#L86

thanks again!

PR #2181, opened 2015-01-07
Refactor serializeData to no longer use toJSON.

### LINKED ISSUE #1476: Stop using toJSON for rendering templates
Right now we're using [toJSON to render templates](https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.itemview.js#L21-L32).

Per [this issue](https://github.com/jashkenas/backbone/issues/2134), and [its resolution](https://github.com/jashkenas/backbone/pull/2135), we shouldn't be.

Instead, we should be cloning the attributes object for our templates explicitly. I'm talking `_.clone(model.attributes)`.

Your next question might be...why does this matter at all? toJSON does the _exact_ same thing...all it does is clone the attributes.

The difference is use cases. toJSON is called when you're persisting data to the server – it's meant to filter out any data you don't want to save. It's not at all meant to be used in views!

This lets you do things like:
1. Get data back from the server that has time stored as a string
2. Store that data as a moment in your model, within attributes
3. Filter that out with toJSON
4. Still have access to everything in render

This is a breaking change, but is necessary to get Marionette consistent w. Backbone. It'll also be a huge help to everyone who is trying to do something like the above, which is otherwise really difficult.

jmeas:
toJSON should only be used for preparing data to be sent to the
server. In addition, serializeData no longer accepts arguments,
distinguishing it further from templateHelpers.

Resolves #1476 

====

This prepares our `serialize` methods for tha' future. There are two changes:

1. No longer uses `toJSON` to serialize the data. `toJSON` should never be used by views.
2. `serializeData` has a single purpose: preparing the special `view.model` and `view.collection` attributes to be sent to a template. They are not for adding arbitrary extra data to your template, as that is what `templateHelpers` are for.

#### Todo:

- [x] Update documentation to more clearly explain what `serializeData` is for.

### DIFF
--- a/docs/marionette.itemview.md
+++ b/docs/marionette.itemview.md
@@ -225,13 +225,29 @@ Backbone.Marionette.ItemView.extend({
 
 ## ItemView serializeData
 
-Item views will serialize a model or collection, by default, by
-calling `.toJSON` on either the model or collection. If both a model
-and collection are attached to an item view, the model will be used
-as the data source. The results of the data serialization will be passed to the template
-that is rendered.
+This method is used to convert a View's `model` or `collection`
+into a usable form for a template.
 
-If the serialization is a model, the results are passed in directly:
+Item Views are called such because they process only a single item
+at a time. Consequently, only the `model` **or** the `collection` will
+be serialized. If both exist, only the `model` will be serialized.
+
+By default, models are serialized by cloning the attributes of the model.
+
+Collections are serialized into an object of this form:
+
+```js
+{
+  items: [modelOne, modelTwo]
+}
+``
+
+where each model in the collection will have its attributes cloned.
+
+The result of `serializeData` is included in the data passed to
+the view's template.
+
+Let's take a look at some examples of how serializing data works.
 
 ```js
 var myModel = new MyModel({foo: "bar"});
@@ -272,19 +288,11 @@ MyItemView.render();
 </script>
 ```
 
-If you need custom serialization for your data, you can provide a
-`serializeData` method on your view. It must return a valid JSON
-object, as if you had called `.toJSON` on a model or collection.
+If you need to serialize the View's `model` or `collection` in a custom way,
+then you should override either `serializeModel` or `serializeCollection`.
 
-```js
-Backbone.Marionette.ItemView.extend({
-  serializeData: function(){
-    return {
-      "some attribute": "some value"
-    }
-  }
-});
-```
+On the other hand, you should not use this method to add arbitrary extra data
+to your template. Instead, use [View.templateHelpers](./marionette.view.md#viewtemplatehelpers).
 
 ## Organizing UI Elements
 
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -46,17 +46,9 @@ Marionette.CompositeView = Marionette.CollectionView.extend({
     return childView;
   },
 
-  // Serialize the model for the view.
-  // You can override the `serializeData` method in your own view
-  // definition, to provide custom serialization for your view's data.
+  // Return the serialized model
   serializeData: function() {
-    var data = {};
-
-    if (this.model){
-      data = _.partial(this.serializeModel, this.model).apply(this, arguments);
-    }
-
-    return data;
+    return this.serializeModel();
   },
 
   // Renders the model and the collection.
@@ -83,9 +75,7 @@ Marionette.CompositeView = Marionette.CollectionView.extend({
   // Render the root template that the children
   // views are appended to
   _renderTemplate: function() {
-    var data = {};
-    data = this.serializeData();
-    data = this.mixinTemplateHelpers(data);
+    var data = this.mixinTemplateHelpers(this.serializeData());
 
     this.triggerMethod('before:render:template');
 
--- a/src/item-view.js
+++ b/src/item-view.js
@@ -12,35 +12,34 @@ Marionette.ItemView = Marionette.View.extend({
     Marionette.View.apply(this, arguments);
   },
 
-  // Serialize the model or collection for the view. If a model is
-  // found, the view's `serializeModel` is called. If a collection is found,
-  // each model in the collection is serialized by calling
-  // the view's `serializeCollection` and put into an `items` array in
-  // the resulting data. If both are found, defaults to the model.
-  // You can override the `serializeData` method in your own view definition,
-  // to provide custom serialization for your view's data.
-  serializeData: function(){
-    if (!this.model && !this.collection) {
-      return {};
-    }
+  // Serialize the view's model *or* collection, if
+  // it exists, for the template
+  serializeData: function() {
+    var data = {};
 
-    var args = [this.model || this.collection];
-    if (arguments.length) {
-      args.push.apply(args, arguments);
-    }
-
+    // If we have a model, we serialize that
     if (this.model) {
-      return this.serializeModel.apply(this, args);
-    } else {
-      return {
-        items: this.serializeCollection.apply(this, args)
+      data = this.serializeModel();
+    }
+
+    // Otherwise, we serialize the collection,
+    // making it available under the `items` property
+    else if (this.collection) {
+      data = {
+        items: this.serializeCollection()
       };
     }
+
+    return data;
   },
 
-  // Serialize a collection by serializing each of its models.
-  serializeCollection: function(collection){
-    return collection.toJSON.apply(collection, _.rest(arguments));
+  // Serialize a collection by cloning each of
+  // its model's attributes
+  serializeCollection: function() {
+    if (!this.collection) { return {}; }
+    return this.collection.map(function(model) {
+      return _.clone(model.attributes);
+    });
   },
 
   // Render the view, defaulting to underscore.js templates.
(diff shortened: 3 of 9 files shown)


samccone:
mind putting all of the comment above in the commit message ^

i think if we basically have a very good
what and why in each major commit .. when it comes release time it will be very simple to create a great changelog :)

src/composite-view.js
```diff
@@ -46,17 +46,9 @@ Marionette.CompositeView = Marionette.CollectionView.extend({
...
   serializeData: function() {
-    var data = {};
-
-    if (this.model){
-      data = _.partial(this.serializeModel, this.model).apply(this, arguments);
-    }
-
-    return data;
+    return this.serializeModel();
   },
```
jmeas: I intentionally left this method – even though it isn't *technically* necessary, so that `CompositeView` has the same "serialize API" as the `ItemView` and `LayoutView` class.

jmeas:
> mind putting all of the comment above in the commit message

The first half is, which is a good summary of the change, I think. If you like the 2nd part better, I can swap it, but they are redundant explanations.

docs/marionette.itemview.md
```diff
@@ -272,19 +288,11 @@ MyItemView.render();
...
-Backbone.Marionette.ItemView.extend({
-  serializeData: function(){
-    return {
-      "some attribute": "some value"
-    }
-  }
-});
-```
+On the other hand, you should not use this method to add arbitrary extra data
+to your template. Instead, use [View.templateHelpers](./marionette.view.md#viewtemplatehelpers).
```
jmeas: No idea if this link will work on the `www`

samccone: it will

src/item-view.js
```diff
@@ -12,35 +12,34 @@ Marionette.ItemView = Marionette.View.extend({
...
   },
 
-  // Serialize a collection by serializing each of its models.
-  serializeCollection: function(collection){
-    return collection.toJSON.apply(collection, _.rest(arguments));
+  // Serialize a collection by cloning each of
+  // its model's attributes
+  serializeCollection: function() {
+    if (!this.collection) { return {}; }
+    return this.collection.map(function(model) {
```
megawac: This can also be `_.clone(this.collection.pluck('attributes'))`

megawac:
:+1: for this change, this always see,ed weird to me

src/item-view.js
jmeas: Willchange

jasonLaster:
Nice. Commit messages will be huge. 

I'm going to update the rename templateHelpers ticket. Seems more important now. Thinking renderData is closer to what we want.

thejameskyle:
So I'm thinking that I want to see the same three methods used by every view class. So that we define `serializeData`, `serializeModel`, and `serializeCollection` on `Mn.View` and inherit everywhere else. Thoughts?

samccone:
interesting idea @thejameskyle why do you think we need three methods?

jmeas:
> Thoughts?

I'm fine with it, though it may conflict with a long-term plan for views.

Collection Views never have templates, and therefore never serialize their collection. As such, there's no need for *any* `serializeX` to be on the `collectionView`. A CompositeView is really just a LayoutView with a nested CollectionView, and I think that it's a good idea for us to move in that direction. Then if we merge LayoutView with ItemView, we end up with those 2 view classes we've been talking about for awhile – ItemView and CollectionView. Only the ItemView would need the `serializeX` methods.

If we go this direction, then I'm actually in favor of moving all of these methods to the ItemView itself, then patching in CompositeView-specific versions (which we already have in place) for that Class. Down the road, we would simply pluck CompositeView from the codebase without affecting the ItemView.

What you think?

thejameskyle:
> I'm fine with it, though it may conflict with a long-term plan for views.

So far _every_ discussion I've heard about our future View classes is that we only have two: `Marionette.View` and `Marionette.CollectionView`. In which case it only makes sense for `CollectionView` to extend from `View`.

If that's not what is going to happen then it's an issue for all the View work I've been doing the past few months. But @jmeas, you even mentioned in your talk at BackboneConf about only having two view classes in the future.

jmeas:
I've been lying a bit each time I say two view classes, but only as much as we lie when we say that there are 4 view classes in Marionette, as there are, of course, actually 5. Just as we have `Marionette.View` right now, I envision us having something like [`Marionette.AbstractView`](https://github.com/jmeas/puppets/blob/v0.0.1/src/abstract-view.js), which would contain shared functionality between our "ItemView" and "CollectionView." As you can see from that file, there is very little shared between the two.

You might think, so what, CollectionView **can** extend from ItemView, right? So why not just do that and call it a day? The primary reason I'd be against it is if we end up putting too much unused API on the CollectionView. Let me explain.

The reason there is so little in common between the two is that they **fundamentally different* types of objects. It's nearly violence to the API that we're calling both of them views. This boils down to the difference in how they eventually represent themselves in the DOM. Item Views have templates, whereas Collection Views do not have templates. The ItemView utilizes regions for children views, whereas CollectionViews manage their children through a whole 'nother system. ItemView-type regions interface with Regions; CollectionViews do not.

In my mind, it makes nearly as much sense to say that the ItemView class should extend from a CollectionView as it does the other way around, as neither of them are actually adding any functionality onto the other one.

*With that said*, you **can** do it. [Minionette is an example.](https://github.com/jridgewell/minionette/blob/master/src/collection_view.js) 

I'd just be worried about the "unused" API – all of the templating and region code – hiding on our CollectionViews. It might be weird if 60% of the CollectionView's API is simply ignored because it has an entirely different render path.

thejameskyle:
@jmeas Can you put that text in the meeting notes for later and we can figure it out later today?

jmeas:
yup, done

jridgewell:
> With that said, you can do it. Minionette is an example.

Whoa there, Minionette's [`CollectionView`](https://github.com/jridgewell/minionette/blob/master/src/collection_view.js) and [`ModelView`](https://github.com/jridgewell/minionette/blob/master/src/model_view.js) both inherit from `View`, not the other one. It just happens that `ModelView`s is very similar to `View`s, needing just two differences to specialize for rendering models. Notice `collectionEvents` is [defined](https://github.com/jridgewell/minionette/blob/master/src/collection_view.js#L19-L24), analogous to the [`modelEvents`](https://github.com/jridgewell/minionette/blob/master/src/model_view.js#L3-L6). The only reason `#serialize` isn't defined is because the collection rendering already [takes care of it](https://github.com/jridgewell/minionette/blob/master/src/collection_view.js#L37-L112).

> they are fundamentally different types of objects... This boils down to the difference in how they eventually represent themselves in the DOM.

I think this is _almost_ true. Thinking about them as how "they represent themselves in the DOM" is exactly right. If an `ItemView` is just a rendered template using its model's attributes, then a `CollectionView` is just a rendered template using its `ModelView`s templates.

Minionette's `CollectionView`s break one of Minionette's core beliefs: Never render a subview (a `ModelView`, in this case). Calling `collectionView.render()` will `_.invoke(modelViews, 'render')`, ensuring a everything is up to date. That's because a dev isn't really supposed to mess with any of the `ModelView`s. They're transparent and indivisible from their `CollectionView`.

> The ItemView utilizes regions for children views, whereas CollectionViews manage their children through a whole 'nother system.

I think this is where our disconnect is. Minionette's `CollectionView`s don't treat the `ModelView`s as subviews ("children"). They _are_ the template. Further, adding a `#template` in conjunction with `#appendModelView` allows a `CollectionView` to have a default template, regardless of any models. That's why `CollectionView`s can still have regions attached to them: they can have their own templates, and regions are still the subviews. So, a `CollectionView` is every bit as much a `View` as `ModelView` is.

samccone:
ok this has been on the slate for a long time. @jmeas it makes sense what you are saying with the abstract view and I am behind that work.

I think this commit represent a great step in the right direction. I am 100% :+1: behind this work and will merge unless there are objects

great work @jmeas

src/item-view.js
jasonLaster: changed!

jasonLaster:
Congrats @jmeas! It's on major. Great work

samccone:
awesome!

src/item-view.js
jridgewell: Actually, that'll just reference the same attributes object (as well as construct a bunch of throw-away models). `Model#clone` [isn't deep](https://github.com/jashkenas/backbone/blob/master/backbone.js#L589-L591).

megawac: Yeah, neither is `_.clone` so it's more or less the same

```js
var m = new Backbone.Model({a: [], b: 2});
m.clone().attributes !== m.attributes
m.clone().get('a') === m.get('a');

_.clone(m.attributes) !== m.attributes
_.clone(m.attributes).a === m.attributes.a;
```

jridgewell: Well, this one was what I meant:

```javascript
m.clone().attributes === m.attributes; // => true
```

All `Model#clone` does is construct a new `Model` with the same `attributes` object.

megawac: It's a different `attributes` object @jridgewell

jridgewell: Damn, I misspoke. For some reason, I thought it `Model` just used `attributes`, instead of [setting](https://github.com/jashkenas/backbone/pull/873/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR165) it. :disappointed:

PR #2182, opened 2015-01-07
Add `isRendered` property to Views

duncanbeevers:
`isRendered` is falsey before view has been rendered, and during `onBeforeRender`
`isRendered` is true when view has been rendered, and during `onRender`
`isRendered` is truthy before view has been destroyed, and during `onBeforeDestroy`
`isRendered` is false after view has been destroyed, and during `onDestroy`

CompositeView's previously used an `isRendered` property. This property has been renamed to `_isRendering`


samccone:
yeah I am not sure also @duncanbeevers soon we shall know :)

jmeas:
Thanks @duncanbeevers! Dunno what was up before.

PR #2185, opened 2015-01-09
Fixing the naming inconsistency between child view attribute and event prefix

### LINKED ISSUE #2121: Naming inconsistency between child view attribute and event prefix
I noticed that the child view attribute for `CollectionView`/`CompositeView` is camel cased like `childView:`, while child view events are prefixed with `childview` in lower case.

I think this needs to be fixed to be consistent. I've tripped on this a couple of times causing bugs when I've camel cased the event prefix.

SauloSilva:
Replacing the event name of prefix `childview` to `childView`

Fix #2121


thejameskyle:
@SauloSilva Since this is a breaking change and will need to have a major version bump could you reopen this against the `major` branch?

SauloSilva:
@thejameskyle

thejameskyle:
Thanks!

PR #2186, opened 2015-01-09
Update Application documentation.

jmeas:
This is a rough draft of some much-needed updates to the app docs that should hopefully clarify the purpose of the Class.

Do note that there are some things we might want to fix up. Grammar and such. I wrote this pretty quickly just to get the idea on paper.

### DIFF
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -2,21 +2,34 @@
 
 # Marionette.Application
 
-The `Backbone.Marionette.Application` object is the hub of your composite
-application. It organizes, initializes and coordinates the various pieces of your
-app. It also provides a starting point for you to call into from your HTML
-script block, or directly from your JavaScript files if you prefer to go that
-route.
+The **Application** is a container for the rest of your code. It is recommended
+that every Marionette app have at least one instance of Application.
 
-The `Application` is meant to be instantiated directly, although you can extend
-it to add your own functionality.
+By creating an Application you get three important things:
 
-```js
-var MyApp = new Backbone.Marionette.Application();
-```
+- A `start` method to kick off your application.
+  This allows you an opportunity to do things that may need to occur before, say, you
+  begin routing. An example would be making an AJAX call to request data that your app
+  needs before starting.
+
+- A namespace to keep things off of the `window`.
+  If you are not using a module loader like ES6 modules, CommonJS, or AMD, then
+  you can use the Application to store your Javascript objects. And if you are
+  using one of those module systems, then you can still attach things to the
+  application to aid in debugging.
+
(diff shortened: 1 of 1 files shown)


docs/marionette.application.md
```diff
@@ -2,21 +2,34 @@
...
+
+- A namespace to keep things off of the `window`.
+  If you are not using a module loader like ES6 modules, CommonJS, or AMD, then
+  you can use the Application to store your Javascript objects. And if you are
+  using one of those module systems, then you can still attach things to the
+  application to aid in debugging.
+
+- Integration with the Marionette Inspector. The Marionette Inspector is a fantastic tool
+  that makes it easy to understand and debug your application. Using the Application Class
+  will automatically hook up your application to that extension.
```
jmeas: Is this true? Or does it need to be on `window.app`? //cc @jasonLaster

docs/marionette.application.md
```diff
@@ -2,21 +2,34 @@
 
 # Marionette.Application
 
-The `Backbone.Marionette.Application` object is the hub of your composite
-application. It organizes, initializes and coordinates the various pieces of your
-app. It also provides a starting point for you to call into from your HTML
-script block, or directly from your JavaScript files if you prefer to go that
-route.
+The **Application** is a container for the rest of your code. It is recommended
```
megawac: > The **Application** is a container

& bootstrapper?

jmeas: I like the sentiment, but I wonder if I can communicate it in a different way. 'Bootstrapper' seems too colloquial. Nawm sayin?

docs/marionette.application.md
jasonLaster: nope - just needs to exist. And honestly, I'd like to remove the `app` dependency entirely. So, I'm not sure I really need this paragraph

jasonLaster:
I like where this is headed @jmeas - I'll provide more thoughts later

docs/marionette.application.md
```diff
@@ -2,21 +2,34 @@
...
+that every Marionette app have at least one instance of Application.
```
thejameskyle: Isn't it only possible to have one application today?

samccone: you can have multiple, just not multiple routers :)

thejameskyle: ...but i have multiple routers...

jmeas: If you look through the source, it's easy to confirm that you can have any number of Applications or Routers. The only thing to be wary of is a channel conflict, but changing the channel name is documented, and using the channel is deprecated.

thejameskyle: Aha! That's what I was trying to remember

samccone:
:ship: 'd

PR #2189, opened 2015-01-09
LayoutView: Support @ui hash selectors in region definition objects

mattj65817:
LayoutView appears to support @ui-style selectors in the `regions` config object, but only if simple *string* configuration is used:

    var MyLayout = Marionette.LayoutView.extend({
        // ...
        ui: {
            dialog: "div.dialog"
        },
        regions: {
            dialogRegion: "@ui.dialog"  // <-- works fine
        }
    });

However, this does not work if *object* configuration is used, because `Marionette.normalizeUIValues()` does not recurse into objects:

    var MyLayout = Marionette.LayoutView.extend({
        // ...
        ui: {
            dialog: "div.dialog"
        },
        regions: {
            dialogRegion: {
                regionClass: MyCustomRegionClass,
                selector: "@ui.dialog"  // <-- does not work; goes to jQuery.find() without normalization
            }
        }
    });`

I believe (sorry, I am quite new to Marionette) that this means I cannot use UI hash references if I am using a custom region class with a LayoutView. This is somewhat inconvenient.

I have modified LayoutView to support UI hashes in region definition objects by adding a `properties` argument to `Marionette.normalizeUIValues()`. I then modified `Marionette.normalizeUIValues()` to recurse into *object* values, if the `properties` argument is present and is an array. If so, properties of the definition object whose names are included in the `properties` array are normalized if their values are strings.

Not sure that this is the best way to accomplish what I need, but I do feel as though handling should be consistent regardless of whether a region is defined using a simple string value, or an object value with a `selector` property. This is simply one way to accomplish that...

I attempted to create a spec and run tests and all of that, but I am not a Node user and couldn't get the build process to run to completion (bunch of ENOENT errors for whatever reason.) Got tired of trying to figure that stuff out and just decided to submit the pull request. I apologize for the breach of protocol.


mattj65817:
Might have been better for me to leave Marionette.View.normalizeUIValues() alone and just override it for LayoutView. Unless a similar issue exists for other view types.

jasonLaster:
Thanks for the work / research @mattj65817. At first glance this looks good. 

Mind getting some unit tests in there:
+ https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/normalize-ui.spec.js
+ https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/region.spec.js

I'll do a read through later today

mattj65817:
I failed at my last attempt to get Node running, which is why there are no
tests, but I will take another stab at it over the weekend. Thanks.

On Fri, Jan 9, 2015 at 1:38 PM, Jason Laster <notifications@github.com>
wrote:

> Thanks for the work / research @mattj65817 <https://github.com/mattj65817>.
> At first glance this looks good.
>
> Mind getting some unit tests in there:
>
>    -
>    https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/normalize-ui.spec.js
>    -
>    https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/region.spec.js
>
> I'll do a read through later today
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/marionettejs/backbone.marionette/pull/2189#issuecomment-69386889>
> .
>



-- 

Matt Jensen
Developer
matt.jensen@sansio.com

Physio-Control Data Solutions
11 E Superior St, Ste 310, Duluth MN 55802
*Website:* www.physio-control.com/ds

-- 
[CONFIDENTIALITY AND PRIVACY NOTICE]

Information transmitted by this email is proprietary to Physio-Control and 
is intended for use only by the individual or entity to which it is 
addressed, and may contain information that is private, privileged, 
confidential or exempt from disclosure under applicable law. If you are not 
the intended recipient or it appears that this mail has been forwarded to 
you without proper authority, you are notified that any use or 
dissemination of this information in any manner is strictly prohibited. In 
such cases, please delete this mail from your records.
 
To view this notice in other languages you can either select the following 
link or manually copy and paste the link into the address bar of a web 
browser: http://www.physio-control.com/emaildisclaimer/

jasonLaster:
No problem @mattj65817 - note you can run the tests in the browser as well `open SpecRunner.html`

mattj65817:
@jasonLaster I've added tests to normalize-ui.spec.js and layout-view.spec.js.

src/helpers.js
```diff
@@ -88,11 +88,20 @@ Marionette.normalizeUIKeys = function(hash, ui) {
...
-Marionette.normalizeUIValues = function(hash, ui) {
+Marionette.normalizeUIValues = function(hash, ui, properties) {
   _.each(hash, function(val, key) {
     if (_.isString(val)) {
       hash[key] = Marionette.normalizeUIString(val, ui);
     }
+    else if (_.isObject(val) && _.isArray(properties)) {
+      /* Value is an object, and we got an array of embedded property names to normalize. */
+      _.each(properties, function(property) {
+        var propertyVal = val[property];
```
samccone: I think we can be really sneaky here and use recursion 

```js
val = Marionette.normalizeUIValues(_.pick(val, properties), ui)
```

thus cleaning this up just a touch :dancer:

samccone:
hey @mattj65817 After this comments would you mind squashing this commit into a single commit.

Thanks so much for this awesome PR and really great work!

src/helpers.js
mattj65817: Looking for something like this? I believe the above would cause us to lose anything not in `properties`--unless I am misreading something.

    _.extend(val, Marionette.normalizeUIValues(_.pick(val, properties), ui));

samccone: yep you are correct :dancer:

samccone:
:+1: on this (we can squash when we merge) @marionettejs/marionette-core thoughts?

thejameskyle:
@mattj65817 Why do you need to use the same element for both a Region and UI element?

mattj65817:
@thejameskyle I don't *need* to use the UI hash at all, it just seems more consistent (and dry) if I do. The region acts as a container for a dialog. But none of that is really germane to the issue--perhaps it's a poor choice by a Marionette newbie (me)--the reason I submitted this pull request is because the behavior *as it now exists* is inconsistent. If I define my region as a simple string selector, I can refer to the UI hash. If I define my region using an object with an embedded selector, I cannot refer to it. So whether what I'm doing in this case is philosophically right or wrong, my feeling is that the inconsistency in Marionette needed to be addressed.

samccone:
@mattj65817 I am in agreement with you, I think that the current behavior is inconsistent. as for @thejameskyle 's point the UI hash is for more than just using elements one off, it is also for consolidating your DOM interactions into an abstraction layer. So again totally on board with this :ship:

samccone:
moved to #2222

PR #2194, opened 2015-01-12
Fixing the naming inconsistency between child view attribute and event prefix

### LINKED ISSUE #2121: Naming inconsistency between child view attribute and event prefix

SauloSilva:
Replacing the event name of prefix childview to childView

Fix #2121


.editorconfig
```diff
@@ -4,6 +4,7 @@ root = true;
 
 [*]
 trim_trailing_whitespace = true
+trim_trailing_white_space_on_save = false
```
samccone: this is only true for markdown files

lib/backbone.marionette.js
```diff
@@ -2,7 +2,7 @@
 // ----------------------------------
 // v2.3.0
 //
-// Copyright (c)2014 Derick Bailey, Muted Solutions, LLC.
+// Copyright (c)2015 Derick Bailey, Muted Solutions, LLC.
```
samccone: no need to commit these files

samccone:
@SauloSilva  mind rebasing off of major and I can put these in.

Thanks so much :heart:

PR #2197, opened 2015-01-14
v2.3.1

### LINKED ISSUE #2108: Be consistent of each iteration
All over the place there are two patterns I've noticed being used

[**Example a**](https://github.com/megawac/backbone.marionette/blob/helpers/src/helpers.js#L70)

``` js
_.each(_.keys(object), func)
```

[**Example b**](https://github.com/megawac/backbone.marionette/blob/helpers/src/helpers.js#L81)

``` js
_.each(object, func)
```

(Note those are 10 lines apart)

---

Of the two I prefer **b** for simplicity and it will usually be good enough, (unless we should allow `events: {length: 10}`, which I don't believe is useful to support).

The other option is to wait for `underscore@2` or `lodash@3` and upgrade the dependencies and use _.forIn (https://github.com/jashkenas/underscore/issues/1590), and use `forOwn` consistently for these cases

samccone:


samccone:
https://docs.google.com/document/d/1cu39wggfCdWPyEX--q5Y2q0koghjZ07KXVaP1W91cyc/edit

samccone:
@marionettejs/marionette-core let me know :+1: or :-1:

thejameskyle:
:+1:

jfairbank:
:+1:

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+##### Features
+
+* Regions can set a `parentEl` as a way of specifying the DOM tree (default `body`) that they are scoped with. (useful for instance in `LayoutView`).
+
+```js
+  var region = new Region({parentEl: $(“#sub-tree”)})
+```
+
+##### BugFixes
```
jmeas: Bug Fixes

samccone: good call

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+* Layout region lookups are now scoped to the layout and not to the entire DOM.
+
+* Calling `delegateEvents` after the `ui` hash has been modified now works.
+
+* Prevent unsetting event listeners on region swap.
```
jmeas: > When a view is swapped out from a region, but not destroyed, its DOM events will not be removed.

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+* A view's `isDestroyed` state is now explicitly set to false.
```
jmeas: > ...when the view is created.

changelog.md
samccone: :+1:

changelog.md
samccone: :+1:

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+##### Refactors
+
+* Added `Marionette._getValue`. This method is similar to `_.result`. If a function is provided we call it with context otherwise just return the value. If the value is undefined return a default value. This method is private and should not be used directly in your code.
+
+* Refactor pass on `collectionView` index tracking to reduce complexity through the use of a ternary.
```
jmeas: I'm not sure if we need to be this specific in the release notes?

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+* Behavior collision resolution was refactored to take advantage of jquery event namespaces.
```
jmeas: same as above

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+* Refactor minor partial => bind optimization in behaviors.
```
jmeas: Same as above

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+* `_.each` iteration is now consistent across the app, there is no need to do `_.each(_.keys(obj), ...)`
```
jmeas: same as above

samccone: I like noting these changes, sure they are not interesting to everyone but that is why I have them under refactors. They act as a historical marker for us to look back on.

jmeas:
Awesome work @samccone. I think you may have been a bit *too* thorough with the refactors section, though. If all that we're doing is simplifying the logic without changing behavior, then I don't think that we need to put that in the documentation.

A simple:

> Other code refactors and clean ups

is typical to describe this sort of thing.

I consider this README as an avenue to provide valuable information to developers who use Marionette. I can't imagine that anyone cares if we use `_.each` or `_.keys` or whatever in the source. Know what I mean?

megawac:
I tend to agree @jmeas

changelog.md
```diff
@@ -1,3 +1,38 @@
...
+
+##### Bug Fixes
+
+* Layout region lookups are now scoped to the layout and not to the entire DOM.
+
+* Calling `delegateEvents` after the `ui` hash has been modified now works.
+
+* Prevent unsetting event listeners on region swap when a view is swapped out from a region, but not destroyed, its DOM events will not be removed.
+
+* A view's `isDestroyed` state is now explicitly set to false when the view is created.
```
megawac: `false` :p

samccone:
kk

rhubarbselleven:
> A simple
>> Other code refactors and clean ups
> is pretty typical to explain these sorts of changes.

Nooooo !  Thinkof it looking from a source control point of view. If I read a one liner like that, I'd have to trawl the commit to see what actually changed.

History is history, generalising it to broad statements is cool - but it still needs to retain some specifics

jmeas:
> Thinkof it looking from a source control point of view. If I read a one liner like that, I'd have to trawl the commit to see what actually changed.
> History is history, generalising it to broad statements is cool - but it still needs to retain some specifics

As you've pointed out, the specifics aren't going anywhere. They're in the commit history.

The point of the changelog is **not** to mirror the commit history. It is to a tool for developers to use when upgrading their applications.

By including every nitpicky detail, even refactors that **could not conceivably impact anyone's application**, you're making the tool worse. It makes it require more time to read and process, and distracts you from the more important changes that might actually matter to you.

I strongly believe that expounding on refactors is not a good idea for changelogs.

PR #2204, opened 2015-01-15
Enable passing preventDestroy to region.empty.

samccone:


samccone:

src/region.js
```diff
@@ -202,20 +202,29 @@ Marionette.Region = Marionette.Object.extend({
...
-    this._destroyView();
+    if (!preventDestroy) {
+      this._destroyView();
+    }
     this.triggerMethod('empty', view);
 
     // Remove region pointer to the currentView
     delete this.currentView;
+
+    if (preventDestroy) {
```
megawac: I would just do `Marionette._getValue(options, 'preventDestroy', this)` if we want to support function prevent destroy otherwise `_.result(options, 'preventDestroy')`.

Otherwise +1

samccone: :+1: great call

jasonLaster:
:+1:

src/region.js
samccone: ah duh we are already doing this on line 208!

samccone:
into minor we go

Issue #2215, opened 2015-01-19
Marionette.bindEntityEvents docs are misleading


okitu:
Hi there,

is just read the docs for Marionette.bindEntityEvents.

http://marionettejs.com/docs/v2.3.0/marionette.functions.html#marionettebindentityevents

the parameter description for the 'entitiy' say

*The second parameter is the entity (Backbone.Model or Backbone.Collection)
to bind the events from.*

But actually, the entity can be any object, that implements Backbone.Events.
Or am i missing something?

jmeas:
Thanks @okitu! You're absolutely right. Would you have time to make a PR to fix this mistake? The docs are in this repository under the `docs` folder.

okitu:
Done =)

samccone:
this has been fixed :dancer:

PR #2217, opened 2015-01-20
Docfix for #bindEntityEvents

okitu:
see #2215


samccone:
manually picked in

thanks so much @okitu

PR #2222, opened 2015-01-29
Support @ui selectors in region definition objects for LayoutView.

samccone:
via #2189


samccone:
kk I am 100% :+1: on this, will merge unless any objections

cmaher:
looks good :+1:

PR #2226, opened 2015-01-29
Ensure 'params' is defined (fixes #2225)

### LINKED ISSUE #2225: _getValue fails in IE8
When testing my app in IE8, I noticed an error coming from `_getValue()`. It appears that if you pass `undefined` to `apply()` as the second parameter, IE8 errors out:

image

This is also discussed here: http://stackoverflow.com/questions/11112950/function-apply-causing-jscript-object-expected-error-in-ie

Since there are many instances where `_getValue` is called without a 3rd parameter, this case will come up often.

craigjennings11:


jmeas:
ltgm :+1:

@samccone, if you're good we should merge to patch & release.

paulfalgout:
I think you should update https://github.com/marionettejs/backbone.marionette/blob/master/src/helpers.js#L44 and not the built lib file.

jmeas:
ha good catch @paulfalgout.

samccone:
thanks @craigjennings11 moved to here
https://github.com/marionettejs/backbone.marionette/pull/2226

samccone:
@craigjennings11 thanks for the bug report all fixed!

craigjennings11:
Thanks for getting that in.

samccone:
thanks for the awesome issue :+1: :+1:

PR #2231, opened 2015-01-30
Ensure apply is not called with undefined.

samccone:


samccone:
moved via https://github.com/marionettejs/backbone.marionette/pull/2226

jfairbank:
:+1:

PR #2233, opened 2015-01-30
Use call when there are no params

jridgewell:
Still guards against [IE8 apply undefined](https://github.com/marionettejs/backbone.marionette/issues/2225#issuecomment-72225361) by not `apply`ing without params. :wink:


jridgewell:
Ha. This should be against minor.

megawac:
I do that branch issue all the time. :+1: to change

PR #2234, opened 2015-01-30
Use call when there are no params

jridgewell:
Still guards against [IE8 apply undefined](https://github.com/marionettejs/backbone.marionette/issues/2225) by not `apply`ing without params. :wink:

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -43,11 +43,7 @@ Marionette.proxyGetOption = function(optionName) {
 // undefined return a default value
 Marionette._getValue = function(value, context, params) {
   if (_.isFunction(value)) {
-    // We need to ensure that params is not undefined
-    // to prevent `apply` from failing in ie8
-    params = params || [];
-
-    value = value.apply(context, params);
+    value = params ? value.apply(context, params) : value.call(context);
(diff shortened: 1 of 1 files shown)


samccone:
@jridgewell sorry if you rebase you will be inline

jridgewell:
Marionette's branching strategy is all kinds of confusing.

samccone:
nah not really

master is the release branch
```
master 
 - bugfix <- for hotbugfixes
 - minor <- for minor feature dev
 - major <- major feature dev
```

all sub branches must be fast forwarded on release to prevent rebase hell and to keep things inline.

samccone:
:+1: for this

jridgewell:
I've always thought of long-running branches as the root of all evil (and horrible histories).

samccone:
yep they are not disconnected and long running the way we do it :) 

screen shot 2015-01-30 at 12 04 58 pm


there is nothing wrong with long running branches, it is only when they are no longer easily connected to the main common node

(which is why i fast forward all sub branches on release)

jmeas:
Oh, of course! Good idea @jridgewell.

PR #2240, opened 2015-01-31
Add before:show event to ChildView when added after show

paulfalgout:
Essentially duplicated from 
https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L228

Already the behavior if the `CollectionView` is buffering: 
https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L53

But if adding to a collection using 'collection.add' after the 'CollectionView' is shown only the 'show' event was occurring.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -347,6 +347,12 @@ Marionette.CollectionView = Marionette.View.extend({
 
     this.triggerMethod('before:add:child', view);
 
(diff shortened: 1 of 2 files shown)


samccone:
:+1: good call and nice catch

samccone:
heh such an edge case but this is good

Issue #1940, opened 2014-09-30
Default collection event handlers are not copied on setElement


pimlottc:
If you change a view's element via `setElement`, Backbone handles re-binding your events via `undelegateEvents`/`delegateEvents`.  However, this does not include the "automatic rendering" collection bindings for add/remove/reset.  These should be copied as well.

My particular use case is that I have a table widget handles DOM rendering on its own but I'm still using marionette views to consolidate event handling/behavior for individual cells.

paulfalgout:
So I think https://github.com/marionettejs/backbone.marionette/pull/2241 solves this issue.  Backbone's `setElement` calls `undelegateEvents` which previously unbound the entity events.  But moving entity event binding into the constructor alone I believe removes the need to have the events re-bound on `setElement`

paulfalgout:
Update:  I was way off.  To my defense I'm still up all night with a newborn.  The events in question here are only the events bound in `_initialEvents` on `CollectionView` and they really have nothing to do with any of the event delegation, which clearly aren't in use in the fiddle example.

paulfalgout:
So after looking into this a bit further I am unconvinced that this is a bug.

Essentially what is happening is this:

The DOM is being duplicated and the el is set for the collectionview itself, but the el is _not_ re-set for any of the childViews..  so all of the childViews for the original collection would still have _their_ events/triggers and the duplicated DOM ones would not.

This is important because new children are inserted based on other children.. and those children's el is still located in the original collection.
https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L499

So the problem is, if you setElement on a CollectionView, you also have to setElement on the children as well, but this is really not something we could support out of the box as who knows if the new $el has the same children?

I'm closing for now, but if I'm missing something, certainly chime in.

PR #2241, opened 2015-01-31
Remove entity events from delegateEvents/undelegateEvents

paulfalgout:
Readdresses https://github.com/marionettejs/backbone.marionette/pull/1569

One Backbone behavior I noticed this time around is that `delegateEvents` first calls `undelegateEvents`
https://github.com/jashkenas/backbone/blob/master/backbone.js#L1152

However if someone expected a `Marionette.View` to do the same, they may not realize their entity events are stacking.

This is resolved by removing entity events from `delegateEvents`, however I also added `undelegateEntityEvents` as `delegateEntityEvents`'s first action.  This may not be ideal performance or necessary, but it matches `Backbone` behavior.  Thoughts? 

~~Add a test for https://github.com/marionettejs/backbone.marionette/issues/1940~~

### DIFF
--- a/docs/marionette.abstractview.md
+++ b/docs/marionette.abstractview.md
@@ -331,9 +331,9 @@ bound at the time of instantiation, and an exception will be thrown
 if the handlers on the view do not exist.
 
 The `modelEvents` and `collectionEvents` will be bound and
-unbound with the Backbone.View `delegateEvents` and `undelegateEvents`
-method calls. This allows the view to be re-used and have
-the model and collection events re-bound.
+unbound with the Backbone.View `delegateEntityEvents` and `undelegateEntityEvents`
+method calls. `delegateEntityEvents` is called in the View's `constructor` and
+entity events are unbound during the View's `destroy`.
 
 ### Multiple Callbacks
 
--- a/src/abstract-view.js
+++ b/src/abstract-view.js
@@ -25,6 +25,8 @@ Marionette.AbstractView = Backbone.View.extend({
 
     Backbone.View.apply(this, arguments);
 
+    this.delegateEntityEvents();
+
     Marionette.MonitorDOMRefresh(this);
     this.on('show', this.onShowCalled);
   },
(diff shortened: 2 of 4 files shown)


docs/marionette.view.md
```diff
@@ -333,9 +333,9 @@ bound at the time of instantiation, and an exception will be thrown
 if the handlers on the view do not exist.
 
 The `modelEvents` and `collectionEvents` will be bound and
-unbound with the Backbone.View `delegateEvents` and `undelegateEvents`
-method calls. This allows the view to be re-used and have
-the model and collection events re-bound.
+unbound with the Backbone.View `delegateEntityEvents` and `undelegateEntityEvents`
+method calls. `delegateEntityEvents` is called in the View's `constructor` and
+entity events are unbound during the View's `destroy`.
```
samccone: yeah some people totally wont like this, for long lived views... wonder if this should respect preventDestroy ?

paulfalgout: I guess I'm not sure of the issue you're addressing.
`modelEvents` / `collectionEvents` will only be bound in the `constructor`
and they're only removed when the view is destroyed.
If the view has been `preventDestroy`ed, they'll still be listening.

Are you saying they should stop listening if destroy is prevented?

samccone: derp nm nm sorry just trying to wrap my head around this change

samccone:
@paulfalgout mind fastforwarding this

samccone:
@paulfalgout mind making this one commit?

paulfalgout:
@samccone squished

samccone:
@paulfalgout looks like we are stilling missing the test for  #1940 right?

paulfalgout:
@samccone yeah I just added that.. seems like this should fix it, but should be certain

jasonLaster:
Thanks @paulfalgout - I like you're approach while going through the code. Ya know, separate small things into their own code. It's cool.

So my 2 cents:
+ I like `delegateEntityEvents` as its own method - it gives apps more control
+ I see your point that calling `delegateEntityEvents` everytime `delegateEvents` is unecessary, but I don't see this as a perf hit. @megawac, you would know better than I would. Do you think `bindEntityEvents` is expensive relative to `$.on`.
+ Separating out the functionality be default means having two calls to remember in many cases as opposed to one.
+ I can see this refactor being a `6` on the refactoring scale, with a `2` on the benefits scale for most existing apps. For that reason, I think we can make this functionality opt-in


```js
delegateEvents: function() {
  if this.getOption('bindEntityEventsOnDelegateEvents')) {
     this.delegateEntityEvents();
  }

  this.delegateDomEvents();
}
```

The nice thing here is that if you have the kind of app where you are running into this perf issue, then you can flag on this optimization. But for the vast majority of apps, we don't encourage optimizing for this edge-case.

jmeas:
:+1: on this refactor. As an aside, I don't like `delegateEntityEvents` tbqh. In Puppets I implemented it [like so](https://github.com/jmeas/puppets/blob/v0.0.1/src/abstract-view.js#L22-L35).

paulfalgout:
@jasonLaster What is a use case for needing to call both `delegateEvents` and `delegateEntityEvents` in a single call?  They don't seem to me like they do similar work or would need to happen around a single event..

jasonLaster:
@paulfalgout Thanks for the pushback. I'm plus one. 

@jmeas What I hear you saying is that you've got a slicker bindEntityEvents implementation. PR it up w/ spaces.

PR #2246, opened 2015-02-02
Add lodash to test matrix

samccone:


samccone:
ok looks like 8 real failures here, fairly sure they are all related to deep object equality and lodash setting some prop that is causing the diff to fail. any insight @megawac ?

test/ci.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
megawac: Cheater xD

jmeas: rofl

samccone: lol inlining all the code in the yml file is not really scalable

jmeas: true true

jasonLaster: but.... yaml was designed for inlined codes guyz

samccone: :D

PR #2249, opened 2015-02-03
Add a Feature Flag for behavior trigger proxy

jasonLaster:
This commit introduces two things:

1. Basic support for feature flagging marionette breaking changes. I'm really keen to get feature flagging in so we can do productive breaking changes while on minor

2. behavior triggerMethod proxying. triggerMethod as it stands is kinda
crippled without proxying to the view. Proxying to the view, does make things more complicated, but it also makes the events work the first time... like 90% of the time you want proxying. It definitely could be argued both ways.

---

Lets talk about each addition. i think they're both interesting and it's fun to bundle the too, but lets not let one block the other.

 i'll get tests and docs in order when the time comes


src/features.js
```diff
@@ -0,0 +1,8 @@
+Marionette.Features = {
```
thejameskyle: I'd prefer if these were named like CONSTANTS instead of Classes.

samccone:
@jasonLaster minor newline style nit

src/behavior.js
```diff
@@ -35,5 +35,17 @@ Marionette.Behavior = Marionette.Object.extend({
   proxyViewProperties: function (view) {
     this.$el = view.$el;
     this.el = view.el;
+  },
+
+  // First emit on the behavior, then emit on the view context
+  triggerMethod: function() {
```
samccone: can you expand on the use case here... I am not sure of one just yet

src/features.js
jasonLaster: Agreed. I'll make changes

src/behavior.js
jasonLaster: Sure

thejameskyle:
@jasonLaster rebase?

src/behavior.js
cmaher: I have definitely wanted this before. In my case, I had a special type of view (a grid), that has a special action (show/hide columns in the grid). Using grid-specific behaviors, I would like those behaviors to act on the grid. Currently, I have to do 
```js
// in the behavior
this.triggerMethod('columns:show')
this.view.triggerMethod('columns:show')
```

cmaher:
I like the `Marionette.Features`, and I like the feature itself.

jasonLaster:
Killing this PR in favor of a simpler features PR. I'll revisit proxying behavior events another day

Issue #2201, opened 2015-01-15
App Sub Applications


samccone:
Application objects should be able to have "sub" applications that are started and destroyed on parent app state change.

```js
Mn.Application.extend({
  applications: {
    "myApp1": {
      class: MyApp1
      options: {}
    },
    "myApp2": MyApp2,
    "myApp3": {
      getClass: function(){},
      getOptions: {}
    }
  }
})
```

Each application has a name to make it simple to interact with it from within the parent app.

Related #2200

paulfalgout:
Also related https://github.com/marionettejs/backbone.marionette/issues/2026

jasonLaster:
Ooh pretty

JSteunou:
I can live with a removal of Mn.Module with this sub app feature :)

rafde:
Modules gone.

paulfalgout:
While I dunno that this needs to be a part of Mn, I dunno how it related to modules?

rafde:
I thought you needed Modules to have SubApps

paulfalgout:
I think this is suggesting using Applications or some new object as sub apps.  unrelated to modules... kind of like https://github.com/marionettejs/backbone.marionette/issues/2026 which suggests removing modules and adding subapps.  Either way I think it safe to close both of these for now.  Hopefully, for the moment at least, we can focus more on doing less better, rather than adding stuff.

PR #2256, opened 2015-02-04
Kill callbacks

samccone:
This branch only has 1 breaking change
https://github.com/marionettejs/backbone.marionette/pull/2256/files#diff-b4b5b8205cd99b848eb22b4b431bc568L68

But I think we can all agree this is fine and is so so so weird


thejameskyle:
:+1:

samccone:
__-336 LOC__

test/unit/application.spec.js
```diff
@@ -65,20 +65,6 @@ describe('marionette application', function() {
     });
   });
 
-  describe('when an app has been started, and registering another initializer', function() {
```
derickbailey: +1 on killing Marionette.Callbacks - i wrote those before I knew what jQuery.Deffered and Promises were.

-1 for killing this behavior in the Application object. I use this feature, still. There are a lot of systems out there, that use it. 

that doesn't mean you can't get rid of it, but if you destroy this behavior, what will you recommend people do to replace it?

samccone: To be clear the breaking change here is this is __no longer true__

> when an app/module has been started, and registering another initializer it will __automatically be invoked__

This raises the question of what to do if you want this behavior.

Well I would say since you have to have had a reference to your module/app (before this breaking change) you can simply invoke the function with the app/module as the callee of the method with the options you would like via call or apply.

How does this sound to you @derickbailey

megawac: Here's an example from my own code:

```js
var $ = require("jquery");
var Backbone = require("backbone");

var App = require("./controllers/app");
var appRouter = require("./appRouter");
var appParams = require("./controllers/appParams");

// Async stuff
App.addInitializer(appParams.fetch());
App.addInitializer($.ready.promise());

App.addInitializer(() => App.content.load(App));

// Leave this as the absolute last initializer, so routes can be configured first
App.addInitializer(() => Backbone.history.start({
    pushState: true
}));

App.start();
```

Becomes

```js
var $ = require("jquery");
var Backbone = require("backbone");

var App = require("./controllers/app");
var appRouter = require("./appRouter");
var appParams = require("./controllers/appParams");

App.on("start", () => App.content.load(App));

// Leave this as the absolute last initializer, so routes can be configured first
App.on("start", () => Backbone.history.start({
    pushState: true
}));

Promise.all([$.ready.promise(), appParams.fetch()])
.then(() => App.start());
```

/cc @jmeas

jmeas: Code can always be refactored to work without Callbacks. Just use Promises, as in @megawac's example. If there's any specific code you'd like us to look at @derickbailey, you should post a Gist!

derickbailey: a clarifying question, maybe: what's the plan to support loading sub-apps at runtime, and having them start up? 

that's the primary reason i built initializers, back in the day. each sub-app had a reference to the primary app and would register itself when it was loaded, ensuring the sub-app was started either when the app starts or when the sub-app was loaded.

thejameskyle: Well first, the relationship between app and "sub-app" needs to change. Right now the sub-app referring to a sort of global App object is a bit gross. Instead "sub-apps" need to be required into the main application.

Initializers get in the way of that because you're tacking onto a global App object. The easiest migration is just to move things into the "sub-apps" `initialize` method.

paulfalgout:
I'm :+1: on this.  Reasonable v3.0 change with some moderately easy work-arounds if you still need similar functionality.

test/unit/application.spec.js
derickbailey: that doesn't solve the problem: who calls the sub-app's initialize method? when? that's the question that needs to be answered, and was previously answered with this behavior. yes it was "bad" design having the sub-app know of the parent app and register itself. i'm still looking for a better design, 3 years later.

jmeas: > a clarifying question, maybe: what's the plan to support loading sub-apps at runtime, and having them start up?

It's a hope of mine that we stop promoting 'sub-apps' to users of Marionette in the near future.

This is beyond the scope of this PR, but it's my belief that with a stateful router and services pattern, we would find that there's no need for 'container' objects like "Sub Applications." The Application itself should just be an object you create once to get setup with some of the most common features of a Marionette app, like a router.

thejameskyle: We haven't quite agreed on what the api should look like, but it'd be something like this:


```js
var App = Application.extend({
  initialize: function() {
    console.log('app!');
    this.router = new App.Router();
  }
});

App.Router = Router.extend({
  routes: {
    'colors' : function() { return new ColorsApp(); }
  }
});

var ColorsApp = Application.extend({
  initialize: function() {
    console.log('colors!');
  }
});

Backbone.history.start();
var app = new App();
// app!
app.router.navigate('colors');
// colors!
```

paulfalgout: Perhaps this PR should wait on https://github.com/marionettejs/backbone.marionette/issues/2201 to help alleviate concern.

jmeas: > We haven't quite agreed on what the api should look like, but it'd be something like this:

The problem with this API is that it conflating an 'application' with a 'state.' We should have a single application, and that application should be stateful. The router's role should be to manage the state of that application. Continuing with this thought, things defined on the Router should be states, and not applications.

In the end, the functionality and interface might be similar – even exactly the same – to the end user. But the story we tell to explain the concepts will be **far** easier to understand, I think. The only thing you tap into when you talk about 'sub applications' is maybe getting people excited to think that they're making more modular, less coupled code, when they're really not.

derickbailey: hmmmm #2201 reminds me of older EmberJS code that heavily relied on state machines - which is what i think @jmeas is talking about. i agree that routes represent states. 

even if @thejameskyle's example isn't the "correct" way of managing sub-applications in a router, it gets the point across that what i have been calling a sub-application is probably going to be initialized very explicitly in code. even in cases where you are loading another .js resource to bring in more functionality, that code will likely expose an API that the loading code knows about. 

personally, I don't think @thejameskyle's example is bad, even under the description of routes as a state. in this case, the route happens to require a new Application for this particular state. As long as that is clearly stated in documentation and examples, then I don't think there should be an issue. 

so, ok. +1 for removing this behavior if this is the direction that Marionette is heading - to allow multiple Application objects, as needed. i'm assuming (requesting!!!!) that marionette's module system is also going away? please? :)

thejameskyle: For the record, my example was meant to be a way you could do this today.

>  i'm assuming (requesting!!!!) that marionette's module system is also going away? please? :)

yes

jmeas: > hmmmm #2201 reminds me of older EmberJS code that heavily relied on state machines - which is what i think @jmeas is talking about. i agree that routes represent states.

You nailed it! Our router plans are heavily inspired by Ember's routing system.

> i'm assuming (requesting!!!!) that marionette's module system is also going away? please? :)

Hahaha. Yes, they are going away :) At least, they're being removed from the main codebase. The current code will be moved to a separate repository, so all of the people that follow Sulc's and @brian-mann's guides can easily migrate. @thejameskyle also made a fantastic micro-lib that re-creates the namespacing functionality of Modules in like 10 LoC. That guy's brilliant.

jmeas: > For the record, my example was meant to be a way you could do this today.

Ahh. Makes sense!

derickbailey: > You nailed it! Our router plans are heavily inspired by Ember's routing system.

Ember's *old* router style. They dropped the state-machine API some time ago, because it was confusing and difficult for people that aren't intimately familiar with classic state-machine syntax. The current Ember router looks more like Rails' router.

jasonLaster:
@samccone mind rebasing

samccone:
yep np

samccone:
k @jasonLaster

jmeas:
I'm confused. Why are we keeping the `addInitializer` interface? Sure, we could reproduce it...but that isn't the point of removing callbacks, right? The point is that we can both get rid of Callbacks and that API in favor of, `app.on('start', x);`, which provides the same functionality without duplicating any logic.

Let's leverage Backbone.Events here; not create our own quasi-related abstractions.

I'm for this change if we remove the **entire** initializer/finalizer interface.

samccone:
ok I was trying to keep the same interface while removing the callback behaviors

I will add another PR that just tanks initializers if we want

jmeas:
> I will add another PR that just tanks initializers if we want

If you think two PRs makes sense, then k. I think it makes the most sense to just go in this one tho'

PR 1: Delete the feature

vs.

PR 1: Delete some of the feature, change the rest of the feature
PR 2: Delete the part of the feature that you just changed

samccone:
kk

jmeas:
After chatting with @samccone, we decided it makes sense for this to go in as-is. There's a bit of stuff to change with Modules and start-with-parent and all of that.

samccone:
@jmeas after a bit of digging and work I have decided that it is going to be best to do the full rip out of initializers and finalizers from modules and applications due to some further complexity in the startWithParent behavior that modules currently have.

thejameskyle:
@samccone I'd rather open a pull request that ripped those and modules out at the same time.

samccone:
sounds good to me @thejameskyle we are in agreement tho that it will be a new PR

jmeas:
> @samccone I'd rather open a pull request that ripped those and modules out at the same time.

I'd be cool with that, too.

thejameskyle:
@samccone rebase?

samccone:
ha yeah will do again hahah keeps blowing up whenyou guys merge stuff

samccone:
@thejameskyle rebased

src/build/bundled.js
```diff
@@ -32,16 +32,12 @@
 
   Backbone.Marionette = Marionette;
 
-  // Get the Deferred creator for later use
-  Marionette.Deferred = Backbone.$.Deferred;
```
thejameskyle: Wooooooooo!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

samccone:
@thejameskyle @jmeas

thejameskyle:
:+1:

PR #2261, opened 2015-02-04
(#1596) Add 'filter' option to the CollectionView

### LINKED ISSUE #1596: CollectionView / CompositeView should have an optional filter method
Collections have a comparator, and we now respect that in the view layer. However the next most common tasks for lists is to filter / reduce the set of results shown to the user.

There are currently several strategies, such as resetting with a new collection, using a third party plugin like backbone filtered collection, or to hack into `addChildView`.

We should provide a formal way for people to do this to remove the burden from the developer.

```js
Marionette.CollectionView.extend({
  filter/omit/where: function(model, ChildView, index) {
    return true / false
  }
})
```

* https://github.com/marionettejs/backbone.marionette/issues/183
* https://github.com/marionettejs/backbone.marionette/issues/835

cmaher:
Fixes #1596

### DIFF
--- a/api/collection-view.yaml
+++ b/api/collection-view.yaml
@@ -324,3 +324,18 @@ functions:
       @api public
       @param {Marionette.View} view
     
+   filter:
+    description: |
+      ...
+
+      @api public
+      @param {Backbone.Model} model
+      @param {Number} index
+      @param {Backbone.Collection} collection
+
+   setFilter:
+    description: |
+      ...
+
+      @api public
+      @param {Function} function
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -58,6 +58,7 @@ will provide features such as `onShow` callbacks, etc. Please see
 * [CollectionView's attachHtml](#collectionviews-attachhtml)
 * [CollectionView's resortView](#collectionviews-resortview)
 * [CollectionView's viewComparator](#collectionviews-viewcomparator)
+* [CollectionView's `filter`](#collectionviews-filter)
 * [CollectionView's children](#collectionviews-children)
 * [CollectionView destroy](#collectionview-destroy)
 
@@ -301,6 +302,9 @@ is activated, when you sort your `Collection`, there will be no re-rendering, on
 will be reordered. This can be a problem if your `ChildView`s use their collection's index
 in their rendering. In this case, you cannot use this option as you need to re-render each
 `ChildView`.
+
+If you combine this option with a [filter](#collectionviews-filter) that changes the views that are
+to be displayed, `reorderOnSort` will be bypassed to render new children and remove those that are rejected by the filter.
 
 ## CollectionView's `emptyView`
 
@@ -847,6 +851,51 @@ CollectionView allows for a custom `viewComparator` option if you want your Coll
 ```
 
 The `viewComparator` can take any of the acceptable `Backbone.Collection` [comparator formats](http://backbonejs.org/#Collection-comparator) -- a sortBy (pass a function that takes a single argument), as a sort (pass a comparator function that expects two arguments), or as a string indicating the attribute to sort by.
+
+## CollectionView's `filter`
+
+CollectionView allows for a custom `filter` option if you want to prevent some of the
+underlying `collection`'s models from being rendered as child views.
+The filter function takes a model from the collection and returns a truthy value if the child should be rendered,
+and a falsey value if it should not.
+
+```js
+  var cv = new Marionette.CollectionView({
+    childView: SomeChildView,
+    emptyView: SomeEmptyView,
+    collection: new Backbone.Collection([
+      { value: 1 },
+      { value: 2 },
+      { value: 3 },
+      { value: 4 }
+    ]),
+
+    // Only show views with even values
+    filter: function (child, index, collection) {
+      return child.get('value') % 2 === 0;
+    }
+  });
+
+  // renders the views with values '2' and '4'
+  cv.render();
+
+  // change the filter
+  cv.filter = function (child, index, collection) {
+    return child.get('value') % 2 !== 0;
+  };
+
+  // renders the views with values '1' and '3'
+  cv.render();
+
+  // remove the filter
+  // note that using `delete cv.filter` will cause the prototype's filter to be used
+  // which may be undesirable
+  cv.filter = null;
+
+  // renders all views
+  cv.render();
+```
+
 
 ## CollectionView's children
 
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -92,17 +92,17 @@ Marionette.CollectionView = Marionette.View.extend({
   // Handle a child added to the collection
   _onCollectionAdd: function(child, collection, opts) {
     var index;
-
-    this.destroyEmptyView();
-    var ChildView = this.getChildView(child);
-
     if (opts.at !== undefined) {
       index = opts.at;
     } else {
-      index = this._sortedModels().indexOf(child);
+      index = this._filteredSortedModels().indexOf(child);
     }
 
-    this.addChild(child, ChildView, index);
+    if (this._shouldAddChild(child, index)) {
+      this.destroyEmptyView();
+      var ChildView = this.getChildView(child);
+      this.addChild(child, ChildView, index);
+    }
   },
 
   // get the child view by model it holds, and remove it
@@ -134,17 +134,28 @@ Marionette.CollectionView = Marionette.View.extend({
   // all the collectionView
   reorder: function () {
     var children = this.children;
-
-    // get the DOM nodes in the same order as the models
-    var els = _.map(this._sortedModels(), function (model) {
-      return children.findByModel(model).el;
+    var models = this._filteredSortedModels();
+    var modelsChanged = _.find(models, function (model) {
+      return !children.findByModel(model);
     });
 
-    // since append moves elements that are already in the DOM,
-    // appending the elements will effectively reorder them
-    this.triggerMethod('before:reorder');
-    this.$el.append(els);
-    this.triggerMethod('reorder');
+    // If the models we're displaying have changed due to filtering
+    // We need to add and/or remove child views
+    // So render as normal
+    if (modelsChanged) {
+      this.render();
+    } else {
+      // get the DOM nodes in the same order as the models
+      var els = _.map(models, function (model) {
+        return children.findByModel(model).el;
+      });
+
+      // since append moves elements that are already in the DOM,
+      // appending the elements will effectively reorder them
+      this.triggerMethod('before:reorder');
+      this.$el.append(els);
+      this.triggerMethod('reorder');
+    }
   },
 
   // Render view after sorting. Override this method to
@@ -162,7 +173,7 @@ Marionette.CollectionView = Marionette.View.extend({
   // Internal method. This checks for any changes in the order of the collection.
   // If the index of any view doesn't match, it will render.
   _sortViews: function() {
-    var models = this._sortedModels();
+    var models = this._filteredSortedModels();
 
     // check for any changes in sort order of views
     var orderChanged = _.find(models, function(item, index){
@@ -193,6 +204,11 @@ Marionette.CollectionView = Marionette.View.extend({
       this.showCollection();
       this.endBuffering();
       this.triggerMethod('render:collection', this);
+
+      // If we have shown children and none have passed the filter, show the empty view
+      if (this.children.isEmpty()) {
+        this.showEmptyView();
+      }
     }
   },
 
@@ -200,7 +216,7 @@ Marionette.CollectionView = Marionette.View.extend({
   showCollection: function() {
     var ChildView;
 
-    var models = this._sortedModels();
+    var models = this._filteredSortedModels();
 
     _.each(models, function(child, index) {
       ChildView = this.getChildView(child);
@@ -209,7 +225,7 @@ Marionette.CollectionView = Marionette.View.extend({
   },
 
   // Allow the collection to be sorted by a custom view comparator
-  _sortedModels: function() {
+  _filteredSortedModels: function() {
     var models;
     var viewComparator = this.getViewComparator();
 
@@ -221,6 +237,13 @@ Marionette.CollectionView = Marionette.View.extend({
       }
     } else {
       models = this.collection.models;
+    }
+
+    // Filter after sorting in case the filter uses the index
+    if (this.getOption('filter')) {
+      models = _.filter(models, function (model, index) {
+        return this._shouldAddChild(model, index);
+      }, this);
     }
 
     return models;
@@ -520,6 +543,18 @@ Marionette.CollectionView = Marionette.View.extend({
     this.children.each(this.removeChildView, this);
     this.checkEmpty();
     return childViews;
+  },
+
+  // Return true if the given child should be shown
+  // Return false otherwise
+  // The filter will be passed (child, index, collection)
+  // Where
+  //  'child' is the given model
+  //  'index' is the index of that model in the collection
+  //  'collection' is the collection referenced by this CollectionView
+  _shouldAddChild: function (child, index) {
+    var filter = this.getOption('filter');
+    return !_.isFunction(filter) || filter.call(this, child, index, this.collection);
   },
 
   // Set up the child view event forwarding. Uses a "childview:"
--- a/test/unit/collection-view.filter.spec.js
+++ b/test/unit/collection-view.filter.spec.js
@@ -1,0 +1,404 @@
+describe("collection view - filter", function() {
+  'use strict';
+
+  beforeEach(function() {
+    var spec = this;
+
+    this.passModel = new Backbone.Model({foo: true});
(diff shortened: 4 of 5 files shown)


docs/marionette.collectionview.md
```diff
@@ -54,6 +54,7 @@ will provide features such as `onShow` callbacks, etc. Please see
 * [CollectionView's attachHtml](#collectionviews-attachhtml)
 * [CollectionView's resortView](#collectionviews-resortview)
 * [CollectionView's viewComparator](#collectionviews-viewcomparator)
+* [CollectionView's `filter`](#collectionviews-filter)
```
samccone: add docs to the yml also plz

src/collection-view.js
```diff
@@ -499,6 +503,19 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  // Return a function for determining whether or not to display a given model
+  // The function must match the interface (Model) -> Boolean
+  getFilter: function() {
+    return this.getOption('filter');
+  },
+
+  // Return true if the given child should be shown
+  // Return false otherwise
+  testChild: function (child) {
+    var filter = this.getFilter();
```
samccone: Marionette.getOption

src/collection-view.js
```diff
@@ -499,6 +503,19 @@ Marionette.CollectionView = Marionette.View.extend({
...
+    return !_.isFunction(filter) || filter(child);
```
samccone: i think we will want to pass

`child, index, collection`

src/collection-view.js
cmaher: Following the pattern set by `getEmptyView` and `getViewComparator`, I wrapped `getOption('filter')` in the method `getFilter`.

cmaher:
Travis failing on npm install...

src/collection-view.js
samccone: kk, best to make _getFilter private, I dont see any reason to expose that api

samccone:
I reran

src/collection-view.js
cmaher: eh, I'll just remove the getFilter call then. `testChild` is short enough.

samccone:
The following cases should be tested

* adding a model that matches the filter (thus not rendered)
* reseting a collection (with a filter)
* passing in the filter as an option
* a filter based on the index
* removing a child (of the collection) that is matched by the filter

kurt-integral:
This is great! No more role your own filtering!!

kkemple:
This is great! No more role your own filtering!

cmaher:
So, in my mind, the tests do their job, but I wonder if @thejameskyle will approve.

api/collection-view.yaml
```diff
@@ -7,320 +7,328 @@ description:
...
     description: |
       ...
-      
+
       @api public
-    
+
     examples:
-      
+
```
thejameskyle: I remember @jasonLaster telling me these whitespace changes were a no-no

test/unit/collection-view.spec.js
```diff
@@ -532,6 +585,40 @@ describe('collection view', function() {
...
+  describe('when a model that does not pass the filter is removed from the collection', function() {
+    beforeEach(function() {
+      this.model = new Backbone.Model({foo: 'bar'});
+      this.rejectedModel = new Backbone.Model({ foo: 'rejected' });
+      this.collection = new Backbone.Collection([this.model, this.rejectedModel]);
+
+      this.CollectionView = this.MockCollectionView.extend({
+        indexToReject: 1,
+        filter: function(model, index) {
+          return index !== this.indexToReject;
```
thejameskyle: Why not reject based on the `foo` attribute instead?

```js
this.collection = new Backbone.Collection([{ foo: true }, { foo: false }]);
this.CollectionView = this.MockCollectionView.extend({
  filter: function(model) {
    return model.get('foo');
  }
});
```

api/collection-view.yaml
cmaher: For the life of me, I am trying to avoid these changes. Maybe next try will get it.

jasonLaster: White space sensitive

test/unit/collection-view.spec.js
cmaher: This one specifically was adding a test for filtering by index. I could move it to a separate block, though.

thejameskyle: Oh, could you label it that way then

samccone:
ok this is looking real nice, left some words, just squash and ping me when you are ready for another :eyes:

docs/marionette.collectionview.md
samccone: :heavy_check_mark:

docs/marionette.collectionview.md
```diff
@@ -783,6 +784,51 @@ CollectionView allows for a custom `viewComparator` option if you want your Coll
...
+      { value: 4 }
+    ]),
+    filter: function (child, index, collection) {
+      return child.get('value') % 2 === 0;
+    }
+  });
+
+  cv.render(); // renders the views with values '2' and '4'
+
+  cv.setFilter(function (child) {
```
samccone: lets put all the arguments on each of these, just to be consistent

docs/marionette.collectionview.md
```diff
@@ -783,6 +784,51 @@ CollectionView allows for a custom `viewComparator` option if you want your Coll
...
+    return child.get('value') % 2 !== 0;
+  }); // re-renders the collection view, showing the views with values '1' and '3'
+
+  cv.setFilter(function () { return false; }); // re-renders the collection view, showing the empty view
```
samccone: ^ ditto

docs/marionette.collectionview.md
```diff
@@ -783,6 +784,51 @@ CollectionView allows for a custom `viewComparator` option if you want your Coll
...
+
+  cv.setFilter(function () { return true; }, { preventRender: true }); // update the filter, but does not re-render
```
samccone: ^ ditto

src/collection-view.js
```diff
@@ -499,6 +506,30 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  //  'collection' is the collection referenced by this CollectionView
+  testChild: function (child, index) {
+    var filter = this.getOption('filter');
+    return !_.isFunction(filter) || filter.call(this, child, index, this.collection);
+  },
+
+  // Update the filter for this collection view
+  // The filter will be used by testChild instead of the previous filter
+  // This will cause the view
+  setFilter: function (filter, options) {
```
samccone: Does anyone else think that we should have a `before:set:filter` and a `set:filter` event pair here?

or is this just crazy

test/unit/collection-view.spec.js
```diff
@@ -1276,4 +1276,312 @@ describe('collection view', function() {
...
+          this.collectionView.setFilter(this.inverseFilter);
+        });
+
+        it("should set the filter", function() {
+          expect(this.collectionView.filter).to.equal(this.inverseFilter);
+        });
+
+        it("should not render the view", function() {
+          this.collectionView.setFilter(this.inverseFilter);
+          expect(this.collectionView.render).not.to.have.been.called;
```
samccone: we should put a 

.called.once
.called.with

on all of these. These assertions have caught some nasty issues in the past for us. 

remember @thejameskyle  :dancer:

src/collection-view.js
cmaher: do you have a use case? every time the filter changes, render will be called, so the standard render events will happen.

samccone: what if someone wanted to trigger some sort "ui" action or display some info

or am I just making up crazy ideas?

cmaher: I don't quite see it. The difference between `before:set:filter` and `set:filter` is simply an assignment `this.filter = filter`. Wrapping that in a lifecycle call seems a bit weird to me, and I picture overriding being more common (`if (this.isLocked()) return`). Maybe we should wait until people are using this to see if there's a use case for `before:set:filter` and `set:filter`

samccone: kk

just a thought

samccone:
so close to the :checkered_flag:

test/unit/collection-view.spec.js
cmaher: added it to the remaining methods (that aren't called with any params)

samccone:
:+1:

test/unit/collection-view.spec.js
thejameskyle: **
`.calledOnce`
`.calledWith`

cmaher: it's not being called

megawac:
+1

src/collection-view.js
```diff
@@ -499,6 +506,30 @@ Marionette.CollectionView = Marionette.View.extend({
...
+    if (this.filter === filter) { return; }
+    options = options || {};
+    this.filter = filter;
+    if (this.isRendered && !this.isDestroyed && !options.preventRender) {
```
megawac: just do `_.result(options, 'preventRender')`

test/unit/collection-view.spec.js
```diff
@@ -1276,4 +1276,316 @@ describe('collection view', function() {
       expect(this.childView.$el).to.contain.$text('bar');
     });
   });
+
+  describe("filtering", function () {
```
jasonLaster: I think this deserves it's own file.

src/collection-view.js
```diff
@@ -499,6 +506,30 @@ Marionette.CollectionView = Marionette.View.extend({
...
   },
 
+  // Return true if the given child should be shown
+  // Return false otherwise
+  // The filter will be passed (child, index, collection)
+  // Where
+  //  'child' is the given model
+  //  'index' is the index of that model in the collection
+  //  'collection' is the collection referenced by this CollectionView
+  testChild: function (child, index) {
```
jasonLaster: What do other people think of `testChild` the name...

cmaher: suggestions for something better are certainly welcome. `filterChild` was the only other thing I thought of, but that didn't seem right

src/collection-view.js
jasonLaster: yup

src/collection-view.js
jmeas: I dislike the name `testChild`

jasonLaster: yea or `shouldFilterChild`

cmaher: `shouldRenderChild`?

jridgewell: I've always thought any use of "child" is confusing. "Is it an `ItemView` or a `Model`?"

test/unit/collection-view.spec.js
samccone: :+1:

src/collection-view.js
samccone: we have `childView` in the vernacular of Marionette so I think that language works

what about a predicate name? something like `isChildFiltered`

cmaher: so 'isFiltered' makes me think that it's not going to be in there, even though returning 'true' from filter means the model passes.
...
`doesChildPassFilter`?

cmaher: or `shouldDisplayChild`, which, in the future, might contain more than filter stuff (pagination?)

jasonLaster:
:+1:

samccone:
in we goooooooo

jmeas:
What does this do for the `length` property of the CollectionView? I think it would be prudent for us to officially define what the length is.

Previously, the collection view displayed *exactly* what was in the collection, so it made sense for the `length` to match that. Now we have two lengths: the length of the collection, and the length of what is displayed.

I would think that the length would track the number of displayed child views. Does it?

cmaher:
[ ] tests for children length
[ ] make `checkEmpty` display the empty view if the entire collection is rejected by the filter

src/collection-view.js
```diff
@@ -499,6 +506,30 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  //  'index' is the index of that model in the collection
+  //  'collection' is the collection referenced by this CollectionView
+  shouldAddChild: function (child, index) {
+    var filter = this.getOption('filter');
+    return !_.isFunction(filter) || filter.call(this, child, index, this.collection);
+  },
+
+  // Update the filter for this collection view
+  // The filter will be used by shouldAddChild instead of the previous filter
+  // This will cause the view
```
jmeas: This will cause the view...

jmeas:
How do folks feel about the `setFilter` api? Why not just say:

> Update the `filter` property and call `render`

?

jmeas:
An alternative, shorter name for the option would be `render`. `render: true` is the default. pass `render:false` to turn it off.

I suppose `preventRender` is more consistent w. `preventDestroy`, but I'm not sure if I like that option, either.

src/collection-view.js
rhubarbselleven: I think it should start with `is` for readability. The prefix of `is` denotes a boolean result.

cmaher:
I think `setFilter` is necessary for the purpose of doing something whenever the filter changes. And I was going with consistency with `preventRender`.

samccone:
ok @cmaher and I chatted about this and for the 2.4 release we are going to do a limited version of this feature. basically omitting the setFilter method (and possible removeFilter method)

This lets us discuss the future API without holding up this great addition to 2.4.

jmeas:
:+1: @samccone @cmaher

cmaher:
Hold on this. This definitely needs some work to be compatible with #2302.

cmaher:
Updated tests and code to factor in sorting + reordering

src/collection-view.js
```diff
@@ -522,6 +545,18 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  // Return false otherwise
+  // The filter will be passed (child, index, collection)
+  // Where
+  //  'child' is the given model
+  //  'index' is the index of that model in the collection
+  //  'collection' is the collection referenced by this CollectionView
+  shouldAddChild: function (child, index) {
+    var filter = this.getOption('filter');
+    return !_.isFunction(filter) || filter.call(this, child, index, this.collection);
+  },
```
jmeas: Should this be internal?

samccone: yeah I think it should be, no need to expose this method since we already give people the `filter` method to set

samccone: great catch @jmeas

test/unit/collection-view.filter.spec.js
```diff
@@ -0,0 +1,398 @@
...
+      this.collectionView = new this.CollectionView({
+        childView: this.ChildView,
+        reorderOnSort: true
+      });
+    });
+
+    describe("when the filter does not change between sorts", function() {
+      beforeEach(function() {
+        this.collectionView.render();
+        this.collection.sort();
```
ahumphreys87: Because `render` is called first I dont actually think the `sort` will trigger a reorder as the order hasn't changed. Maybe try silently adding a model to the start of the collection and then calling `sort`

ahumphreys87: you need to change comparator after the initial render, something like:
```js
this.collectionView.render();
this.collection.comparator = 'foo';
this.collection.sort();
```

test/unit/collection-view.filter.spec.js
```diff
@@ -0,0 +1,401 @@
...
+      it("renders the views in the correct order", function() {
+        expect(this.collectionView.$el).to.contain.$text('31');
+      });
+    });
+
+    describe("when the filter changes between sorts", function() {
+      beforeEach(function() {
+        this.collectionView.render();
+        this.collectionView.filter = this.inverseFilter;
+        this.collection.sort();
```
ahumphreys87: change comparator

test/unit/collection-view.filter.spec.js
```diff
@@ -0,0 +1,401 @@
...
+      it("renders the views in the correct order", function() {
+        expect(this.collectionView.$el).to.contain.$text('13');
+      });
+    });
+
+    describe("when the filter changes between sorts", function() {
+      beforeEach(function() {
+        this.collectionView.render();
+        this.collectionView.filter = this.inverseFilter;
+        this.collection.sort();
```
ahumphreys87: change comparator

test/unit/collection-view.filter.spec.js
cmaher: In this case, I want to make sure that only changing the filter will cause the view to re-render. The expectation is that a collectionview will update on a sort, and since the filter can possibly depend on the index, we want that to cause a render

ahumphreys87: Ok I follow now :)

cmaher: The comparator in the `reorder` test still needs to change, though :flushed:

samccone: :clap: to this really indepth review guys

samccone:
@cmaher just ping me when you want another set of :eyes:

ahumphreys87:
:+1: from me, can't wait for this @cmaher!

samccone:
picked into 2.4

cmaher:
woooooooooo :dancers:

Issue #2133, opened 2014-12-13
Remove unused event listener in ItemView


jmeas:
[This handler](https://github.com/marionettejs/backbone.marionette/blob/acb44af7ccd11f841dce0da7f4ea48204430bb93/src/marionette.view.js#L22) is used by CollectionView, but not ItemView. It should be removed.

Issue #1263, opened 2014-04-29
Allow for LayoutView.prototype.render to be non-destructive


jmeas:
Repeatedly calling LayoutView.prototype.render will destroy all of the views within the layout's regions, and those regions themselves, and completely re-set the layout's regions.

Yikes.

That's a lot. I'm not sure if we should do all of that on a re-render. This is to discuss the possibility of changing this behavior or maybe allowing an option as a parameter of render to change the behavior.

The original discussion can be found in #877.

ianmstew:
Leaving child view re-rendering up to the developer as an option makes sense.  This could be done with a detach/re-attach of Regions like discussed here.  I haven't had this problem myself because I write my templates/views in such a way that I never need to re-render a layout, unless it's trivial.

rbu:
Marionette 3.1 now comes with a "detachChildView" function that allows you to re-attach a child view later on, which can be used to build this:

http://marionettejs.com/docs/v3.1.0/marionette.view.html#detaching-a-child-view

paulfalgout:
Agreed.  

There's a little bit of complexity here dealing with how to either reset or rebuild the regions.. and what to do it.. for instance you re-render the parent view, and the child view detaches.. but then the DOM element for the region renders out of the parent view.. I suppose you then destroy the child view, or do you keep it around for later?  then what if you render it again and the region renders back in?  what are the strategies for knowing if child views need to be re-rendered or if they're merely detached

rafde:
moved to https://gitter.im/marionettejs/topics/topic/57fb29d884f1db0614985140/allow-for-layoutviewprototyperender-to-be-non-destructive

paulfalgout:
As we start to look at this in 4.x I think there are 3 scenarios.  
1. marionette managed destructive regions on render (what it is now)
2. marionette managed non-destructive regions (ie: marionette detaches/reattaches on render)
3. marionette non-managed regions (ie: the renderer must know to render around the regions content on render)
I think we need to setup a way to allow for this to be configurable per view, but _not_ per region.

PR #2262, opened 2015-02-04
Remove unused listener from base view class.

samccone:
because good call @jmeas in #2133

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -29,6 +29,7 @@ Marionette.CollectionView = Marionette.View.extend({
 
     Marionette.View.apply(this, arguments);
 
+    this.on('show', this._onShowCalled);
     this.initRenderBuffer();
   },
 
@@ -111,8 +112,7 @@ Marionette.CollectionView = Marionette.View.extend({
     this.checkEmpty();
   },
 
-  // Override from `Marionette.View` to trigger show on child views
-  onShowCalled: function() {
+  _onShowCalled: function() {
     this.children.each(_.partial(this._triggerMethodOnChild, 'show'));
   },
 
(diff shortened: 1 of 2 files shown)


thejameskyle:
Hmm? https://github.com/marionettejs/backbone.marionette/blob/sjs/fix-unused-listener/src/collection-view.js#L116-L118

samccone:
yeah it triggers show on all childviews when the parent is shown

thejameskyle:
oooooh got it

:+1:

paulfalgout:
Hmm I'd be up for renaming `onShowCalled` to something else.. not sure what.. `_onShow` `_proxyShowEvent` I dunno... at least `_onShowCalled`
Also `onShowCalled` should be removed from view https://github.com/marionettejs/backbone.marionette/blob/minor/src/view.js#L138

This is breaking, kind of, since it doesn't follow naming conventions..  I don't think this functionality is documented, but someone somewhere could be doing this:
```
  onShow: function() {
    this.doSomething();
  },
  onShowCalled: function() {
    alert('Why God why?!?');
  }
});
```

samccone:
yeah we say it is internal
https://github.com/marionettejs/backbone.marionette/blob/minor/src/view.js#L137

good catch... deleting

also @paulfalgout i am all for renaming the method

paulfalgout:
Ah!  Ha.  Missed the inline comment :-)

paulfalgout:
:+1:

jmeas:
It should describe what it's doing, so I'm in favor with a name like `_propagateShowEvent`.

jmeas:
Also, I could actually see regular old LayoutViews propagating this event to their children views, too. This will be of particular importance after #1263.

samccone:
yeah i could see that also @jmeas this is mostly just a minor branch fixup

jmeas:
:+1:

thejameskyle:
:+1:

Issue #2242, opened 2015-01-31
View options inconsistency


paulfalgout:
This issue is similar to https://github.com/marionettejs/backbone.marionette/issues/2229

When adding options to a `Marionette.View` you can either add them to the `View` definition:
```
var MyView = Marionette.View.extend({
  options: {}  // can also be a function returning a hash
});
```
or you can pass the options when instantiating:
```
var options = {};  //can also be a function returning a hash
var myView = new Marionette.View(options);
```
In both cases they are added to `myView.options` here: 
https://github.com/marionettejs/backbone.marionette/blob/master/lib/backbone.marionette.js#L1632

**Backbone Views**
Backbone Views used to attach options in a very similar way.  https://github.com/jashkenas/backbone/blob/1.0.0/backbone.js#L1087
But this functionality was removed in `v1.1.0`

**The issue**
Marionette does attach `options` to the `View` as Backbone used to, however it does not pass the `options` hash to the `Backbone.View` `constructor`.  So if you add options to the `View`'s definition as in the first example, or if you return `options` as a function when passing them during instantiation, you cannot expect the `Backbone.View` to receive these `options`.

**Examples**
```
// As Expected

var options = {
  className: '.some-class'
};  

var myView = new Marionette.View(options);

console.log(myView.options.className === '.some-class'); // -> logs true
console.log(myView.className === '.some-class'); // -> logs true
```

```
// Unexpected

var MyView = Marionette.View.extend({
  options: {
    className: '.some-class'
  }  // can also be a function returning this hash
});

var myView = new MyView();

console.log(myView.options.className === '.some-class'); // -> logs true
console.log(myView.className === '.some-class'); // -> logs false
```

As in https://github.com/marionettejs/backbone.marionette/issues/2229
```
// Unexpected

var options = function(){
  return {
    className: '.some-class'
  };
};  

var myView = new Marionette.View(options);

console.log(myView.options.className === '.some-class'); // -> logs true
console.log(myView.className === '.some-class'); // -> logs false
```

**Possible solutions**

1. We could remove the ability for `options` to be a function and/or for `options` to be attached to the `View` definition.  I think I prefer both of these in the long term.  So essentially just `this.options = options || {};`

[...]

PR #2263, opened 2015-02-04
Make view options implementation consistent

paulfalgout:
This addresses fixing the bug related to https://github.com/marionettejs/backbone.marionette/issues/2242

While fixing the issue related to the `sort` option on `CollectionView` when `options` is a function, I ran into an additional bug.

Because `this.sort` is utilized it is possible to set the `sort` for future instantiations of an extended `CollectionView`  I've fixed this bug in this PR as well, but can split it up if necessary.

If this :eyes: :+1: I'll get the tests written / adjusted

### Bug proof
As Expected:
```
var MyCollectionView = Marionette.CollectionView.extend(); // Not setting sort

var myCollectionView = new MyCollectionView({ sort: true });

console.log(myCollectionView.sort); -> logs true
```

Unexpected:
```
var MyCollectionView = Marionette.CollectionView.extend({ sort: false }); 

var myCollectionView = new MyCollectionView({ sort: true });

console.log(myCollectionView.sort); -> logs false
```

**Todo**
- [x] Test for `Backbone.View` `viewOptions` when `options` passed as function
- [x] Test for `Backbone.View` `viewOptions` when `options` is set on `View` definition
- [x] Test for 'sort' when `CollectionView` instantiation overrides `CollectionView` definition `sort`

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -11,6 +11,9 @@ Marionette.CollectionView = Marionette.View.extend({
   // that are forwarded through the collectionview
   childViewEventPrefix: 'childview',
 
+  // flag for maintaining the sorted order of the collection
+  sort: true,
+
   // constructor
   // option to pass `{sort: false}` to prevent the `CollectionView` from
   // maintaining the sorted order of the collection.
@@ -19,10 +22,6 @@ Marionette.CollectionView = Marionette.View.extend({
   // option to pass `{comparator: compFunction()}` to allow the `CollectionView`
   // to use a custom sort order for the collection.
   constructor: function(options){
-    var initOptions = options || {};
-    if (_.isUndefined(this.sort)){
-      this.sort = _.isUndefined(initOptions.sort) ? true : initOptions.sort;
-    }
 
     this.once('render', this._initialEvents);
     this._initChildViewStorage();
@@ -30,6 +29,7 @@ Marionette.CollectionView = Marionette.View.extend({
     Marionette.View.apply(this, arguments);
 
     this.on('show', this._onShowCalled);
+
     this.initRenderBuffer();
   },
 
@@ -83,7 +83,7 @@ Marionette.CollectionView = Marionette.View.extend({
       this.listenTo(this.collection, 'remove', this._onCollectionRemove);
       this.listenTo(this.collection, 'reset', this.render);
 
-      if (this.sort) {
+      if (this.getOption('sort')) {
         this.listenTo(this.collection, 'sort', this._sortViews);
       }
     }
@@ -321,7 +321,7 @@ Marionette.CollectionView = Marionette.View.extend({
   // Internal method. This decrements or increments the indices of views after the
   // added/removed view to keep in sync with the collection.
   _updateIndices: function(view, increment, index) {
-    if (!this.sort) {
+    if (!this.getOption('sort')) {
       return;
     }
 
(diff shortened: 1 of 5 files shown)


jasonLaster:
I like the change. Let's get the tests in. Btw, I think Travis failed

jmeas:
:+1:

jfairbank:
Nice! Let's get those tests.

src/collection-view.js
```diff
@@ -19,16 +19,16 @@ Marionette.CollectionView = Marionette.View.extend({
...
-      this.sort = _.isUndefined(initOptions.sort) ? true : initOptions.sort;
-    }
 
     this.once('render', this._initialEvents);
     this._initChildViewStorage();
 
     Marionette.View.apply(this, arguments);
 
+    if (_.isUndefined(this.sort)) {
+        _.defaults(this.options, { sort: true });
```
thejameskyle: Do we do this sort of thing anywhere?

paulfalgout: I'm not sure?
Originally with it being `this.sort = _.isUndefined(initOptions.sort) ? true : initOptions.sort;`  all I've really changes is `this.sort` -> `this.options.sort` but `_.defaults` seemed more readable.

samccone:
yes yes yes

funny we did something similar with behaviors, what a nasty bug indeed

test/unit/view.spec.js
```diff
@@ -192,6 +192,29 @@ describe('base view', function() {
     });
   });
 
+  // http://backbonejs.org/#View-constructor
```
jasonLaster: Yep, this is great.

src/collection-view.js
```diff
@@ -19,10 +19,6 @@ Marionette.CollectionView = Marionette.View.extend({
   // option to pass `{comparator: compFunction()}` to allow the `CollectionView`
   // to use a custom sort order for the collection.
   constructor: function(options){
-    var initOptions = options || {};
-    if (_.isUndefined(this.sort)){
-      this.sort = _.isUndefined(initOptions.sort) ? true : initOptions.sort;
-    }
```
jasonLaster: Ugly :)

src/collection-view.js
```diff
@@ -30,6 +26,11 @@ Marionette.CollectionView = Marionette.View.extend({
     Marionette.View.apply(this, arguments);
 
     this.on('show', this._onShowCalled);
+
+    if (_.isUndefined(this.sort)) {
+        _.defaults(this.options, { sort: true });
+    }
```
jasonLaster: So a couple thoughts. 
1) this boilerplate rubs me the wrong way
2) I think sort: true as a prototype property would work, subclasses could override in their proto or initialize
3) getOption should take a default as a last param. It's been discussed.

jasonLaster:
Really nice work. I'd like to get this in soon!

src/collection-view.js
paulfalgout: Yeah I agree.  This part of this PR was a bit of a rabbit trail, so I was attempting to keep it similar, but I could help making at least more readable.
I'm all for putting sort on the prototype. That would replace this entire block.  
As for #3, I think that'd be useful assuming it sticks around, though we wouldn't need it here.

samccone:
@paulfalgout just a little but of git foo needed on this.
can you pull the tests for the view constructor into the first commit, and then the tests for sort onto the second commit.

in the end there should be two commits. After that :shipit: :rocket:

paulfalgout:
Git foo'd :shipit: :rocket:

PR #2264, opened 2015-02-04
Eliminate _setLength for micro optimization

jfairbank:
Micro optimization to remove `_setLength` from `RegionManager`. It could be called multiple times thanks to `addRegions`, meaning we were calling `_.keys` multiple times when we just need to increment by one for each iteration of `addRegions`.

### DIFF
--- a/src/region-manager.js
+++ b/src/region-manager.js
@@ -5,6 +5,7 @@
 Marionette.RegionManager = Marionette.Controller.extend({
   constructor: function(options) {
     this._regions = {};
+    this.length = 0;
 
     Marionette.Controller.call(this, options);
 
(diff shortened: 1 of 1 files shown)


paulfalgout:
So if you call `addRegion` twice with the same name `this.length !== _.size(this._regions)`

jfairbank:
@paulfalgout Uggh good point.

jmeas:
I'd rather see more drastic overhauls to the RegionManager / Babysitter / CollectionView classes. We really need a unified List/Enumerable/whatever-you-want-to-call-it Class/Mixin that manages all of this stuff for us in a smarter way.

jfairbank:
@jmeas completely agree. I saw this as a low-hanging fruit while working on the initial mixin. We should get together and discuss the overhaul ideas. I see at first abstracting out region manager completely to the mixin as a first step at least.

thejameskyle:
:+1:

samccone:
nice find @jfairbank

PR #2269, opened 2015-02-04
Remove Channels from the Application. Remove Wreqr as a dependency.

jmeas:
:wave: wreqr

:wave: app channel


paulfalgout:
:tada: :balloon: :birthday:

api/application.yaml
```diff
@@ -216,7 +177,7 @@ functions:
           
           MyApp.start(options);
           ```
-
```
samccone: watch the deleting of WS here .. because in yaml it is needed

bower.json
```diff
@@ -37,7 +37,6 @@
     "backbone": "1.0.0 - 1.1.2",
     "underscore": "1.4.4 - 1.6.0",
     "backbone.babysitter": "^0.1.0",
-    "backbone.wreqr": "^1.0.0",
```
samccone: :wave:

samccone:
this commit should add in radio, because the tests really will not change so much

jmeas:
> this commit should add in radio, because the tests really will not change so much

Let's hold off on adding Radio until we introduce a PR that actually requires that we use Radio. I think that if the library doesn't **need** it, then it should be an extension that we officially promote, like `ngRoute`. We can document it in our documentation and go the whole nine yards with telling people about it, but just not bundling it up by default.

megawac:
^ +100

samccone:
yeah I gotcha @jmeas  after a chat

samccone:
:+1:

thejameskyle:
:+1:

PR #2271, opened 2015-02-04
Backbone Metal

thejameskyle:
This is as close as I can get before hundreds of tests start breaking for no apparent reason.


jasonLaster:
Mind writing up some next steps so that we can see how you plan on upgrading the rest of the lib

thejameskyle:
I've been creating other issues for it

jasonLaster:
cool

samccone:
reabase

thejameskyle:
done

src/metal.js
```diff
@@ -0,0 +1,42 @@
...
+  return Marionette.Class.extend(
+    _.extend({ constructor: obj }, _.omit(obj.prototype, _.keys(Backbone.Events))),
+    _.omit(obj, _.keys(Marionette.Class))
+  );
+}
+
+Marionette.deprecate = Metal.deprecate;
+Marionette.Mixin      = Metal.Mixin;
+Marionette.Events     = Backbone.Events     = new Marionette.Mixin(Backbone.Events);
+Marionette.Model      = Backbone.Model      = classify(Backbone.Model);
```
jasonLaster: Tjk, mind adding some docs for this. 

Also lets setup a call to discuss it as well. I'd like to make sure we're all on the same page here.

thejameskyle: I will add docs. We also spoke about this in the last call.

blikblum: It'll be required to patch base Backbone objects (Backbone.Events and Backbone.Model) to get Marionette features working in v3?

If not, i suggest to not patch base Backbone objects. Those that need such features could, optionally, use Marionette.Events and Marionette.Model.

jasonLaster:
:+1:

src/metal.js
paulfalgout: @blikblum this is an issue on metal https://github.com/marionettejs/backbone-metal/issues/6 but I agree, I'd like to see that issue resolved prior to merging this.

paulfalgout:
Yeah.  Love this. :+1:

samccone:
I would like some language around this still, before we merge to let people know what they can do withthis

src/metal.js
jmeas: > Those that need such features could, optionally, use Marionette.Events and Marionette.Model.

:+1:

thejameskyle: @paulfalgout that issue was already fixed.

thejameskyle: @blikblum Yes it is required to maintain the correct prototype chain when we extend from Backbone.

jasonLaster:
bump

samandmoore:
As an outsider, is there anywhere to find some history about why this is happening and what Backbone.Metal is about?

paulfalgout:
It seems to be inspired by ember-metal?  Very base functions not meant to be used directly.  I don't really know the organizational rational entirely.  In my opinion, the most exciting thing this adds is [_super](https://github.com/marionettejs/backbone-metal#super)

samccone:
xref #2278

still waiting on a critical writeup of this.

I would like the following questions answered before we decide how to proceed.

* What does this add to marionette?
* Why it this is a good thing for marionette users?
* How will this change/improve developers experience with marionette?

PR #2272, opened 2015-02-04
Deprecate getOption

thejameskyle:
Reopening this from #2073.

I think this should go in


jmeas:
Yes :+1:

megawac:
I'm not so sure if `getOption` should be deprecated. It is kind of handy and makes managing options simpler. Also it will be a pain for people to change when upgrading as it's evaluated under the hood of Marionette (not a ctrl+f problem).

jasonLaster:
:-1: 

`getOption` is something that we use extensively within Marionette. It is also something that many people use throughout their apps. I think we have much bigger battles to fight than the one where we tell people that the way they access object properties is wrong.

samccone:
I am all for holding on this one. As @jasonLaster said, much larger battles to fight on this one

thejameskyle:
We don't have to remove it from the codebase right away. By making it deprecated we're telling people it's a bad idea.

We should push people to move away from doing shit like this:

```js
Marionette.Foo.extend({
  options: { /* gross... */ }
});
```

```js
new Marionette.Foo(function() {
  return { /* why are you doing this to me */
});
```

jasonLaster:
@thejameskyle I'm not persuaded that these things are *bad*.

In both of those cases, that's functionality that a developer opted into using and has decided that would be useful. The most common case is obviously `new ItemView({foo:foo})` which is okay. 

If we weren't consolidating our views, introducing a new architecture, and inserting a better class system I think this would be worth considering. But at the moment, this just doesn't seem like a priority.

thejameskyle:
> But at the moment, this just doesn't seem like a priority.
That's fine

paulfalgout:
I agree with @thejameskyle that passing view options as a function or defining them on the view definition is weird.

```
var myView = new Mn.View.extend({
  someOption: true
});
console.log(myView.getOption('someOption'));  -> logs true
```
```
var myView = new Mn.View.extend({
  options: {
    someOption: true
  }
});
console.log(myView.getOption('someOption'));  -> logs true
```

Also note that neither of these functionalities have really worked since `Backbone 1.0.0`

thejameskyle:
@jasonLaster let's leave this open for the future. Last time we spoke about this we were all agreed on it.

paulfalgout:
To do think `getOption` is important to prevent bugs like: https://github.com/marionettejs/backbone.marionette/pull/2263

But I think `options` can be a simple hash always, just like Backbone.

jmeas:
`getOptions` exists to solve a problem that no longer exists in Backbone applications. The problem was that `options` used to be attached to the instance directly, and it caused confusion between accessing a property through `this.x` or `this.options.x`.

Marionette's solution to this problem was introducing `this.getOption`. Backbone's solution to this problem was no longer attaching options to the instance.

Marionette continues to attach the options to the instance, so we continue to 'need' `getOption`, and are now inconsistent with Backbone primitives.

Going forward I would like to see us move away from attaching options to the instance for consistency. When that day comes, we could get rid of `getOption`.

A solution I've been using that is really baller is `mergeOptions`, where you pass it options and a list of acceptable keys and it picks the right values to merge onto the instance. Then, later, you access those props via `this.x`.

thejameskyle:
@jmeas Agreed. I'm assuming you mean for mergeOptions to look _something_ like this?

```js
Class.extend({
  initialize: function(options) {
    this.mergeOptions(options, ['foo', 'bar']);
  },

  mergeOptions: function(options, merge) {
    _.extend(this, _.pick(options, merge));
  }
});
```
and not:
```js
Class.extend({
  constructor: function(options) {
    this._mergeOptions(options, this.mergeOptions);
  },
  mergeOptions: ['foo', 'bar'],
  _mergeOptions: function(options, merge) {
    _.extend(this, _.pick(options, merge));
  }
});
```
for the reasons I outlined before related to subclassing

jmeas:
Correct-a-mundo @thejameskyle.

jasonLaster:
Hmm, so I think we can move forward this way. We can add mergeOptions in minor and then transition our class properties over to mergeOptions on major or behind a flag. 

I don't want to be opposed to better things. We can encourage mergeOptipns and start using them without stressing our community out. The typical member won't be thrilled to refactor their working code, they'll see it as wasted time. Thoughts?

thejameskyle:
I pushed the devs at my company to stop using it and they're on board.

samccone:
yep.. just will want some path with mergeOptions so people can get the old behavior even tho it is "bad"

(i.e) merge all the options

jmeas:
> yep.. just will want some path with mergeOptions so people can get the old behavior even tho it is "bad"
> 
> (i.e) merge all the options

`this.options = options` would be the upgrade path, assuming we remove `getOptions`. But we're keeping `getOptions` around. The idea of `mergeOptions` is to move away from these anti-patterns, so I don't think that we should add anything to the API that enables it.

paulfalgout:
Relates to https://github.com/marionettejs/backbone.marionette/pull/2286

samccone:
rebase sorry

samccone:
cherry-picked onto minor

jasonLaster:
Hey, all. The merge was reverted. While there was agreement in using mergeOptions, there was not agreement on deprecating getOptions. Here's a summary of the rationale:

----

At the moment, we're moving ahead with adding mergeOptions and using it internally. Deprecating getOption is really a PR question:

do we want to take away the functionality
strongly discourage it
encourage mergeOptions

PR #2277, opened 2015-02-04
Clairfy 'onAttach' documentation

### LINKED ISSUE #2208: onAttach documentation clarity
The phrasing of the following statement, [found here](http://marionettejs.com/docs/marionette.view.html#view-attach--onattach-event), could use some work:

> Because the attach event is only fired when the view is a child of the document, it is a requirement that the Region you're showing it in be a child of the document at the time that you call show.

The statement seems to imply that if a user calls "region.show" before the region is a child of the document then Marionette will throw an error. In reality, the onAttach event will not fire for the child view being shown until the region becomes a child of the DOM.

MeoMix:
This change is intended to resolve #2208. The original language used was misleading to me.

### DIFF
--- a/docs/marionette.view.md
+++ b/docs/marionette.view.md
@@ -137,8 +137,7 @@ the view in a Region causes it to be attached to the `document`. Like other Mari
 executes a callback method, `onAttach`, if you've specified one. The `"attach"` event is great for jQuery
 plugins or other logic that must be executed *after* the view is attached to the `document`.
 
-Because the `attach` event is only fired when the view is a child of the `document`, it is a requirement
-that the Region you're showing it in be a child of the `document` at the time that you call `show`.
+The `attach` event is only fired when the view becomes a child of the `document`. If the Region you're showing the view in is not a child of the `document` at the time that you call `show` then the `attach` event will not fire until the Region becomes a child of the `document`.
 
 This event is unique in that it propagates down the view tree. For instance, when a CollectionView's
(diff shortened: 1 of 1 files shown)


jmeas:
:+1:

Thanks @meomix!

docs/marionette.view.md
```diff
@@ -137,8 +137,7 @@ the view in a Region causes it to be attached to the `document`. Like other Mari
 executes a callback method, `onAttach`, if you've specified one. The `"attach"` event is great for jQuery
 plugins or other logic that must be executed *after* the view is attached to the `document`.
 
-Because the `attach` event is only fired when the view is a child of the `document`, it is a requirement
-that the Region you're showing it in be a child of the `document` at the time that you call `show`.
+The `attach` event is only fired when the view becomes a child of the `document`. If the Region you're showing the view in is not a child of the `document` at the time that you call `show` then the `attach` event will not fire until the Region becomes a child of the `document`.
```
thejameskyle: I think it'd be clearer like this:
```diff
- view becomes a child of the `document`
+ view is inserted into the `document`
```

MeoMix: Whatever works! I was just trying to maintain the original language of the docs.

thejameskyle: Thanks!

samccone:
cherry-picked onto minor

thanks @MeoMix !

Issue #1887, opened 2014-09-14
Consider deprecating `onDomRefresh`


jamesplease:
`onAttach` will serve the needs of most people, but is it enough to fully replace `onDomRefresh`? This issue is for us to think about that more.

An interesting thing to note is that if we get rid of `onDomRefresh`, we might be able to get rid of the internal `_rendered` and `_shown` properties.

PR #2281, opened 2015-02-05
Refactor onDomRefresh

jmeas:
WIP

Todo:

- [ ] Add tests back in
- [ ] Refactor onDomRefresh

ref https://github.com/marionettejs/backbone.marionette/issues/1887


thejameskyle:
:tada:

jasonLaster:
Nice. Let's make sure this is deprecated on minor

src/abstract-view.js
```diff
@@ -25,7 +27,10 @@ Marionette.AbstractView = Backbone.View.extend({
 
     Backbone.View.apply(this, arguments);
 
-    Marionette.MonitorDOMRefresh(this);
+    this.on({
+      render: this._setRendered,
+      show: this._setShown
+    }, this);
```
jmeas: I know, I know. This is odd. The thing is, `onDomRefresh` also set internal state of Views that the CollectionView relies on. We need to continue tracking that internal state for the views to continue functioning in the same way.

jmeas:
Updated!

jmeas:
Updated @paulfalgout! Good catch.

docs/marionette.abstractview.md
```diff
@@ -17,7 +17,6 @@ AbstractView directly.
 * [AbstractView onBeforeDestroy](#abstractview-onbeforedestroy)
 * [AbstractView "attach" / onAttach event](#abstractview-attach--onattach-event)
 * [AbstractView "before:attach" / onBeforeAttach event](#abstractview-beforeattach--onbeforeattach-event)
-* [AbstractView "dom:refresh" / onDomRefresh event](#abstractview-domrefresh--ondomrefresh-event)
```
samccone: not in the .yaml?

src/abstract-view.js
```diff
@@ -9,6 +9,8 @@
 
 Marionette.AbstractView = Backbone.View.extend({
   isDestroyed: false,
+  _isShown: false,
```
samccone: who uses these...

jmeas: CollectionView. [Comment here](https://github.com/marionettejs/backbone.marionette/pull/2281#discussion-diff-24139170)

samccone: ah cra

paulfalgout: `CompositeView` uses _isRendered  
There's really not much reason these couldn't move into those two.
`CollectionView` has that '_onShowCalled` method and `CompositeView` has a `render` function.  I don't think there'd be need for the listeners.

jmeas: :+1: @paulfalgout.

samccone: yep yep @paulfalgout want to open a new issue for this?

paulfalgout: https://github.com/marionettejs/backbone.marionette/issues/2287

jmeas:
Rebased – let's get this in!

lcwork:
I need to use `onDomRefresh` in a few places in my app. `onAttach` does not work in these cases. Will there be an alternate way to handle a refresh of an existing view by 3.0? (Or one that exists already?) 

I did see a comment about `onRefresh`, but only saw it mentioned in one place. Thanks.

jmeas:
What's the use case?

samccone:
wI could invent one... every render of a view you need to calc the bounding rect of the view to do some jquery plugin goop. (epic editor for instance does this) 
on attach would only work once and not on each rerender, but perhaps you could just use onRender after onAttach fired once... just some thoughts :)

lcwork:
Saving something to the server and the server is massaging the data a bit, and we want that to be shown the user. After I get the data back I can show the new data by rendering, but I need to call a plugin (X-Editable) to make the data "editable" again. I tried a bunch of different things and onDomRefresh was the the only thing that worked.

paulfalgout:
@lcwork We played with a similar situation and was able to reproduce that functionality by doing what @samccone mentioned.   Essentially:

```
onAttach: function(){
  this.doSomething();
  this.on('render', this.doSomething);
}
```

lcwork:
@paulfalgout okay thanks. I will try that.

jmeas:
I could see us adding that back into the core library, if people were interested. Maybe the name `onDomRefresh` could stick around even. But it can be added in fewer LoC, and be less complicated.

Also, it should work with the view being detached from a region, then reattached – i have a feeling the current implementation does not.

samccone:
:+1: yep @paulfalgout that is what i was thinking

jmeas:
@lcwork see #2299. Thanks for the comments :)

docs/marionette.abstractview.md
jmeas: Nope. It's just a stub with not much content.

jmeas:
Ready for review

samccone:
@jmeas I think your commit message description is messed up ha

other than that :shipit:

jmeas:
lol vimfail. Updated.

Should I add a new `onDomRefresh` and leave the tests in this same PR?

samccone:
I would add it in a new PR since it is a "new" feature, what do you think?

jmeas:
Fine by me!

samccone:
cool, one more set of eyes and we can get this in!

jasonLaster:
I think it'd be better to do the refactor in one PR. That way, the refactor is *atomic*

samccone:
kk

samccone:
ping ping @jmeas ok looks like we still need the following

* [ ] reintroduce refactored onDomRefresh
* [ ] put tests back in place around this behavior
* [ ] add a onDomRefresh test that would have failed before and now wont :speak_no_evil: 
* [ ] add back in documentation


Let me know if you want a hand doing this @jmeas thanks again for this amazing :construction_worker:

jasonLaster:
@jmeas mind if I close this and we can return to it when there's interest in picking it up again?

We'll still have the open issue

jasonLaster:
Closing as it has gone stale, but it's referenced in open issues so can be picked up later

PR #2284, opened 2015-02-05
Deprecate the Controller Class.

jmeas:


docs/marionette.controller.md
```diff
@@ -2,6 +2,9 @@
 
 # Marionette.Controller
 
+> Warning: deprecated. The Controller object is deprecated. Instead of using the Controller
+> class with the AppRouter, you should specify your callbacks on a plain Javascript object.
```
thejameskyle: I guess this line has a space at the end of it and somehow that breaks the build

PR #2285, opened 2015-02-05
Fix issue with bundled root.Mn alias.

jmeas:
Our bundled version of Marionette doesn't replace `root.Mn`. Now it does.


jmeas:
This should actually go out as a bug release come to think of it

thejameskyle:
:+1:

samccone:
good catch

2.4 is coming up super soon so this can go out with that

jmeas:
k @samccone.

PR #2286, opened 2015-02-05
Add mergeOptions helper.

jmeas:
Will add docs once I get approval of this implementation.

Usage:

```js
  initialize: function(options) {
    this.mergeOptions(options, ['color', 'size', 'location']);
  },

  someMethod: function() {
    if (this.color) {}
  }
});
```

This would replace `getOptions` as the preferred way to access options passed into your views.

The way I use it is:

```js
  myViewOptions: ['color', 'size', 'location'],

  initialize: function(options) {
    this.mergeOptions(options, this.myViewOptions);
  },

  someMethod: function() {
    if (this.color) {}
  }
});
```

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -17,6 +17,11 @@ Marionette.isNodeAttached = function(el) {
   return Backbone.$.contains(document.documentElement, el);
 };
 
+// Merge `keys` from `options` onto `this`
+Marionette.mergeOptions = function(options, keys) {
+  if (!options) { return; }
+  _.extend(this, _.pick(options, keys));
+};
 
 // Marionette.getOption
 // --------------------
--- a/api/helpers.yaml
+++ b/api/helpers.yaml
@@ -78,6 +78,33 @@ functions:
           // => true
           ```
 
+  mergeOptions:
+    description: |
+      Merge `keys` from `obj` onto `this`.
+      
+      @api public
+      @static
+      @param {Object} obj
+      @param {Array} keys
+
+    examples:
+      -
+        name: mergeOptions
+        example: |
+          By setting the keys on your prototype, you can create self-documenting classes. Take
+          the example below.
+          
+          ```js
+          var MyView = Marionette.ItemView.extend({
+            myViewOptions: ['color', 'size', 'country'],
+          
+            initialize: function(options) {
+              this.mergeOptions(options, this.myViewOptions),
+              console.log('These are my options:', this.color, this.size, this.country);
+            }
+          });
+          ```
+
   getOption:
     description: |
       Retrieve an object, function or other value from a target object or its `options`, with
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -41,6 +41,7 @@ instead of these deprecated features.
   * [Overriding the default RegionManager](#overriding-the-default-regionmanager)
   * [Get Region By Name](#get-region-by-name)
   * [Removing Regions](#removing-regions)
+* [Application.mergeOptions](#applicationmergeoptions)
 * [Application.getOption](#applicationgetoption)
 * [Adding Initializers (deprecated)](#adding-initializers)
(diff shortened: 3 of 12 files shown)


jasonLaster:
I like it.

src/helpers.js
```diff
@@ -17,6 +17,11 @@ Marionette.isNodeAttached = function(el) {
   return Backbone.$.contains(document.documentElement, el);
 };
 
+// Merge `keys` from `options` onto `this`
+Marionette.mergeOptions = function(options, keys) {
+  if (!options) { return; }
+  _.extend(this, _.pick(options, keys));
```
samccone: my question is what is this really adding? other than a way for people to run _.pick and _.extend

jmeas: It's a more expressive API. Writing `mergeOptions` vs these two lines in every constructor makes a big difference, I think.

samccone: yep fair, do you think we should have a mergeOptions key on all Marionette+ classes?

jmeas: I did for some time. @thejameskyle convinced me otherwise because of the fact that it gets lost in the prototype chain.

thejameskyle: Yeah, this is already a headache for deeply extended view classes for stuff like `events` and `ui`. They get clobbered really easily.

cmaher: We could also override extend and merge events + ui, etc, as well as mergeOptions (though that would be breaking so it would have to come in v3)

jasonLaster: I can see a `abstractViewProperties: ['_isShown']` or whatever for properties that `Marionette.AbstractView` uses. Subsequent child classes wouldn't override it, yet it would provide some degree of documentation and introspection for the curious of heart (think inspector).

jasonLaster:
:+1:

src/helpers.js
jmeas: ^ precisely.

paulfalgout:
:+1:

src/helpers.js
thejameskyle: > We could also override extend

No no no, that would completely break compatibility with ES6

samccone:
This is still missing yaml documentation

jmeas:
will do tonight

jasonLaster:
Tests and docs plz

jmeas:
ILL GET 2 IT

jmeas:
Updated

paulfalgout:
:+1: All looks clear to me.

jmeas:
Btw I'm still not sold on this method myself. I just think that it's better than `getOptions`. I could see it being made more general, like `mergeKeys`, so that it could be used in other contexts, too. @megawac / @jridgewell what do you two think?

megawac:
I'm not convinced its much better than `getOption` and it has the potential to be slightly breaking

Though it does make it easier to reason about things like this: https://github.com/marionettejs/backbone.marionette/blob/minor/src/layout-view.js#L14

jmeas:
> I'm not convinced its much better than getOption and it has the potential to be slightly breaking

I dislike the fact that we merge options directly onto the instance. Just taken alone I dislike that pattern, and it's doubly bad given that it's now inconsistent w. Backbone. Do you feel differently?

paulfalgout:
I think there is a need for some sort of utility.  `mergeOptions` feels more transparent than `getOption` which is a tiny bit magical unless you inspect what it's doing.  `getOption` is anything passed when instantiating, anything already attached to the prototype or possibly anything in an `options` hash on the prototype.

Many of my `initialize`s already begin with a `_.extend ... _.pick` A part of me thinks that neither are really necessary, but having the utility does help take some of the concern off of new users for how they should store `options`.  But it's not too much of a baked in feature, you can certainly ignore it if it's not a good fit for your project.

samccone:
:+1: from me

PR #2288, opened 2015-02-05
Rename ItemView to be View.

### LINKED ISSUE #2254: Rename ItemView as View
Pulled out of #2251 as it's sorta a separate thing.

jmeas:
Resolves #2254 

I know that the LayoutView docs weren't updated. I still need to merge those into the View docs. That will be a separate PR. This updates everything else tho'


thejameskyle:
:+1:

PR #2295, opened 2015-02-05
Revert "Deprecate getOption"

samccone:
Going to revert the merge on #2272

It is one thing to encourage people to use mergeOptions and another to deprecate getOption
(needs more discussion)


jasonLaster:
Lets continue the discussion on #2272 about `getOption`.

At the moment, we're moving ahead with adding `mergeOptions` and using it internally. Deprecating `getOption` is really a PR question:
+ do we want to take away the functionality
+ strongly discourage it
+ encourage mergeOptions

Issue #1979, opened 2014-10-08
CollectionView: Reorder DOM elements after sorting


ptbrowne:
Hi,

I have just set up a proof of concept to increase the performance of re-rendering large CollectionView after sorting. The default behavior is to re-render but this seems very inefficient since the templates are already rendered and inserted in the DOM. My approach is simple but it seems to deliver good results: I basically detach all children' $el and reinsert them in the order of the collection.

I have had more than 10x perf increase and it should be more if the template is complicated. If you click on profile in the jsBin, you will be able to see that in Chrome Profiler.

I wonder what do you think, if this technique has any flaws (beside the fact that if the template or itemViewOptions use the index, it will not work as expected). I am open to critics.

Here is the jsBin: http://jsbin.com/vecobada/13/edit

Thanks for Marionette !

samccone:
@ptbrowne any word on the PR?

ptbrowne:
I am pretty busy at the moment but I will try to do that this week. Sorry
for not sending updates...

ahumphreys87:
fixed in 2.4

PR #2301, opened 2015-02-07
perf: reorderOnSort option for collectionView

ptbrowne:
## Why

When the collectionView need to resort, the default implementation is to re-render. This can have performance problems on large CollectionViews where ChildViews are a bit complex.

## How

ChildViews have already been rendered once so instead of re-rendering them, reuse them and only sort DOM nodes.

## Possible problems

ChildViews that knows their index and use it in their rendering will know be able to use this technique since they need to be re-rendered after sorting. This is why this performance improvement is not enabled by default.

See https://github.com/marionettejs/backbone.marionette/issues/1979

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -128,12 +128,36 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
     return this;
   },
 
+  // Reorder DOM after sorting. When your element's rendering
+  // do not use their index, you can pass reorderOnSort: true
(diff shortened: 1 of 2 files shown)


samccone:
hi @ptbrowne thanks for opening this PR!

It looks like since this is behind a feature flag we can get this sucker into minor. Would you mind reopening this PR into the minor branch?

Thanks

ptbrowne:
Done in #2302

samccone:
:+1:

PR #2302, opened 2015-02-07
perf: reorderOnSort option for collectionView

ptbrowne:
## Why

When the collectionView need to resort, the default implementation is to re-render. This can have performance problems on large CollectionViews where ChildViews are a bit complex.

## How

ChildViews have already been rendered once so instead of re-rendering them, reuse them and only sort DOM nodes.

## Possible problems

ChildViews that knows their index and use it in their rendering will know be able to use this technique since they need to be re-rendered after sorting. This is why this performance improvement is not enabled by default.

See https://github.com/marionettejs/backbone.marionette/issues/1979

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -128,12 +128,35 @@ Marionette.CollectionView = Marionette.View.extend({
     return this;
   },
 
+  // Reorder DOM after sorting. When your element's rendering
+  // do not use their index, you can pass reorderOnSort: true
+  // to only reorder the DOM after a sort instead of rendering
+  // all the collectionView
+  reorder: function () {
+    var children = this.children;
+
+    // get the DOM nodes in the same order as the models
+    var els = _.map(this._sortedModels(), function (model) {
+      return children.findByModel(model).el;
+    });
+
+    // since append moves elements that are already in the DOM,
+    // appending the elements will effectively reorder them
+    this.triggerMethod('before:reorder');
+    this.$el.append(els);
+    this.triggerMethod('reorder');
+  },
+
   // Render view after sorting. Override this method to
   // change how the view renders after a `sort` on the collection.
   // An example of this would be to only `renderChildren` in a `CompositeView`
   // rather than the full view.
   resortView: function() {
-    this.render();
+    if (Marionette.getOption(this, 'reorderOnSort')) {
+      this.reorder();
+    } else {
+      this.render();
+    }
   },
 
   // Internal method. This checks for any changes in the order of the collection.
--- a/test/unit/sorted-views.spec.js
+++ b/test/unit/sorted-views.spec.js
@@ -391,4 +391,86 @@ describe('collection/composite view sorting', function() {
       });
     });
   });
+
+  describe('when using `{ reorderOnSort: true }`', function () {
+    var texts = {
+      asOption: 'as an option',
+      onPrototype: 'on the prototype',
+      viewComparator: 'with viewComparator'
+    };
+
+    var getSpecTitle = function(options) {
+      return _.map(options, function (v, k) {
+        return texts[k];
+      }).join(' ');
+    };
+
+    var describeSpec = function (specOptions) {
+      describe(getSpecTitle(specOptions), function () {
+        beforeEach(function () {
+          var commonAttrs = {
+            childView: this.ChildView,
+            collection: this.collection
+          };
+
+          if (specOptions.viewComparator) {
+            commonAttrs.viewComparator = 'foo';
+          }
+
+          if (specOptions.asOption) {
+            this.collectionView = new this.CollectionView(_.extend({}, {
+              reorderOnSort: true
+            }, commonAttrs));
+          } else if (specOptions.onPrototype) {
+            var ReorderedCollectionView = this.CollectionView.extend({
+              reorderOnSort: true,
+              onReorder: this.sinon.spy(),
+              onBeforeReorder: this.sinon.spy()
+            });
+            this.collectionView = new ReorderedCollectionView(commonAttrs);
+          }
+
+          this.collectionView.render();
+          this.sinon.spy(this.collectionView, 'reorder');
+          this.sinon.spy(this.collectionView, 'render');
+          this.sinon.spy(this.collectionView, 'trigger');
+
+          var cmp = function (m) {
+            return m.get('bar');
+          };
+          if (specOptions.viewComparator) {
+            this.collection.comparator = 'foo';
+            this.collectionView.options.viewComparator = cmp;
+          } else {
+            this.collection.comparator = cmp;
+          }
+          this.collection.sort();
+        });
+
+        it('should call reorder instead of render', function () {
+          expect(this.collectionView.render).not.to.have.been.called;
+          expect(this.collectionView.reorder).to.have.been.calledOnce;
+        });
+
+        it('should reorder the DOM', function () {
+          expect(this.collectionView.$el).to.have.$text('321');
+        });
+
+        it('should triggerMethods events', function () {
+          var cv = this.collectionView;
+          if (specOptions.onPrototype) {
+            expect(cv.onBeforeReorder).calledBefore(cv.onReorder);
+          }
+          expect(cv.trigger).to.have.been.calledWith('before:reorder');
+          expect(cv.trigger).to.have.been.calledWith('reorder');
+          expect(cv.trigger).to.have.been.calledTwice;
+        });
+      });
+    };
+
+    describeSpec({ asOption: true });
+    describeSpec({ asOption: true, viewComparator: true });
+    describeSpec({ onPrototype: true });
+    describeSpec({ onPrototype: true, viewComparator: true });
+  });
 });
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -25,12 +25,15 @@ will provide features such as `onShow` callbacks, etc. Please see
   * [CollectionView's `childEvents`](#collectionviews-childevents)
   * [CollectionView's `buildChildView`](#collectionviews-buildchildview)
   * [CollectionView's `addChild`](#collectionviews-addchild)
+  * [CollectionView's `reorderOnSort`](#collectionviews-reorderonsort)
 * [CollectionView's `emptyView`](#collectionviews-emptyview)
   * [CollectionView's `getEmptyView`](#collectionviews-getemptyview)
   * [CollectionView's `emptyViewOptions`](#collectionviews-emptyviewoptions)
 * [Callback Methods](#callback-methods)
   * [onBeforeRender callback](#onbeforerender-callback)
   * [onRender callback](#onrender-callback)
+  * [onBeforeReorder callback](#onbeforereorder-callback)
+  * [onReorder callback](#onreorder-callback)
   * [onBeforeDestroy callback](#beforedestroy-callback)
   * [onDestroy callback](#ondestroy-callback)
   * [onBeforeAddChild callback](#onbeforeaddchild-callback)
@@ -40,6 +43,7 @@ will provide features such as `onShow` callbacks, etc. Please see
 * [CollectionView Events](#collectionview-events)
   * ["before:render" event](#beforerender-event)
   * ["render" event](#render-event)
+  * ["before:reorder" / "reorder" event](#beforereorder--reorder-event)
   * ["before:destroy" event](#beforedestroy-event)
   * ["destroy" / "destroy:collection" event](#destroy--destroycollection-event)
   * ["before:add:child" / "add:child" event](#beforeaddchild--addchild-event)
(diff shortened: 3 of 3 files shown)


ptbrowne:
I am not quite happy with the name of the feature flag, tell me if you have any suggestion.

megawac:
I'm :-1: for this without some decent benchmarks as you are still detaching all the elements in the tree and then readding them in proper order. If you were diff swapping nodes I could see how that would make a difference in some cases, but this method just avoids some of the render overhead.

ptbrowne:
I am all for doing benchmark if you like it but I could not find the jsLitmus files that were there before ? How can I include a benchmark ?

You can see http://jsbin.com/vecobada/13/edit?html,output for an example. Try with the option checked and without check. On my MBPr, the difference is cleary noticeable.

For what it’s worth, here is Chrome profile output of my example where I click on sort.

Here without the reorder option:
screen shot 2015-02-07 at 16 53 15

Here with the reorder option (same scale):
screen shot 2015-02-07 at 16 53 29

“just avoids some of the render overhead” -> If you have moderately complex elements, this render overhead can make all the difference between “laggy” and “instant”.

samccone:
yep @ptbrowne I am all for this, I think regardless of the final "optimized" implementation this is an amazing first step.

As far as perf, I think we can do something on jsperf. It will just require someone to run grunt build on this branch and use that for one test and the normal dist for the other.

:clap: for this great work so far

src/collection-view.js
```diff
@@ -128,12 +128,36 @@ Marionette.CollectionView = Marionette.View.extend({
...
+    // get the views in the same order as the models
+    var arr = this.collection.models.map(function (model) {
+      var viewid = idx[model.cid];
+      var view = views[viewid];
+      return view.$el;
+    });
+    // transform into a jQuery array
+    var allEls = $(arr).map(function () { return this.toArray(); });
+    allEls.detach();
+    this.$el.append(allEls);
```
megawac: Correct me if there is a gotcha here, but I'm pretty sure this can just be

```diff
-    // transform into a jQuery array
-    var allEls = $(arr).map(function () { return this.toArray(); });
-    allEls.detach();
-    this.$el.append(allEls);
+    this.$el.append(arr);
```

ptbrowne:
Another thing: If you have images in your ChildViews, they will be refetched if you re-render (normally the cache will take care of that but still, it is some work for the browser). When reordering nodes, they are not re-fetched.

http://jsbin.com/soxume/1/edit?html,output (check the network tab of Chrome Developer Tools)

src/collection-view.js
ptbrowne: Seems to work, I do not remember why I wrote it in this convoluted way :+1:

src/collection-view.js
```diff
@@ -128,12 +128,34 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  // to only reorder the DOM after a sort instead of rendering
+  // all the collectionView
+  reorder: function () {
+    var children = this.children;
+    var idx = children._indexByModel;
+    var views = children._views;
+    // get the views in the same order as the models
+    var arr = this.collection.models.map(function (model) {
+      var viewid = idx[model.cid];
+      var view = views[viewid];
```
cmaher: `children.findByModel` would simplify this code

test/unit/sorted-views.spec.js
```diff
@@ -391,4 +391,30 @@ describe('collection/composite view sorting', function() {
       });
     });
   });
+
+  describe('when using `{ reorderOnSort: true }`', function () {
+    beforeEach(function () {
+      this.collectionView = new this.CollectionView({
+        childView: this.ChildView,
+        collection: this.collection,
+        reorderOnSort: true
```
cmaher: testing that `reorder` is called when `reorderOnSort` is set to true on the prototype would be good too.

cmaher:
This is looking pretty good. Some docs would also be nice.

src/collection-view.js
ptbrowne: you're right :)

test/unit/sorted-views.spec.js
ptbrowne: `+      expect(this.collectionView.reorder).to.have.been.calledOnce;`

is already there. Did you mean something else ?

megawac: I think @cmaher means setting `MArionette.CollectionView.prototype.reorderOnSort = true;`

samccone: yeah so 

you would make a new collectionView extension

```js
MyView = Mn.CollectionView.extend({
  reorderOnSort: true
});
```
```js
new MyView({collection: this.collection})
```

ptbrowne: understood :) thanks

src/collection-view.js
```diff
@@ -128,12 +128,33 @@ Marionette.CollectionView = Marionette.View.extend({
...
 
+  // Reorder DOM after sorting. When your element's rendering
+  // do not use their index, you can pass reorderOnSort: true
+  // to only reorder the DOM after a sort instead of rendering
+  // all the collectionView
+  reorder: function () {
+    var children = this.children;
+
+    // get the DOM nodes in the same order as the models
+    var els = this.collection.models.map(function (model) {
```
samccone: If we really want to be fancy

```js
_(this.collection.models)
.map(children.findByModel)
.pick(el)
.value()
```

src/collection-view.js
```diff
@@ -128,12 +128,33 @@ Marionette.CollectionView = Marionette.View.extend({
...
+      return children.findByModel(model).el;
+    });
+
+    // since append moves elements that are already in the DOM,
```
samccone: this is a cool behavior that i did not know about!

test/unit/sorted-views.spec.js
```diff
@@ -391,4 +391,53 @@ describe('collection/composite view sorting', function() {
...
     });
   });
+
+  describe('when using `{ reorderOnSort: true }`', function () {
+    var instances = {
+      asOption: 'as an option',
+      onPrototype: 'on the prototype'
+    };
+
+    var describeSpec = function (instance) {
```
samccone: #meta spec :pager:

samccone:
@ptbrowne thanks for the work! sorry we are getting into the tedious part of a PR of this nature. thanks for sticking it out!

ptbrowne:
No problem, it's normal. Happy to contribute. What do you mean by :pager: ?

samccone:
oh nothing, just a meta spec :) nothing wrong with it

samccone:
@ptbrowne on your next pass would you mind squashing this down to a single commit?

ptbrowne:
OK, no problem. How do I do that ? I squash on my machine and then `push -f` ?

samccone:
you got it!

src/collection-view.js
ptbrowne: I guess it would have to be `.map(this.children.findByModel, this.children)` and `.pick('el')` ;-) Anyway, I think until `underscore` becomes lazy, the chaining is uglified by the `.value()` at the end (that I always forgot). Also, I try to avoid `_`'s chaining because it is too easy to shoot yourself in the foot with the `this`.

samccone: hmm not sure if it would be slower or not.  above is a more functional approach ^ that required less internal state. 
but not a real big deal one way or the other

samccone:
:+1: on this

src/collection-view.js
megawac: It would be slower but I wouldn't expect it to be an important issue

paulfalgout:
So thoughts on adding events?  A normal resort would trigger `render` events.
`this.triggerMethod('before:reorder', this);`
`this.triggerMethod('reorder', this);`

Also this wouldn't work until after a `render` occurs.. it's probably ok because the only thing directly calling this function is a `sort` event that is setup after the first `render`.  Not sure if we need to handle or indicate that somewhere?

Also in the comments for `CollectionView`  sort and comparator are mentioned up at the constructor.  Do we want to keep those option comments together?

samccone:
ah good call on the events @paulfalgout , I think since it is a new "path" it would be wise to do this..

ptbrowne:
I agree with @paulfalgout and I added the events. Updated the docs too.

About the need for docs about the need to specify that a first render call is necessary, I think that it is both unnecessary since as you said the first `sort` event is sent after the first `render`. And also because in the docs, it is already stated that this option is useful when `resorting` the collection, and that it uses pre-existing DOM nodes. At least I can't find a formulation that would be clear and not confusing apart from what is already stated.

Thanks for the comments :)

ptbrowne:
Mmh I have just seen this https://github.com/marionettejs/backbone.marionette/commit/7fcdc351099b5b54edf32a2c4aa3097dc3c95b7f, it seems that `reorder` should support `viewComparator` right ?

samccone:
yep thats correct @ptbrowne , should be as simple as calling into `_sortedModels`

samccone:
@marionettejs/marionette-core one more set of eyes?

this is :+1: for me

test/unit/sorted-views.spec.js
```diff
@@ -391,4 +391,85 @@ describe('collection/composite view sorting', function() {
...
   });
+
+  describe('when using `{ reorderOnSort: true }`', function () {
+    var texts = {
+      asOption: 'as an option',
+      onPrototype: 'on the prototype',
+      viewComparator: 'with viewComparator'
+    };
+
+    var getSpecTitle = function(options) {
```
cmaher: is there a reason you're passing an object instead of an array?

src/collection-view.js
```diff
@@ -128,12 +128,39 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  // to only reorder the DOM after a sort instead of rendering
+  // all the collectionView
+  reorder: function () {
+    var children = this.children;
+
+    // get the DOM nodes in the same order as the models
+    var viewComparator = Marionette.getOption(this, 'viewComparator');
+    var models = this.collection.models;
+    var mapper = (viewComparator ?
+      _.partial(_.map, this._sortedModels()) : _.bind(models.map, models));
```
cmaher: `models.map` will break IE8

samccone: dope great catch totally missed that!

test/unit/sorted-views.spec.js
```diff
@@ -391,4 +391,85 @@ describe('collection/composite view sorting', function() {
...
+
+          if (specOptions.asOption) {
+            this.collectionView = new this.CollectionView(_.extend({}, {
+              reorderOnSort: true
+            }, commonAttrs));
+          } else if (specOptions.onPrototype) {
+            var ReorderedCollectionView = this.CollectionView.extend({
+              reorderOnSort: true,
+              onReorder: sinon.spy(),
+              onBeforeReorder: sinon.spy()
```
cmaher: `this.sinon`

test/unit/sorted-views.spec.js
```diff
@@ -391,4 +391,85 @@ describe('collection/composite view sorting', function() {
...
+          expect(this.collectionView.$el).to.have.$text('321');
+        });
+
+        it('should triggerMethods events', function () {
+          var cv = this.collectionView;
+          if (specOptions.onPrototype) {
+            expect(cv.onBeforeReorder).calledBefore(cv.onReorder);
+          }
+          expect(cv.trigger).to.have.been.calledWith('before:reorder');
+          expect(cv.trigger).to.have.been.calledWith('reorder');
```
cmaher: `.and.calledOnce`

ptbrowne: `.and.calledOnce` does not follow the “filter” on args done by `calledWith` so I put `         expect(cv.trigger).to.have.been.calledTwice;`

src/collection-view.js
ptbrowne: perfect, it simplifies the code :)

test/unit/sorted-views.spec.js
ptbrowne: `getSpecTitle` gets an object because the `specOptions` is an object meant to imitate a `set`. I could do `getSpecTitle(_.keys(specOptions))` and have:
```
getSpecTitle: function (options) {
   return _.map(options, function (k) { return texts[k]; })
}
```

but I do not think its better. Anyway, at this point, it really is a matter of choice.

test/unit/sorted-views.spec.js
ptbrowne: ok. Is there a reason why you do not use the global `sinon` object ? Did you have some problems with it being polluted during the tests ?

ptbrowne:
Travis failed `git clone git://github.com/jasonLaster/dox.git github.com[0: 192.30.252.129]: errno=Connection timed out`. Can someone restart the Travis job ?

jmeas:
Restarted

test/unit/sorted-views.spec.js
```diff
@@ -391,4 +391,86 @@ describe('collection/composite view sorting', function() {
...
+            this.collection.comparator = 'foo';
+            this.collectionView.options.viewComparator = cmp;
+          } else {
+            this.collection.comparator = cmp;
+          }
+          this.collection.sort();
+        });
+
+        it('should call reorder instead of render', function () {
+          expect(this.collectionView.render).not.to.have.been.calledOnce;
```
cmaher: `not.to.have.been.called` is probably safer that `calledOnce` (in case it ever gets called twice)

cmaher:
One last tiny nit (sorry), and then I am :+1:  to this. It's looking great!

test/unit/sorted-views.spec.js
ptbrowne: good catch :)

cmaher:
:+1: :shipit:

jmeas:
Nice work @ptbrowne!

PR #2310, opened 2015-02-10
Make sure before:render is triggered before modifying layout view.

scottaj:
To me, the implications of the `before:render` event is that it runs before rendering any changes in the DOM. This is true for ItemView, but LayoutView was resetting its regions before triggering the event. This means that the DOM has already been modified because the subviews have been removed.

I ran into this while trying to write some code that would find the currently focused element on the page (which could be in a view or its subviews) and record that element so it could be re-focused after re-rendering. The code didn't work because the subviews were already gone, and I had to override `render` instead.

### DIFF
--- a/src/item-view.js
+++ b/src/item-view.js
@@ -53,11 +53,7 @@ Marionette.ItemView = Marionette.View.extend({
 
     this.triggerMethod('before:render', this);
 
-    this._renderTemplate();
-    this.isRendered = true;
-    this.bindUIElements();
-
-    this.triggerMethod('render', this);
+    this._renderItem();
 
     return this;
   },
@@ -89,6 +85,14 @@ Marionette.ItemView = Marionette.View.extend({
     this.attachElContent(html);
 
     return this;
+  },
+
+  _renderItem: function() {
+    this._renderTemplate();
+    this.isRendered = true;
+    this.bindUIElements();
+
+    this.triggerMethod('render', this);
   },
 
   // Attaches the content of a given view.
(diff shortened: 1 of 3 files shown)


jasonLaster:
Thanks @scottaj. I think you found something really good. Also thanks for the clear write up and PR. 

This will be fun because on major we've merged LayoutView into ItemView. @jme how do you want to proceed here?

jmeas:
This is a good idea, but breaking, so it should go in major methinks.

Luckily, because LayoutView and ItemView are the same thing in major, this should be a much less destructive change. @scottaj would you want to checkout major branch and add this change there? You'll want to look in `Marionette.View` for the code.

jasonLaster:
@jmeas @samccone do you think we can get two different commits, one minor and one major? His way we can get a fix out to people on 2.foo?

jmeas:
We *could* classify this as a bug, but I'm sure many applications depend on this behavior. It would be irresponsible for us to just cut this as a bug fix.

If we release it on minor, I'd suggest that it should go behind a feature flag. What do you think @jasonLaster?

scottaj:
@jasonLaster @jmeas I'll open another PR against the master branch, and then you guys can decide if you want to decline this, put it behind a feature flag, or something else. LMK when/if I need to make changes to this one.

jmeas:
> I'll open another PR against the master branch

Major, not master.

And thanks!

scottaj:
That's what I meant :-X

samccone:
> We could classify this as a bug, but I'm sure many applications depend on this behavior. It would be irresponsible for us to just cut this as a bug fix.

Yeah I am pretty sure although this is a `bug` it would break peoples applications. For that reason I vote to only put it on major and simply note it in the changelog.

great catch @scottaj

samccone:
merged into major, we are not going to put this into minor just because we are fairly sure this is going to cause issues for people with their applications.

Again, at some point bugs that have been coded against become breaking changes. :8ball:

PR #2315, opened 2015-02-11
Add Marionette feature flags

jasonLaster:
Feature flags are added onto the Marionette.FEATURES
object:

```js
Marionette.FEATURES = {
  fooFlag: false
};
```

and are checked in the library code via a simple `isEnabled` check:

```js
Marionette.isEnabled('fooFlag')
```

The goal of feature flags is to be able to add *breaking* features in
a minor release behind a feature flag. There are two benefits to this
strategy:
1) users can access the new feature immediately
2) the code can be merged onto master much sooner, which prevents higher
risk merges down the road.

### DIFF
--- a/src/features.js
+++ b/src/features.js
@@ -1,0 +1,6 @@
+Marionette.FEATURES = {
+};
+
+Marionette.isEnabled = function(name) {
+  return !!Marionette.FEATURES[name];
+};
--- a/test/unit/features.spec.js
+++ b/test/unit/features.spec.js
@@ -1,0 +1,20 @@
+describe('features', function() {
+
+  beforeEach(function() {
+    Marionette.FEATURES = {};
+  });
+
+  it('enabled when its present and true', function() {
+    Marionette.FEATURES.foo = true;
+    expect(Marionette.isEnabled('foo')).to.be.true;
+  });
+
+  it('disabled when its present and false', function() {
+    Marionette.FEATURES.foo = false;
+    expect(Marionette.isEnabled('foo')).to.be.false;
+  });
+
+  it('disabled when not present', function() {
+    expect(Marionette.isEnabled('foo')).to.be.false;
+  });
+});
--- a/SpecRunner.html
+++ b/SpecRunner.html
@@ -77,6 +77,7 @@
     <script src="src/renderer.js"></script>
     <script src="src/template-cache.js"></script>
     <script src="src/object.js"></script>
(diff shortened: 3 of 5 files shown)


jmeas:
:+1:

test/unit/features.spec.js
```diff
@@ -0,0 +1,26 @@
+describe('features', function() {
+
+  beforeEach(function() {
+    _.extend(Marionette.FEATURES, {
```
cmaher: maybe use `this.sinon.stub(Marionette, 'FEATURES', { ... })`?

cmaher:
:+1:

src/features.js
```diff
@@ -0,0 +1,6 @@
+Marionette.FEATURES = {
+};
+
+Marionette.isEnabled = function(name) {
+  return !!Marionette.FEATURES[name];
+};
```
thejameskyle: Does this need it's own file?

cmaher: Why not?

test/unit/features.spec.js
jasonLaster: Naa I think thats over thinking it.

src/features.js
thejameskyle: Shrug

test/unit/features.spec.js
cmaher: as long as we're careful not to introduce a 'foo' feature tag later :stuck_out_tongue::

test/unit/features.spec.js
```diff
@@ -0,0 +1,26 @@
...
+  });
+
+  it('disabled when its present and false', function() {
+    Marionette.FEATURES.foo = false;
+    expect(Marionette.isEnabled('foo')).to.be.false;
+  });
+
+  it('disabled when not present', function() {
+    delete Marionette.FEATURES.foo;
+    expect(Marionette.isEnabled('foo')).to.be.false;;
```
cmaher: this extra semicolon seems to be the one breaking travis

jasonLaster: oh my. that could be it. thought that was an es6 feature. like ,,

jasonLaster:
All good.

@cmaher - the bug was that I wasn't including the source file in node.js so grunt would miss it. I run my tests in SpecRunner which was fine

thejameskyle:
:+1:

PR #2316, opened 2015-02-11
Add child events for layout children views

jasonLaster:
Region events are one of the most common requets we've received:
events in a layout like you can listen for child events in a collection
view.

```js
LayoutView.extend({

  childEvents: {
    'submit': 'onChildViewSubmit'
    'fail': 'onChildViewFail'
  },

  onChildViewSubmit: function(childView, successMsg) {},
  onChildViewFail: function(childView, failMsg) {}
})
```

### Workarounds:
The first is share a radio channel,
the second is share a common eventable object (think model). Messaging
through a channel can be ugly because it encourages an explosion of
channels that can be hard to keep track of. Think lots of little wires
through the app. Messaging through a shared eventable object, is ugly
because it's a source of indirection that's unexpected. It's not obvious
that the model is being used in this way.

### Event Bubbling:
This is not event bubbling where an event will go all the way up the
view tree. I do not want to rule that out, as it might be nice, but
there are more design considerations there.

### Utility Methods
When we introduced the `_parent` reference, we discussed that there was
support for getting nested children for proxying events, but there was
no support for getting ancestor views for proxying events and other
internal utilities. We now have support for this, but will obviously
discourage end users from tapping into the power. With great power
comes...

### Docs
I'll add them when the time comes 

### Major
This should be mostly trivial to get onto major. There's one ref to LayoutView, but besides that everything should be fine

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -245,7 +245,7 @@ var ChildView = new Marionette.ItemView.extend({
   showMessage: function () {
     console.log('The button was clicked.');
 
-    this.trigger('show:message');
+    this.triggerMethod('show:message');
   }
 });
 
--- a/docs/marionette.layoutview.md
+++ b/docs/marionette.layoutview.md
@@ -27,6 +27,7 @@ will provide features such as `onShow` callbacks, etc. Please see
 
 * [Basic Usage](#basic-usage)
 * [Region Options](#region-options)
+* [LayoutView.childEvents](#layoutview-childevents)
 * [Specifying Regions As A Function](#specifying-regions-as-a-function)
 * [Overriding the default RegionManager](#overriding-the-default-regionmanager)
 * [Region Availability](#region-availability)
@@ -101,6 +102,74 @@ new Marionette.LayoutView({
  }
 })
 ```
+
+### LayoutView childEvents
+
+You can specify a `childEvents` hash or method which allows you to capture all
+bubbling `childEvents` without having to manually set bindings.
+
+The keys of the hash can either be a function or a string
+that is the name of a method on the layout view.
+
+The function is called in the context of the view. The first parameter is
+the child view, which emitted the event, the remainder are the arguments
+associated with the event.
+
+```js
+// childEvents can be specified as a hash...
+var MyLayoutView = Marionette.LayoutView.extend({
+
+  // This callback will be called whenever a child is rendered or emits a `render` event
+  childEvents: {
+    render: function(childView) {
+      console.log("a childView has been rendered");
+    }
+  }
+});
+
+// ...or as a function that returns a hash.
+var MyLayoutView = Marionette.LayoutView.extend({
+
+  childEvents: function() {
+    return {
+      render: this.onChildRender
+    }
+  },
+
+  onChildRender: function(childView) {
+  }
+});
+```
+
+This also works for custom events that you might fire on your child views.
+
+```js
+  // The child view fires a custom event, `show:message`
+  var ChildView = new Marionette.ItemView.extend({
+    events: {
+      'click .button': 'showMessage'
+    },
+
+    showMessage: function (e) {
+      console.log('The button was clicked.');
+      this.triggerMethod('show:message', msg);
+    }
+  });
+
+  // The parent uses childEvents to catch that custom event on the child view
+  var ParentView = new Marionette.LayoutView.extend({
+    childEvents: {
+      'show:message': function (childView, msg) {
+        console.log('The show:message event bubbled up to the parent.');
+      }
+    },
+
+    onChildviewShowMessage: function (childView, msg) {
+      console.log('The show:message event bubbled up to the parent.');
+    }
+  });
+```
+
 
 ### Specifying Regions As A Function
 
--- a/src/layout-view.js
+++ b/src/layout-view.js
@@ -14,6 +14,10 @@ Marionette.LayoutView = Marionette.ItemView.extend({
     destroyImmediate: false
   },
 
+  // used as the prefix for child view events
+  // that are forwarded through the layoutview
+  childViewEventPrefix: 'childview',
+
   // Ensure the regions are available when the `initialize` method
   // is called.
   constructor: function(options) {
--- a/src/view.js
+++ b/src/view.js
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
   // import the `triggerMethod` to trigger events with corresponding
   // methods if the method exists
   triggerMethod: function() {
+    var ret = Marionette._triggerMethod(this, arguments);
+
+    this._triggerEventOnBehaviors(arguments);
+    this._triggerEventOnParentLayout(arguments[0], _.rest(arguments));
+
+    return ret;
+  },
+
+  _triggerEventOnBehaviors: function(args) {
     var triggerMethod = Marionette._triggerMethod;
-    var ret = triggerMethod(this, arguments);
     var behaviors = this._behaviors;
     // Use good ol' for as this is a very hot function
     for (var i = 0, length = behaviors && behaviors.length; i < length; i++) {
-      triggerMethod(behaviors[i], arguments);
+      triggerMethod(behaviors[i], args);
+    }
+  },
+
+  _triggerEventOnParentLayout: function(eventName, args) {
+    var layoutView = this._parentLayoutView();
+    if (!layoutView) {
+      return;
     }
 
-    return ret;
+    // invoke triggerMethod on parent view
+    var eventPrefix = Marionette.getOption(layoutView, 'childViewEventPrefix');
+    var prefixedEventName = eventPrefix + ':' + eventName;
+
+    Marionette._triggerMethod(layoutView, [prefixedEventName, this].concat(args));
+
+    // call the parent view's childEvents handler
+    var childEvents = Marionette.getOption(layoutView, 'childEvents');
+    var normalizedChildEvents = layoutView.normalizeMethods(childEvents);
+
+    if (!!normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
+      normalizedChildEvents[eventName].apply(layoutView, [this].concat(args));
+    }
   },
 
   // This method returns any views that are immediate
@@ -298,6 +325,28 @@ Marionette.View = Backbone.View.extend({
       if (!view._getNestedViews) { return memo; }
       return memo.concat(view._getNestedViews());
     }, children);
+  },
+
+  // Internal utility for building an ancestor
+  // view tree list.
+  _getAncestors: function() {
+    var ancestors = [];
+    var parent  = this._parent;
+
+    while (parent) {
+      ancestors.push(parent);
+      parent = parent._parent;
+    }
+
+    return ancestors;
+  },
+
+  // Returns the containing parent view.
+  _parentLayoutView: function() {
+    var ancestors = this._getAncestors();
+    return _.find(ancestors, function(parent) {
+      return parent instanceof Marionette.LayoutView;
+    });
   },
 
   // Imports the "normalizeMethods" to transform hashes of
--- a/test/unit/view.spec.js
+++ b/test/unit/view.spec.js
@@ -198,7 +198,7 @@ describe('base view', function() {
       this.MyView = Marionette.View.extend({
         options: {
           className: '.some-class'
-        } 
+        }
       });
       this.myView = new this.MyView();
       expect(this.myView.className).to.equal('.some-class');
@@ -209,7 +209,7 @@ describe('base view', function() {
         return {
           className: '.some-class'
         };
-      };  
+      };
       this.myView = new Marionette.View(options);
       expect(this.myView.className).to.equal('.some-class');
     });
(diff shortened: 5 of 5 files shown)


jasonLaster:
After sitting on this for a day, there are two additional things that I'd like to consider:
1. adding a `childEvents` property to layoutView that works the same as collectionView
2. deleting [proxyChildEvents](https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L454) in favor of using this new approach 
3. cleaning up some variable/method names

jasonLaster:
Just pushed updates that make use of `childEvents` and remove `proxyChildEvents` which is huge. There are some broken tests, but things are looking pretty cool. I think you guys could also help me clean up the internals a lot

jasonLaster:
Alright, so I think this is in a good place. I went ahead with unifying implementation logic between layouts and collection views and that work is [here](https://github.com/jasonLaster/backbone.marionette/compare/marionettejs:minor...region-events2). I decided that while this works and is super cool, there is one really big drawback for unifying the. collectionView proxies `trigger` and I was only going to proxy `triggerMethod`, which is how `behaviors` work. If we want to make `CollectionView` only proxy w/r/t `triggerMethod` we *could* do it in minor I think because it passes my, it's so esoteric/rarely used nobody will notice test, but ofcourse that's more of a judgement call (semver rrrrr).

---

So, what should you look for w/ this PR:
1. layouts have a `childEvents` property
2. child view events emitted w/ `triggerMethod` will be called on the layout so `child:event` and `onChildEvent` will work.

src/layout-view.js
```diff
@@ -14,6 +14,10 @@ Marionette.LayoutView = Marionette.ItemView.extend({
     destroyImmediate: false
   },
 
+  // used as the prefix for child view events
+  // that are forwarded through the layoutview
+  childViewEventPrefix: 'childView',
```
jasonLaster: this will be changed to `childview` to be consistent

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
   triggerMethod: function() {
+    var ret = Marionette._triggerMethod(this, arguments);
+
+    this._triggerEventOnBehaviors(arguments);
+    this._triggerEventOnParentLayout(arguments[0], _.rest(arguments));
+
+    return ret;
+  },
+
+  _triggerEventOnBehaviors: function(args) {
```
jasonLaster: small cleanup

paulfalgout:
So I'm starting to have reservations about this.

Correctly me if I'm wrong, but this only proxies `triggerMethod` and not `trigger` ?

This is a big inconsistency with how `childEvents` works on `CollectionView`  Below is our code example from that `CollectionView` documentation which uses `trigger`  It just seems like users will expect `trigger` to work, and I don't know why we wouldn't want that too.

```
// The child view fires a custom event, `show:message`
var ChildView = new Marionette.ItemView.extend({
  events: {
    'click .button': 'showMessage'
  },

  showMessage: function () {
    console.log('The button was clicked.');

    this.trigger('show:message');
  }
});

// The parent uses childEvents to catch that custom event on the child view
var ParentView = new Marionette.CollectionView.extend({
  childView: ChildView,

  childEvents: {
    'show:message': function () {
      console.log('The show:message event bubbled up to the parent.');
    }
  }
});
```

jasonLaster:
We could. Same mechanics. Trigger is funny cause I think onChildeventsFoo works too. I believe proxy does that for you.

test/unit/view.spec.js
```diff
@@ -242,4 +242,74 @@ describe('base view', function() {
...
+        },
+
+        childEvents: {
+          'boom': 'onBoom'
+        },
+
+      });
+
+      this.ChildView = Marionette.ItemView.extend({
+        template: _.template('')
```
samccone: heh just use `template: false`

jasonLaster:
NOTE: we discussed in channel adding support for `trigger` so that `this.trigger` would be handled by a layout's childEvents. We all agree that it's useful, but I'm hesitant to personally add it at this time because this PR is already complicated and I've got a lot on my plate.

---
I'll be adding some docs in a little bit after breakfast

docs/marionette.collectionview.md
```diff
@@ -245,7 +245,7 @@ var ChildView = new Marionette.ItemView.extend({
   showMessage: function () {
     console.log('The button was clicked.');
 
-    this.trigger('show:message');
+    this.triggerMethod('show:message');
```
jasonLaster: we should be encouraging people to use `triggerMethod`

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
     // Use good ol' for as this is a very hot function
     for (var i = 0, length = behaviors && behaviors.length; i < length; i++) {
-      triggerMethod(behaviors[i], arguments);
+      triggerMethod(behaviors[i], args);
     }
+  },
 
-    return ret;
+  _triggerEventOnParentLayout: function(eventName, args) {
+    var layoutView = this._parentLayoutView();
```
jasonLaster: I think this should be `layoutView`

test/unit/view.spec.js
```diff
@@ -265,4 +265,74 @@ describe('base view', function() {
...
+    describe('when there is a containing layout', function() {
+      beforeEach(function(){
+        this.layoutView.showChildView('child', this.childView);
+        this.childView.triggerMethod('boom', 'foo', 'bar');
+      });
+
+      it('emits the event on the layout', function() {
+        expect(this.layoutEventHandler)
+          .to.have.been.calledWith(this.childView, 'foo', 'bar')
+          .and.CalledOnce;
```
samccone: add context and arguments plz

test/unit/view.spec.js
```diff
@@ -265,4 +265,74 @@ describe('base view', function() {
...
+      });
+
+      it('invokes the layout on handler', function() {
+        expect(this.layoutEventOnHandler)
+          .to.have.been.calledWith(this.childView, 'foo', 'bar')
+          .and.CalledOnce;
```
samccone: add context and arguments plz

test/unit/view.spec.js
```diff
@@ -265,4 +265,74 @@ describe('base view', function() {
...
+      });
+
+      it('invokes the layout childEvents handler', function() {
+        expect(this.layoutViewOnBoomHandler)
+          .to.have.been.calledWith(this.childView, 'foo', 'bar')
+          .and.CalledOnce;
```
samccone: add context and arguments plz

src/view.js
samccone: can we split this change into another commit, just to keep things atomic

docs/marionette.view.md
```diff
@@ -152,6 +152,7 @@ For more on efficient, deeply-nested view structures, refer to the LayoutView do
 This is just like the attach event described above, but it's triggered right before the view is
 attached to the document.
 
+
```
samccone: :x

docs/marionette.collectionview.md
samccone: can we split this into another commit just to be atomic

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
+    if (!layoutView) {
+      return;
+    }
+
+    // invoke triggerMethod on parent view
+    var eventPrefix = layoutView.childViewEventPrefix;
```
samccone: `getChildViewEventPrefix`

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
+    var prefixedEventName = eventPrefix + ':' + eventName;
+
+    Marionette._triggerMethod(layoutView, [prefixedEventName, this].concat(args));
```
samccone: so 

Just to be clear
`(eventName, ViewInstance, restOfArguments....)`

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
+
+    // call the parent view's childEvents handler
+    var childEvents = _.result(layoutView, 'childEvents');
```
samccone: ya gona want to use
https://github.com/marionettejs/backbone.marionette/blob/master/src/helpers.js#L26

here

src/view.js
samccone: https://github.com/marionettejs/backbone.marionette/blob/master/src/helpers.js#L26

samccone: since can be on the class, statically, a fn, or an option, static of fn

src/view.js
samccone: since can be on the class, statically, a fn, or an option, static of fn

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
+    var normalizedChildEvents = layoutView.normalizeMethods(childEvents);
+
+    if (!!normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
```
samccone: ha why do we need to type case here :) either it is a hash or it is not :)

src/view.js
```diff
@@ -300,6 +327,28 @@ Marionette.View = Backbone.View.extend({
...
 
+  // Internal utility for building an ancestor
+  // view tree list.
+  _getAncestors: function() {
+    var ancestors = [];
+    var parent  = this._parent;
+
+    while (parent) {
+      ancestors.push(parent);
+      parent = parent._parent;
```
samccone: Since this chain can be quite large for complex apps and this is a "hot" method, should we investigate a recursive trampolined implementation?

docs/marionette.layoutview.md
```diff
@@ -102,6 +103,64 @@ new Marionette.LayoutView({
...
+  This also works for custom events that you might fire on your child views.
+
+  ```js
+  // The child view fires a custom event, `show:message`
+  var ChildView = new Marionette.ItemView.extend({
+    events: {
+      'click .button': 'showMessage'
+    },
+
+    showMessage: function () {
```
samccone: going to want to put the arguments in this.. because people like to see them

docs/marionette.layoutview.md
```diff
@@ -102,6 +103,64 @@ new Marionette.LayoutView({
...
+      this.triggerMethod('show:message');
+    }
+  });
+
+  // The parent uses childEvents to catch that custom event on the child view
+  var ParentView = new Marionette.LayoutView.extend({
+    childView: ChildView,
+
+    childEvents: {
+      'show:message': function () {
```
samccone: going to want to put the arguments in this.. because people like to see them

docs/marionette.layoutview.md
```diff
@@ -102,6 +103,64 @@ new Marionette.LayoutView({
...
+        console.log('The show:message event bubbled up to the parent.');
+      }
+    },
+
+    onChildviewShowMessage: function () {
```
samccone: going to want to put the arguments in this.. because people like to see them

docs/marionette.layoutview.md
```diff
@@ -102,6 +103,64 @@ new Marionette.LayoutView({
 })
 ```
 
+### LayoutView childEvents
+
+You can specify a `childEvents` hash or method which allows you to capture all bubbling `childEvents` without having to manually set bindings. The keys of the hash can either be a function or a string that is the name of a method on the layout view.
```
samccone: we will want to mention what these are called with and what context

src/view.js
jasonLaster: yup that's it

docs/marionette.collectionview.md
jasonLaster: I think this should be grouped because it's part of the rationale.

src/view.js
jasonLaster: meh - not sure... copy pasta

src/view.js
jasonLaster: a wha????

docs/marionette.layoutview.md
jasonLaster: gotta make the people happy

src/view.js
samccone: yep it is not faster
http://jsperf.com/while-vs-recursion-mn

docs/marionette.collectionview.md
samccone: k

src/view.js
samccone: http://raganwald.com/2013/03/28/trampolines-in-javascript.html

src/view.js
jasonLaster: same argument - i'd prefer to keep this refactor together.

src/view.js
jasonLaster: i see

samccone:
:+1: lgtm

samccone:
@marionettejs/marionette-core can i get one more set of eyes on this... pretty sure it is good to go

samccone:
@marionettejs/marionette-core eyes plz

docs/marionette.layoutview.md
ahumphreys87: @jasonLaster still needs the args :)

docs/marionette.layoutview.md
```diff
@@ -102,6 +103,77 @@ new Marionette.LayoutView({
...
+
+    showMessage: function () {
+      console.log('The button was clicked.');
+      this.triggerMethod('show:message', msg);
+    }
+  });
+
+  // The parent uses childEvents to catch that custom event on the child view
+  var ParentView = new Marionette.LayoutView.extend({
+    childView: ChildView,
```
paulfalgout: I don't think this is really helping?

test/unit/view.spec.js
```diff
@@ -265,4 +265,77 @@ describe('base view', function() {
...
+      this.layoutViewOnBoomHandler = this.sinon.spy();
+      this.layoutView.onBoom = this.layoutViewOnBoomHandler;
+    });
+
+    describe('when there is not a containing layout', function() {
+      beforeEach(function(){
+        this.childView.triggerMethod('boom', 'foo', 'bar');
+      });
+
+      it('emits the event on the layout', function() {
```
paulfalgout: Should this be `it('should not emit the event on the layout, ` ?

paulfalgout:
Couple of small notes/questions.. and I think there's still an outstanding comment about arguments in the documentation for `showMessage`

otherwise LGTM :+1:

docs/marionette.layoutview.md
jasonLaster: nope copy pasta bug. nice catch

test/unit/view.spec.js
jasonLaster: yep

docs/marionette.layoutview.md
jasonLaster: got em

jasonLaster:
@marionettejs/marionette-core should be all good

paulfalgout:
:shipit: :+1:

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
+
+    Marionette._triggerMethod(layoutView, [prefixedEventName, this].concat(args));
+
+    // call the parent view's childEvents handler
+    var childEvents = Marionette.getOption(layoutView, 'childEvents');
+    var normalizedChildEvents = layoutView.normalizeMethods(childEvents);
+
+    if (!!normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
+      normalizedChildEvents[eventName].apply(layoutView, [this].concat(args));
+    }
```
jmeas: so much code to support this :(

jmeas:
Lots of code to add this, but it does make Mn more consistent.

:+1:

src/view.js
jasonLaster: Haha this is what we do. We write code. We can make it better though

ianmstew:
:clap: Thank you, thank you, thank you for this.

src/view.js
```diff
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
...
     var behaviors = this._behaviors;
     // Use good ol' for as this is a very hot function
     for (var i = 0, length = behaviors && behaviors.length; i < length; i++) {
-      triggerMethod(behaviors[i], arguments);
+      triggerMethod(behaviors[i], args);
     }
+  },
 
-    return ret;
+  _triggerEventOnParentLayout: function(eventName, args) {
```
megawac: @jasonLaster Any reason to do this on each `triggerMethod` rather than registering the events at view intitializtion and binding the functions there?

paulfalgout: In a future version of Mn I think I'd prefer to only use https://github.com/marionettejs/backbone.marionette/pull/2818 and remove the auto proxying of events.  we throw sooooo many events with no listeners and the auto proxying cause edge case bugs where combining collectionview proxying and layout child proxying doubles events.

Issue #1570, opened 2014-07-03
CompositeView's HTML state is lost when sorting collection.


MeoMix:
I have a CompositeView with the following logic:

```
onShow: function () {
    //  TODO: Marionette 2.0 broke this UI when calling .sort() -- need to fix!
    this.ui.panel.transition({
        x: this.ui.panel.width()
    }, 300, 'snap');
}
```

This logic provides the CompositeView with a nice slide-in effect when it is being shown.

The CompositeView's ChildViewContainer is able to be organized using jQuery UI's sortable functionality:

```
this.ui.childContainer.sortable({
    update: function (event, ui) {
        var playlistId = ui.item.data('id');

        var playlist = this.collection.get(playlistId);
        playlist.set({
            sequence: ui.item.index() 
        });

        //  Collections with a comparator will not automatically re-sort if you later change model attributes
        this.collection.sort();
    }.bind(this)
});
```

When sort is called -- the CompositeView is re-rendered which causes the transform on my panel to be lost. This is bad and unexpected! How should I fix?

Here's a video: http://screencast.com/t/hcTSyp6rGhJk

PR #2325, opened 2015-02-13
Don't rerender the entire CompositeView on sorting

Dremora:
Currently, every time a collection is sorted, the `CollectionView` calls `resortView` which in turn calls `_renderChildren`. It seems logical to do so, since `render` doesn't do much more than calling `_renderChildren`.

The same behavior is used for `CompositeView`: it calls `_sortView` which calls `_renderChildren` which in turn calls `render`. Out of these 3 methods, only `render` is reimplemented in `CompositeView`. Here it does much more - it re-renders not only the children, but also the entire template.

It doesn't make much sense to do so, re-rendering just children seems to be enough. And this is exactly what the `resortView` description suggests us to do (while incorrectly referring to `_renderChildren` as `renderChildren`).

I have replaced the call to `render` with the call to `_renderChildren` and removed part of the method description.

All tests still pass.


samccone:
Seems quite reasonable to me :+1: 

@Dremora great research and write up. This is a big win for consistency as you said and makes a ton of sense when viewed in the larger context of the other view implementations and code paths.

samccone:

ahumphreys87:
yup im good with this as well, nice catch @Dremora :+1:

samccone:
~~great research and write up. This is a big win for consistency as you said and makes a ton of sense when viewed in the larger context of the other view implementations and code paths.~~

As pointed out by @ahumphreys87 this is actually a breaking change and something that we have talked about before

> CompositeViews are tricky tho..... since they render into the root element by default... So this change makes sense only if you are rendering into a custom container that we can clear out.

https://github.com/marionettejs/backbone.marionette/issues/1570#issuecomment-48221863

This change will be breaking to people who are not rendering into a container :|

ahumphreys87:
@samccone although this does throw up an issue with our tests :) Ill create an issue and work on it

Dremora:
I was suspecting there'll be a backwards compatibility issue, but wasn't sure where exactly.
With this taken into account, a possible solution could be, instead of modifying `resortView` in `CollectionView`, to reimplement it in `CompositeView` like this:
```javascript
  resortView: function() {
    if (Marionette.getOption(this, 'reorderOnSort')) {
      this.reorder();
    } else if (Marionette.getOption(containerView, 'childViewContainer')) {
      this._renderChildren();
    } else {
      this.render();
    }
  }
```

I can't however understand what exactly will break if we go ahead with the original solution. I couldn't extract a valid explanation from #1570. I've tried writing some tests and everything seems work work as expected (rendering a collection into root with a template present, then sorting, then comparing the HTML with the expected one). I'm hoping that the test implemented for #2334 will clarify it for me.

Also, the question asked by @MeoMix: should this be submitted as a PR for the `major` branch?

ahumphreys87:
@Dremora Ive also had a go at creating a test but `_renderChildren` seems to work fine. Im :+1: for this change but would you mind adding a couple of things before we merge:

[] Update docs
[] Test to prove `CompositeView` still works without a `childViewContainer`

Something like this should do for the test:

```js
  describe('when sorting a CompositeView without a childViewContainer', function() {
    beforeEach(function() {
      this.compositeView = new this.CompositeView({
        childView: this.ChildView,
        collection: this.collection,
        template: _.template('<p>Hello</p>')
      });

      this.compositeView.render();

      this.model = new Backbone.Model({foo: '0', bar: '5'});
      this.collection.add(this.model);

      this.collection.comparator = 'foo';
      this.collection.sort();
    });

    it('should still render the markup of the template', function() {
      expect(this.compositeView.$el).to.have.$text('Hello0123');
    });
  });
```

ahumphreys87:
@Dremora any progress on this?

Dremora:
@ahumphreys87 not yet — but I'll try to finish it soon.

jasonLaster:
rebase please @Dremora.

where's this at?

ahumphreys87:
#closing in favour of #2411

PR #2330, opened 2015-02-14
when template function is called, it passes view context

airinterface:


jmeas:
Duplicate of #2081

thejameskyle:
since when did coveralls start commenting on issues again?

samccone:
yep so this seems reasonable enough and I know we talked about it in #2081

any opposition @jmeas @thejameskyle from cherry-picking in the work done on this PR and the work in #2081 into major?

jmeas:
I don't think that this is a good idea. The whole point of templates is that you pass them data, and that data is used to render the template. The context of the template should not matter.

If someone wants to do this (they shouldn't) then they can override the method in their own application. But this should def. not be the default behavior imo.

samccone:
you are right it should "not" matter. That said it does no harm exposing the correct context to the view template render method. There have been a few people that have asked for this behavior over time, and we always kinda say... just hack it in if you need it.

It is for these reasons that I am not so anti it.

jmeas:
If we agree that templates relying on their context is a bad idea, then this is also a bad idea. It's dead simple to add into your own application if you want to write crummy templates.

:-1:

samccone:
kk fair enough @jmeas 

Thanks again for the work @airinterface . 

We are going to not move forward with this at this time. If you would like to chat more about this please come join us in chat https://gitter.im/marionettejs/backbone.marionette/

PR #2331, opened 2015-02-15
Update marionette.regionmanager.md

FlorentD:
'getRegion' doesn't seem to exist with RegionManager in 2.3.2. We rather use 'get'.


samccone:
ah good catch, I really screwed that one up back in september

https://github.com/marionettejs/backbone.marionette/commit/f3aa2912b1817f457241f29d5265e18b96fa8c79

great catch :+1:

samccone:
manually merged

PR #2332, opened 2015-02-16
Updating the appRoutes object when the appRoute() function is called

cjmyles:
Backbone version #1.1.2
Underscore version #1.7.0
jQuery version #1.11.1
Marionette version #2.2.0
Operating System / version #OS X Yosemite 10.10.1
Browser and version #Chrome 40.0.2214.111 (64-bit)

Why: The problem - In my version of Marionette.AppRouter I do not define the 'appRoutes' object but I do define 'controller'. I instantiate Marionette.AppRouter (var router = new Router()) and then I call the router.appRoute() method to dynamically associate a route with a controller method. I noticed that if I create an onRoute function within the Router class the path is being returned as undefined, rather than the path I created. I believe this is because the appRoutes object is not being updated when the appRoutes() function is being called dynamically, and therefore the path cannot be found when the route event is fired.

What: I propose to instantiate the appRoutes object as an empty object if it doesn't already exist, or leave it as is if it does exist. In addition to this I propose to add the new route to the appRoutes object whenever appRoutes() is called.

How: This solves the problem as the _processOnRoute() function now has access to the routes that were dynamically added and can find the associated path.

Given my lack of knowledge of the inner workings of backbone.marionette I admit there may be a more elegant way of solving this. I also wonder if this will cause problems if someone is using the 'routes' object of their Router rather that 'appRoutes', in other words using methods on the Router rather than using a controller.

I ran grunt test without any issues.


samccone:
@cjmyles this seems to be a reasonable addition. Would you be willing to write a spec for this new functionality 

(add it here https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/app-router.spec.js)

Basically write a test that would have failed before this addition and write a test that proves that this patch does what you say it does.

Thanks for the great write up and PR. As always you can come join us in chat https://gitter.im/marionettejs/backbone.marionette/ if you have any questions.

jfairbank:
@cjmyles can you rebase this please?

cjmyles:
Hey @jfairbank does that look ok now?

jfairbank:
Still seeing a ton of other files. How did you do your rebase? Something like this should work:

cjmyles:
Ok tried that - let me know if there's any issues. First time doing this so you'll have to bear with me!

jfairbank:
I'm still seeing a lot of files. When you have your branch checked out what are you running to rebase and what do you see when you run the interactive rebase (`$ git rebase -i upstream/minor` for example)? Can you share some screenshots by chance?

jasonLaster:
Closing in favor of working off of this branch https://github.com/marionettejs/backbone.marionette/pull/2451

PR #2337, opened 2015-02-17
Allow an underscore as a splitter in the method string name that is passed in

jeremyhicks:
This would allow a string "method_name" to properly convert to "onMethodName" for the function.

This is helpful to me as the keys that are being walked through my application also link up with css class names and item views that are properly named. I believe this would be consistent with similar string manipulation.

### DIFF
--- a/src/trigger-method.js
+++ b/src/trigger-method.js
@@ -6,7 +6,7 @@
 
 Marionette._triggerMethod = (function() {
   // split the event name on the ":"
-  var splitter = /(^|:)(\w)/gi;
+  var splitter = /(^|:|_)(\w)/gi;
 
   // take the event section ("section1:section2:section3")
   // and turn it in to uppercase name


jmeas:
Thanks for the PR, @jeremyhicks! This is breaking, so it's a candidate for the v3.0 release.

I'm hesitant to include this in the core library. The colon is there because of a precedent set by Backbone – colons are separators in the event names. Marionette follows the same convention. I don't want people to use colons, separators, and hyphens all willy-nilly, which is what could result if we made this in the core library.

What we could do is expose the regex so that it's easier for people to modify, if they choose. This approach seems more in line with our stance on convention vs. configuration.

samccone:
Hi @jeremyhicks thanks for the PR. 

I will be the first to say that I dislike the magic that the event splitter adds to Marionette. To many it seems like a black box that just somehow works. It is in that automagic land, that we have been trying to avoid as we grow Marionette into a more mature and stable tool.

Your pr adds more complexity to this regex, and adds another branch of logic for someone to have to wrap their head around. If we were to merge something like this I think it most likely be a replacement and not an addition. 

That said I am :-1: on this direction, it is a syntactical decision and one that make it look "odd" which is exactly how the event splitter works .. oddly.

Let me know what others think @marionettejs/marionette-core  Obviously I am but one voice in a sea of many. 


Thanks again for the work @jeremyhicks, come join us in chat if you want to talk more 
https://gitter.im/marionettejs/backbone.marionette/

src/trigger-method.js
```diff
@@ -6,7 +6,7 @@
 
 Marionette._triggerMethod = (function() {
   // split the event name on the ":"
-  var splitter = /(^|:)(\w)/gi;
+  var splitter = /(^|:|_)(\w)/gi;
```
megawac: regex can be improved to `[^:_](\w)/g` if accepted

megawac:
I love event splitter but i dislike this alternative :-1:

jmeas:
Gonna close for now. @jeremyhicks, if you open a PR that exposes this regex, then it's more likely to be implemented.

Thanks!

Issue #2339, opened 2015-02-17
Availability of Views ui.hash in Behaviors is inconsistent


benmccormick:
From my discussion with @samccone on Gitter:

The way Behaviors mixin Views ui.hashes is inconsistent right now.

On the events and trigger hashes, Behaviors mixin `view.ui` to their own independent ui hash.  In callback functions, only the Behavior's ui hash is available.

So this works.

```javascript
let ExampleBehavior = Mn.Behavior.extend({
  events: {
    'click @ui.test' : 'doStuff'
  },

  doStuff: function() {
   //...
  }
});

let ExampleView = Mn.ItemView.extend({
  ui: {
   test: '.test'
  },
  behaviors: {
    example: {
      behaviorClass: ExampleBehavior
    }
  }

});
```

and this works

```javascript
let ExampleBehavior = Mn.Behavior.extend({
  events: {
    'click @ui.test' : 'doStuff'
  },
  
  ui: {
    test: '.test'
  },

  doStuff: function() {
    this.ui.test.addClass('example');
  }
});

let ExampleView = Mn.ItemView.extend({
 
  behaviors: {
    example: {
      behaviorClass: ExampleBehavior
    }
  }

});
```

but this does not

```javascript
let ExampleBehavior = Mn.Behavior.extend({
  events: {
    'click @ui.test' : 'doStuff'
  },
  
  doStuff: function() {
    this.ui.test.addClass('example');
  }
});

let ExampleView = Mn.ItemView.extend({
   ui: {
    test: '.test'
  },
  behaviors: {
    example: {
      behaviorClass: ExampleBehavior
    }
  }

});
```

It would be great to make this behavior more consistent, and either mix in `view.ui` consistently, or ignore it entirely.

Related to #2338

samccone:
:+1: 

the only thing... I dislike about the coupling is that ideally behaviors should be "dumb" and not know about your views too much... this encourages coupling but.. I suppose we are already quite far down that hole

benmccormick:
I see this as making them **less knowledgeable** about views because they don't have to know the selectors/structure of the view (which is needed if you're going to use selectors in a separate behavior hash).  It does mean views need to know more about the behaviors, but I'd rather have the views knowing about the behaviors than vice-versa.

samccone:
yep i'm with you :)

ahumphreys87:
Fixed in 2.4 via #2344

Issue #2144, opened 2014-12-17
calling delegateEvents on a view that includes a behavior with a ui binding breaks


jonknapp:
I ran into a possible bug when trying to call `delegateEvents` after the `ui` hash has been modified for a view's `Behavior`.  It appears that the `Marionette.normalizeUIKeys` method is calling the `Marionette.normalizeUIString` whose purpose is to return the jQuery selector.  This works fine before the view is rendered, however after `ui` elements start to point to jQuery objects, we can no longer call `delegateEvents` again.

It will try to create a `_behaviorsEvents` hash that contains items like:

```
{
  ...
  "keypress [object Object] ": function bound(),
  ...
}
```

`[object Object]` is an invalid selector, which then throws an error in jQuery once the event is caught.

Issue https://github.com/marionettejs/backbone.marionette/issues/1408 seems to relate to it, though on the `View` side.  I haven't dived as deep on that side yet, but it looks like there were steps to try and work with a cached version of the `ui` hash so there was access to the original selectors: https://github.com/marionettejs/backbone.marionette/pull/1731

samccone:
This has been merged and is going out with the 2.4.0 release

jasonLaster:
Great work @jonknapp!

PR #2342, opened 2015-02-18
Improve mixin of view ui hash to behaviors

### LINKED ISSUE #2338: Add a helper method to abstract the merging of behavior ui elements
Basically, we allow for ui interpolation on events, and triggers

https://github.com/marionettejs/backbone.marionette/blob/master/src/behaviors.js#L157
https://github.com/marionettejs/backbone.marionette/blob/master/src/behaviors.js#L52

We do the work to merge the view's ui hash and the behaviors ui hash.. and then we do the interpolation, this is some semi-complex behavior (the merging that is) that needs some refactor love. 

Let's isolate the work into a single method, and then rejoice.

Thanks for asking a good question @benmccormick

------
Related
https://github.com/marionettejs/backbone.marionette/pull/2308

benmccormick:
This commit makes the view ui hash available consistently in
Behaviors, making it available in callbacks as well as the
events and triggers hashes.  There is also some refactoring to simplify
and centralize how the views ui hash is mixed into the Behaviors
reducing duplication

This should close #2338 and #2339

### DIFF
--- a/src/behaviors.js
+++ b/src/behaviors.js
@@ -37,23 +37,15 @@ Marionette.Behaviors = (function(Marionette, _) {
 
     behaviorEvents: function(behaviorEvents, behaviors) {
       var _behaviorsEvents = {};
-      var viewUI = this._uiBindings || _.result(this, 'ui');
 
       _.each(behaviors, function(b, i) {
         var _events = {};
         var behaviorEvents = _.clone(_.result(b, 'events')) || {};
-        var behaviorUI = b._uiBindings || _.result(b, 'ui');
 
-        // Construct an internal UI hash first using
-        // the views UI hash and then the behaviors UI hash.
-        // This allows the user to use UI hash elements
-        // defined in the parent view as well as those
-        // defined in the given behavior.
-        var ui = _.extend({}, viewUI, behaviorUI);
 
         // Normalize behavior events hash to allow
         // a user to use the @ui. syntax.
-        behaviorEvents = Marionette.normalizeUIKeys(behaviorEvents, ui);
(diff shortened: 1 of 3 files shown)


src/behaviors.js
```diff
@@ -154,10 +145,9 @@ Marionette.Behaviors = (function(Marionette, _) {
 
     // Internal method to build all trigger handlers for a given behavior
     _buildTriggerHandlersForBehavior: function(behavior, i) {
-      var ui = _.extend({}, this._viewUI, _.result(behavior, 'ui'));
       var triggersHash = _.clone(_.result(behavior, 'triggers')) || {};
```
jasonLaster: GetBehaviorsUI looks like it's doing something slightly different from before. 
Why, _uiBindings or ui?

jasonLaster:
Mind reopening against minor. I believe this is technically not breaking.

samccone:
yep this sucker can go into minor!

src/behaviors.js
samccone: since uiBindings are the denormalized version of the UI hash I think... right?

benmccormick:
This is replaced by #2344 which is going against minor

src/behaviors.js
benmccormick: I'll be honest I don't have a rock solid understanding of it, but its related to this #2144, and seems to be a way to make sure that the UI hash can be updated after the original setup?

It was only implemented for events before.  This change makes it work for triggers.  Didn't work for behavior methods before and won't now.  I'm not sure if thats a problem or not, I don't understand the use case for changing the ui hash after the views been initialized.

PR #2346, opened 2015-02-18
doc(collectionview): bump attachHtml default impl

jiaweihli:
Copying the old snippet results in the `childViewContainer` option
getting ignored.


jasonLaster:
nice find

samccone:
good catch

samccone:
putting source code in our docs... seem fragile but this is a battle for another day and this is objectively better. thanks so much @jiaweihli !

PR #2350, opened 2015-02-20
return this for destroy for consistency

ahumphreys87:


src/controller.js
```diff
@@ -25,6 +25,7 @@ _.extend(Marionette.Controller.prototype, Backbone.Events, {
 
     this.stopListening();
     this.off();
+
```
samccone: :x: drop change

samccone:
We will want to update the docs, and the .yaml stubs

ahumphreys87:
@samccone updated docs and yaml

samccone:
:clap: 

:+1: from me

cmaher:
:+1:

PR #2352, opened 2015-02-21
Added isDestroyed to Marionette.Controller and Marionette.Object

rmaffitsancsoft:
Added an isDestroyed property to the Marionette.Controller. When the controller is destroyed, isDestroyed is set to true after the `before:destroy` event is triggered. If the controller has already been destroyed, the destroy method returns the controller.


samccone:
Hey @rmaffitsancsoft thanks for the PR, it seems to be correct, just a note, you should use the `Marionette.Object` class, controller is a remnant that we are killing in 3.0. 

The quick once over seems like everything is pretty darn close to being good to go. Will give this the in-depth review treatment.

Thanks for the great work here. As always please come join us in chat
https://gitter.im/marionettejs/backbone.marionette/

jmeas:
Thanks for the PR! But I don't think that we should merge this.

First and foremost, Controller is deprecated. We should not add additional features to a Class that we're planning to remove.

If we *did* want to add these changes, then we should add them to Object. But I don't think that we should. It's easy enough to create a new 'base' Object in your application with this extra functionality.

samccone:
Ah interesting perspective @jmeas . 

I agree this class is deprecated, however it is still there and in use on minor. This PR makes it more consistent with the rest of classes. I agree with you that people should be using Object or migrating to it. 

meh I think it is ok but I see where you are coming from

rmaffitsancsoft:
I added the functionality to the Marionette.Object, if you want I can create a new PR for just the Object change.

samccone:
Ah thinking about this more this is "new" lifecycle behavior on the Object / Controller level. 

We only enforce this destroy behavior on the View layer currently, but I think this is a wise thing to enforce on any class that has a destroy lifecycle method....

so I am still :+1: to adding this, open to thoughts either way tho

rmaffitsancsoft:
The use case I ran into where having the isDestroyed on the controller would be useful was when dealing with fetching data before showing a view in a region. I use a similar method to the `when:synced` method described in the backbone rails video series, and I was working on extending it so you could pass in another (optional) parameter that it could check to see if it had been destroyed or not. If it (the controller in my case) had already been destroyed, it wouldn't execute the callback.

jfairbank:
Thanks for the PR. I understand the consistency with views this may provide, but I'm not sold on it. Maybe it's the way I've built my apps before, but I don't see a frequent use case for this.

I think building a base object for your app like @jmeas suggested is a better way to go.

rmaffitsancsoft:
Gotcha, I hadn't considered using a base object for this specific use case. I also wasn't aware of the controller being deprecated. Ignoring the addition to controller, thinking about the consequences of adding isDestroyed to Marionette.Object:

If the goal of the object is to be as generic as possible, then I agree that the change is unnecessary. An application-specific object could be created that sets the property to false in it's initialize method and true in the onBeforeDestroy event handler. The only caveat I can see with this approach is that there is no way to prevent the destroy method body from executing twice (thus internally triggering two destroy and before:destroy events) if the object was already destroyed. I realize that the second destroy's `before:destroy` and `destroy` triggers wouldn't actually do anything because, assuming my understanding is correct, `this.stopListening()` removes all handlers listening to the object.

With that in mind, I kind of agree that adding this directly to Marionette is not necessary, but I do still think it could be useful to know if something has already been destroyed. Whether or not it should go in an application-specific base object or Marionette itself really just depends on what the desired behavior of calling destroy on an already destroyed object is.

jasonLaster:
rebase please @rmaffitsancsoft

jasonLaster:
@rmaffitsancsoft - I like the idea of adding `isDestroyed` to Object. Closing this PR because `Controller` is deprecated and we're working a major release

PR #2355, opened 2015-02-21
Define view event order when showing a view in a region.

scottaj:
Conceptually, it seems like the first thing that should happen when showing a view in a region is that the `before:show` event gets fired.

Right now, a view is rendered before the `before:show` event is fired. So the the event order is `[before:render, render, before:show, show]`. I think it should wrap the rendering of the view so the order is `[before:show, before:render, render, show]`

This has bit me the other day when I wanted to use `before:show` to do some pre-render setup in a view that should only happen the first time it is shown in a region.

I don't know the intent when designing of all the `before:...` events, but it seems to me like they are most useful if they fire before anything relating to that action has happened.

Let me know if you agree with this idea conceptually, and if so I will work on making this spec pass.


paulfalgout:
I don't know that I necessarily disagree, however, something to consider.. the original add of the  `before:show`  was to give people a way to add views to `LayoutView` prior to it being show.  The issue here is that in order to do so, the layout would need to be rendered.

jmeas:
@paulfalgout sums up my thoughts on this. This change may in fact make a lot of sense, but it would break **a whole lotta** people's apps.

I think it's possible that one day we do make the change, but we would need to market it well.

@paulfalgout shed some light on the history, but let me go through it in a bit more in-depth. Right now you can use the `onBeforeShow` callback to render a deeply nested layout in 1 reflow – guaranteed. We've been pushing people to do this for well over a year now, praising the pattern for its performance benefits over composing a view tree in, say, `onShow`. Because of this, I would wager most folks use `onBeforeShow` to build their nested view tree. If we made this change, then all of those applications would break.

It turns out that our performance argument doesn't hold up to much scrutiny. Unless your code causes a reflow, browsers will batch all of the DOM operations into a single reflow even if you use `onShow` to build your layout. And even if you **do** cause a reflow, then the loss of speed generally won't be that much worse. Browsers are pretty quick at doing a handful of reflows in a row (not 60fps fast, but folks generally aren't animating entire view tree transitions).

Ember's entire nested view/route system is built on taking advantage of not building your tree in one reflow. In fact, there are actually some pretty awesome reasons to not do it that way.

scottaj:
Thanks for the background, I hadn't known any of that. I have never seen anything recommending building subviews trees in `onBeforeShow`, I have always used `onShow`.

Sounds like this would be a major breaking change though.

jasonLaster:
@paulfalgout @scottaj what do we want to do with this?

I'm tempted to close this as the order of events is probably tested elsewhere extensively...

scottaj:
@jasonLaster @paulfalgout @jmeas 

I didn't see any tests for event order actually when I looked. I still think this change makes sense, but obviously there is a lot of historical baggage riding on it. Do you guys want me to finish out the PR and see what the actual change would look like? I could get that done sometime this week I expect. Otherwise I guess just close it.

jasonLaster:
@scottaj could we get this test to pass as is and then create a new PR for the change?

scottaj:
So you want me to change the test to check for the currently defined order, and then open a new PR with a it changed back to the order I proposed as well as the fix to make that pass?

jasonLaster:
@scottaj yup - it'd be good to lock down the current behavior, before we "improve it".

jasonLaster:
@scottaj Mind reopening this PR against next. Keeping minor and major in sync is too hard.

scottaj:
Yup, I can move it. Pretty busy this week though so will try to get to it this weekend.

scottaj:
Closing so I can reopen against the `next` branch

Issue #2357, opened 2015-02-22
Use template as el


designermonkey:
One thing I find increasingly annoying about Backbone is it's penchant for wrapping everything in a `div`. While using `LayoutManager` for a short time, I liked how a developer could set the `el` of a `View` to `false`, so as to tell `LayoutManager` to use the provided template as the `View`s `el`.

IMO, this should be a part of `Marionette` also. There is already the option to pass false to `template` to disable it, so would follow suit in that regard.

It would be optional and defaulted to the current behaviour to not break any backwards compatibility.

From what I can see of the source code, it's not easily something that can be monkey patched in an extension of `View` either, although I am really going to try.

Any thoughts?

samccone:
xref
https://github.com/marionettejs/backbone.marionette/issues/2041
https://github.com/marionettejs/backbone.marionette/issues/815
https://github.com/marionettejs/backbone.marionette/issues/431
https://github.com/jashkenas/backbone/issues/2329

I still stand by what I said in #2041

designermonkey:
Thanks for replying so fast.

After reading all those issues (thanks for the links, I did try looking), I can see how they are related. I am also astounded at some of the responses regarding why it should be accepted to allow arbitrarily wrapping anything added by JavaScript in a div, because, well, just because.

I know this isn't a responsibility of Marionette, which goes a hell of a way to patch up the great simplicity of Backbone to be a useable framework for JavaScript, but I have to say, it is a problem that needs addressing. If it can be proven to work in an extension of Backbone, then maybe it can be back-ported into Backbone.

Frameworks like this, need to be as agnostic about how a developer chooses to implement their HTML and CSS as possible. Personally, I don't tag every element with a classname because it looks horrible, and as stated in one response, it looks auto-generated, and doesn't adhere to DRY and KISS.

Wrapping everything in divs forces a certain way of working, and when the JavaScript is the last piece of a very large project-puzzle to be done, having to go back through to re-write all of the CSS can ruin morale, budgets, deadlines, and client expectations, especially when the developers have written HTML and CSS in a very light manner to keep markup clean.

Also, think about this markup...

```html
<ul>
    <div>
        <li></li>
    </div>
    <div>
        <li></li>
    </div>
    
    ...
    
</ul>
```

You would never allow this to be output by a server side script, so it should never become acceptable for JavaScript.

samccone:
[...]

PR #2358, opened 2015-02-22
First pass at optional root element

designermonkey:
1. Adds ability to pass `el: false` via View.
2. Handles errors if both `template: false` and `el: false`.
3. Changes `attachElContent` in `ItemView` and `CompositeView` to allow replacing default `$el`.

This is my very naive first attempt at this. I have more than likely not considered some aspects, so please feel free to tell me where I've not looked.

One thing that jumped out at me is that `CompositeView` doesn't handle errors when `template` is false. Is that expected behaviour?

I am new to Marionette, so please help me get this PR up to scratch.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -271,15 +271,42 @@ Marionette.View = Backbone.View.extend({
   // import the `triggerMethod` to trigger events with corresponding
   // methods if the method exists
   triggerMethod: function() {
+    var ret = Marionette._triggerMethod(this, arguments);
+
+    this._triggerEventOnBehaviors(arguments);
+    this._triggerEventOnParentLayout(arguments[0], _.rest(arguments));
+
+    return ret;
+  },
+
+  _triggerEventOnBehaviors: function(args) {
     var triggerMethod = Marionette._triggerMethod;
-    var ret = triggerMethod(this, arguments);
     var behaviors = this._behaviors;
     // Use good ol' for as this is a very hot function
     for (var i = 0, length = behaviors && behaviors.length; i < length; i++) {
-      triggerMethod(behaviors[i], arguments);
+      triggerMethod(behaviors[i], args);
+    }
+  },
+
+  _triggerEventOnParentLayout: function(eventName, args) {
+    var layoutView = this._parentLayoutView();
+    if (!layoutView) {
+      return;
     }
 
-    return ret;
+    // invoke triggerMethod on parent view
+    var eventPrefix = Marionette.getOption(layoutView, 'childViewEventPrefix');
+    var prefixedEventName = eventPrefix + ':' + eventName;
+
+    Marionette._triggerMethod(layoutView, [prefixedEventName, this].concat(args));
+
+    // call the parent view's childEvents handler
+    var childEvents = Marionette.getOption(layoutView, 'childEvents');
+    var normalizedChildEvents = layoutView.normalizeMethods(childEvents);
+
+    if (!!normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
+      normalizedChildEvents[eventName].apply(layoutView, [this].concat(args));
+    }
   },
 
   // This method returns any views that are immediate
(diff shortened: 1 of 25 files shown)


designermonkey:
#2357

samccone:
I would start with getting the tests to pass and adding some specs. We can review the implementation details after :)

thejameskyle:
This is not a pattern we absolutely should not be changing

src/view.js
```diff
@@ -17,6 +18,10 @@ Marionette.View = Backbone.View.extend({
     // at some point however this may be removed
     this.options = _.extend({}, _.result(this, 'options'), options);
 
+    if (this.options && ('el' in this.options ? this.options.el === false : this.el === false)) {
```
megawac: You can do `this.getOption('el') === false` here

jmeas:
@tbranyen told me that he's happy with the way LayoutManager handles this. If we get his :+1: on an implementation then I am +1.

samccone:
@thejameskyle I kinda want to see where this or goes. I view it as an exploration on top of the issue.  If you have comments regarding the feature I would voice your thoughts over on the origin issue

src/view.js
designermonkey: Thanks for the hint, I'll update.

designermonkey:
@samccone this is where I start to fail myself. I have no idea how testing in JS works I'm afraid, and honestly haven't got a lot of time to learn right now (way too much client work on). Would you be able to help?

@thejameskyle I can understand the reservation, but this would be an optional thing that wouldn't change any default behaviour.

tbranyen:
@thejameskyle care to elaborate why? Especially given the absoluteness of your comment.  I don't know exactly how Marionette Views are architected, but the only issue we've faced w/ LayoutManager and optional root elements are when developers try and use two root elements, which would obviously not work and breaks the conventions in place w/ Backbone.  Element:View are 1:1.

It can totally work (and does).  We are very well tested for this feature so I have hope it'll make it into Marionette as well.

The only fear we had was that Backbone would change the semantics of `el` in the future, but that was a risk we were willing to take, believing naively in semver.  Recently though, I've grown more and more concerned, since it seems like Backbone could break this in the future at a patch level.

designermonkey:
I have been naive here, as I've completely overlooked how `CollectionView` works. Sorry. I don't know enough to make this work without guidance.

thejameskyle:
The element to view pattern is one of the very few contracts you have with any Backbone View. From the moment it's created until destroyed a View has a single element that everything can be based upon, from event delegation to a lot of Marionette's features like the `ui` hash. This makes it fairly simple to author mixins, plugins, behaviors, etc. There needs to be a _really_ good reason to break this contract. The `<dl>` case is unfortunate but I'm not sure that's a good enough reason to make this change.

thejameskyle:
Also, this is pretty common in other frameworks' views/components. Both Ember and React need a root element. I believe angular directives need some kind of root node as well (I believe this can be any type of node?).

jmeas:
Random thought: I wonder if a document fragment could do some magic here. Imagine if a user specified that they didn't want an element, and their template contains many root nodes. What if we rendered it into a document fragment, and did all of the event delegation / ui binding / etc. there. By keeping a reference to the fragment around, we could manage everything like normal. When it comes time to append it to the DOM, the fragment would just vanish.

I admit I'm not sure how much of the HTMLElement API fragments have tho' / how well they work with jquery / and so on. But I'd like to investigate the possibility!

tbranyen:
@thejameskyle both Ember and React enforce the rules trying to be added by this PR.  With React you don't define the root element on the class definition, you define it in the template.  This allows you to do much more modifications to the root element via the template syntax instead of programmatically in the instance methods.

In the case of LayoutManager we warn the user when they use two top level elements, and React has done the same.  It's a workable pattern.

thejameskyle:
@tbranyen I meant in the sense that they enforce that rule, not that it's defined in the same way. I don't think a view should have multiple root elements and I `this.($)el` should always refer to that single root element.

jasonLaster:
Closing this PR as it stale and I don't see the idea moving along. 

Thankyou for the work @designermonkey and really good discussionguys @thejameskyle, @tbranyen, @thejameskyle.

PR #2365, opened 2015-02-24
Improve Behavior documentation

benmccormick:
Two main goals

1. Make it clear that View life cycle functions are available on
Behaviors
2. Improve grammar/consistency of docs when referencing Views and
Behaviors

Specifically the docs had alternated between `behavior`, `view`, Behavior, View, and behavior, view
to refer to Behaviors and Views.

In addition to some spelling/grammar stuff, I attempted to fix this by following the convention that:

- `View` and `Behavior` refer to the Marionette.View and Marionette.Behavior classes
- View and Behavior refer to a View or Behavior class (possibly a subclass of `View` or `Behavior`)
- view and behavior refer to instances of a View or Behavior class.

PR #2368, opened 2015-02-24
Adding an array-based syntax for Behaviors

### LINKED ISSUE #2366: Add alternative array syntax for Behaviors
As I've started to use Behaviors more, I've noticed that the current syntax for binding Behaviors to Views is optimized for a namespace based method of storing Behaviors.  As somebody who uses a module system (AMD and ES6 modules in my case), and uses several Behaviors that don't accept options, I end up writing extra code every time.  Specifically in situations like this

```javascript
import * as Mn from 'marionette';
import * as ValidationBehavior from 'behaviors/validation';
import * as KeymapBehavior from 'behaviors/keymap';
import * as WarnOnDestroyBehavior from 'behaviors/warn';

let ExampleView = Mn.LayoutView.extend({
    behaviors: {
        validation: {
            behaviorClass: ValidationBehavior,
        },
        keymap: {
            behaviorClass: KeymapBehavior
        },
        warn: {
           behaviorClass: WarnOnDestroyBehavior,
           message: 'Are you sure you want to delete this?'
        }
    },
});

export default ExampleView
```

So in this case validation, keymap, and warn are all unnecessary names that aren't used and don't need to be anything in particular.  They're basically just boilerplate/fluff.  I also have to keep typing behaviorClass and the object syntax even though classes are my primary way of defining behaviors, and I often don't need additional options.  It would be much nicer if I could write that code like this instead:

```javascript
import * as Mn from 'marionette';
import * as ValidationBehavior from 'behaviors/validation';
import * as KeymapBehavior from 'behaviors/keymap';
import * as WarnOnDestroyBehavior from 'behaviors/warn';

let ExampleView = Mn.LayoutView.extend({
    behaviors: [ValidationBehavior, KeymapBehavior, {
      behaviorClass: WarnOnDestroyBehavior,
      message: 'Are you sure you want to delete this?'
    }],
});

export default ExampleView
```

Instead of a big object and a bunch of boilerplate, I just pass the Behaviors as an array that accepts either a Behavior class, or an object that is defined the same way Behaviors are defined now.  So I avoid the unnecessary keys, and Behaviors that don't receive options can avoid all of the boilerplate altogether.

I'm happy to whip up a PR for this, I just wanted to get feedback first.  Note I'm proposing this as an additional syntax, not a breaking change.

benmccormick:
Add an array based syntax for attaching Behaviors to Views.  

Would close #2366.

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -55,7 +55,8 @@ var MyView = Marionette.ItemView.extend({
 
 Interaction points, such as tooltips and warning messages, are generic concepts. There is no need to recode them within your Views. They are prime for abstraction into a higher level non-coupled concept, which is exactly what Behaviors provide you with.
 
-Here is the syntax for declaring which behaviors get used within a View.
+Here is the syntax for declaring which behaviors get used within a View.  You can pass behaviors either as a set of key-value pairs where the keys are used to
+lookup the behavior class, or as an array.
 The keys in the hash are passed to `getBehaviorClass` which looks up the correct `Behavior` class.
 The options for each Behavior are also passed through to the Behavior during initialization. The options are then stored within each Behavior under `options`.
 
--- a/docs/marionette.behaviors.md
+++ b/docs/marionette.behaviors.md
@@ -25,50 +25,48 @@ Marionette.Behaviors.behaviorsLookup = function() {
 }
 ```
 
-By default the Behaviors are looked up by their key value in a given View's behavior hash.
+There are 2 different syntaxes for attaching Behaviors to a View.  The first is an object syntax where the Behaviors are looked up by their key value in a given View's behavior hash. The second is an array syntax where you can pass the Behavior class directly.
 
-In this sample (using the default `getBehaviorClass` implementation) your code will expect the following Behaviors to be present in `window.Behaviors.DestroyWarn` and `window.Behaviors.ToolTip`
+In this sample, which uses the object syntax, your code will expect the following Behaviors to be present in `window.Behaviors.DestroyWarn` and `window.Behaviors.ToolTip`
 
 ```js
 var MyView = Marionette.ItemView.extend({
 	behaviors: {
+    ToolTip: {},
 		DestroyWarn: {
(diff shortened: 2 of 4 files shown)


benmccormick:
I included my improve behavior documentation PR in here as well since I didn't want to deal with the merge conflicts involved in that, but I can pull it out or smash it into the other commit if people prefer.  That was a separate PR #2365

jasonLaster:
behavior docs are in! great job

benmccormick:
@jasonLaster thanks!   So that got merged into master, and this is against minor, so whats the procedure for how stuff migrates from master to minor?  It will make it easier for others to review this pull request if that stuff gets pulled out of the diff.

samccone:
I keep minor on top of master. We can cherrypick the SHA onto minor and it will go away when i rebase of of master

JSteunou:
Ho god +1, so sad it did not reach 2.4 :p

samccone:
oh man @benmccormick mind getting a rebase on this for us?

benmccormick:
@samccone yep, rebased and updated to comply with jscs.

jasonLaster:
:+1: Nice work @benmccormick!

samccone:
in it goes

thanks again @benmccormick

benmccormick:
Awesome!  Thanks all.

thejameskyle:
Hmm, sorry I didn't tune into this sooner. This might be a bad idea for inheritance, since there is no way you can easily inherit now without implementaion knowledge about the parent class.

This is already complicated enough:
```js
var MyView = AnotherView.extend({
  behaviors: function() {
    return _.extend(_.result(AnotherView.prototype, 'behaviors'), {
      // more behaviors
    });
  }
});
```

This is a complete disaster:
```js
var MyView = AnotherView.extend({
  behaviors: function() {
    var behaviors = _.result(AnotherView.prototype, 'behaviors');
    var myBehaviors = {
      // more behaviors
    };

    if (_.isArray(behaviors)) {
      return behaviors.concat(_.values(myBehaviors));
    } else {
      return _.extend(behaviors, myBehaviors);
    }
  }
});
```

PR #2376, opened 2015-02-26
allow CompositeView to override how reordered children are appended

### LINKED ISSUE #2367: CompositeView reorderOnSort ignores childViewContainer
Using a CompositeView with:

```
childViewContainer: 'tbody'
reorderOnSort: true
```
Moves the children outside of the table element on the first sort.  The sort is performed properly; the elements just aren't in the table afterward.  Issue appears to be that the reordering is being performed by the base CollectionView, not the CompositeView.

ahumphreys87:
Fixes #2367 

Added `_appendReorderedChildren` method to allow `CompositeView` to override where the reordered children are appended.

Added a test around this behaviour which fails without the new fix.

### DIFF
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -134,6 +134,14 @@ Marionette.CompositeView = Marionette.CollectionView.extend({
     $container.append(childView.el);
   },
 
+  // Internal method. Append reordered childView'.
+  // Overidden from CollectionView to ensure reordered views
+  // are appended to childViewContainer
+  _appendReorderedChildren: function(children) {
+    var $container = this.getChildViewContainer(this);
+    $container.append(children);
+  },
+
   // Internal method to ensure an `$childViewContainer` exists, for the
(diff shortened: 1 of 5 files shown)


src/composite-view.js
```diff
@@ -134,6 +134,14 @@ Marionette.CompositeView = Marionette.CollectionView.extend({
     $container.append(childView.el);
   },
 
+  // Internal method. Append reordered childView'.
+  // Overidden from CollectionView to ensure reordered views
+  // are appended to childViewContainer
+  _appendReorderedChildren: function(children) {
```
samccone: :boom: 
nice fix

samccone:
:+1:

cmaher:
:+1:

jasonLaster:
:+1: 2.4.1 here we go

samccone:
in minor now

PR #2383, opened 2015-02-27
Start the guides section

jasonLaster:
Lets write some guides
@samccone, @jmeas, @jdaudier, @benmccormick

+ **What are guides?** Guides are the educational content compliment to the API. 
+ **Why write guides now?** The API will be built soon. The website is in a much better place.
+ **What should be the outline for guide content?** @jmeas has a lot of ideas
+ **Why start with the intro?** It's a well written overview of what Mn is about


samccone:
Do we want this to live in the www? Or is there an advantage of having it be versioned behind the main repo?

jdaudier:
i like it in the main repo and pulled in to the www since i think people that want to look at the guides are probably going to be on that repo instead of the www one

jasonLaster:
I also like it being tied to versions and added with PRs. It feels closer to the source.

jasonLaster:
Closing this for now until we give it more thought

@2448

PR #2384, opened 2015-02-27
add checkEmpty flag to destroyChildren in CollectionView

### LINKED ISSUE #2313: CollectionView#destroyChildren should not call checkEmpty
```js
  destroyChildren: function() {
    var childViews = this.children.map(_.identity);
    this.children.each(this.removeChildView, this);
    this.checkEmpty();  // this is a problem
    return childViews;
  },
```
`checkEmpty` will show an emptyView if the collection is empty. `destroy` calls `destroyChildren`, which removes all children, and will then add one back (if the collection is empty). This is a leak.

Other methods calling `destroyChildren`:

* `destroyEmptyView` -- probably shouldn't be immediately showing another empty view
* `_renderChildren` -- manually calls `showEmptyView` when appropriate

Labeling as 'minor' because `destroyChildren` isn't publicly exposed in the docs.

ahumphreys87:
Fixes #2313 

I've added a `checkEmpty` flag to `destroyChildren` to maintain BC for `CompositeView` where users may actually want to destroy the children and show an emptyView. The only places where the flag is passed are: `renderChildren` and `destroy` as it makes no sense to `checkEmpty` in those 2 places.

ping @cmaher

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -200,7 +200,7 @@ Marionette.CollectionView = Marionette.View.extend({
   // process
   _renderChildren: function() {
     this.destroyEmptyView();
-    this.destroyChildren();
+    this.destroyChildren({checkEmpty: false});
 
     if (this.isEmpty(this.collection)) {
       this.showEmptyView();
@@ -538,7 +538,7 @@ Marionette.CollectionView = Marionette.View.extend({
     if (this.isDestroyed) { return this; }
 
     this.triggerMethod('before:destroy:collection');
-    this.destroyChildren();
+    this.destroyChildren({checkEmpty: false});
     this.triggerMethod('destroy:collection');
 
     return Marionette.View.prototype.destroy.apply(this, arguments);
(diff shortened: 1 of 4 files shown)


src/collection-view.js
```diff
@@ -544,10 +544,15 @@ Marionette.CollectionView = Marionette.View.extend({
 
   // Destroy the child views that this collection view
   // is holding on to, if any
-  destroyChildren: function() {
+  destroyChildren: function(checkEmpty) {
```
cmaher: since this is a public api, we should probably accept options instead of just a boolean field.

cmaher:
Good approach. Just one comment

src/collection-view.js
ahumphreys87: Yeah good spot, I'll sort that

ahumphreys87:
@cmaher updated, had to refactor the `shouldCheckEmpty` assignment as the ternary ended up being too long of a line - damn you JSCS! :stuck_out_tongue:

cmaher:
:+1:

cmaher:
@ahumphreys87 it looks like this needs a rebase
@marionettejs/marionette-core it would be cool to get this in 2.4.1, making it a nice bug-fix release.

ahumphreys87:
@cmaher rebased. We've missed the boat on 2.4.1 :disappointed:

cmaher:
I am behind :(

samccone:
heh @ahumphreys87 lets get some docs on this one, and then :ship:

cmaher:
I'm not sure this needs docs, since `destroyChildren` is entirely undocumented (and should probably be made internal in v3)

samccone:
i am all for marking it for a v3 pass, I think we should just be consistent though with our private api. This is in the "public" namespace so we have to support it for now :) might as well add docs you know?

jasonLaster:
looks like the `checkEmpty` param is documented. :+1: on merging

cmaher:
:raised_hands:

Issue #2395, opened 2015-03-04
CollectionView attachBuffer not invoked internally as documented


dkurucz:
Line 2171 invokes attachBuffer with only one parameter (this), while the documentation notes that this should have a second parameter as well, the buffer itself.  It is no longer possible to add the children in the overridden method without manually constructing the buffer out of _bufferedChildren.  Not sure if this is a new design, but this was working correctly in 2.0.1.  Thanks!

dkurucz:
please see proposed solution at https://github.com/marionettejs/backbone.marionette/pull/2396

ahumphreys87:
fixed via #2503

Issue #2388, opened 2015-03-01
Provide a standard mechanism for tweaking collection/compositeView HTML output before attaching to DOM.


sentientnebula:
Example use case:  A View similar to a bowling score sheet with a couple of twists.

A standard bowling score sheet can be easily represented by a sheetView CompositeView containing a collection of gameView rows.  Each gameView, in turn can be represented by a CompositeView consisting of a playerName, a Collection of sets, and a total box.

The twists:  Number of sets is variable, and the width of the score sheet adjusts to fit user device display size and/or user preference.  This is done by setting maxSetsPerRow low enough to not exceed the constraints on the sheet width.  If the actual number of sets per game exceed maxSetsPerRows the sets wrap within their container to form multiple “subrows” per game and extra empty sets are injected as needed to ensure everything divides out evenly and grid remains nicely aligned.  There is no totals box to worry about.

So what is the problem?  In Theory, there is no problem, and in practice there is USUALLY no problem.  At standard 100% zoom there is no problem on any target browser, but certain combinations of device/browser/custom-zoom-level create a 1px rounding error that make the grid render less than perfectly.

The  solution:
Fortunately, all you need to do to prevent this rounding error is to wrap a div around each of the “subrows” that occur when setsPerGame > maxSetsPerRow.

In Marionette 2.3.2 and prior, I achieved this by overriding attachHtml, buffering the setView childViews into a div node until once each maxSetsPerRow iteration, at which point I appended the div node (div-wrapped set of childViews) in place of a childView.

My app broke after upgrading to Marionette 2.4 because attachHtml now passes the childView to attachBuffer, which in turn passes it to _buildBuffer.  buildBuffer wants to extract the node itself from  a childView's el itself rather than taking a node directly.  In addition to overriding attachHTML, I also had to override _buildBuffer to append plain b rather than b.el in order for it to work.

I didn't like this because it was starting to feel a bit convoluted.  Then I thought, with the code changes in 2.4, wouldn't it be better to remove the attachHTML override, let childViews buffer normally, and then apply custom logic at buffer building time?  So I just overrode _buildBuffer and nothing else, and it feels like a much cleaner solution.

The only problem is that the underscore in _buildBuffer implies that it is a private internal function for use by attachBuffer, and not something that should be overridden.

[...]

PR #2396, opened 2015-03-04
bug fix for #2395

dkurucz:


sentientnebula:
I just adjusted my code to 2.4's new buffering logic.   Your proposed patch would break me again by bypassing _buildBuffer.  This kind of reinforces my point about needing a standardized method/approved point for tweaking Collection/CompositView output so you don't need to worry about having the rug pulled out from under you with every update/patch.

See:   <a href="https://github.com/marionettejs/backbone.marionette/issues/2388">#2388</a>

dkurucz:
Overriding methods starting with underscore is dangerous and it would be unwise to assume that version updates will continue to support such things.  The documentation hinted at a convention for changing the content of a CollectionView by overriding the attach, so I figured unless marionette actually gains efficiency by changing the buffer logic, they should just stick to the standard that worked well and then encourage its use in the documentation.  Currently I don't see either happening.

sentientnebula:
Too late for for sticking with the standard.  The previous recommendation was to override AttachHTML, but the new buffer logic added in 2.4 makes that no longer a good place for the reasons described in my post.

_buildBuffer, however, becomes a good spot for custom buffer building logic.  That is why I recommended it be renamed buildBuffer (no underscore).

samccone:
hey @dkurucz @sentientnebula I will dig into this, been really slammed with things at the moment, if you would not mind could you possibly find which commit caused this regression and we can get a fix out instead of layering on "patches"

Thanks so much for the great find and PR @dkurucz 

I can assure you we don't want to break peoples code, and we want to be consistent, the last thing we want to do is to cause you guys issues.

sentientnebula:
Might need to think about this one a bit :)  Buffering code was changed in the context of adding a major feature.  There are also a couple of other open issues related to this change, so you'll want to look at everything together.

*For some reason, I keep referring to _createBuffer as _buildBuffer.  The new function in question is _createBuffer.

I actually like the change so long as it is renamed createBuffer (no underscore), so it can be reliably overridden to tweak how collections are formatted just prior to attachment.  It's a very nice spot for it.
 
<a href="https://github.com/marionettejs/backbone.marionette/commit/0dd7775ae6e261eb3666aa482f9660462065c756">#0dd7775 Allow custom sort comparator for CollectionView</a>

dkurucz:
Just throwing in my $0.02, devs should probably not be responsible for creating and changing the buffer itself, this should be handled by marionette.  The developer should use the resulting buffer content as they see fit inside the attachBuffer override during the render phase.  This was a good idea from the beginning, although I am not sure if there were plans to enhance this part of the framework.  I'm not sure which major feature you are referring to, but it is at the moment inconclusive whether my changes would negatively affect any plans for its implementation.

The problem with changing _createBuffer to createBuffer is that the framework will then need to support two exclusive entry points into the CollectionView content generation phase (render), when really, one should be all that is needed, and ideally, if a better idea/feature/view comes into play down the road, none at all would be needed.  I included my changes into a 2.4.1 build and into my project, and the fixes are working great and are indeed backwards compatible.

@sentientnebula would there be problems going back to your original design if this fix were to be rolled in?

jasonLaster:
ping @cmaher, @ahumphreys87 mind taking a look

ahumphreys87:
LGTM, @dkurucz nice work! Could you a some tests that prove the bug is there without this fix? This will stop this issue creeping back in again :)

sentientnebula:
I think someone needs to line up all the issues together and make a top to bottom review of the collectionView/CompositeView code as a whole, instead of continuing to make out of context changes to bits and pieces of code with each point release.  That how this whole thing got started.

If you accept this request for 2.4.2, without thinking too hard, for example, there will be a request to change it back again in 2.4.3 to match the collectionView's implementation--or vice versa.  At the very least, change them both together.

Has anyone even asked cpjolicoeur why he decided to rebuild the buffer?

I still think there should be a mechanism provided to wrap/sub-group itemviews before attachment. attachement.

sentientnebula:
Looking over cpjolicoeur's commit, it appears that his changes were to intended to delay creating the buffer until right before attachment in order to maximize flexibility to manipulate the children without having to duplicate effort (i.e.sorting, or in my case, grouping).  ChildViews were previously being buffered on the fly as children arrived, so any manipulation would require a rebuffer.

This proposed commit would pull the trigger a bit earlier and bake in the buffer right before it gets to attachBuffer--the place where most people would want to make their changes.

paulfalgout:
> I think someone needs to line up all the issues together and make a top to bottom review of the collectionView/CompositeView code as a whole

:+1:

jasonLaster:
Mind reopening this issue against 2.4.2 so it can go in as a bug fix? 

Also let's add a test to demonstrate the previous bug and the fix.

ahumphreys87:
@dkurucz any word on this?

dkurucz:
Yeah I will get to them this weekend.

sentientnebula:
What exactly is being accomplished by calling _createBuffer, a step earlier other than reducing flexibility?

Actually, I'm not going to be effected by the change because I moved all the customization into _createBuffer, so it won't matter at what point _createBuffer is called.  This, however, is definitely not the recommended customization point.  If this change is accepted what will be the new recommended point to customize the buffer creation process without building the buffer twice?

The documentation says to override attachBuffer.

Now:

Just call your own custom method in place of the _createBuffer.
Your custom buffer method is called.
Standard _createBuffer is not.
Buffer is only built once.

After this patch:

`````
// You might need to override this if you've overridden attachHtml
-  attachBuffer: function(compositeView) {
+  attachBuffer: function(compositeView, buffer) {
     var $container = this.getChildViewContainer(compositeView);
-    $container.append(this._createBuffer(compositeView));
+    $container.append(buffer);
   },
`````
Standard Buffer is created BEFORE recommended override points of attachHTML/attachBuffer.  Too late to customize there now without duplicating effort.

New Recommendation:  ???
Create your own buffer and throw away the already created buffer (buffer created twice).   This seems like a big step backward.

Unless, of course, you override _createBuffer instead.  The problem with this is that _createBuffer is marked as a private method.  I don't think this patch should be accepted unless  _createBuffer is also changed to a public createBuffer method.

ahumphreys87:
Closing in favour of #2494

PR #2402, opened 2015-03-08
Update controller to be inline with Object.

samccone:


samandmoore:
So, this was undone and now it's being redone? :trollface:

samccone:
@samandmoore I don't think this was ever done before... just looked at the blame on the file to refresh my memory, link me if you can find where we flip flopped on this before.
https://github.com/marionettejs/backbone.marionette/blame/master/src/controller.js

Controller is already going to :skull: but I wanted to cut down on the LOC of the lib a bit where we could.

samandmoore:
In version 2.1, I think it used to look like this. Not really relevant, was
just wondering the why and you've answered that.

samccone:
looked back at 2.1 and I don't think it was ever like this before :tokyo_tower:

ahumphreys87:
:+1: from me

jasonLaster:
Nice cleanup!

samandmoore:
weird, i coulda sworn it was like that. my bad :P

PR #2407, opened 2015-03-10
Set regionManager to extend from Object

### LINKED ISSUE #1966: Refactor RegionManager to extend from Object
We recently refactored Region, Application, and Behavior to extend from Object.

We should do the same for RegionManager
https://github.com/marionettejs/backbone.marionette/tree/master/src

Thanks @paulovieira for pointing it out

samccone:
Fixes #1966

### DIFF
--- a/src/region-manager.js
+++ b/src/region-manager.js
@@ -2,12 +2,12 @@
 // --------------
 
 // Manage one or more related `Marionette.Region` objects.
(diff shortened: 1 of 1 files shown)


jasonLaster:
nice fix - I like all these small wins moving us in the right spot

samccone:
:bread: :bride_with_veil: :bridge_at_night:

thejameskyle:
:+1:

PR #2408, opened 2015-03-10
Fix broken markdown interpolation.

samccone:


paulfalgout:
:+1:

samccone:
going to merge... I think we are safe here

PR #2381, opened 2015-02-27
Test for #2373 which appears to be fixed in Marionette 2.4.0

scott-w:
Added a test to confirm #2373 which seems to be fixed in Marionette 2.3.2 and 2.4.0 so that issue can be closed. I'm just including the PR to document the intended behaviour in test format.

PR #2410, opened 2015-03-10
add test for binding UI on views with template: false

ahumphreys87:
Thanks to @scott-w for the work over at #2381


jasonLaster:
:+1:  I'll always merge new tests

PR #2416, opened 2015-03-13
Remove redundant throw error in ItemView._renderTemplate

paulovieira:
In `ItemView._renderTemplate` the `template` property is checked. If it is undefined or null, an error is thrown. But in `Renderer.render` (which is called immediately after) that same check is done. The only thing that changes is error name: `'UndefinedTemplateError'` in the first case, and `'TemplateNotFoundError'` in the second. 

This check can be removed from `_renderTemplate`. The method would be a little cleaner without it.

### DIFF
--- a/src/item-view.js
+++ b/src/item-view.js
@@ -64,21 +64,12 @@ Marionette.ItemView = Marionette.View.extend({
 
   // Internal method to render the template with the serialized data
   // and template helpers via the `Marionette.Renderer` object.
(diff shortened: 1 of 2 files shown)


jasonLaster:
Nice. Fix the failing test and were good. Thanks

paulovieira:
Sorry for not running the tests. The commit is now updated.

jasonLaster:
:+1: nice

ahumphreys87:
:+1: manually merged into minor

PR #2419, opened 2015-03-15
Mn.Controller must come after Mn.Object

paulovieira:
The current build from minor is broken because `Mn.Controller` comes before `Mn.Object`. But with the recent changes introduced in #2402 `Mn.Controller` now depends on `Mn.Object`.

### DIFF
--- a/src/build/bundled.js
+++ b/src/build/bundled.js
@@ -47,8 +47,8 @@
 
   // @include ../error.js
   // @include ../callbacks.js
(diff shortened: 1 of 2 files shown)


samccone:
ah good catch! heh I wonder what we can change in CI to make sure to catch these things :)

jasonLaster:
Heh maybe another test runner w the build.

Issue #1846, opened 2014-09-10
Improve CollectionView API: accept ChildView as a function and make getChildView private


paulovieira:
Many properties accept two forms for its value: 

1) as a direct object; 
2) as a function (which is executed at runtime and returns an object that will override the value for that property).

For instance:

* `events` (for any view type)
* `templateHelpers` (for any view type)
* `regions` (for LayoutView)

The second form is very useful and in my opinion should be generalized as much as possible.

In `CollectionView` we have a similar feature for `childView`, but instead of accepting the two forms we have to use another property if we want to define it as a function: `getChildView`.

```javascript
MyCV = M.CollectionView.extend({
    childView: SpringView,

    getChildView: function(child){
        if(child.get("season")==="summer"){ return SummerView; }

        return WinterView;
    }
});
```

Note that in the above snippet the `childView` property will never be used (because `getChildView` was overriden, and it doesn't use `this.childView`). In practice, by overriding `getChildView` we are also overriding `childView` (in the sense that the only reference to `this.childView` is inside `getChildView`)

The API would be more consistent if we could define `childView` also as function:

```javascript
MyCV = M.CollectionView.extend({
    childView: function(child){
        if(child.get("season")==="summer") { return SummerView; }

        return WinterView;
    }
});
```

or 

```javascript
var myCV = new MyCV({
    childView: function(){
        if(child.get("season")==="summer"){ return SummerView; }

        return WinterView;
    }
});
```

The `getChildView` method would still make sense, but I think it would be better as a private method:

```javascript
    _getChildView: function(child) {
        var childView = this.getOption('childView');
  
        if (!childView) {
            throwError('A "childView" must be specified', 'NoChildViewError');
        }
  
        // <new-stuff>
        if(childView.prototype instanceof Backbone.View){ 
            return childView; 
        }
        else if(_.isFunction(childView)){
            return childView(child)
        }
        else{
            throwError('...');
        }
        // </new-stuff>
    }
```
We can also make this backwards compatible by keeping `getChildView` as a public method. Nothing would change for those who are overriding it. But ideally, instead of overriding `getChildView` we should be defining `childView` as a function. It would make the API cleaner.

I can submit a PR if there is interest in going on with this idea.

samccone:
@paulfalgout would you like to open a PR to implement getChildView? still very open to this feature

paulfalgout:
I think @samccone meant @paulovieira

samccone:
[...]

Issue #1861, opened 2014-09-10
Allow bypassing of Marionette.Collection's ensure initOptions.collection instanceof Backbone.Collection


MeoMix:
Hey,

The following was added in Marionette 2.2:

```
if (initOptions.collection && !(initOptions.collection instanceof Backbone.Collection)) {
    throw new Marionette.Error('The Collection option passed to this view needs to be an instance of a Backbone.Collection');
}
```

This breaks my chrome extension because the Backbone.Collection is instantiated on a background page and then accessed through chrome.extension.getBackgroundPage().

See screenshot: http://i.imgur.com/aLeZwqO.png

Could a means of bypassing this check be provided? Thanks

PR #2420, opened 2015-03-15
Accept ChildView as a function and make getChildView private (issue #1846)

paulovieira:
#### The problem
The `getChildView` property is unnecessary in `Mn.CollectionView`. We could instead allow the `childView` property to be either a view class or a function that returns a view class. This way the role of `getChildView` (which is that of "a function that returns a view class") could be played directly by `childView`. 

This would make the API more simple and more coherent as other properties also allow this duality ("thing" / "function that returns the thing").

Please read #1846 for more details.

#### The solution
1) Remove getChildView from the public API. It still makes sense as a private method though. It's a good place to have the logic that checks if `childView` is a view class or a "regular function" (which should return a view class).
2) add a `_getChildView` private method that is essentially the same as the old `getChildView`

#### Upgrade effort 
This breaking change is very easy to fix: the property `getChildView` should simply be renamed to `childView`. Everything will work as before. Note that in the tests I essentially had to search-and-replace `getChildView` to `_getChildView`.

Note also that having both `childView` and `getChildView` was not possible anyway. If that happened, the view class in `childView` would never be used (that is, the view class to be used would always be the one returned by `getChildView`).

### The bonus
`Mn.CompositeView` will also benefit from this PR: the new functionality of `childView` also works for `CompositeView`. Besides this, if `childView` is undefined, the recursive feature will be then used, as happens currently. The `_getChildView` method in `CompositeView` takes care of this.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -93,7 +93,7 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
     var index;
 
     this.destroyEmptyView();
-    var ChildView = this.getChildView(child);
+    var ChildView = this._getChildView(child);
 
     if (opts.at !== undefined) {
       index = opts.at;
@@ -180,7 +180,7 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
     var models = this._sortedModels();
 
     _.each(models, function(child, index) {
-      ChildView = this.getChildView(child);
+      ChildView = this._getChildView(child);
       this.addChild(child, ChildView, index);
     }, this);
   },
@@ -282,10 +282,11 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
   // Retrieve the `childView` class, either from `this.options.childView`
   // or from the `childView` in the object definition. The "options"
   // takes precedence.
-  // This method receives the model that will be passed to the instance
-  // created from this `childView`. Overriding methods may use the child
-  // to determine what `childView` class to return.
-  getChildView: function(child) {
+  // The `childView` property can be either a view class or a function that
+  // returns a view class. If it is a function, it will receive the model that
+  // will be passed to the view instance (created from the returned view class)
+
+  _getChildView: function(child) {
     var childView = this.getOption('childView');
 
     if (!childView) {
@@ -295,7 +296,20 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
       });
     }
 
-    return childView;
+    // first check if the `childView` is a view class (the common case)
+    if(childView.prototype instanceof Backbone.View || childView === Backbone.View){
+      return childView;
+    }
+    // then check if it's a function (which we assume that returns a view class)
+    else if(_.isFunction(childView)){
+      return childView.call(this, child);
+    }
+    else{
+      throw new Marionette.Error({
+        name: 'InvalidChildViewError',
+        message: '"childView" must be a view class or a function that returns a view class'
+      });
+    }
   },
 
   // Render the child's view and add it to the
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -19,7 +19,6 @@ will provide features such as `onShow` callbacks, etc. Please see
 ## Documentation Index
 
 * [CollectionView's `childView`](#collectionviews-childview)
-  * [CollectionView's `getChildView`](#collectionviews-getchildview)
   * [CollectionView's `childViewOptions`](#collectionviews-childviewoptions)
   * [CollectionView's `childViewEventPrefix`](#collectionviews-childvieweventprefix)
   * [CollectionView's `childViewEvents`](#collectionviews-childviewevents)
@@ -72,8 +71,7 @@ Backbone.Marionette.CollectionView.extend({
 ```
 
 Child views must be defined before they are referenced by the
-`childView` attribute in a collection view definition. Use `getChildView`
-to lookup the definition as child views are instantiated.
+`childView` attribute in a collection view definition.
 
 Alternatively, you can specify a `childView` in the options for
 the constructor:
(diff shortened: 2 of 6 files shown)


ahumphreys87:
@paulovieira I'm :+1: on this PR however I do think itll be a good feature to add into 2.5 rather than waiting for 3.0 - could we keep `getChildView` public to avoid the breaking change? Or is it still breaking?

paulovieira:
Yeah, we could still have `getChildView` public. In that case it wouldn't be breaking, but would be bad design since `getChildView` and `childView` as function-that-returns-a-view-class would be doing the same thing, which would make the users confused. But we could always say `getChildView` would be deprecated for v3 and tell users to use `childView` instead.

If `childView` and `getChildView` are both defined, `getChildView` wins. That is, `childView` would always be ignored, whether it is a view class or a function-that-returns-a-view-class.

jasonLaster:
@paulovieira docs can help here. Also, no reason not to make the `major` PR right afterwards so we don't forget.

paulovieira:
Sure, I'll update the docs as well. But I didn't understand what to do in relation to the branches. Should I make a new PR for minor (keeping `getChildView` public, eventually adding the deprecated status)?

src/collection-view.js
```diff
@@ -295,7 +296,20 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
       });
     }
 
-    return childView;
```
jasonLaster: This could be problematic actually. I remember us trying something like this earlier and running into bugs with identifying constructor functions. 

@MeoMix @samccone @jmeas recollections?

samccone: oh i remember it was with collection and was a total fail instant revert

MeoMix: Might be worth a more explicit comment then :p Cuz to me it just looks like an unused param.

paulovieira: >This could be problematic actually. I remember us trying something like this earlier and 
>running into bugs with identifying constructor functions.

Sorry, I'm not following.

The checks in line 300 and 304 are enough to identify if `childView` is a view constructor function (or "view class", I don't know the term that is usually used in Marionette), or just a regular function (which we assume that returns a constructor function).

jasonLaster: I think @MeoMix ran into an isssue where there was a different Mn lib on the page whose views did not match

paulovieira: Sorry, I don't understand. Did not match with what? You mean the the views were some exotic object, not derived from `Backbone.View`?

jasonLaster: there was a second Backbone lib on the page so those views did not match

paulovieira: I see. But that seems to be a very particular case. The user can still override `_getChildView` for those corner cases. For the vast majority of people this would not be a problem.

The PR has been updated with the docs.

paulfalgout: So the issue in question was https://github.com/marionettejs/backbone.marionette/issues/1861 with passing collection as an array or a backbone.collection.  Rather than solve it, the revert also considered that it might not be a super useful feature on its own.

But in this edge case, really the user should not have to override `_getChildView`, but in the case that there are multiple instances of Backbone, simply always make `childView` a function that returns the class rather than passing the class directly... simple enough.

It might be useful to have some reference on the error though?  So that someone can immediately determine what is wrong, as this edge case might be a major headache to debug otherwise.

@jasonLaster thoughts?

src/collection-view.js
```diff
@@ -282,10 +282,11 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
...
-  // to determine what `childView` class to return.
-  getChildView: function(child) {
+  // The `childView` property can be either a view class or a function that
+  // returns a view class. If it is a function, it will receive the model that
+  // will be passed to the view instance (created from the returned view class)
+
+  _getChildView: function(child) {
     var childView = this.getOption('childView');
 
     if (!childView) {
```
paulfalgout: I don't think this check or error is valuable now.  Its fairly redundant.
`InvalidChildViewError` will be thrown later if there's an issue.

paulovieira: I agree and I think we shouldn't have "custom errors".

On the other hand, for this case I actually think it is useful because I'd say one of the most common mistakes in using `CollectionView` is when the key "childview" is used instead of `childView`. Having a explicit message for this case can save some time.

src/collection-view.js
jasonLaster: Thanks for doing the research Paul!

I think multiple versions of BB on the page is definitely an edge case so, we should probably find a good work around.

I think it's a little bit of a regression if we have to tell everyone who has two versions to always do

```js
extend({
  childView: function() {return View}
})
```

Maybe a good compromise, would be to move the check to a global function that could be overriden?

```js
Marionette.isBackboneView 
```

paulovieira: I think it's a good compromise. I have tried to add it in the past but was rejected: #1877

jasonLaster: Haha - that's true. Lets play it safe and make the check a CollectionView method, that way for those who have two bb libs, they can override this one function to do the check appropriately and be happy.

src/collection-view.js
jasonLaster: Agreed

jasonLaster:
I think this is close!

Quick heads up, we're going to a simpler branch scheme. Mind reopening against next

paulovieira:
> Mind reopening against next

Will do soon!

ahumphreys87:
we gonna leave this as breaking and push to v3 or we wanna push into 2.5? Im easy either way but I think we need to keep `getChildView` if we go for the 2.5 route

jasonLaster:
My vote is to get this sucker into v3. This is s a cool feature, but not something where it needs to land in a minor.

ahumphreys87:
This looks good to me, @paulovieira could you open against next and we get this merged :smile:

ahumphreys87:
closing in favour of #2472

Ive cherry picked onto next - thanks for the good work @paulovieira (you'll still get the contribution points :smile: )

paulovieira:
Sorry for being slow with the new PR. I was going to do it tonight, but @ahumphreys87 has taken care. Thank you all for this improvement to CollectionView!

PR #2421, opened 2015-03-15
Make Travis start build faster with Docker

kahwee:
More information here: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/


jasonLaster:
Thanks @kahwee. I don't have much context here, but seems good

samccone:
Lgtm. Thanks so much and nice find!

PR #2422, opened 2015-03-15
reassign view index on reorder

### LINKED ISSUE #2387: reorderOnSort: true only works once when reversing collection order
When using
```
reorderOnSort: true
```
in a collection or composite view, the reordering will only work once when reversing the order of a collection.  Problem is here:

https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L179

Basically, this is running the list of models in the collection against the BabySitter's view of things.  The first time one changes the comparator such that the model order in the collection is reversed and sorts, we'll land here and the model order will be compared against the BabySitter's view of things.  They won't match, and so the view will be reordered.

However, nothing in the BabySitter has changed at this point.  If we then change the comparator again, once again reversing the order and sorting, we're once again back here, but now the model order matches what the BabySitter regards as correct.  As a result, orderChanged will be undefined; so far as the BabySitter is concerned, the model order and the view order are identical, so there's nothing to do.

Thus, one can reverse the sort order of a collection once, but subsequently, the view will remain unchanged.

ahumphreys87:
Fixes #2387 

Issue was that on a reorder the index on the view in babysitter wasn't reassigned. Now doing this in the `_.map` and added extra tests around reversing the sort order which fail without this fix.


jasonLaster:
:+1:

jasonLaster:
@ahumphreys87 mind rebasing?

jasonLaster:
Mind opening against next?

ahumphreys87:
moved to #2455

PR #2424, opened 2015-03-16
typo

caseycs:


jasonLaster:
Nice fix. Not sure why one of the travis' died. 

@samccone - thoughts?

samccone:
rerunning, we need to switch to wget intead of curl so we get retries for free

PR #2427, opened 2015-03-17
use initRenderBuffer to reset the _bufferedChildren array

paulovieira:
We are already using `initRenderBuffer` to reset the `_bufferedChildren` array in other parts of CollectionView (`startBuffering`, `endBuffering`). 

We should do it in `_triggerShowBufferedChildren` as well.


jasonLaster:
@cmaher, @ahumphreys87 thoughts

cmaher:
:+1:

ahumphreys87:
Yup good catch @paulovieira :+1:

PR #2429, opened 2015-03-18
Don't assume that options is always defined; instead just don't rely on it

jiskiras:
If options is undefined when a module gets instantiated, an error gets thrown:

    Uncaught TypeError: Cannot read property 'initialize' of undefined

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -10,7 +10,7 @@ Marionette.Module = function(moduleName, app, options) {
   this.options = _.extend({}, this.options, options);
   // Allow for a user to overide the initialize
   // for a given module instance.
-  this.initialize = options.initialize || this.initialize;
+  this.initialize = this.options.initialize || this.initialize;
 
   // Set up an internal store for sub-modules.
   this.submodules = {};
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -55,7 +55,7 @@ var MyView = Marionette.ItemView.extend({
 
 Interaction points, such as tooltips and warning messages, are generic concepts. There is no need to recode them within your Views. They are prime candidates for abstraction into a higher level, non-coupled concept, which is exactly what Behaviors provide you with.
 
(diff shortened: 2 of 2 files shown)


cmaher:
It looks like that entire line is unnecessary, since we still have this terrible line above it

`this.options = _.extend({}, this.options, options);`

jiskiras:
The problem is that a few lines down, `this.initialize` wasn't using it. I've actually found another problem. When using Modules as a class, you can't define submodules because they get bricked over on line 16.

jasonLaster:
Mind rebasing off of minor

jiskiras:
All done

src/module.js
```diff
@@ -10,7 +10,7 @@ Marionette.Module = function(moduleName, app, options) {
   this.options = _.extend({}, this.options, options);
   // Allow for a user to overide the initialize
   // for a given module instance.
-  this.initialize = options.initialize || this.initialize;
```
jasonLaster: yup. great fix.

jasonLaster:
:+1: 

---
by the way, the travis fail is just bad luck. doesn't look like you

cmaher:
:+1:

samccone:
I would like to see a test here :grimacing:

jasonLaster:
good call. @jiskiras, can we bother you for another PR where you demonstrate the failure w/ a test?

jiskiras:
I've got this created as a test:

          it('should be instantiated as an app.module when created with new keyword', function() {
            var module = new this.ModuleClass('Mod1', this.app);
            expect(this.app.module('Mod1')).to.equal(module.app.module('Mod1'));
          });

Is this a valuable test, or is there something else/more I need to test for. It does demonstrate the problem, however, I'm not sure about the wording or the test itself.

jasonLaster:
If this would have errored before, that's the right test. It's hard to believe the test case was that simple.

Issue #2324, opened 2015-02-12
Add api for binding an object to a radio channel


jasonLaster:
Radio is a fantastic eventbus that we should make as easy to use as possible from within Mn. The api, I have in mind is `bindToChannel`, with some great defaults for things like Routes and Application

```js
var FormService = Marionette.Class.extend({
  logChannel: {
    'event info' :'onInfo'
  },

  formChannel: {
    'event invalid': 'onInvalid',
    'command submit' : 'submitForm'
  },

  initialize: function() {
    this.bindToChannel('log', this.logChannel);
    this.bindToChannel('form', this.formChannel);
  }
})
```

```js
headerRoute = Marionette.Route.extend({
  channel: {
    'command user:show' : 'userShow'
  }
});
```

jasonLaster:
maybe types could be `on`, `reply`, `comply` instead of event, command, request

thejameskyle:
Why not have something like:

```js
Whatever.extend({
  channelName: 'whatever',
  reply: {
    'foo': 'foo'
  },
  comply: {
    'bar': 'bar'
  }
});
```

jasonLaster:
@thejameskyle definitely an option, my thought there is that it's actually kinda common to listen to multiple channels... so i like a flatter api there... also with the api for radio pretty set, i feel confident you can follow the pattern "type event"

thejameskyle:
What about the inverse then:

```js
Class.extend({
  requests: {
    'myChannel foo': 'onFoo'
  },
  commands: {
    'myChannel bar': 'onBar'
  }
});
```

jfairbank:
@thejameskyle I like that a little better for avoiding calling `this.bindToChannel`, but what do we call events? This wouldn't work on a view if we used `events` as the key.

jasonLaster:
@thejameskyle - weird i'm surprised you're not a fan of grouping by channel.

```
channelName: { '(event|command|request) event': 'handler', ... }
```

feels a lot like the inspector radio view here:

![](http://f.cl.ly/items/3i233A2G250N2027341V/Image%202015-02-12%20at%206.22.00%20PM.png)

jmeas:
I'm :+1: for this **if** we end up bundling Radio with Marionette v3.

thejameskyle:
>  like that a little better for avoiding calling this.bindToChannel, but what do we call events? This wouldn't work on a view if we used events as the key.

@jfairbank I'm not sure this should be on every object anyways.

[...]

PR #2431, opened 2015-03-19
Add first class support for Backbone.Radio in Mn.Object

benmccormick:
This is a very initial PR to try and handle #2324.  It needs more tests, any documentation at all, and the code could probably be cleaned up considerably.  But I wanted to keep that issue moving and give everybody a concrete proposal to discuss.  

This PR adds support to Mn.Object and its subclasses for the following API:

```
radioEvents: {
  'app start': 'onAppStart',
  'app finish': 'onAppFinish',
  'books start': 'onBooksStart',
  'books finish': 'onBooksFinish',
},

radioCommands: {
  'app doFoo': 'makeFooHappen',
},

radioRequests: {
  'resources bar': 'getBar',
},
```

The biggest change that it would involve is making Backbone.Radio a dependency of Marionette.  In this PR, I've made it a hard dependency and something that the UMD modules try to pull in automatically.  I could also see this being written in a way though that only activated the feature if Radio was present on the Backbone object.


src/object.js
```diff
@@ -37,5 +37,62 @@ _.extend(Marionette.Object.prototype, Backbone.Events, {
   bindEntityEvents: Marionette.proxyBindEntityEvents,
 
   // Proxy `unbindEntityEvents` to enable unbinding view's events from another entity.
-  unbindEntityEvents: Marionette.proxyUnbindEntityEvents
+  unbindEntityEvents: Marionette.proxyUnbindEntityEvents,
```
jasonLaster: I don't want to add too much more work for you, but I think this deserves it's own file.

```
src/proxyRadioHandlers.js
```

benmccormick: I can get on board with that. It certainly is doubling the size of Mn.Object right now. So you're suggesting just leaving the _.proxyRadioHandlers call in the constructor, and then adding a separate extend call in a different file?

src/object.js
```diff
@@ -37,5 +37,62 @@ _.extend(Marionette.Object.prototype, Backbone.Events, {
...
+      startMethod: 'comply',
+      stopMethod: 'stopComplying'
+    },
+    'radioRequests' : {
+      startMethod: 'reply',
+      stopMethod: 'stopReplying'
+    }
+  },
+  //store the radio channels being used
+  _radioChannels: [],
```
jfairbank: I'm not seeing in the code where you add channels to this array?

src/object.js
```diff
@@ -37,5 +37,62 @@ _.extend(Marionette.Object.prototype, Backbone.Events, {
...
+      }, this);
+  }, this);
+},
+
+  _proxyRadioHandler: function(channel, radioType, messageName, handler) {
+    var method = this._radioTypes[radioType].startMethod;
+    Backbone.Radio[method](channel, messageName, handler, this);
+  },
+
+  _unproxyRadioEvents: function() {
```
jfairbank: This function will probably need called in the `destroy` function too.

src/object.js
```diff
@@ -37,5 +37,62 @@ _.extend(Marionette.Object.prototype, Backbone.Events, {
...
+      }
+      _.each(hash, function(handler, radioMessage) {
+        if (!_.isFunction(handler)) {
+          handler = this[handler];
+        }
+        if (!handler) {
+          return;
+        }
+        var messageComponents = radioMessage.split(' '),
+          channel = messageComponents[0],
```
jfairbank: Can you use individual `var` lines here instead of comma-separated?

src/object.js
jasonLaster: Nah, I'm suggesting adding two functions: Marionette.bindRadioHandlers(context) and Marionette.unbindRadioHandlers(context). I think we could use these two functions in the Object class, but atleast we don't clutter Object source w/ tons of radio knowledge. And, other external classes get this logic for free in 3.0

paulfalgout:
So I think maybe we need to take a step back on this for a sec and look at either Radio additions or Marionette/Radio helpers that need to be introduced prior to this work.

One concern in this PR is that we're using `on/off` for the listeners..  correct me if I'm missing something, but this sets a bunch of `on` listeners on a channel, and when it gets cleaned up, `off` is called on the channel, but you may have other listeners out there getting inadvertently cleaned up.  This is much less of a concern with command/reply as you only have one actor..  

But so we should really be switching to `listenTo`/`stopListening` which fortunately Object already handles quite well.  Similarly I would suggest adding `replyTo` / `complyTo` helpers for Object such that the object's destroy only needs `stopReplying/stopComplying` on itself to handle any `reply/comply` events registered with itself, rather than on the channel directly.

Plus any inline `Object.replyTo` `Object.complyTo` would be automatically cleaned up by the `Object.destroy` as well.

Second,
Is it really the common case that any object would want to do this for multiple channels?  For my use cases, I generally have one channel that the object is mainly concerned with, and occasionally another or two that I might use for a one-off here or there.

I think it would be better to suggest a primary channel for an object, giving the user the freedom to `replyTo` / `complyTo` other channels inline as necessary.  I also think this would simplify the implementation quite a bit.
However if necessary it seems reasonable to allow for the `'channelName replyName' : 'handler'`  but for almost all of my known use-cases I'd be duplicating the same channelName over and over again, rarely using an additional channel.  If that's the case would anyone be interested in a primary channel that is implied such that:

```js
  channelName: 'someChannel',
  radioEvents: {
    'some:event', 'someHandler',
    'otherChannel other:event', 'someOtherHandler'
  }
```
etc.

paulfalgout:
Ah I jumped the gun a bit.. Radio is working on `replyFor` / `complyFor` here:  https://github.com/marionettejs/backbone.radio/pull/157

TL;DR for my last post.  Get this Radio API in first, and utilize it for Object.

jasonLaster:
haha - @paulfalgout. This stuff is confusing, no? 

I definitely like the idea of RadioHelpers and radio helper tests, which will expose the api we offer. Lets try and get the smallest thing in which in my mind is `radioEvents`, `radioRequests`, `radioCommands` and the bind and unbind methods that set them up.

paulfalgout:
I don't know how you accomplish `radioEvents`, `radioRequests`, `radioCommands` without them being leaky or overly destructive without better base helpers that attach listeners to the object instead of the channel.

src/object.js
benmccormick: Whoops, that seems to have been typoed out of existence.  Good catch.  Will clean that up.

src/object.js
```diff
@@ -37,5 +37,62 @@ _.extend(Marionette.Object.prototype, Backbone.Events, {
...
+
+  _proxyRadioHandler: function(channel, radioType, messageName, handler) {
+    var method = this._radioTypes[radioType].startMethod;
+    Backbone.Radio[method](channel, messageName, handler, this);
+  },
+
+  _unproxyRadioEvents: function() {
+    _.each(this._radioChannels, function(channel) {
+      _.each(this._radioTypes,function(commands) {
+        Backbone.Radio[commands.stopMethod](channel,this);
```
jfairbank: Passing in `this` as the second argument will prevent unbinding I believe. The second argument needs to be the callback and then the context. Obviously, `listenTo/stopListening` and the forthcoming command/reply counterparts would be preferred here. Otherwise, we'll need to track the callbacks we add, which is really just reimplementing `listenTo`-type behavior.

benmccormick:
> One concern in this PR is that we're using on/off for the listeners.. correct me if I'm missing something, but this sets a bunch of on listeners on a channel, and when it gets cleaned up, off is called on the channel, but you may have other listeners out there getting inadvertently cleaned up. 

>I don't know how you accomplish radioEvents, radioRequests, radioCommands without them being leaky or overly destructive without better base helpers that attach listeners to the object instead of the channel.

on and off accept a context argument that only deletes events and commands in the context of that object.  See: http://codepen.io/ben336/pen/zxyPQj  I'm currently not doing this right in the code.  The whole unregistering stuff got written too quickly and is screwed up.  I'll clean it up and resubmit soon.  But basically you can have pretty fine-grained control of what you're unsubscribing from.  I was trying (a little too carelessly sadly) to deregister only events in context of the current object.  That would possibly remove any manual listeners on that object, but wouldn't interfere with anything else.

src/object.js
benmccormick: Yep.  I think I actually need to add more arguments here, Backbone events isn't smart enough to figure out that this is supposed to be the context if you only pass one argument.  So this is broken.  But thats the general idea of what I'm trying to achieve.  A small fix will prevent over-unsubscribing on everything but manually subscribed methods on the same object.   Like @jfairbank said, if we want to be airtight we will need to keep track of the callbacks or get the updated radio functions.  It's hardly a disaster as is though.

paulfalgout: https://github.com/marionettejs/backbone.radio/issues/162 There's also a confusing issue with radio around clean-up not being as selective as it should.

@jmeas you anywhere where you could weigh in?

benmccormick:
So, thanks for the feedback everyone.  I'm going to take some of that back and try to improve the PR.  Specifically:

- [] Clean up the _unproxy code and fix my dumb typos
- [] add tests to make sure we're not over-unsubscribing when we do unsubscribe
- [] Pull the radio stuff out into a separate file
- [] Add _unproxy to destroy

I'll aim for that over the weekend and hopefully this can continue to get pushed forward.

paulfalgout:
:clap: thanks a ton for taking this on!  I'm pretty excited to start using it actually.  It will clean up a lot of boilerplate for my various services.

jasonLaster:
too late @paulfalgout 

https://twitter.com/jasonlaster11/status/578954960420769794

benmccormick:
Ok I worked in the feedback from @jasonLaster @jfairbank and @paulfalgout.

The radio code is now in its own file. The unsubscribe stuff has been cleaned up so that it works correctly and happens in the destroy function.  And there are tests to verify that and also make sure it doesn't "over unsubscribe".  

The only cases I can see where this might interfere with existing code:

- If a user has been manually wiring up object methods to radio events and doesn't want them to be unsubscribed when the object is destroyed.  
- if a user is using a `_radioChannels` property on an existing object (seems possible, I think the other functions being added are long enough/specific enough that they're unlikely to cause collisions)

The first one could be mitigated by doing more manual tracking of which callbacks are being attached.  I don't think thats worth it.

The second could be mitigated by creating a central object => channel tracking object rather than sticking that directly on the object.  I'm indifferent here, I can see advantages either way.

Other than that, the only big issue I'm aware of with this code is that it currently makes Radio a hard dependency for Mnv3.  If that is not something the core team wants to do, I'll probably have to put some thought into the best way to make this an optional behavior depending on Radio availability, similar to Backbone.View.$.

test/unit/radio-helpers.spec.js
```diff
@@ -0,0 +1,84 @@
...
+
+  describe('when creating a Marionette.Object', function() {
+
+    beforeEach(function() {
+      this.clickStub1 = this.sinon.stub();
+      this.clickStub2 = this.sinon.stub();
+      this.clickStub3 = this.sinon.stub();
+      this.Object = Marionette.Object.extend({
+        radioEvents: {
+          'foo bar' : this.clickStub1
```
jasonLaster: This is weird because you run into the `object2` problem w/ prototypes.

The easier thing is

```js
```
O = extend({
  radioEvents: {'foo bar': 'barHandler'}

  barHandler: function() {}
});

o = new O();
o.barHandler = this.barHandler;
```

File: test/unit/radio-helpers.spec.js [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+      this.object.destroy();
+      Backbone.Radio.channel('foo').request('bar');
+      expect(this.clickStub3).to.have.been.calledOnce;
+    });
+
+    it('shouldn\'t overunsubscribe events when the object is destroyed', function() {
+      this.object2 = new this.Object();
+      this.object.destroy();
+      Backbone.Radio.channel('foo').trigger('bar');
+      expect(this.clickStub1).to.have.been.calledTwice;
```
jasonLaster: Hmm. why do we need these?

src/radio-helpers.js
```diff
@@ -0,0 +1,64 @@
...
+      if (!hash) {
+        return;
+      }
+      _.each(hash, function(handler, radioMessage) {
+        if (!_.isFunction(handler)) {
+          handler = this[handler];
+        }
+        if (!handler) {
+          return;
+        }
```
jasonLaster: Could we use a `normalizeX` method?

src/radio-helpers.js
```diff
@@ -0,0 +1,64 @@
...
+          channel = messageComponents[0],
+          messageName = messageComponents[1];
+        proxyRadioHandler.call(this,channel, radioType, messageName, handler);
+      }, this);
+    }, this);
+  }
+
+  function proxyRadioHandler(channel, radioType, messageName, handler) {
+    var method = radioTypes[radioType].startMethod;
+    if(this._radioChannels) {
```
jasonLaster: mayble? 
```js
if (!this._radioChannels) {
  this._radioChannels = [];
}
```

test/unit/radio-helpers.spec.js
benmccormick: Mostly that was to assuage the concerns from my original version.  I'm ok with not having the tests, but there was a concern raised that my code might over-unsubscribe and we should wait for Radio to add listenTo equivalents.

test/unit/radio-helpers.spec.js
benmccormick: ok, I can switch that up.

jasonLaster:
Really excited about this @benmccormick!

+ API is solid
+ Left some comments on test and code
+ Lets add Radio as hard dep to Mn for V3.
+ _radioChannels is fine for V3. We can get the upgrader to check for that terrm as a safety.
+ not worried about over subscribing, apps can progressively switch to `radioEvents` etc
+ I'm -1 central registry. It makes sense in theory, but it's not something I'd expect w/ BB.events and other patterns.
+ Looking forward to docs


:100: :palm_tree:

src/radio-helpers.js
ahumphreys87: :+1: Id prefer this over the `else`

ahumphreys87:
@benmccormick this is looking great! This is gonna be such a nice feature for v3!

test/unit/radio-helpers.spec.js
benmccormick: @jasonLaster So after looking at this again, I don't see the problem here.  I pass 2 functions directly and one by reference, to make sure that we handle both of those cases.  Yes object2 shares the stub on the prototype, but I think the testing is still pretty unambiguous.  I just wrote it out the other way, and it makes the code a lot more verbose and (in my mind) less clear.  So I'm going to leave it unless you strongly object.

benmccormick:
@jasonLaster Addressed the comments other than the testing one, which I replied to.  Added a first pass at docs.  

Do we want to try to include this in Views as well?  Or just Object?  Behaviors and Application get it by default now, though I'm not sure commands or requests make any sense at all on Behaviors, since they have to be unique.  But if they're going to get included there, probably makes sense to include them on Views as well.

benmccormick:
Also, I threw together a quick library for anyone who wants to use this feature with Mn 2.1.x.  
https://github.com/benmccormick/marionette-service

ahumphreys87:
@benmccormick docs looking good. Maybe provide a link to the radio docs to help users who are new

test/unit/radio-helpers.spec.js
jasonLaster: i'm cool with this

jasonLaster:
:+1:  (one to go)

---

@benmccormick Lets create a separate PR for adding Radio to Views. It should be easy / un-controversial, but still nice to break up the discussions.

This looks great. Huge Huge. 3.0 contribution.

ahumphreys87:
:+1: let's add the link to radio docs when we add to views

jasonLaster:
:radio: :radio_button: :thought_balloon: 

Yesss!
@jmeas, it's in!

benmccormick:
:+1:  Thanks for the feedback everyone.  I'll toss up a separate PR for the Views/ more documentation sometime later this week.  Is there anywhere else in this repo where I should document that Radio will now be a dependency (instead of wreqr?) Or is that more of a www thing?

jasonLaster:
That's a good question. Man it'd be great to pull in radio docs.

PR #2434, opened 2015-03-19
Propagate before:show to CollectionView children

### LINKED ISSUE #2428: When a CollectionView is shown by a Region, its children's before:render is never fired
Consider the following nested structure:

```js
var ChildView = Marionette.LayoutView({
  regions: {
    grandchild: '.js-grandchild'
  },
  // never called
  onBeforeShow: function () {
    // never shown
    this.showChildView('grandchild', new GrandChildView());
  }
});

var ChildrenView = Marionette.CollectionView({
  childView: ChildView
});

var TopLayout = Marionette.LayoutView.extend({
  regions: {
    children: '.js-children'
  },
  onBeforeShow: function () {
    var childrenView = new ChildrenView();
    this.showChildView('children', childrenView);
  }
});
```

When a CollectionView is rendered by a Region, the Region calls [_view.render()_](https://github.com/marionettejs/backbone.marionette/blob/1b2222e2cf61e70fc2f8a513894b334c59867485/src/region.js#L90) before [triggering _show_](https://github.com/marionettejs/backbone.marionette/blob/1b2222e2cf61e70fc2f8a513894b334c59867485/src/region.js#L134).  A CollectionView's _render()_ method ultimately calls _\_triggerBeforeShowBufferedChildren()_ before returning, which [checks _\_isShown_](https://github.com/marionettejs/backbone.marionette/blob/1b2222e2cf61e70fc2f8a513894b334c59867485/src/collection-view.js#L59).  The problem is that _\_isShown_ is only set when [_show_ is fired](https://github.com/marionettejs/backbone.marionette/blob/1b2222e2cf61e70fc2f8a513894b334c59867485/src/dom-refresh.js#L33), which hasn't yet.  Thus, before:render is never fired on the instances of ChildView.

ianmstew:
In addition to covering the _before:show_ test case, I also added a test for the existing propagation of _show_. Fixes #2428.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -28,6 +28,7 @@ Marionette.CollectionView = Marionette.View.extend({
 
     Marionette.View.apply(this, arguments);
 
+    this.on('before:show', this._onBeforeShowCalled);
     this.on('show', this._onShowCalled);
 
     this.initRenderBuffer();
@@ -110,6 +111,10 @@ Marionette.CollectionView = Marionette.View.extend({
     var view = this.children.findByModel(model);
     this.removeChildView(view);
     this.checkEmpty();
+  },
+
+  _onBeforeShowCalled: function() {
+    this.children.each(_.partial(this._triggerMethodOnChild, 'before:show'));
   },
 
   _onShowCalled: function() {
--- a/test/unit/collection-view.spec.js
+++ b/test/unit/collection-view.spec.js
@@ -1318,4 +1318,46 @@ describe('collection view', function() {
       expect(this.childView.$el).to.contain.$text('bar');
     });
   });
+
+  describe('when a collection view has been rendered but not shown', function() {
+    beforeEach(function() {
+      this.ChildView = Backbone.Marionette.ItemView.extend({
+        onBeforeShow: function() {},
+        onShow: function() {},
+        template: _.template('<%= foo %>')
+      });
+
+      this.CollectionView = Backbone.Marionette.CollectionView.extend({
+        childView: this.ChildView
+      });
+
+      this.sinon.spy(this.ChildView.prototype, 'onBeforeShow');
+      this.sinon.spy(this.ChildView.prototype, 'onShow');
+
+      this.model1 = new Backbone.Model({foo: 1});
+      this.model2 = new Backbone.Model({foo: 2});
+      this.collection = new Backbone.Collection([this.model1, this.model2]);
+      this.collectionView = new this.CollectionView({
+        collection: this.collection
+      });
+      $('body').append(this.collectionView.el);
+
+      this.collectionView.render();
+      this.collectionView.trigger('before:show');
+      this.collectionView.trigger('show');
+
+      this.childView1 = this.collectionView.children.findByIndex(0);
+      this.childView2 = this.collectionView.children.findByIndex(1);
+    });
+
+    it('onBeforeShow should propagate to each child view', function() {
+      expect(this.childView1.onBeforeShow).to.have.been.called;
+      expect(this.childView2.onBeforeShow).to.have.been.called;
+    });
+
+    it('onShow should propagate to each child view', function() {
+      expect(this.childView1.onShow).to.have.been.called;
+      expect(this.childView2.onShow).to.have.been.called;
+    });
+  });
 });
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -55,7 +55,7 @@ var MyView = Marionette.ItemView.extend({
 
 Interaction points, such as tooltips and warning messages, are generic concepts. There is no need to recode them within your Views. They are prime candidates for abstraction into a higher level, non-coupled concept, which is exactly what Behaviors provide you with.
 
-Here is the syntax for declaring which behaviors get used within a View.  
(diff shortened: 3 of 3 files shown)


jasonLaster:
This is awesome. I love seeing you scratch your itch. 

Mind looking into the test fail

ianmstew:
Oh no way!  I didn't run a comprehensive test, but I will in the future.  Odd, because I didn't even commit that file.

jasonLaster:
:+1: looks great.

src/collection-view.js
```diff
@@ -28,6 +28,7 @@ Marionette.CollectionView = Marionette.View.extend({
 
     Marionette.View.apply(this, arguments);
 
+    this.on('before:show', this._onBeforeShowCalled);
```
jasonLaster: going to go on record to say this is a little weird in the `constructor`, but it's consistent with what we're doing with `show`. Maybe we re-evaluate or at a minimum add a comment describing the rationale.

test/unit/collection-view.spec.js
```diff
@@ -1318,4 +1318,46 @@ describe('collection view', function() {
...
+      this.sinon.spy(this.ChildView.prototype, 'onBeforeShow');
+      this.sinon.spy(this.ChildView.prototype, 'onShow');
+
+      this.model1 = new Backbone.Model({foo: 1});
+      this.model2 = new Backbone.Model({foo: 2});
+      this.collection = new Backbone.Collection([this.model1, this.model2]);
+      this.collectionView = new this.CollectionView({
+        collection: this.collection
+      });
+      $('body').append(this.collectionView.el);
```
samccone: we have been trying in the tests to not touch the DOM directly like this perhaps putting it in a detached node might be nicer

test/unit/collection-view.spec.js
```diff
@@ -1318,4 +1318,46 @@ describe('collection view', function() {
...
+      this.collectionView.render();
+      this.collectionView.trigger('before:show');
+      this.collectionView.trigger('show');
+
+      this.childView1 = this.collectionView.children.findByIndex(0);
+      this.childView2 = this.collectionView.children.findByIndex(1);
+    });
+
+    it('onBeforeShow should propagate to each child view', function() {
+      expect(this.childView1.onBeforeShow).to.have.been.called;
```
samccone: let's check the context and the arguments also

samccone: and call count

src/collection-view.js
ianmstew: Ask @samccone?  I think `this.on('show'...)` came from #d1afdcd6e7aa33c6234912d4684760915bc32178.

test/unit/collection-view.spec.js
ianmstew: Ahh, good to know.  I was following precedent from earlier tests in the collection view spec.  I can refactor.  In the case of a refactor like that, should I refactor the earlier tests while I'm at it?

test/unit/collection-view.spec.js
ianmstew: K

test/unit/collection-view.spec.js
samccone: in another commit that is cool, the call count, context, and argument check is something we started doing a while ago, it makes our tests much more explicit and hardened against regressions.

ahumphreys87:
I think a global find/replace for `called` to `calledOnce` and this is :+1:

test/unit/collection-view.spec.js
ianmstew: I like it.

jasonLaster:
@samccone - nice comments. @ianmstew welcome to the Mn review cycle, don't get discouraged, this is the best part!

ianmstew:
:D  An A+ paper is never written on the first pass.  Review and revision is where good becomes excellent.  I'm happy to be in that process!

joelleibow:
I just wanted to add that I just ran into this problem myself when using a CompositeView. I've thrown together a JSFiddle here http://jsfiddle.net/0tpc8vwL/

Since CompositeView is a child of CollectionView it seems as if the PR here will address this but I just wanted to chime in :)

jasonLaster:
Nice @joelleibow 

Mind opening against next

ianmstew:
Re-opening against **next**.

joelleibow:
Apologies for not jumping on this after I was asked to. Been away from a computer and under the weather. Thanks Ian.

ianmstew:
Ahh no problem.  I figured "opening against next" was directed at me so I didn't think anything of it.  I just realized I should be more methodical about the CompositeView case you brought up, so I will at least add tests and probably update #2428 to include the CompositeView case, since they are so closely related.  @joelleibow, thanks for reporting this!

ianmstew:
@jasonLaster In the case of the same issue with CompositeView and CollectionView like we have here, do we add tests for both or just to the CollectionView spec?  I went to investigate the CompositeView spec and noticed it doesn't repeat any child view event lifecycle tests found in CollectionView.

jasonLaster:
Just collectionView probably. Use your judgement though. Not as familiar

ianmstew:
K, thanks Jason.

PR #2435, opened 2015-03-20
Add triggerMethodMany

jasonLaster:
triggerMethodMany invokes triggerMethod on many targets from a source
it's useful for standardizing a pattern where we propogate an event from a source
to many targets.

For each target we want to follow the pattern
target.triggerMethod(event, target, source, ...other args)
e.g childview.triggerMethod('attach', childView, region, ...args)

@ianmstew i'm adding this now so that hopefully `attach` and `_triggerMethodChild` can be standardized

### DIFF
--- a/src/trigger-method.js
+++ b/src/trigger-method.js
@@ -66,3 +66,18 @@ Marionette.triggerMethodOn = function(context) {
 
   return fnc.apply(context, _.rest(arguments));
 };
+
+// triggerMethodMany invokes triggerMethod on many targets from a source
+// it's useful for standardizing a pattern where we propogate an event from a source
+// to many targets.
+//
+// For each target we want to follow the pattern
+// target.triggerMethod(event, target, source, ...other args)
+// e.g childview.triggerMethod('attach', childView, region, ...args)
+Marionette.triggerMethodMany = function(targets, source, eventName) {
+  var args = _.drop(arguments, 3);
+
+  _.each(targets, function(target) {
+    Marionette.triggerMethodOn.apply(target, [target, eventName, target, source].concat(args));
+  });
+};
--- a/docs/marionette.functions.md
+++ b/docs/marionette.functions.md
@@ -189,6 +189,21 @@ Marionette.triggerMethodOn(ctx, "foo", bar);
 // will trigger "foo" on ctx
 ```
 
+## Marionette.triggerMethodMany
+
+Invokes `triggerMethod` on many contexts.
+
+This is useful when you want to trigger an event on many different objects.
+
+```js
+var views = getManyViews();
+var context = this;
+Marionette.triggerMethodMany(views, context, "foo", bar);
+// will call `onFoo: function(view, context, bar){...})` for each view
+// will trigger "foo" on each of the views
+```
+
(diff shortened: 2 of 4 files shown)


ianmstew:
Cool, once this is in we can use it on [CollectionView->_triggerXXXBufferedChildren](https://github.com/marionettejs/backbone.marionette/blob/2d94cdce2c6df84876d9c620b88de4e3da7c4c0b/src/collection-view.js#L58-L70) and [Region->_triggerAttach(views, prefix)(https://github.com/marionettejs/backbone.marionette/blob/2d94cdce2c6df84876d9c620b88de4e3da7c4c0b/src/region.js#L145-L150).

ianmstew:
(Among others)

jasonLaster:
This is why supporting N versions of everything sucks!

now we have to do it.

ianmstew:
:(

jasonLaster:
@samccone, @megawac any good poointers on how to test lodash 3.0 fails in an automated fashion?

megawac:
Locally you mean?

megawac:
You can do this if you want to test locally

```sh
export LODASH=3
curl -L https://rawgit.com/marionettejs/travis-ci-setup/master/ci.sh | bash
npm test
```

src/trigger-method.js
```diff
@@ -66,3 +66,18 @@ Marionette.triggerMethodOn = function(context) {
...
+
+// triggerMethodMany invokes triggerMethod on many targets from a source
+// it's useful for standardizing a pattern where we propogate an event from a source
+// to many targets.
+//
+// For each target we want to follow the pattern
+// target.triggerMethod(event, target, source, ...other args)
+// e.g childview.triggerMethod('attach', childView, region, ...args)
+Marionette.triggerMethodMany = function(targets, source, eventName) {
+  var args = _.rest(arguments, 3);
```
megawac: Use `tail` or `drop` instead of `rest` for lodash 3 support

src/trigger-method.js
```diff
@@ -66,3 +66,18 @@ Marionette.triggerMethodOn = function(context) {
...
+
+  _.each(targets, function(target) {
+    Marionette.triggerMethodOn.apply(target, [target, eventName, target, source].concat(args));
+  });
+};
```
megawac: No `return`?

jasonLaster: Which gets the returns? My thinking is establish the pattern when we have more of a need

src/trigger-method.js
jasonLaster: Thanks

jasonLaster:
:eyes: please
@marionettejs/marionette-core

src/trigger-method.js
paulfalgout: yikes.. `Mn` uses `rest` in quite a few places..  twice in `trigger-method.js`

paulfalgout:
LGTM,  simple enough.  needs tests and docs.

jasonLaster:
Added tests and docs!

ahumphreys87:
:+1:

Issue #2209, opened 2015-01-18
Trigger onAttach on child views during CollectionView re-render


MeoMix:
The current implementation of onAttach does not trigger when an ItemView is attached to the DOM during re-rendering of a CollectionView. Although no region is being used I believe it makes sense to trigger onAttach in a fashion similar to how onShow is triggered.

jmeas:
:+1: Makes sense to me.

My implementation of `onAttach` is *closely* linked to the `onShow` event. I (incorrectly) assumed that only regions trigger the `onShow` event. But Collection Views trigger it on their children, as well!

bazineta:
Further to the discussion here, a rehash of Gitter discussion on this topic.

I have a view hierarchy LayoutView -> CollectionView -> LayoutView -> ItemView.  The final ItemView contains a div that hosts a Highcharts chart; for the chart to render properly, the width metric must be valid when the ItemView is shown.

To date, I've used onShow() in both LayoutViews to instantiate the CollectionView and ItemView, respectively, and onShow() in the ItemView to display the Highcharts chart.  This works correctly, and the width in the ItemView is valid in its onShow().

Attempting to use the new onAttach() function to perform the same function works, sort of.  If the chart is displayed in onAttach() and the LayoutViews instantiate their children in onRender() instead of onShow(),  the ItemView ends up with a valid width in onAttach() on first render.  Using this setup, it's got an invalid width in onShow().

However, if the collection is reset, onAttach() won't be called, so the effect is that the ItemViews are destroyed and recreated, but the charts don't get drawn, as onAttach() is never called.

musterknabe:
:+1:  Bumped into this aswell

jasonLaster:
Thanks for the bump @musterknabe. @ianmstew want to become our collection view events :older_man:? Half joking, no pressure

ianmstew:
Ahhh, I'll look into it... might as well while I'm down there :)

ahumphreys87:
Fixed with #2481

PR #2441, opened 2015-03-21
onBeforeAttach and onAttach for CollectionView child views

ianmstew:
This pull request satisfies the expectation a CollectionView should trigger _onBeforeAttach_ and _onAttach_ on child views it attaches to the DOM, e.g, when calling `collectionView.render()`, in the case where the CollectionView itself is already shown.  Prior to this, as documented in #2209, _onBeforeAttach_ and _onAttach_ were **only** triggered on child views by `Region#show()` during  `region.show(collectionView)`.

Tests cover the new case as well as backfill tests for the case of `region.show(collectionView)` that was already working.

Discussion remains for the issue of "opting out" of onBeforeAttach and onAttach for newly created child views by an already shown CollectionView.  Ideally, we would have a parallel opt-out to `Region#show()`.  However, when a CollectionView is creating new views outside of a `region.show()`, there is no exact parallel to `Region#show()`'s _showOptions_.  This discussion does not necessarily need to block merging, because this option is not a direct expectation (is not yet documented).

Note: complexity and maxstatements had to be increased to accommodate `addEmptyView()` and `_addChildView()`, for much the same reason that `Region#show()` requires higher complexity and maxstatements.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -80,11 +80,11 @@ Marionette.CollectionView = Marionette.View.extend({
   },
 
   _triggerMethodOnAll: function(views, event) {
-    _.each(views, _.partial(this._triggerMethodOn, _, event));
+    _.each(views, _.partial(this._triggerMethodOn, event));
   },
 
   // Internal method for _.each loops to call `Marionette.triggerMethodOn` on a view
-  _triggerMethodOn: function(view, event) {
+  _triggerMethodOn: function(event, view) {
     Marionette.triggerMethodOn(view, event);
   },
 
@@ -126,7 +126,7 @@ Marionette.CollectionView = Marionette.View.extend({
   },
 
   _onShowCalled: function() {
-    this.children.each(_.partial(this._triggerMethodOn, _, 'show'));
+    this.children.each(_.partial(this._triggerMethodOn, 'show'));
   },
 
   // Render children views. Override this method to
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -1,4 +1,4 @@
-/* jshint maxstatements: 14 */
+/* jshint maxcomplexity: 8, maxstatements: 20 */
 
 // Collection View
 // ---------------
@@ -14,6 +14,11 @@ Marionette.CollectionView = Marionette.View.extend({
   // flag for maintaining the sorted order of the collection
   sort: true,
 
+  // TODO: Discuss how to handle option to disable attach/beforeAttach on child view trees.
+  // These properties are a placeholder for now.
+  triggerBeforeAttach: true,
+  triggerAttach: true,
+
   // constructor
   // option to pass `{sort: false}` to prevent the `CollectionView` from
   // maintaining the sorted order of the collection.
@@ -22,7 +27,6 @@ Marionette.CollectionView = Marionette.View.extend({
   // option to pass `{comparator: compFunction()}` to allow the `CollectionView`
   // to use a custom sort order for the collection.
   constructor: function(options) {
-
     this.once('render', this._initialEvents);
     this._initChildViewStorage();
 
@@ -46,33 +50,42 @@ Marionette.CollectionView = Marionette.View.extend({
   },
 
   endBuffering: function() {
+    // Only trigger attach if already shown and attached, otherwise Region#show() handles this.
+    var canTriggerAttach = this._isShown && Marionette.isNodeAttached(this.el);
+    var nestedViews;
+
     this.isBuffering = false;
-    this._triggerBeforeShowBufferedChildren();
+
+    if (this._isShown) {
+      this._triggerMethodOnAll(this._bufferedChildren, 'before:show');
+    }
+    if (canTriggerAttach && this.triggerBeforeAttach) {
+      // Recursively retrieve nested views.
+      nestedViews = this._getNestedViews();
+      this._triggerMethodOnAll(nestedViews, 'before:attach');
+    }
 
     this.attachBuffer(this);
 
-    this._triggerShowBufferedChildren();
+    if (canTriggerAttach && this.triggerAttach) {
+      // Don't recursively retrieve nested views unnecessarily.
+      nestedViews = nestedViews || this._getNestedViews();
+      this._triggerMethodOnAll(nestedViews, 'attach');
+    }
+    if (this._isShown) {
+      this._triggerMethodOnAll(this._bufferedChildren, 'show');
+    }
+
     this.initRenderBuffer();
   },
 
-  _triggerBeforeShowBufferedChildren: function() {
-    if (this._isShown) {
-      _.each(this._bufferedChildren, _.partial(this._triggerMethodOnChild, 'before:show'));
-    }
+  _triggerMethodOnAll: function(views, event) {
+    _.each(views, _.partial(this._triggerMethodOn, _, event));
   },
 
-  _triggerShowBufferedChildren: function() {
-    if (this._isShown) {
-      _.each(this._bufferedChildren, _.partial(this._triggerMethodOnChild, 'show'));
-
-      this.initRenderBuffer();
-    }
-  },
-
-  // Internal method for _.each loops to call `Marionette.triggerMethodOn` on
-  // a child view
-  _triggerMethodOnChild: function(event, childView) {
-    Marionette.triggerMethodOn(childView, event);
+  // Internal method for _.each loops to call `Marionette.triggerMethodOn` on a view
+  _triggerMethodOn: function(view, event) {
+    Marionette.triggerMethodOn(view, event);
   },
 
   // Configured the initial events that the collection view
@@ -113,7 +126,7 @@ Marionette.CollectionView = Marionette.View.extend({
   },
 
   _onShowCalled: function() {
-    this.children.each(_.partial(this._triggerMethodOnChild, 'show'));
+    this.children.each(_.partial(this._triggerMethodOn, _, 'show'));
   },
 
   // Render children views. Override this method to
@@ -294,6 +307,9 @@ Marionette.CollectionView = Marionette.View.extend({
   // but "add:child" events are not fired, and the event from
   // emptyView are not forwarded
   addEmptyView: function(child, EmptyView) {
+    // Only trigger attach if already shown and attached, otherwise Region#show() handles this.
+    var canTriggerAttach = this._isShown && Marionette.isNodeAttached(this.el);
+    var nestedViews;
 
     // get the emptyViewOptions, falling back to childViewOptions
     var emptyViewOptions = this.getOption('emptyViewOptions') ||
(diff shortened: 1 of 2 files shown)


ianmstew:
Ah, I was afraid of that.  I will revert a too-smart feature of `_.partial()`--using `_` as an argument placeholder.

ianmstew:
Hmm, I'll have to see why coverage decreased.  I did have to introduce some new conditionals, but I only added bdd tests, not line-by-line unit tests.  I would love to have your all's opinion on this.  @jasonLaster @samccone

src/collection-view.js
```diff
@@ -46,33 +50,42 @@ Marionette.CollectionView = Marionette.View.extend({
...
+      this._triggerMethodOnAll(nestedViews, 'before:attach');
+    }
 
-    this._triggerShowBufferedChildren();
-    this.initRenderBuffer();
-  },
+    this.attachBuffer(this);
 
-  _triggerBeforeShowBufferedChildren: function() {
+    if (canTriggerAttach && this.triggerAttach) {
```
jasonLaster: Do we need to check to see if it's attached
```js
 var attachedRegion = Marionette.isNodeAttached(this.el);
```

src/collection-view.js
```diff
@@ -46,33 +50,42 @@ Marionette.CollectionView = Marionette.View.extend({
...
+  _triggerMethodOnAll: function(views, event) {
+    _.each(views, _.partial(this._triggerMethodOn, event));
   },
 
-  // Internal method for _.each loops to call `Marionette.triggerMethodOn` on
-  // a child view
-  _triggerMethodOnChild: function(event, childView) {
-    Marionette.triggerMethodOn(childView, event);
+  // Internal method for _.each loops to call `Marionette.triggerMethodOn` on a view
+  _triggerMethodOn: function(event, view) {
```
jasonLaster: I don' think we need this

ianmstew: I was wondering about that.  Delegate straight to Marionette.triggerMethodOn, then?  Or you mean leave `_triggerMethodOnChild` alone?

src/collection-view.js
ianmstew: I'm checking that on line 54, because it is possible be "shown" and not attached.  If it's _not_ shown it's certainly not attached, so I'm skipping the `isNodeAttached` check in that case.

src/collection-view.js
jasonLaster: yea just use this Marionette.triggerMethodOn

src/collection-view.js
```diff
@@ -46,33 +50,42 @@ Marionette.CollectionView = Marionette.View.extend({
...
   endBuffering: function() {
+    // Only trigger attach if already shown and attached, otherwise Region#show() handles this.
+    var canTriggerAttach = this._isShown && Marionette.isNodeAttached(this.el);
+    var nestedViews;
+
     this.isBuffering = false;
-    this._triggerBeforeShowBufferedChildren();
 
-    this.attachBuffer(this);
+    if (this._isShown) {
```
jasonLaster: Dumb question: why do we check this?

src/collection-view.js
```diff
@@ -321,9 +337,19 @@ Marionette.CollectionView = Marionette.View.extend({
     // remove and/or close it later
     this.children.add(view);
 
+    if (canTriggerAttach && this.triggerBeforeAttach) {
+      nestedViews = [view].concat(view._getNestedViews());
+      // Trigger before:attach after render to honor Region#show() lifecycle order.
+      view.once('render', _.partial(this._triggerMethodOnAll, nestedViews, 'before:attach'), this);
```
jasonLaster: why `once` render?

could we  put this in a helper so that a lot of this duplicative code is shared?

src/collection-view.js
ianmstew: k

jasonLaster:
Looks good, added some questions because this is complicated stuff

test/unit/on-attach.spec.js
```diff
@@ -749,4 +771,64 @@ describe('onAttach', function() {
       });
     });
   });
+
+  describe('when showing an empty CollectionView', function() {
```
jasonLaster: Could we add some tests where the `CollectionView` is also attached to a region that's atached to the document?

src/collection-view.js
ianmstew: So, I put a lot of thought/research into reasoning this one out.  It seems like the previous code was written such that a collection view does _not_ behave like a region (firing lifecycle events on child views) _unless_ it's situated within a Region, for one.  For two, if a view lifecycle event is the Region's responsibility (`attach`, `before:show`, `show) and we _don't_ have that check, view lifecycle events could be triggered twice on child views.  One question to table is that perhaps a collection view _should_ behave like a region even when it's not within a region, but then we would have to solve the double-triggered child view events.

src/collection-view.js
ianmstew: Ahh, I thought about refactoring addEmptyView and _addChildView to eliminate duplicated code, but decided not to take that on for pull request size and without more research (why are they duplicated in the first place).  I think that would be a good idea.

I'm glad you asked about `once` render.  The very next line is `this.renderChildView`:

```js
    view.render();
    this.attachHtml(this, view, index);
    return view;
  },
```

`renderChildView` is part of the public API and a suggested override for custom child view rendering.  I noticed a pattern within collection-view.js that override-friendly public API methods do not trigger lifecycle events, which seemed reasonable because then the developer would be responsible for lifecycle.  The issue here that _before:attach_ comes after _render_ but before the html is attached to the DOM.  The only other way to code this would be:

```js
    view.render();
    Marionette.triggerMethodOn(view, 'before:attach');
    this.attachHtml(this, view, index);
    return view;
  },
```

Triggering _before:attach_ seemed like too much responsibility for the overridable public method in my gut, but also based on the pattern I noticed elsewhere.

test/unit/on-attach.spec.js
ianmstew: I think we are doing that with `this.region.show(this.collectionView)`, because `this.region` is [attached to a fixture](https://github.com/ianmstew/backbone.marionette/blob/collection-view-attach/test/unit/on-attach.spec.js#L6-L9).

ianmstew:
@jasonLaster Agree this is complicated, definitely more than I thought going into it.  It's like we have a collection-view behaving like a Region, _sometimes_.  I had to really understand Region#show.

What docs do you think?  The issues triggering this PR are based on unexpected behavior so I assumed this is more of a bugfix than a feature add requiring docs.  Unless we just want to be perfectly explicit in our docs that collection view behaves like a region while shown in a region.  Thoughts?

src/collection-view.js
```diff
@@ -14,6 +14,11 @@ Marionette.CollectionView = Marionette.View.extend({
   // flag for maintaining the sorted order of the collection
   sort: true,
 
+  // TODO: Discuss how to handle option to disable attach/beforeAttach on child view trees.
+  // These properties are a placeholder for now.
```
jasonLaster: What concerns do you have? I'm thinking these flags work like a regions flags and should be the same...

src/collection-view.js
```diff
@@ -321,9 +337,19 @@ Marionette.CollectionView = Marionette.View.extend({
     // remove and/or close it later
     this.children.add(view);
 
+    if (canTriggerAttach && this.triggerBeforeAttach) {
+      nestedViews = [view].concat(view._getNestedViews());
+      // Trigger before:attach after render to honor Region#show() lifecycle order.
```
jasonLaster: Are there nested views when it's empty?

src/collection-view.js
jasonLaster: I'm fine adding life cycle events to the public API. I hear your concern that they could lose them, but I think that's okay if you're customizing. Beats unnecessary events that are tough to reason about imo

test/unit/on-attach.spec.js
jasonLaster: Got it. Cool

src/collection-view.js
jasonLaster: Yes you are. Nice

src/collection-view.js
jasonLaster: Mind explaining the double event fire?

My instinct is for the CV to be region like. Also, favor consistency.

jasonLaster:
Just did another pass. This is big, but I really like it. 

Andrew, Christian.  Let's get eyes here. Sam you too :)


Mind opening against next?

src/collection-view.js
ianmstew: During the show cycle of a CollectionView being shown in a Region, the following occurs in sequence:
1. collectionView.render()
  * child `before:render`
  * child `render`
1. trigger collectionView `before:show`
  * childView `before:show`
1. trigger on collectionView and descendants `before:attach` if Region is attached to DOM
  * childView `before:attach`
1. append collectionView html to region
1. trigger on collectionView and descendants `attach` if Region is attached to DOM
  * childView `attach`
1. trigger collectionView trigger `show`
  * childView `show`
1. _isShown is set to `true` [here](https://github.com/marionettejs/backbone.marionette/blob/08bfcbea6000901d8a268ad572883090c0c85868/src/dom-refresh.js#L13)

It is expected that a CollectionView that is already rendered will fire `before:show`/`show` and `before:attach`/`attach` on its children as they are added to a CollectionView that is already showing, or the same CollectionView is re-rendered manually.  It is also expected that a Region will instigate these events when showing a CollectionView, and in the attach case the Region gains a hard reference to every descendant view.  Without the _isShown check, these two cases have an overlap: during a Region's show cycle, both the CollectionView and Region would be responsible for calling attach/show events.

src/collection-view.js
ianmstew: I see your point about the lifecycle event in a public API method, although I'm not _perfectly_ comfortable with it.  I guess it's a lesser of two evils, eh?  But I _am_ concerned that adding not only a lifecycle event, but a lifecycle event contingent on a private variable, could throw some people for a loop--especially since changing the private variable would mean legitimate overrides could break.

src/collection-view.js
ianmstew: Well, we can reproduce half of [this line in Region#show](https://github.com/ianmstew/backbone.marionette/blob/9907c2982fa57138d1a29596cd8dc41925ff32b2/src/region.js#L113) with a class option, but there isn't a direct equivalent to `showOptions` in any CollectionView method.  We might have to either not implement it, or add it to `render()`.  This goes back to my comment earlier that refactoring a CollectionView to be more like a Region might be beneficial, otherwise we will have to compromise on slight inconsistencies like this.

src/collection-view.js
ianmstew: Good question!  I didn't think about the default return value for _getNestedView().  It appears [we can rely on an empty array in the empty case, however](https://github.com/ianmstew/backbone.marionette/blob/9907c2982fa57138d1a29596cd8dc41925ff32b2/src/view.js#L321-L337).

ianmstew: On second thought, I might as well add a test case for this.

ianmstew:
Sure thing, my next submission will be against next.

src/collection-view.js
ahumphreys87: Nice reasoning, I'd very much like to see your thoughts on `CollectionView` for v3

ahumphreys87:
Yup I like this too. Could you take a look at the test coverage? https://coveralls.io/builds/2165784 should give details of what isn't covered. I think your right, it looks like the conditionals

src/collection-view.js
ianmstew: Thanks @ahumphreys87, I will try to put those together :)

ianmstew:
I'll take a look!  Thanks for the link.  I would definitely like to keep our 100%.

jasonLaster:
Hey @ianmstew where's this at? We're at 19 days and at a month I start to worry :frowning: 

seriously, I think this is really good work and I don't want to see it go stale

ianmstew:
Thanks for the nudge!  I actually just rebased onto next and I'm running down those coverage deficits...  You'll see a new PR shortly :)

paulfalgout:
:clap:

src/collection-view.js
ianmstew: Hey @jasonLaster, final call on this?  I'm worried about putting a bunch of logic in renderChildView, which right now is so straightforward.  Not to mention logic based on a private variable.  From that point of view I think using a "once render" is not as messy.

PR #2443, opened 2015-03-21
Add missing files/tests and remove unused/non-existent tests from specRunner on major

benmccormick:
While I was working on the Radio PR, I noticed that specRunner hadn't been kept up to date in terms of tests and files used in the major branch.  So this just gets it up to date with the various changes

- removed src/spec files for deprecated code like controllers and layoutView
- added in files for new/renamed stuff like AbstractView and View
- alphabetized the spec files so its easier to see whats missing.


jasonLaster:
Thanks. Looks fine to me

samccone:
:+1:  thx

PR #2444, opened 2015-03-22
Fix module definition when array is provided as additional argument

kestutis-stalmokas:
I guess this is not intended behaviour:

if array is provided as additional argument to module definition it gets flattened into separate arguments.

    MyApp.module("MyModule", function(MyModule, MyApp, Backbone, Marionette, $, _, arg1, arg2){
        // arg1 === 1;
        // arg2 === 2;
    }, [1, 2]);

### DIFF
--- a/src/module.js
+++ b/src/module.js
@@ -117,7 +117,7 @@ _.extend(Marionette.Module.prototype, Backbone.Events, {
       Marionette,
       Backbone.$, _,
       customArgs
-    ]);
+    ], true);
 
     definition.apply(this, args);
   },
--- a/src/module.js
+++ b/src/module.js
@@ -114,7 +114,7 @@ _.extend(Marionette.Module.prototype, Backbone.Events, {
     // make flattening consistent
     // in lodash (v3) it's flatten(collection, deep)
     // and in underscore it's flatten(collection, shallow)
-    var deepFlag = !!_.flattenDeep;
+    var deepFlag = !_.flattenDeep;
     var args = _.flatten([
       this,
       this.app,
--- a/src/module.js
+++ b/src/module.js
@@ -110,6 +110,11 @@ _.extend(Marionette.Module.prototype, Backbone.Events, {
     if (!definition) { return; }
 
     // build the correct list of arguments for the module definition
+
+    // make flattening consistent
+    // in lodash (v3) it's flatten(collection, deep)
+    // and in underscore it's flatten(collection, shallow)
+    var deepFlag = !!_.flattenDeep;
     var args = _.flatten([
       this,
       this.app,
@@ -117,7 +122,7 @@ _.extend(Marionette.Module.prototype, Backbone.Events, {
       Marionette,
       Backbone.$, _,
       customArgs
-    ], true);
+    ], deepFlag);
 
     definition.apply(this, args);
   },
--- a/test/unit/module.spec.js
+++ b/test/unit/module.spec.js
@@ -25,7 +25,7 @@ describe('application modules', function() {
 
       describe('and the define function is provided', function() {
         beforeEach(function() {
-          this.additionalParam = [1, 2];
+          this.additionalParam = [[1], 2];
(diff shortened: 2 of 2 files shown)


jasonLaster:
Nice idea @kestutis-stalmokas looks like there's a lodash inconsistency https://travis-ci.org/marionettejs/backbone.marionette/jobs/55339975

@megawac - eyes please.

@samccone - what do you think about additonal args for Modules? I never do this?

src/module.js
```diff
@@ -117,7 +117,7 @@ _.extend(Marionette.Module.prototype, Backbone.Events, {
       Marionette,
       Backbone.$, _,
       customArgs
-    ]);
+    ], true);
```
megawac: `lodash@3` has the opposite behaviour of `underscore` and `lodash@2` for this flag

test/unit/module.spec.js
```diff
@@ -25,7 +25,7 @@ describe('application modules', function() {
 
       describe('and the define function is provided', function() {
         beforeEach(function() {
-          this.additionalParam = {};
+          this.additionalParam = [1, 2];
```
megawac: can you make these params `[[1], 2]` or something

src/module.js
jasonLaster: Mind elaborating? what do you need to do here to make it pass?

megawac: It's tricky because in lodash its `flatten(collection, deep)` and in underscore its `flatten(collection, shallow)`

jasonLaster: funny - do we do a lodash check anywher?

kestutis-stalmokas: I can think of only two options:
 1. test for lodash `_.name === 'lodash'`
 2. use `_.noConflict()` and override `flatten`

Second option seems to be more future proof as there might be more inconsistencies in the future.

What are your thoughts?

megawac: You can always do `[].concat.apply([], [[
   <stuff to include>
]])`

jdalton: > kestutis-stalmokas  wrote: test for `lodash _.name === 'lodash'`

totally don't do the `.name` check as that will be borked on minified code.

> megawac: It's tricky because in lodash its `flatten(collection, deep)` and in underscore its `flatten(collection, shallow)`

Yap you can do `_.flatten(collection, !!_.flattenDeep)`. In Underscore it will be `false` so `_.flatten` will be deep and in lodash it will be `true` so also a deep flatten.

test/unit/module.spec.js
```diff
@@ -25,7 +25,7 @@ describe('application modules', function() {
 
       describe('and the define function is provided', function() {
         beforeEach(function() {
-          this.additionalParam = {};
+          this.additionalParam = [[1], 2];
           this.module = this.app.module('Mod', this.defineSpy, this.additionalParam);
         });
```
ahumphreys87: what happens if you do want to pass multiple additional params - I can imagine some people might be using this. Can we add a test to see if it passes?

src/module.js
jasonLaster: Thanks @megawac & @jdalton

ahumphreys87:
@kestutis-stalmokas any update on this one, just 1 question from me. This is good idea so keen to get this one in

jasonLaster:
@kestutis-stalmokas this looks nice. 

Would you mind filing this bug against https://github.com/marionettejs/marionette.module? Module is a deprecated class in `major`, so we won't be accepting any changes here.

megawac:
> Would you mind filing this bug against https://github.com/marionettejs/marionette.module? Module is a deprecated class in major, so we won't be accepting any changes here.

Awesome!

kestutis-stalmokas:
No problem @jasonLaster. Do you need any additional changes to my PR before I send it there?

jasonLaster:
@ahumphreys87 had a question up top. Thanks!

kestutis-stalmokas:
Issue has been moved to https://github.com/marionettejs/marionette.module/issues/2

Thanks everyone for helping me out.

Issue #2437, opened 2015-03-20
CollectionView.emptyView not shown when filtered collection is empty


jackocnr:
When **not** using CollectionView.filter, if you remove the last model from the collection, the CollectionView will automatically display the emptyView. When you **are** using CollectionView.filter however, if you remove the last model that matches the filter, it doesn't display the emptyView, presumably because the pre-filtered collection is not empty. This leaves you with a broken empty state (no childViews are rendered, but no emptyView is rendered either).

Demo here: http://codepen.io/jackocnr/pen/wBRyrE?editors=101

Workaround:
```js
onRemoveChild: function() {
if (!this.children.length) { this.showEmptyView(); }
}
```

cmaher:
It looks like `isEmpty` needs to be updated to account for the filter: https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L462

```js
if (!this.collection) { return true; }
var models = this.collection.filter(function(model, index) {
        return this._shouldAddChild(model, index);
      }, this);
return models.length === 0;
```

cmaher:
resolved by #2437

PR #2445, opened 2015-03-23
Support empty view on removal from filtered CollectionView

cmaher:
As per #2437, currently, the CollectionView doesn't consider the filter when checking for emptiness.

In order to prevent two complete filter operations when rendering (once for checking emptiness, and once when displaying models), this implementation has an option to skip checking based on filtering (see comments).

I wouldn't say this PR is complete, but I wanted to get the rest of the team's opinion, since I can multiple less-than-ideal solutions to this problem.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -146,8 +146,10 @@ Marionette.CollectionView = Marionette.View.extend({
       this.render();
     } else {
       // get the DOM nodes in the same order as the models
-      var els = _.map(models, function(model) {
-        return children.findByModel(model).el;
+      var els = _.map(models, function(model, index) {
+        var view = children.findByModel(model);
+        view._index = index;
+        return view.el;
       });
 
       // since append moves elements that are already in the DOM,
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -204,36 +204,32 @@ Marionette.CollectionView = Marionette.View.extend({
     this.destroyEmptyView();
     this.destroyChildren({checkEmpty: false});
 
-    if (this.isEmpty(this.collection)) {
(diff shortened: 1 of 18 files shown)


src/collection-view.js
```diff
@@ -457,9 +460,13 @@ Marionette.CollectionView = Marionette.View.extend({
     return view;
   },
 
-  // check if the collection is empty
-  isEmpty: function() {
-    return !this.collection || this.collection.length === 0;
+  // check if the collection is empty, taking the filter into account unless 'skipFilter' is true
+  isEmpty: function(collection, options) {
```
cmaher: Alternatives to the given approach

1. Change this method to reference the collection that's passed in, and have  `_renderChildren` pass in a collection with the filtered models
2. Pass in an arbitrary array of models  as an option instead of `skipFilter`
3. Only consider the filter in `checkEmpty`

I'm thinking that option 1 might be the best, but I'm also concerned about how it might change the contract, if at all.

jasonLaster:
@cmaher I'll review this tomorrow, but the tests look good and the src makes sense (hard to believe right). Ian and Andrew are the ppl with context now.

src/collection-view.js
ahumphreys87: how about a variant of option 1...

Only reference the collection if its passed in else  use `this.collection`. I think this would maintain BC?

cmaher: I'm pretty sure we pass in `this.collection` everywhere in CollectionView.  But I'll switch it up to use option 1.

jasonLaster:
Mind reopening on next

cmaher:
closing in favor of #2458

PR #2447, opened 2015-03-24
Remove unneeded files from bower package

JustBlackBird:
`.jscsrc` file is useless within the bower package, thus it should be removed from there.


samccone:
it is indeed, good catch @JustBlackBird

samccone:
:+1:

PR #2450, opened 2015-03-25
Refactor serializeData to no longer use toJSON.

### LINKED ISSUE #1476: Stop using toJSON for rendering templates

jasonLaster:
Cherry-picks https://github.com/marionettejs/backbone.marionette/commit/f14b7eb55f7b1bd28243c31e8ec9f74da58aad28 onto next.

Here's what my setup looks like 

I'm sorry in advance if this proecess kinda sucks and you've got to squint real hard to double check everything. 

On a meta level, I could just merge these cherry-picks into next w/o a review, but honestly w/ big commits like this one i feel better sharing it.

---
toJSON should only be used for preparing data to be sent to the
server. In addition, serializeData no longer accepts arguments,
distinguishing it further from templateHelpers.

Resolves #1476

### DIFF
--- a/api/item-view.yaml
+++ b/api/item-view.yaml
@@ -40,9 +40,14 @@ constructor:
 functions:
   serializeData:
     description: |
-      Serialize the model or collection for the view. If a model is found, the view's `serializeModel` is called. If a collection is found, each model in the collection is serialized by calling the view's `serializeCollection` and putting into an `items` array in the resulting data. If both are found, the model is used.
+      Serialize the view's model or collection for rendering the view's template. If a model is found, the view's `serializeModel` is called. If a
+      collection is found, `serializeCollection` will be called. The collection will be available in your template as `items`. If both a model and a
+      collection are found, the model will be used.
 
-      You can override the `serializeData` method in your own view definition, to provide custom serialization for your view's data. These serializations are then passed into the template function if one is set.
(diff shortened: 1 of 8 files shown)


ahumphreys87:
lgtm :+1:

jasonLaster:
Merged it because why not.... Feel free to ping if you see anything weird

Issue #2297, opened 2015-02-05
Mn.deprecate is broken


thejameskyle:
https://github.com/marionettejs/backbone-metal/issues/9

paulfalgout:
resolved

cmaher:
yep, resolved by #2459

PR #2459, opened 2015-03-27
(#2297) Apply console context to deprecate warnings

cmaher:
To ensure console compatibility

Had to split out console so I could stub it and test it.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -145,5 +145,9 @@ var deprecate = Marionette.deprecate = function(message, test) {
   }
 };
 
-deprecate._warn = typeof console !== 'undefined' && (console.warn || console.log) || function() {};
+deprecate._console = typeof console !== 'undefined' ? console : {};
+deprecate._warn = function() {
+  var warn = deprecate._console.warn || deprecate._console.log || function() {};
+  return warn.apply(deprecate._console, arguments);
+};
 deprecate._cache = {};
--- a/test/unit/helpers.spec.js
+++ b/test/unit/helpers.spec.js
@@ -61,8 +61,24 @@ describe('normalizeUIKeys', function() {
 
 describe('Marionette.deprecate', function() {
   beforeEach(function() {
-    this.sinon.stub(Marionette.deprecate, '_warn');
+    this.sinon.spy(Marionette.deprecate, '_warn');
+    this.sinon.stub(Marionette.deprecate, '_console', {
+      warn: this.sinon.stub()
+    });
     Marionette.deprecate._cache = {};
+  });
+
+  describe('Marionette.deprecate._warn', function() {
+    beforeEach(function() {
+      Marionette.deprecate._warn('foo');
+    });
+
+    it('should `console.warn` the message', function() {
+      expect(Marionette.deprecate._console.warn)
+        .to.have.been.calledOnce
+        .and.calledOn(Marionette.deprecate._console)
+        .and.calledWith('foo');
+    });
   });
 
   describe('when calling with a message', function() {


paulfalgout:
@cmaher jscs error in the spec.

Otherwise :+1:

src/helpers.js
```diff
@@ -145,5 +145,9 @@ var deprecate = Marionette.deprecate = function(message, test) {
   }
 };
 
-deprecate._warn = typeof console !== 'undefined' && (console.warn || console.log) || function() {};
+deprecate._console = typeof console !== 'undefined' ? console : {};
+deprecate._console.warn = deprecate._console.warn || deprecate._console.log || function() {};
```
jridgewell: You're overwriting `console.warn` here. Not horrible, but it doesn't seem like the best solution.

cmaher: Only setting it if it's not a thing... but it's still not great to write to a global.

jasonLaster: would something like this do the trick?

```js
deprecate._warn = function() {
  var fnc = (console && console.warn || console.warn) || function() {};
  fnc.apply(console || window, arguments);
}
```

cmaher: that's what I'm thinking of, but still using `deprecate._console`.  The indirection is really just there for testing purposes

cmaher:
something seems wrong with travis and jscs

jasonLaster:
indentation lol Expected indentation of 8 characters at test/unit/collection-view.spec.js :

src/helpers.js
jridgewell: What's wrong with just binding to `console`? I don't think it's really necessary to test that `deprecate._warn` calls `console.warn`, only that `deprecate._warn` is called.


```javascript
deprecate._warn = typeof console !== 'undefined' ? _.bind(console.warn || console.log, console) : function() {};
```

cmaher: Because `deprecate._warn` is called on `deprecate`, and you can't test that `console` is bound.

jridgewell: But this is a line that should never change? I see the benefit of mocking as 1% and the cost to support as 99%.

cmaher: Yep, there's just that bad feeling about a bugfix without a test.

cmaher:
updated to not modify console object and for jscs

test/unit/collection-view.spec.js
```diff
@@ -310,7 +311,8 @@ describe('collection view', function() {
           foo: 'biz'
         }, {
           foo: 'baz'
-      }]);
+        }
+      ]);
```
jasonLaster: Fucking jscs

test/unit/helpers.spec.js
```diff
@@ -61,10 +61,26 @@ describe('normalizeUIKeys', function() {
 
 describe('Marionette.deprecate', function() {
   beforeEach(function() {
-    this.sinon.stub(Marionette.deprecate, '_warn');
+    this.sinon.spy(Marionette.deprecate, '_warn');
+    this.sinon.stub(Marionette.deprecate, '_console', {
+      warn: this.sinon.stub()
+    });
```
jasonLaster: Nice dude!

jasonLaster:
:thumbsup:

jasonLaster:
@marionettejs/marionette-core needs another thumb

ahumphreys87:
:+1:

ahumphreys87:
manually merged

megawac:
Shame this didn't make it into 2.4.3 or 2.4.4

samccone:
it can make it into the next release right?

paulfalgout:
It's in for v3.  I assume this was merged in the days where we thought v3 was sooner than it has been.  Currently there is no `2.4.5`

megawac:
I don't understand why we wouldn't do very simple maintenance (such as
bumping backbone versions) on old major versions. No need to force users to
upgrade so aggressively

paulfalgout:
I don't disagree.

paulfalgout:
:cherries: 'd on to `patch`

PR #2460, opened 2015-03-29
Fix code style error that is failing build.

scottaj:
Just started working tonight and this is stopping me from building. Also saw it failing on travis.


jasonLaster:
Thanks!

PR #2461, opened 2015-03-29
Establish existing render/show event order.

scottaj:


scottaj:
See discussion on #2355.

test/unit/region.spec.js
```diff
@@ -247,6 +246,22 @@ describe('region', function() {
...
 
+    describe('Render and show event ordering', function() {
+      beforeEach(function() {
+        this.viewEvents = _.map(this.triggerSpy.getCalls(), function(call) {
+          return call.args[0];
+        });
+      });
+
+      it('triggers the before render event before the before show event', function() {
+        expect(_.indexOf(this.viewEvents, 'before:render')).to.be.below(_.indexOf(this.viewEvents, 'before:show'));
```
jasonLaster: There's a spy calledBefore assertion that could be coo here

scottaj: I'll give it a shot. I tried that when I was first writing these and was having trouble getting it to work properly.

jasonLaster: yup. we've got some examples in the codebase
https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/collection-view.spec.js#L171

ahumphreys87:
@scottaj hows it going on this one? Would be great to push this through if you get time

jasonLaster:
Ping.

scottaj:
Sorry, been really busy and lost track of this. I'll try and take a look at it tonight.

scottaj:
@ahumphreys87 @jasonLaster - Updated

test/unit/region.spec.js
```diff
@@ -147,9 +146,11 @@ describe('region', function() {
       this.regionBeforeEmptySpy = this.sinon.spy();
 
       this.view = new this.MyView();
+      this.viewOnBeforeRenderSpy = this.sinon.spy(this.view, 'onBeforeRender');
       this.viewRenderSpy = this.sinon.spy(this.view, 'render');
       this.viewOnBeforeShowSpy = this.sinon.spy(this.view, 'onBeforeShow');
       this.viewOnShowSpy = this.sinon.spy(this.view, 'onShow');
+      this.triggerSpy = this.sinon.spy(this.view, 'trigger');
```
jasonLaster: do we need this?

jasonLaster:
one comment, otherwise good

test/unit/region.spec.js
scottaj: Nope, it's a leftover.

scottaj:
Updated.

ahumphreys87:
:+1:  @marionettejs/marionette-core needs :eyes:

test/unit/region.spec.js
```diff
@@ -122,18 +122,17 @@ describe('region', function() {
         onSwapOut: function() {}
       });
 
-      this.MyView = Backbone.View.extend({
+      this.MyView = Marionette.View.extend({
```
samccone: Why not add a new test as compared to changing this one?

This test had the intent to test compatibility with a vanilla Backbone view, and now this test is gone :crying_cat_face:

scottaj: I thought it was just a Backbone view for simplicity, not to actually test that case. It seemed to fit in with these existing tests, and I didn't know conventions around making new test blocks.

samccone: ah yeah it is a bit confusing in the pits of test land, but this change has a few side effects as to coverage of non Mn Views in regions :fire:

scottaj: Ah, is there coverage data somewhere?

I can work on reverting it. I guess I'd like to see where everyone stands on my other outstanding patch as if that is likely to get merged I'll just fix it there.

samccone: there is coverage metics, however coverage of external lib support does not exist. It is more of a boolean does it work y/n kind of thing currently

paulfalgout: Ah yeah interesting.  `onBeforeRender` would not have worked on a vanilla View.  Seems like the vanilla view test would be much less involved as well.  Would it be easier just to make a simpler test case well documented for testing the vanilla view, or is that just too much extra stuff?

samccone: I am fine for any solution that keeps the vanilla tests around

PR #2411, opened 2015-03-10
Don't rerender the entire CompositeView on sorting

ahumphreys87:
Opening in favour of #2325 - rebased on minor branch

PR #2467, opened 2015-03-31
Don't rerender the entire CompositeView on sorting

ahumphreys87:
Replaces #2457 and #2411

@marionettejs/marionette-core this just needs eyes, should be good to go


jasonLaster:
:+1:

cmaher:
This is potentially breaking (consider the potentially common case in the CompositeView template that references the first/last item in the collection).  Unfortunately, this needs a flag to opt-in.

Issue #2397, opened 2015-03-05
Dependency issues in Node


FishbaitHarry:
- Backbone works fine in Nodejs, it doesn't depend on jquery and doesn't list it in package.json
 - Backbone.Marionette depends on Backbone and does list it in package.json
 - Backbone.Marionette depends on jquery (or a similar library), but does not list it in package.json
 - The line: `Marionette.Deferred = Backbone.$.Deferred;` throws an error in Nodejs environment regardless of whether jquery is available or not, because Backbone doesn't require it and doesn't provide it.

Possible solution: when Marionette detects being in CommonJS environment, it should probably require jquery explicitly. What do you think about it?

samccone:
See for the jquery dependency
https://github.com/marionettejs/backbone.marionette/issues/2300

For how to use in commonjs (browserify) see https://github.com/marionettejs/marionette-integrations/tree/master/browserify

PR #2468, opened 2015-04-01
Put Backbone.$.Deferred in a function to enable node dependency

samjoch:
I had a problem using Marionette when i use it with `mochify`. I want that only one file require shared depencies like Backbone, jQuery but `mocha` seems to include the deeper nested module and when it come to Martionette it hasnt require `Backbone`, nether `Backbone.$.Deferred`.

cf https://github.com/marionettejs/backbone.marionette/issues/2397

### DIFF
--- a/lib/core/backbone.marionette.js
+++ b/lib/core/backbone.marionette.js
@@ -40,7 +40,9 @@
   };
 
   // Get the Deferred creator for later use
-  Marionette.Deferred = Backbone.$.Deferred;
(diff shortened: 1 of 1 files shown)


paulfalgout:
You might want to open an issue for this.  As it is, this is pulled of master and modified on the build file.

However, this would be a breaking change and in the next version `Marionette.Deferred` is removed entirely.

Issue #1852, opened 2014-09-10
Get rid of region.reset


jmeas:
Ahh...`region.reset`. It's been a reliable ol' method on Regions since [before v1.0.0 was even released](https://github.com/marionettejs/backbone.marionette/commit/ae6f134206eec514e4b75a1368158b31d4a8027a). But, alas, it might now be time to discuss its removal. This is breaking in a big way, so it can't happen until a major release if we decide to do it.

So what's the matter with region.reset? Well, it's incompatible with where I'd like to see Regions go. The point of it is to reset the Region's state from being `ensureEl` was ever even called. This only works, of course, when the Region is defined with a selector. In a recent update, we added the ability to define a Region as an already-existing element, too. This led to a number of bugs, several of which are still outstanding, all surrounding region.reset.

I like the ability to define regions with existing elements because it makes them more consistent with Views, and I think there is an incredible amount of power in that. Both views and regions operate with elements, and it makes Marionette a lot easier to learn if you can use the same knowledge you know about a view's element and bring it right over to Regions. (The difference between a Region and View, of course, is what they *do* with that element once they've got it.)

A potential problem is that LayoutViews rely on this method when the layoutView is re-rendered. What for? If we get rid of it, can we still have LayoutViews work the way they need to?

As a cross-reference, there are a handful of [v3 issues](https://github.com/marionettejs/backbone.marionette/issues/1796) regarding LayoutView rendering.

paulfalgout:
Reopening for greater visibility for v3

PR #2473, opened 2015-04-07
bugfix region.reset can't set el in some case

SenYu:
// When this.$el is Object and this.el is DOM selector string, it doesn't works,
  // because this.$el.selector contains parent's selector,
  // for example 
  //   this.$el.selector = '#parent #child',
  //   this.el = '#child', it will be override so that this.el can't be find in DOM...

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -263,12 +263,11 @@ Marionette.Region = Marionette.Object.extend({
   // clearing out the cached `$el`. The next time a view
   // is shown via this region, the region will re-query the
   // DOM for the region's `el`.
-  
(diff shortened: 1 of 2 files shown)


jasonLaster:
@SenYu mind adding some tests that demonstrate the regression.

also if you want to run locally you can do `npm i; grunt test`

SenYu:
@jasonLaster  Glad to get response. Ok, I will add that tests later

paulfalgout:
I think this is mentioned here: https://github.com/marionettejs/backbone.marionette/issues/1852

Issue #2482, opened 2015-04-10
Update bower and component with new Radio dependency


paulfalgout:
Related to https://github.com/marionettejs/backbone.marionette/commit/33467f2bcbfad2477d91d70467b88836d9cdd2db

Issue #2273, opened 2015-02-04
Remove bundled bundle


thejameskyle:
In the future when we remove babysitter we should remove the bundled version of Marionette. Right now (in v3) it only exists for babysitter.

jmeas:
One idea would be pulling it **directly** into the code in v3. Then, throughout v3, we can begin building the list abstraction that will power Regions/RegionManagers/CollectionView and slowly phase babysitter out.

We could just as easily not do it this way, but if we pull it in, then we have no external dependencies, which simplifies the installation process.

thejameskyle:
No external dependencies?

jmeas:
Ya. But if Metal lands, it may be a moot point.

thejameskyle:
Are we not going to include Radio as well?

jmeas:
> Are we not going to include Radio as well?

[comment](https://github.com/marionettejs/backbone.marionette/pull/2269#issuecomment-72884374)

thejameskyle:
Ah okay

jasonLaster:
We've gone back and forth on this, and often come out in favor of keeping it because it can be helpful to devs who use one script tag

thejameskyle:
Clearing out old issues, closing for inactivity

Issue #2426, opened 2015-03-17
Backbone-Metal


paulfalgout:
:metal: 

As the champion of Metal has moved on to greener pastures,  we need to take a hard look at Metal, understand what it means for Marionette and figure out how to proceed.  @marionettejs/marionette-core 

https://github.com/marionettejs/backbone-metal seems to be inspired by http://emberjs.com/api/modules/ember-metal.html

The purpose of Ember-metal is not completely clear http://numberofpeoplewhounderstandembermetal.com/

So what does Backbone-Metal do, and what is its purpose in regards to Marionette?  I'll attempt a first pass at some explanations.  No doubt I'll misrepresent something.. so please chime in.

Very generally:
* It adds a generic `Class` to replace Marionette's `Object` formerly known as `Controller`.  
* It adds `Mixin` functionality which is very similar to `extend` but it promotes a pattern of sharing functionality vs deep inheritance. 
* It wraps each function so that you can call _super which is a generalized way to call the same function from the inherited Class.

In general I support these features, but some cons might be:
* `Marionette.Class` and `Marionette.Object` are not all that different in providing a generic base element to extend from.  Is there really that much benefit in renaming it yet again?
* The suggested implementation of metal on Marionette patches Backbone as well for `_super` to work with Backbone classes.
* ~~To get `_super` to work, _every_ function _everywhere_ must be wrapped.~~  Are there expensive implications to this extra wrap and step in the scope chain?

Regarding the general purpose of extracting the generic object from Marionette as well as errors, deprecation and other helpers, I believe this is the first step in making Marionette's code specifically more concerned with Views alone.  I am on the fence as far as the necessity of making this a stand-alone library.  The fact that it is also ES6, in the current javascript climate, I believe, really makes it less accessible to the general population of developers.  And I am :-1: on making Marionette more difficult to understand from soup to nuts.  I am also :-1: on having a separate library [pick](https://github.com/marionettejs/backbone-metal/issues/3) and [choose](https://github.com/marionettejs/backbone-metal/issues/4) what base Marionette features to support.

Thoughts?

capricube:
Where did the champion go to?
Am 17.03.2015 07:15 schrieb "Paul Falgout" <notifications@github.com>:

jasonLaster:
Thanks for the writeup. I'll share more thoughts later, but perf wise I believe we're only wrapping the functions who override ancestor functions. It's slower, but not too much.

cmaher:
[...]

Issue #2312, opened 2015-02-10
Migrate away from Babysitter


jmeas:
We unanimously agree that Babysitter is no good. We can move toward removing it **now** by proxying its features directly onto the collection view.

- [ ] Add length to the collectionView.
- [ ] Add those findBy methods to the collectionView
- [ ] Make a plan for a more general [enumerable object](http://emberjs.com/api/classes/Ember.Enumerable.html) that we can use in Regions, CollectionViews, and any time we need a list

This issue extends into breaking-compatibility territory, but for now I want to focus on backwards-compatible solutions toward us getting rid of Babysitter.

samccone:
:+1: :+1:

jasonLaster:
Yup. The babysitter spirit will always be with us. Actually think it was a great first version for what it does.

bazineta:
It'd be nice to have the third option, i.e., an enumerable object, which allows for implementation of CollectionView-esque things using a tested implementation.  For example, I've got a drag and drop dashboard view that uses BabySitter today, and it was easy to instrument with Inspector as a result.

Thus, at least in my case, I think it'd be nice to say 'if you want an inspectable view that manages subviews in some unusual manner, use one of these enumerable objects to manage the subviews, and Inspection will be yours, oh yes.....'

ianmstew:
https://github.com/joefitter/backbone.enumerable/blob/master/src/backbone.enumerable.js

PR #2485, opened 2015-04-11
Remove bundled build

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2273

Essentially after removing `wreqr` and adding `radio` as an external dependency, we are requiring people loading with ye-ole-script-src to load other dependencies that do not get bundled.  Similarly if backbone-metal (https://github.com/marionettejs/backbone.marionette/issues/2426) is added it will be an additional base dependency.  And with https://github.com/marionettejs/backbone.marionette/issues/2312 babysitter may also get removed or swallowed, and thus the bundled `Mn` is not necessary or useful.

Initially my only goal was to remove the bundled build and traces of it, but I ran into a build issue, and I found some left over wreqr -> radio updates that were missing.

- [x] Remove some left-over Wreqr stuff from application
- [x] Remove controller.js from the build
- [x] Update bower/component to match (Resolves https://github.com/marionettejs/backbone.marionette/issues/2482)


ahumphreys87:
Lgtm :+1:

jasonLaster:
So first, Iike the three todos. And we should get those in there.

I'm not sure how I feel about removing a bundled option. The way I understand it, if we're using Mn specific 3rd parties (babysitter, metal, radio) then it's nice to offer a bundle of that stuff so that the lazy man can do

```html
<script src="mn.js"</script></script>win()</script>
```

thoughts?

ahumphreys87:
Those were my initial concerns as well then I thought about how many people
actually load in the scripts like this? I don't think there will be many.

That said I'm now thinking even when using browserify would I rather just
load mn or load everything individually - Id probably go for the lazy
option.

I'm torn on this one, i don't think either way would effect my apps too
much and removing it will shrink the library down a bit but I can see why
it's useful to keep.

@paulfalgout what's you opinion? Do you have a motivation for pushing this
through?

jasonLaster:
Yup, I like the marketing around browserify etc. like ppl just using it, not having to be too confused. With that said, great docs would solve this.

paulfalgout:
We are already requiring people load dependencies.. underscore, backbone, jquery.

I don't know that we are helping too much by hiding away the other dependencies.  I think we are making it easier to not have to understand what is going on.

Specifically with a library like radio which appears to be gaining its own popularity, it is quite possible someone could be adding radio to their backbone project, prior to deciding to add marionette.  It is good to build that community as a separate entity that complements backbone.

While there may be reason to load backbone-metal as an included submodule vs an external dependency (I'd probably say the same for baby-sitter if it was a long-term solution) other libraries like radio that are not building blocks of Marionette should remain fairly external, which will make it easier if we do a switch in the future like Wreqr -> Radio.

Furthermore our libraries are put out on CDN.  For those that do load via script tags, this is a fairly likely scenario, but we're diluting the usefulness of the CDN by offering 2 versions.

Additionally we are complicating the build process which would make moving to modules and/or ES6 more difficult. 

I personally picked and use the bundled version at work as we don't yet have our build system ready for every project.  I picked it I suppose at the time because it seemed easier to not need to know what to upgrade when..  but it didn't make my getting started any faster.. I had to ask why it was bundled.. and figure out why there were multiple downloads.. if anything it was confusing.  I picked the bundled version initially because I assumed that the bundled version was safer and would make sure I did not have an incompatible version of wreqr or babysitter.  In reality Marionette follows semver and that was a mistaken concern.  

I think the multiple options for <script> tag users, while attempting to be helpful, are initially an added cognitive load for new users and only result in some incorrect assumptions.

The biggest reason I can see for having a bundled version is for js sandboxes.

Plus now that we are deprecating Marionette Modules and suggesting AMD/CJS etc, people not using a package manager will have to do something custom, or use the Marionette.Module shim.. yet another external dependency.

If we remove this, the download portion of the website becomes.
Download Marionette  (minified)

Dependncies you'll need:
jQuery
Underscore
Backbone
Backbone-Metal
Backbone.Radio

ahumphreys87:
Ok I'm persuaded again, this gets my :+1: again :)

jasonLaster:
@samccone thoughts?

Thanks for the writeup Paul. That helps a lot. I agree that it is good to simplify the build process.

samccone:
I think many many people use the bundled build. It is simple and does not require multiple script tags. 

Sure we assume people have backbone (which requires jquery and underscore) but after that you just need Marionette (bundled) 

I think the install process for most people gets pretty gross when you start forcing them to put 3+ scripts for Marionette.

Clearly the more advanced users do not use the bundled version however I think that many people still do use that version, since it requires less thinking and is "easier" (it also does not require a build process which many people do not have on their projects)

Making things easier however is not always a good argument in this JS world of "quick start, easy, fast ect ect ect..." 

I am on the fence for this one.

:volcano:

Issue #2229, opened 2015-01-30
Is view options as a function useful?


paulfalgout:
This line in the base `Marionette.View` https://github.com/marionettejs/backbone.marionette/blob/master/src/view.js#L12
lets you pass options as a function.  Neat feature, except that the options function is passed to the `Backbone.View`

Backbone doesn't process the function, so if you use options as a function none of the view options such as model or collection would be applied to the view.
http://backbonejs.org/docs/backbone.html#section-128

If `Backbone.View` isn't given the result of the option function, is it a useful feature?

jmeas:
I think that this behavior should be removed.

paulfalgout:
Yeah I think you're right.

It looks like this may have worked with Backbone 1.0.0 https://github.com/jashkenas/backbone/blob/1.0.0/backbone.js#L1087

But since options are no longer attached by Backbone, it no longer gets the result of options before picking the viewOptions  https://github.com/jashkenas/backbone/blob/1.1.0/backbone.js#L988

ahumphreys87:
fixed with #2263

paulfalgout:
Yeah kinda fixed...  it works now in all cases.. so it was a bug fix for v2, though I think it might be better if it was removed.  It's a fairly confusing, fairly useless feature.

PR #2486, opened 2015-04-11
Remove passing view options as a function

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2229

Essentially Marionette.Views allow you to pass in options as a function like:
```js
var optionsFunc = function(){
  return {
    foo: 'bar'
  };
};
var myView = new Marionette.View(optionsFunc);
```

For one thing it is inconsistent with option handling on anything that is not a view: https://github.com/marionettejs/backbone.marionette/blob/master/src/object.js#L7

And another thing.. it is really pretty pointless and probably promotes bad patterns.

Also it was a largely [broken](https://github.com/marionettejs/backbone.marionette/issues/2242) "feature" for most of its existence.

### DIFF
--- a/src/abstract-view.js
+++ b/src/abstract-view.js
@@ -13,8 +13,6 @@ Marionette.AbstractView = Backbone.View.extend({
   constructor: function(options) {
     _.bindAll(this, 'render');
 
-    options = Marionette._getValue(options, this);
-
     // this exposes view options to the view initializer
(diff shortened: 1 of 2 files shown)


jasonLaster:
:thumbsup:

ahumphreys87:
:+1:

Issue #2379, opened 2015-02-26
Destroying a Region should empty it as well (v3)


cmaher:
Currently, destroying a region will call `Marionette.Object#destroy`, which triggers events and stops listening.  It makes a lot of sense to also `empty` the region, thus destroying the contained view.  Some people might be depending on the current behavior, so this should be scheduled for whatever Region-type-thing we have in v3.

ahumphreys87:
im :+1: for this for consistency, maybe `reset` is a better option over `empty` though, it does a bit more destroying! :boom:

ahumphreys87:
fixed with #2869

PR #2487, opened 2015-04-11
Reset region when destroying

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2379

This makes sense.  I think it was just an oversight when making `Region` extend `Object` as previous to that you could not `destroy` a `Region`

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -282,6 +282,13 @@ Marionette.Region = Marionette.Object.extend({
 
     delete this.$el;
     return this;
+  },
+
+  // Handle cleanup and other destroying needs for the region
+  destroy: function() {
+    this.reset();
+
+    return Marionette.Object.prototype.destroy.apply(this, arguments);
   }
 
 },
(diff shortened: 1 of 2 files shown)


paulfalgout:
Considering https://github.com/marionettejs/backbone.marionette/issues/1852 is `reset` correct?  @cmaher @ahumphreys87 thoughts?

ahumphreys87:
Ah I didn't realise we were remove reset.

I guess if this PR won't land till v3.0 it should call empty instead. I've always used reset though and haven't had problems - can you explain what the issues with it are?

paulfalgout:
So I think I'm going to close this for now.. perhaps the best step forward is to just `empty` for now.

The bugs around `reset` have to do mainly with using existing jQuery DOM elements rather than simply using a selector string to define the region.

If `empty` is the best course of action, I can easily reopen and swap that out.. although I'm wondering if this is part of a bigger issue with regions for v3.

PR #2491, opened 2015-04-14
Stop deleting $childViewContainer

### LINKED ISSUE #2488: resetChildViewContainer should not delete $childViewContainer
when the property is deleted, it causes code that expects the property to fail... We should just null out the property as opposed to delete it

```js
    // Internal method to reset the `$childViewContainer` on render
    resetChildViewContainer: function () {
        if (this.$childViewContainer) {
            delete this.$childViewContainer;
        }
    }
```

jasonLaster:
fixes #2488.

ResetChildViewContainer should not delete $childViewContainer 

I kinda like that we're getting rid of the `in` operator use. It always bothered me


ahumphreys87:
:+1:

PR #2498, opened 2015-04-15
add docs for re-rendering on sort

### LINKED ISSUE #2247: Documentation: CollectionView / CompositeView sort rerender
At present, the CollectionView and CompositeView documentation speaks to the view re-rendering on a collection 'reset' event, but does not speak to it re-rendering on a 'sort' event.  In that the views do re-render on 'sort', it would be helpful to document this.

Additionally, given the relative functions of these two classes, it should be made clear that CompositeView completely re-renders on sort, rather than just re-rendering the child view container.

While completely re-rendering in CollectionView is obvious, I think that it's less so in CompositeView, and so it'd not be unexpected that only the CompositeView's childViewContainer would be re-rendered, when in fact the opposite is true.

ahumphreys87:
Fixes #2247

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -879,6 +879,9 @@ one model in to the collection of child views.
 When a model is removed from a collection (or destroyed / deleted), the collection
 view will destroy and remove that model's child view.
 
+When the collection for the view is sorted, the view will automatically re-sort its child views.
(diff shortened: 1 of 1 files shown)


ianmstew:
Should we address the concern from #2247 that re-rendering not just the childViewContainer but also the entire compositeView is re-rendered without `reorderOnSort`?

ahumphreys87:
I debated that but we have changed this for v3 with this PR: https://github.com/marionettejs/backbone.marionette/pull/2467 when we cherry pick onto 2.4.2 we should add a warning or something for CompositeView

ianmstew:
Ahh, very nice!

ianmstew:
:+1:

PR #2500, opened 2015-04-15
improve _createBuffer and attachBuffer in CollectionView

dkurucz:
opening PR against "next" instead of "master" #2396 

fix for #2395

PR #2503, opened 2015-04-16
improve _createBuffer and attachBuffer in CollectionView and CompositeView

dkurucz:
follow up PR for #2500, sourcing branch "next" for these changes

...

opening PR against "next" instead of "master" #2396

fix for #2395

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -52,7 +52,7 @@ Marionette.CollectionView = Marionette.AbstractView.extend({
       this._triggerShowMultiple(this._bufferedChildren, 'before:');
     }
 
-    this.attachBuffer(this);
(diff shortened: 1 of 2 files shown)


ahumphreys87:
@dkurucz 1 last thing could you squash those 2 commits into 1? then this gets my :+1:

dkurucz:
consolidated into one commit and pushed, no failures locally, so the build shouldn't be failing with underscore 1.6, not sure if the environment has a quirk that needs to be resolved (perhaps with the workspace?)

dkurucz:
Hm, okay, there it goes

jasonLaster:
:thumbsup:

PR #2504, opened 2015-04-17
Remove new keyword in front of childEvents example

fay-jai:
The example doesn't work with the 'new' keyword. Removing the 'new' keyword allows the instances of the ChildView and ParentView to be created.

No other changes were made.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -239,7 +239,7 @@ This also works for custom events that you might fire on your child views.
 
 ```js
 // The child view fires a custom event, `show:message`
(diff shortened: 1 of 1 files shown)


samccone:
good catch :+1:

fay-jai:
Thanks!

paulfalgout:
This is also wrong on the `next` branch.  Should we cherry-pick this there now, or do we think this change will merge ok when `next` gets into master?  https://github.com/marionettejs/backbone.marionette/blob/next/docs/marionette.collectionview.md

jasonLaster:
yikes, lets rebase next asap

Issue #2059, opened 2014-11-06
Marionette classes should have a cid field


jasonLaster:
It's common practice to check a View or Model's `cid` field when debugging. The `cid` field is also an important way to hyperlink objects in the Inspector. e.g click on a view and jump to the UI pane and that view's info panel. 

While working on the inspector, I found that many of Marionette's classes do not have a cid field. This should be relatively easy to add:

Classes in need of a cid
- [x] Object
- [ ] Controller
- [ ] Module
- [x] Application
- [x] Behavior

paulfalgout:
Related: https://github.com/marionettejs/backbone-metal/issues/2

SauloSilva:
All, regions and region-manager also extend from Object too.

ahumphreys87:
Fixed with #2524. Controller is gone as is Module

PR #2510, opened 2015-04-18
Add cid field to object, controller, module, application, behavior, region and region-manager

SauloSilva:
ref #2059

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -6,6 +6,7 @@
 // A view that iterates over a Backbone.Collection
 // and renders an individual child view for each model.
 Marionette.CollectionView = Marionette.View.extend({
+  cidPrefix: 'mncv',
 
   // used as the prefix for child view events
   // that are forwarded through the collectionview
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -6,7 +6,6 @@
 // A view that iterates over a Backbone.Collection
 // and renders an individual child view for each model.
 Marionette.CollectionView = Marionette.View.extend({
-  cidPrefix: 'mncv',
 
   // used as the prefix for child view events
   // that are forwarded through the collectionview
--- a/src/object.js
+++ b/src/object.js
@@ -5,7 +5,7 @@
 // Object borrows many conventions and utilities from Backbone.
 Marionette.Object = function(options) {
   this.options = _.extend({}, _.result(this, 'options'), options);
-  this.cid = _.uniqueId(this.cidPrefix || 'mno');
+  this.cid = _.uniqueId(this.cidPrefix);
   this.initialize.apply(this, arguments);
 };
 
@@ -16,6 +16,8 @@ Marionette.Object.extend = Marionette.extend;
 
 // Ensure it can trigger events with Backbone.Events
 _.extend(Marionette.Object.prototype, Backbone.Events, {
+  cidPrefix: 'mno',
+
   //this is a noop method intended to be overridden by classes that extend from this base
   initialize: function() {},
 
--- a/src/object.js
+++ b/src/object.js
@@ -16,6 +16,7 @@ Marionette.Object.extend = Marionette.extend;
 
 // Ensure it can trigger events with Backbone.Events
 _.extend(Marionette.Object.prototype, Backbone.Events, {
+  cid: _.uniqueId('mno'),
 
   //this is a noop method intended to be overridden by classes that extend from this base
   initialize: function() {},
--- a/src/object.js
+++ b/src/object.js
@@ -5,7 +5,7 @@
 // Object borrows many conventions and utilities from Backbone.
 Marionette.Object = function(options) {
   this.options = _.extend({}, _.result(this, 'options'), options);
-  this.cid = _.uniqueId(this.cidPrefix);
+  this.cid = _.uniqueId(this.cidPrefix || 'mno');
   this.initialize.apply(this, arguments);
 };
 
--- a/src/object.js
+++ b/src/object.js
@@ -5,7 +5,7 @@
 // Object borrows many conventions and utilities from Backbone.
 Marionette.Object = function(options) {
   this.options = _.extend({}, _.result(this, 'options'), options);
-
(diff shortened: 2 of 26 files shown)


paulfalgout:
So I think something like:
```js
this.cid = _.uniqueId(this._cid);
```
should be added to `Marionette.Object`
and `_cid` should be defined as "mno" "mnb" on the various extensions of `Object`.

As mentioned here:  https://github.com/marionettejs/backbone.marionette/issues/2059#issuecomment-63196548

Also I think at this point `2.x` releases other than patches are locked.  This should be pulled against `next` which is currently a replacement for `major`  Sorry for the confusion there, but we're almost through to `v3`

SauloSilva:
@paulfalgout ok. Added the  `_cid` on extension of `Object`. 
After, I close here  this `PullRequest` and open for the branch `next` ok?

src/region.js
```diff
@@ -7,6 +7,8 @@
 // http://lostechies.com/derickbailey/2011/12/12/composite-js-apps-regions-and-region-managers/
 
 Marionette.Region = Marionette.Object.extend({
+  _cid: 'mnr',
```
jasonLaster: I think this could be `cidPrefix` 
https://github.com/jashkenas/backbone/blob/master/backbone.js#L393

test/unit/region.build-region.spec.js
```diff
@@ -70,6 +77,8 @@ describe('Region', function() {
         beforeEach(function() {
           this.definition = {selector: this.fooSelector};
           this.region = Marionette.Region.buildRegion(this.definition, this.DefaultRegionClass);
+
+          delete this.region.cid;
```
jasonLaster: ugh, is this because of the deep equalities? 

I had to deal with this too at one point, I wonder if we could change the checks in regions.spec, to be a little friendlier...
like 
```js
sameRegion = function(ra, rb) {var props = _.partial(_.pluck, _, [list_of_props]), return props(ra) == props(rb) }
```

src/region.js
SauloSilva: @jasonLaster you thing change `_cid` to `_cidPrefix` in all?

test/unit/region.build-region.spec.js
SauloSilva: @jasonLaster yes. It's one option, I'll verify and change it.

src/region.js
paulfalgout: hmm "private" or public..

I suppose there's some arguments that others may want to change the `cid` for their extensions of things.. but at the same time this is not how view cids or model cids work.  if it is public would view / collectionview then follow suit with `mnv_` `mncv_`?

SauloSilva: Looking again, I think `cidPrefix` is better because can be changed in the extensions of things. @paulfalgout I do not understand, why `_` on the end prefix instead `mnv` and `mncv`?

src/collection-view.js
```diff
@@ -6,6 +6,7 @@
 // A view that iterates over a Backbone.Collection
 // and renders an individual child view for each model.
 Marionette.CollectionView = Marionette.View.extend({
+  cidPrefix: 'mncv',
```
jasonLaster: I don't think we should overide the view's cid. It's nice to have a format for that

src/object.js
```diff
@@ -5,7 +5,7 @@
 // Object borrows many conventions and utilities from Backbone.
 Marionette.Object = function(options) {
   this.options = _.extend({}, _.result(this, 'options'), options);
-
+  this.cid = _.uniqueId(this.cidPrefix || 'mno');
```
jasonLaster: `cidPrefix: 'mno'`

would work as well as the subclass will override that property

test/unit/application.app-regions.spec.js
```diff
@@ -24,8 +24,10 @@ describe('application regions', function() {
     });
 
     it('should initialize the regions', function() {
-      expect(this.app.fooRegion).to.deep.equal(this.fooRegion);
-      expect(this.app.barRegion).to.deep.equal(this.barRegion);
+      this.checkProperties(function(props) {
+        expect(props(this.app.fooRegion)).to.deep.equal(props(this.fooRegion));
```
jasonLaster: what's props?

test/unit/application.app-regions.spec.js
```diff
@@ -33,11 +35,11 @@ describe('application regions', function() {
     });
 
     it('should trigger a before:add:region event', function() {
-      expect(this.beforeAddRegionStub).to.have.been.calledWith('fooRegion', this.fooRegion);
+      expect(this.beforeAddRegionStub).to.have.been.calledWith('fooRegion');
     });
 
     it('should trigger a add:region event', function() {
-      expect(this.addRegionStub).to.have.been.calledWith('barRegion', this.barRegion);
```
jasonLaster: why's this no longer true?

test/unit/setup/helpers.js
```diff
@@ -15,6 +15,14 @@ var clearFixtures = function () {
   $body.empty();
 };
 
+before(function() {
+  this.checkProperties = function(block, blacklist) {
+    blacklist = blacklist ? blacklist.push('cid') : 'cid'
+    var props = _.partial(_.omit, _, blacklist);
+    block.call(this, props);
+  }
```
jasonLaster: I think i'm cool with this, but I'd be curious to see what other @marionettejs/marionette-core think?

src/collection-view.js
SauloSilva: @jasonLaster you think should be remove?

src/object.js
SauloSilva: @jasonLaster I changed

test/unit/application.app-regions.spec.js
SauloSilva: @jasonLaster I changed for manually verification of the args in the stubs, ok?

jasonLaster:
@SauloSilva mind re-opening against next? I noticed this because module is not a thing

SauloSilva:
@jasonLaster of course.

PR #2511, opened 2015-04-18
Backbone Metal :metal:

jasonLaster:
Hey y'all, posting thins pretty early in the process. Mind chipping in and letting me know what you think of this PR and what the next steps should be?


paulfalgout:
- [ ] Remove `Mn.Object` and update builds
- [ ] Update `Mn.Class` to include `Mn.Object` functionality
  - [ ] `Mn.mergeOptions`
  - [ ] Attach `options` to `this.options` in `constructor`
  - [ ] `proxyRadioHandlers` in `constructor`
  - [ ] Feature parity `Mn.Object.destroy`
- [ ] Switch anything extending from `Mn.Object` to `Mn.Class`
- [ ] Update `Object` related tests
- [ ] If not https://github.com/marionettejs/backbone.marionette/pull/2485 then `Metal` needs to be unwrapped and added to the bundle and should not be an external dependency to the bundle.

jasonLaster:
Nice writeup Paul. That's a good list of tasks, some blocking some not. I'd also add docs to the list. 

In general, I like small PRs. What do y'all think I should add/remove to this PR for it to be ready minus comments on the work? I'd like to make docs a separate ticket cause that'll be a big writeup worthy of its own set of critiques.

paulfalgout:
The smallest useful PR I think would be to remove this for now:
https://github.com/jasonLaster/backbone.marionette/commit/934ca52999e0dabd99239e4c3d5626dcf7c030e9#diff-929df048617c3352d04339471700c81eR1
And 
- [ ] If not #2485 then Metal needs to be unwrapped and added to the bundle and should not be an external dependency to the bundle.

Then the rest is implementing Class replacing Object and "classifying" everything else

ahumphreys87:
Yeah lets remove `Mn.Class` for now and the `classify` stuff at the bottom. Then I'm happy to merge this and tackle object => class as a separate PR/discussion

blikblum:
In metal master classify is removed: See https://github.com/marionettejs/backbone-metal/commit/3aaa657d032eeefd70ee3d8dfe0ea751b9845009

paulfalgout:
True, but essentially moved to the commit here:
https://github.com/jasonLaster/backbone.marionette/commit/934ca52999e0dabd99239e4c3d5626dcf7c030e9#diff-929df048617c3352d04339471700c81eR37

jasonLaster:
I'll remove Mn.Class for now

hey @paulfalgout , mind elaborating on the unwrapping business.

jasonLaster:
cool, just learned how to unwrap things...

notes:
+ we should now be unwrapping. 
+ the class stuff is gone
+ i kept some of the error stuff because it was minimal

Couple unrelated updates:
1. i'm not sure if we're testing the bundle
2. i'm not sure if we're adding the radio to the bundle

paulfalgout:
Correct we're not testing the bundled.. it is actually quite broken.  doesn't include radio at all or even list radio as a dependency in the UMD.. and it still has controller.js which make the bundling break entirely.

jasonLaster:
lol - i think we should have atleast a small integration test

test/unit/trigger-method.spec.js
```diff
@@ -160,12 +160,12 @@ describe('trigger event and method name', function() {
...
-        this.view = new Backbone.View();
-        this.view.onFoo = this.methodHandler;
-        this.view.on('foo', this.eventHandler);
+        this.obj = _.extend({}, Backbone.Events);
+        this.obj.onFoo = this.methodHandler;
+        this.obj.on('foo', this.eventHandler);
         this.triggerMethodSpy = this.sinon.spy(Marionette, 'triggerMethod');
 
-        Marionette.triggerMethodOn(this.view, 'foo');
+        Marionette.triggerMethodOn(this.obj, 'foo');
```
jasonLaster: This is some great TJK code that uses B.Events directly because he was re-mixing in the B.Events after he classified it. I kept this in because well, when I go back i don't want to miss it :)

src/metal.js
```diff
@@ -0,0 +1,3 @@
+// Add Backbone.Metal extensions here
+// 1. Marionette.Class
+// 2. Marionette.Error and Marionette.Deprecate
```
jasonLaster: can't get more simple than some comments

jasonLaster:
Hey @marionettejs/marionette-core I just removed all of the Error stuff because that was get in the way of the beautiful bundle and npm and specRunner soup.

Hopefully this will be a case-study in how to add a new dependency to Mn. There are a lot of touch points.

ahumphreys87:
:+1: from me. Looking forward to getting class in after this

paulfalgout:
The component.json and bower.json files need to have their dependencies updated.

src/build/bundled.js
```diff
@@ -2,23 +2,27 @@
...
-    define(['backbone', 'underscore'], function(Backbone, _) {
-      return (root.Marionette = root.Mn = factory(root, Backbone, _));
+    define(['backbone', 'underscore', 'backbone-metal'], function(Backbone, _, Metal) {
+      return (root.Marionette = root.Mn = factory(root, Backbone, _, Metal));
     });
   } else if (typeof exports !== 'undefined') {
     var Backbone = require('backbone');
     var _ = require('underscore');
-    module.exports = factory(root, Backbone, _);
+    var Metal = require('backbone-metal');
```
paulfalgout: Actually I don't think we want to introduce Metal in the AMD or CJS here since it is bundled.  The dependencies here should just be backbone and underscore.

src/build/core.js
```diff
@@ -1,18 +1,19 @@
...
     var Backbone = require('backbone');
     var _ = require('underscore');
     var BabySitter = require('backbone.babysitter');
     var Radio = require('backbone.radio');
-    module.exports = factory(root, Backbone, _);
+    var Metal = require('backbone-metal');
+    module.exports = factory(root, Backbone, _, Metal);
   } else {
-    root.Marionette = root.Mn = factory(root, root.Backbone, root._);
+    root.Marionette = root.Mn = factory(root, root.Backbone, root._, root.Metal);
```
paulfalgout: I'm not sure root.Metal is necessary?  I assume it'll be available via Backbone.Metal by this point..  does metal even export a global?  For instance if you `<script>` in Radio it'll be available on `Backbone.Radio`, but not `root.Radio` by default.

jasonLaster: Good thought. We need to require/define them in but no need to pass to factory

src/build/bundled.js
jasonLaster: yup. true, it's coming in below. no need to get it twice

package.json
```diff
@@ -43,6 +43,7 @@
     "backbone.babysitter": "^0.1.0",
     "backbone.radio": "^0.9.0",
     "backbone": "1.0.0 - 1.1.2",
+    "backbone-metal": "^0.5.0",
```
jasonLaster: is it 5 or 3?

ahumphreys87: current is 0.5.0

jasonLaster: yea, i like that. we'll have a chance to do a new release at somepoint and make sure the next is legit

jasonLaster:
@marionettejs/marionette-core this is getting so close to great.

paulfalgout:
:clap: :+1:

ahumphreys87:
:boom: :+1:

PR #2513, opened 2015-04-20
Fix inconsistency between addRegion & addRegions

codespawner:
`addRegion` now behaves like `addRegions` and adds the new region to internal `this.regions`.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -22,7 +22,7 @@ Marionette.View = Marionette.AbstractView.extend({
     this._firstRender = true;
     this._initializeRegions(options);
 
-    this.regions = {};
+    this.regions = this.regions || {};
 
     Marionette.AbstractView.apply(this, arguments);
   },
--- a/src/view.js
+++ b/src/view.js
@@ -22,6 +22,8 @@ Marionette.View = Marionette.AbstractView.extend({
     this._firstRender = true;
     this._initializeRegions(options);
 
+    this.regions = {};
+
     Marionette.AbstractView.apply(this, arguments);
   },
 
@@ -132,9 +134,19 @@ Marionette.View = Marionette.AbstractView.extend({
     return this.addRegions(regions)[name];
   },
 
-  // Add multiple regions as a {name: definition, name2: def2} object literal
-  addRegions: function(regions) {
+  // Add multiple regions as a {name: definition, name2: def2} object literal or
+  // a function evaluation to such literal
+  addRegions: function(regions, defaults) {
+    // Enable regions to be a function
+    regions = Marionette._getValue(regions, this, arguments);
+
+    // Normalize region selectors hash to allow
+    // a user to use the @ui. syntax.
+    regions = this.normalizeUIValues(regions, ['selector', 'el']);
+
+    // Add the regions definitions to the regions property
     this.regions = _.extend({}, this.regions, regions);
+
     return this._buildRegions(regions);
   },
 
@@ -187,24 +199,10 @@ Marionette.View = Marionette.AbstractView.extend({
   // Internal method to initialize the regions that have been defined in a
   // `regions` attribute on this layoutView.
   _initializeRegions: function(options) {
-    var regions;
     this._initRegionManager();
 
-    regions = Marionette._getValue(this.regions, this, [options]) || {};
-
-    // Enable users to define `regions` as instance options.
-    var regionOptions = this.getOption.call(options, 'regions');
-
-    // enable region options to be a function
-    regionOptions = Marionette._getValue(regionOptions, this, [options]);
-
-    _.extend(regions, regionOptions);
-
-    // Normalize region selectors hash to allow
-    // a user to use the @ui. syntax.
-    regions = this.normalizeUIValues(regions, ['selector', 'el']);
-
-    this.addRegions(regions);
+    this.addRegions(this.regions);
+    this.addRegions(this.getOption.call(options, 'regions'));
   },
 
   // internal method to build regions
--- a/src/view.js
+++ b/src/view.js
@@ -135,10 +135,14 @@ Marionette.View = Marionette.AbstractView.extend({
   },
 
   // Add multiple regions as a {name: definition, name2: def2} object literal or
-  // a function evaluation to such literal
-  addRegions: function(regions, defaults) {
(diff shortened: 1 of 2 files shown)


src/view.js
```diff
@@ -129,7 +129,7 @@ Marionette.View = Marionette.AbstractView.extend({
   addRegion: function(name, definition) {
     var regions = {};
     regions[name] = definition;
-    return this._buildRegions(regions)[name];
+    return this.addRegions(regions)[name];
```
jasonLaster: maybe go even further:

```js
return this.addRegions({name: definition})[name];
```

jasonLaster:
great find @codespawner 

mind also adding a test to `test/unit/view.spec.js` that demonstrates that `addRegion` ought to update the regions property

codespawner:
Ok, I'm on it.

jasonLaster:
yea, everyone avoids the tests, but they're the most fun. you get to see how you're changing / fixing the spec

src/view.js
paulfalgout: And FWIW, with @jasonLaster's comment, line 130 and 131 are no longer needed as it'll happen on 137.

codespawner: This would create a new object with a 'name' attribute, which is not what's intended, right?

paulfalgout: ha! you're absolutely correct..

codespawner:
By messing with the unit tests for addRegion and addRegions I stumbled on #2516.

jasonLaster:
Nice investigative work. I love our tests. 

My instinct is to say we should align these functions with the regionManager equivalents. Would have to dig deeper though to remember trade offs. I'm curious how far you can go and not have tests break

codespawner:
Is it actually intended to keep the regions attribute in sync with the regions in the region manager or could a solution just be to forget about it and only use it for initialisation (see http://marionettejs.com/docs/v2.4.1/marionette.layoutview.html#region-options).

I don't see it mentioned anywhere in the docs and the recommended way is to use the getRegion accessor. So maybe we could fix the issues like this:
```
  // Internal method to initialize the regions that have been defined in a
  // `regions` attribute on this layoutView.
  _initializeRegions: function(options) {
    this._initRegionManager();

    this.addRegions(this.regions);
    this.addRegions(this.getOption.call(options, 'regions'));
  },

  // Add a single region, by name, to the layoutView
  addRegion: function(name, definition) {
    var regions = {};
    regions[name] = definition;
    return this.addRegions(regions)[name];
  },

  // Add multiple regions as a {name: definition, name2: def2} object literal
  addRegions: function(regions, defaults) {
    // Enable regions to be a function
    regions = Marionette._getValue(regions, this, arguments);

    // Normalize region selectors hash to allow
    // a user to use the @ui. syntax.
    regions = this.normalizeUIValues(regions, ['selector', 'el']);

    // Could add to `this.regions` here

    return this._buildRegions(regions);
  },
```

jasonLaster:
nice research @codespawner. 

1. I think we should keep the regions prop in sync. I agree it's kinda ugly, but I've seen a lot of `this.regions.header` code out there and we'd have to have a good reason to change that.
2. Furthermore, I'd like to see a test that adds a region and checks to see if it's in the region's prop. Sorry to give you work :)
3. `    this.addRegions(this.regions); this.addRegions(this.getOption.call(options, 'regions'));` the first addRegions i think is unecessary
4. Feel free to PR your gist so that it's easier to see the green and red. 

Thanks for the work

src/view.js
```diff
@@ -22,6 +22,8 @@ Marionette.View = Marionette.AbstractView.extend({
     this._firstRender = true;
     this._initializeRegions(options);
 
+    this.regions = {};
```
codespawner: The property needs to be initialised to prevent errors when calling `removeRegion`when no regions property exists yet.

ahumphreys87:
@jasonLaster I'd be pro removing the `regions` property we are shooting for a major release after all - our upgrade tool could make it easier to migrate old code to use `getRegion`

paulfalgout:
ping @jfairbank  I know you were doing a lot of looking into Regions.. care to weigh in here

jasonLaster:
@ahumphreys87, i think think the `regions` prop discussion should happen in it's own issue/PR. There may be one already

src/view.js
```diff
@@ -187,24 +203,10 @@ Marionette.View = Marionette.AbstractView.extend({
...
-
-    _.extend(regions, regionOptions);
-
-    // Normalize region selectors hash to allow
-    // a user to use the @ui. syntax.
-    regions = this.normalizeUIValues(regions, ['selector', 'el']);
-
-    this.addRegions(regions);
+    this._addRegions(this.regions, [options]);
+    this._addRegions(this.getOption.call(options, 'regions'), [options]);
```
jasonLaster: why do we need both calls here?

ahumphreys87:
ok good shout

src/view.js
```diff
@@ -129,12 +131,26 @@ Marionette.View = Marionette.AbstractView.extend({
...
     regions[name] = definition;
-    return this._buildRegions(regions)[name];
+    return this.addRegions(regions)[name];
   },
 
-  // Add multiple regions as a {name: definition, name2: def2} object literal
+  // Add multiple regions as a {name: definition, name2: def2} object literal or
+  // a function that evaluates to such literal
   addRegions: function(regions) {
+    return this._addRegions(regions, arguments);
```
jasonLaster: this strikes me as weird...  even `_.rest(arguments)` would seeme better

src/view.js
codespawner: Because the regions can both be supplied directly in the `regions` attribute or in the `options`?

src/view.js
codespawner: It's the same as in the region manager `addRegions` and what is written in the docs:

> If you supply a function to addRegions, it will be
called with the RegionManager instance context and
*all the arguments* passed to addRegions.

https://github.com/marionettejs/backbone.marionette/blob/8b32480adb2e432352ea1f29224f125be677f6d7/src/region-manager.js#L22

ahumphreys87:
lgtm - @codespawner could you squash the commits please then Im happy with this

ahumphreys87:
cherry picked onto `next`

PR #2514, opened 2015-04-20
Sorting composite view docs

### LINKED ISSUE #2247: Documentation: CollectionView / CompositeView sort rerender

ahumphreys87:
Fixes #2247

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -759,6 +759,9 @@ one model in to the collection of child views.
 When a model is removed from a collection (or destroyed / deleted), the collection
 view will destroy and remove that model's child view.
 
+When the collection for the view is sorted, the view will automatically re-sort its child views.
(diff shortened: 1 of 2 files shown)


jasonLaster:
looks good @ahumphreys87!

ahumphreys87:
@samccone updated

Issue #2517, opened 2015-04-21
Add Radio to the bundle


jasonLaster:
If we keep the bundle around, we should add Radio to it!

paulfalgout:
.. and remove wreqr :-)

jasonLaster:
Ya know

PR #2525, opened 2015-04-26
Fix dependencies and build

paulfalgout:
This is a portion of the fixes from https://github.com/marionettejs/backbone.marionette/pull/2485 applied to the bundle.

Currently `next` does not build.  This is because the build files have not been maintained well.
Also Radio was not added to the bundled file, but wreqr was still attempting to be "unwrapped" breaking the build.

I removed the rest of the wreqr related issues and updated them to radio.

So for what is currently in `next` this fixes the build and dependencies so `next` is in a releasable state.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -26,8 +26,8 @@ module.exports = function(grunt) {
         ' * Includes BabySitter\n' +
         ' * https://github.com/marionettejs/backbone.babysitter/\n' +
         ' *\n' +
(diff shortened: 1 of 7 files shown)


paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2517

ahumphreys87:
:+1:

jasonLaster:
Easy merge. Looks like we got sloppy with the deprecations :)

PR #2527, opened 2015-04-26
onBeforeAttach and onAttach for CollectionView child views

### LINKED ISSUE #2526: Trigger onAttach on child views during CollectionView re-render
After a CollectionView has been shown into a Region, re-rendering the CollectionView does not trigger onAttach/onBeforeAttach on its child views, and similarly when adding new child views.

Back port of #2209 for 2.4.2.

ianmstew:
Fixes #2526.  Cherry-pick of #2481 for 2.4.2.

### DIFF
--- a/SpecRunner.html
+++ b/SpecRunner.html
@@ -119,6 +119,7 @@
     <script src="test/unit/item-view.spec.js"></script>
     <script src="test/unit/layout-view.dynamic-regions.spec.js"></script>
     <script src="test/unit/layout-view.spec.js"></script>
+    <script src="test/unit/merge-options.spec.js"></script>
     <script src="test/unit/mixin-underscore-collection.spec.js"></script>
     <script src="test/unit/module.spec.js"></script>
     <script src="test/unit/module.stop.spec.js"></script>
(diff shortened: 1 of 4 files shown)


ianmstew:
Pinging @ahumphreys87 and @jasonLaster--you guys thumbs-upped the original PR against `next`.  :)

ianmstew:
The main difference between this and the original PR is that, instead of calling `Marionette.triggerMethodMany()`, I'm calling `this._triggerMethodMany()`, which is actually Jason's code from `next` that I dropped into CollectionView for backward-compatibility.

The second difference is the `once('render'` approach to firing `before:show` within `_addChildView()` that we agreed on strictly for 2.4.2 here: https://github.com/marionettejs/backbone.marionette/pull/2481#issuecomment-92442727.

jasonLaster:
Is it safe to drop addEmptyView? Sorry Ian if yhis is more work.

ahumphreys87:
That's my only questions also, tricky as its one of those methods that probably should never have been "public" as it does so much that people shouldn't mess with but because it was it will be a breaking change :(

ianmstew:
Ahh, I overlooked that.  No problem at all.

ianmstew:
There we go.  I wanted to cry bumping `maxstatements` and `maxcomplexity` just because of `addEmptyView()`...

ahumphreys87:
yeah it's ok we can kill it in 3.0 :)

:+1:

jasonLaster:
:boom:

PR #2529, opened 2015-04-27
Support false triggerBeforeAttach/triggerAttach show() options

### LINKED ISSUE #2484: Support "false" value for Region#show() options triggerBeforeAttach/triggerAttach
The Region gives the developer power to use `show()` option `{triggerBeforeAttach: true}` but not `{triggerBeforeAttach: false}`, and similarly with `triggerAttach`.

While the ability to use `show()` options set to `false` is not documented one way or another, Backbone `Collection#set()` methods establish a precedent of [disabling a default behavior with `false` method options](http://backbonejs.org/#Collection-set).

In this example, onBeforeAttach will fire, [as documented](http://marionettejs.com/docs/marionette.region.html#onbeforeattach-onattach):

```js
var FooRegion = Marionette.Region.extend({
  triggerBeforeAttach: false
});
region.show(view, {triggerBeforeAttach: true});
```

In this example, onBeforeAttach will _also_ fire (no documentation for this case):

```js
var FooRegion = Marionette.Region.extend({
  triggerBeforeAttach: true  // The default value, but presented here for clarity
});
region.show(view, {triggerBeforeAttach: false});
```

The second example does not match my expectations, partly set up by Backbone, because the explicit `false` option is ignored.

The fix for this is small (two lines).  The question is whether we want to support explicit false options, particularly in this case.

ianmstew:
Fixes #2484.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -112,10 +112,12 @@ Marionette.Region = Marionette.Object.extend({
       // as it's a potentially-slow method
       var displayedViews = [];
 
-      var triggerBeforeAttach = showOptions.triggerBeforeAttach || this.triggerBeforeAttach;
-      var triggerAttach = showOptions.triggerAttach || this.triggerAttach;
+      var attachOptions = _.extend({
+        triggerBeforeAttach: this.triggerBeforeAttach,
+        triggerAttach: this.triggerAttach
+      }, showOptions);
 
-      if (attachedRegion && triggerBeforeAttach) {
+      if (attachedRegion && attachOptions.triggerBeforeAttach) {
         displayedViews = this._displayedViews(view);
         this._triggerAttach(displayedViews, 'before:');
       }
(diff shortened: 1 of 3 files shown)


src/region.js
```diff
@@ -112,10 +112,12 @@ Marionette.Region = Marionette.Object.extend({
       // as it's a potentially-slow method
       var displayedViews = [];
 
-      var triggerBeforeAttach = showOptions.triggerBeforeAttach || this.triggerBeforeAttach;
-      var triggerAttach = showOptions.triggerAttach || this.triggerAttach;
+      var attachOptions = _.extend({
+        triggerBeforeAttach: this.triggerBeforeAttach,
+        triggerAttach: this.triggerAttach
+      }, showOptions);
```
ahumphreys87: neat!

ahumphreys87:
:+1:

src/region.js
ianmstew: thanks!  first I tried checking explicitly for `showOptions.triggerAttach === false` but it became a pretty ugly logic statement.

jasonLaster: this will do. nice!

ianmstew: haha, i can see that since I'm pulling `this` into the extend target.

Issue #2386, opened 2015-03-01
Add possibility to changing default Behavior during View initialization.


vovchok:
That would be cool to add the possibility to change Behavior during View initialization. It gives us the flexibility to initialize a view with the different kinds of behavior. The following changes will add the desired result.

In the src/view.js

``` javascript
this._behaviors = Marionette.Behaviors(this);
```

replace to

``` javascript
this._behaviors = Marionette.Behaviors(this,  this.getOption('behaviors'));
```

and in the src/behavior.js

``` javascript
if (!_.isObject(view.behaviors)) {
```

replace to

``` javascript
if (!_.isObject(view.behaviors) && _.isEmpty(behaviors)) {
```

Example of usage:

``` javascript
var Behaviors = {};

Mn.Behaviors.behaviorsLookup = function() {
    return Behaviors;
};

var DefaultBoxBehavior = Mn.Behavior.extend({
    ui: {
        box: ".box"
    },
    events: {
        "click @ui.box": "onClick"
    },
    onClick: function() {
        console.log("default behavior");
    }
});
Behaviors.DefaultBoxBehavior = DefaultBoxBehavior;

var NewBoxBehavior = Mn.Behavior.extend({
    ui: {
        box: ".box"
    },
    events: {
        "click @ui.box": "onClick"
    },
    onClick: function() {
        console.log("new behavior");
    }
});
Behaviors.NewBoxBehavior = NewBoxBehavior;

var BoxView = Mn.ItemView.extend({
    template: '<div class="wrapper"><a href="#" class="box">Click me</a></div>',
    behaviors: {
        DefaultBoxBehavior: {}      
    }
});

// default 
var befaultBox = new BoxView();
befaultBox.render().$el.appendTo("body");

// new
var newBox = new BoxView({
    behaviors: {
        NewBoxBehavior: {}      
    }
});
newBox.render().$el.appendTo("body");
```

jasonLaster:
good call. Would be happy to :+1: a pr

adorsk:
Another use case to consider for this is updating behaviors on view instance that has already been created.

For example, something like this (referring to @wovchok 's example above)
````
// default 
var befaultBox = new BoxView();
befaultBox.render().$el.appendTo("body");

// new
var newBox = new BoxView();

// Update behaviors on existing newBox instance.
newBox.updateBehaviors({NewBoxBehavior: {}})
````

I'll take a look into this.

adorsk:
[...]

PR #2537, opened 2015-04-29
Adorsk/behaviors as constructor opts

adorsk:
PR for #2386 .

Additional specs added to test/unit/behaviors.spec.js .

### DIFF
--- a/src/behaviors.js
+++ b/src/behaviors.js
@@ -20,8 +20,7 @@ Marionette.Behaviors = (function(Marionette, _) {
 
     // Behaviors defined on a view can be a flat object literal
     // or it can be a function that returns an object.
-    behaviors = Behaviors.parseBehaviors(view, behaviors ||
-        _.result(view.options, 'behaviors') || _.result(view, 'behaviors'));
+    behaviors = Behaviors.parseBehaviors(view, behaviors);
 
     // Wraps several of the view's methods
     // calling the methods first on each behavior
--- a/src/behaviors.js
+++ b/src/behaviors.js
@@ -20,7 +20,8 @@ Marionette.Behaviors = (function(Marionette, _) {
 
     // Behaviors defined on a view can be a flat object literal
     // or it can be a function that returns an object.
-    behaviors = Behaviors.parseBehaviors(view, behaviors || _.result(view, 'behaviors'));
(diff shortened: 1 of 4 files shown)


src/behaviors.js
```diff
@@ -20,7 +20,8 @@ Marionette.Behaviors = (function(Marionette, _) {
 
     // Behaviors defined on a view can be a flat object literal
     // or it can be a function that returns an object.
-    behaviors = Behaviors.parseBehaviors(view, behaviors || _.result(view, 'behaviors'));
+    behaviors = Behaviors.parseBehaviors(view, behaviors ||
+        _.result(view.options, 'behaviors') || _.result(view, 'behaviors'));
```
jasonLaster: Hmm maybe Marionette.getOption

adorsk: Roger that, updated in PR to use Marionette.getOption .

jasonLaster:
Two things:
1. This PR should be reopened against next
2. The behavior tests are bloody. Try running the tests with grunt test. Or open SpecRunner in the browser

paulfalgout:
The latest issue is trying to use `_.result`  with `getOption`

functionally you probably want something like: `Mn._getValue(Mn.getOption(view, 'behaviors'), view);`

But I think this might be cleaner if we utilize the 2nd argument of Behaviors.  So essentially what is now:
```js
function Behaviors(view, behaviors) {

    if (!_.isObject(view.behaviors)) {
      return {};
    }

    // Behaviors defined on a view can be a flat object literal
    // or it can be a function that returns an object.
    behaviors = Behaviors.parseBehaviors(view, behaviors || _.result(view, 'behaviors'));
```
 would become:

```js
function Behaviors(view, behaviors) {

    if (!_.isObject(behaviors)) {
      return {};
    }

    behaviors = Behaviors.parseBehaviors(view, behaviors);
```
And then over in the View https://github.com/marionettejs/backbone.marionette/blob/next/src/abstract-view.js#L22 you'd have :
```js
var behaviors = Marionette._getValue(this.getOption('behaviors'), this);
this._behaviors = Marionette.Behaviors(this, behaviors);
```

adorsk:
@jasonLaster: Roger that, I'll close here and re-open against next.

@paulfalgout: Thx for the tips, your suggestion looks good to me. I'll update and use it when I re-open the PR against next.

PR #2543, opened 2015-05-05
Remove redundant return statement

natalan:
Remove redundant `return` in region's `show` method.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -132,8 +132,6 @@ Marionette.Region = Marionette.Object.extend({
 
       this.triggerMethod('show', view, this, options);
       Marionette.triggerMethodOn(view, 'show', view, this, options);
-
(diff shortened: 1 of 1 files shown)


jasonLaster:
congratulations making it to the end of the `show` method :)
seriously, nice catch. 

---
+ just rebooted the failed test
+ do we want to merge onto master or `next`. Generally we don't put new src code into master w/ the policy that head should be identitcal to the last release.

paulfalgout:
I'd vote `next`  Could certainly cherry pick it if there is another patch for v2.4

ahumphreys87:
:+1: for `next`

ahumphreys87:
cherrypicked onto next

ahumphreys87:
nice spot @natalan

PR #2549, opened 2015-05-06
Metal - Re-Classify Backbone classes

### LINKED ISSUE #2533: Metal - Classify backbone primatives
We should classify backbone primatives so that `instanceof` `typeof` all work

```diff
+Marionette.Error = Metal.Error.extend({
+  urlRoot: 'http://marionettejs.com/docs/v' + Marionette.VERSION + '/'
+});
+
+function classify(obj) {
+  return Marionette.Class.extend(
+    _.extend({ constructor: obj }, _.omit(obj.prototype, _.keys(Backbone.Events))),
+    _.omit(obj, _.keys(Marionette.Class))
+  );
+}
+
+Marionette.deprecate = Metal.deprecate;
+Marionette.Mixin      = Metal.Mixin;
+Marionette.Events     = Backbone.Events     = new Marionette.Mixin(Backbone.Events);
+Marionette.Model      = Backbone.Model      = classify(Backbone.Model);
```

jasonLaster:
fixes #2533

### DIFF
--- a/src/metal.js
+++ b/src/metal.js
@@ -1,3 +1,23 @@
-// Add Backbone.Metal extensions here
-// 1. Marionette.Class
-// 2. Marionette.Error and Marionette.Deprecate
+// Missing Metal Items
+// 1. Marionette.Class parity
+// 2. Marionette.Deprecate
+
+Marionette.Class = Metal.Class.extend({});
+
+function classify(obj) {
+  return Marionette.Class.extend(
+    _.extend(
+      {constructor: obj},
+      _.omit(obj.prototype, _.keys(Marionette.Class.prototype))
+    )
+  );
+}
+
+Backbone.View         = classify(Backbone.View);
+Marionette.Mixin      = Metal.Mixin;
+Marionette.Events     = Backbone.Events     = new Marionette.Mixin(Backbone.Events);
+Marionette.Model      = Backbone.Model      = classify(Backbone.Model);
+Marionette.Collection = Backbone.Collection = classify(Backbone.Collection);
+Marionette.Router     = Backbone.Router     = classify(Backbone.Router);
+Marionette.History    = Backbone.History    = classify(Backbone.History);
+Marionette.history    = Backbone.history    = new Marionette.History();
--- a/SpecRunner.html
+++ b/SpecRunner.html
@@ -116,6 +116,7 @@
     <script src="test/unit/get-option.spec.js"></script>
     <script src="test/unit/helpers.spec.js"></script>
     <script src="test/unit/merge-options.spec.js"></script>
+    <script src="test/unit/metal.spec.js"></script>
     <script src="test/unit/mixin-underscore-collection.spec.js"></script>
     <script src="test/unit/normalize-methods.spec.js"></script>
     <script src="test/unit/normalize-ui.spec.js"></script>
--- a/test/unit/metal.spec.js
+++ b/test/unit/metal.spec.js
@@ -1,0 +1,27 @@
+describe('metal', function() {
+  describe('classifying backbone primitives', function() {
+    it('view instances are instances of classes', function() {
+      expect(new Marionette.View()).to.be.instanceof(Backbone.View);
+      expect(new Marionette.View()).to.be.instanceof(Marionette.View);
+      expect(new Marionette.View()).to.be.instanceof(Backbone.Metal.Class);
+    });
+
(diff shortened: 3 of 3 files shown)


jasonLaster:
@marionettejs/marionette-core :eyeglasses:

src/metal.js
```diff
@@ -1,3 +1,21 @@
...
+// 1. Marionette.Class parity
+// 2. Marionette.Deprecate
+
+Marionette.Class = Metal.Class.extend({});
+
+function classify(obj) {
+  return Marionette.Class.extend(
+    _.extend({constructor: obj}, _.omit(obj.prototype, _.keys(Backbone.Events))),
+    _.omit(obj, _.keys(Marionette.Class))
+  );
```
thejameskyle: It would be good to omit `constructor` and `initialize` as well since those have the potential of getting wrapped with a function for `_super` in browsers with bad `function.toString`

paulfalgout: I know there's not much going on in this classify function, but all the various extends and omits are hard to keep straight, maybe a bit more wordy for some clarity?
```js
var initKeys = ['constructor','initialize'];
var eventKeys = _.keys(Backbone.Events);
var classKeys = _.keys(Marionette.Class);

function classify(obj) {
  var protoProps = _.extend({constructor: obj}, _.omit(obj.prototype, eventKeys, initKeys));
  var staticProps = _.omit(obj, classKeys, initKeys);

  return Marionette.Class.extend(protoProps, staticProps);
}
```

src/metal.js
```diff
@@ -1,3 +1,21 @@
...
+}
+
+Marionette.Mixin      = Metal.Mixin;
+Marionette.Events     = Backbone.Events     = new Marionette.Mixin(Backbone.Events);
+Marionette.Model      = Backbone.Model      = classify(Backbone.Model);
+Marionette.Collection = Backbone.Collection = classify(Backbone.Collection);
+Marionette.View       = Backbone.View       = classify(Backbone.View);
```
paulfalgout: This doesn't seem right.  `Marionette.View` will get defined elsewhere.

src/metal.js
thejameskyle: You should be able to just do the following since `Backbone.Events` eventually just gets added to `Metal.Class.prototype` directly. I think that will work for `initialize` and `constructor` too. 

```js
_.extend({constructor: obj}, _.omit(obj.prototype, _.keys(Marionette.Class.prototype)));
```

src/metal.js
```diff
@@ -1,3 +1,21 @@
...
+Marionette.Router     = Backbone.Router     = classify(Backbone.Router);
+Marionette.History    = Backbone.History    = classify(Backbone.History);
+Marionette.history    = Backbone.history    = new Marionette.History();
```
thejameskyle: It would be cool if you could use this to check the order is correct.

```js
if (Backbone.history.handlers.length || Backbone.history.started) {
  throw new Error('You must import Marionette before adding routes or starting history`);
}
```

src/metal.js
jasonLaster: yea, that's wrong

src/metal.js
jasonLaster: I basically agree, but I think that's unrelated to this ticket

src/metal.js
jasonLaster: Alright, went with the last approach

src/metal.js
```diff
@@ -1,3 +1,23 @@
...
+// 2. Marionette.Deprecate
+
+Marionette.Class = Metal.Class.extend({});
+
+function classify(obj) {
+  return Marionette.Class.extend(
+    _.extend(
+      {constructor: obj},
+      _.omit(obj.prototype, _.keys(Marionette.Class.prototype))
+    )
```
paulfalgout: Is this line necessary?  https://github.com/thejameskyle/backbone-metal-classify/blob/master/index.js#L13

paulfalgout:
OK  :+1:

ahumphreys87:
:+1: v3.0.0 progress!!! :dancer:

paulfalgout:
I don't know that classify is working as it should.

Attempting to implement `Marionette.Class` it appears that adding a `destroy` function is clobbering the `destroy` function for anything that gets `classify`ed

jasonLaster:
@paulfalgout hmm, mind creating a PR for everything including destroy and then we can debug together?

PR #2552, opened 2015-05-07
Change show render event ordering

scottaj:
This is the change I originally proposed, making before:show actually being the first thing to happen when showing a view.

This fails a bunch of tests (I'm sure travis will show the exact ones) related to showing subviews in the `onBeforeShow` hook. Obviously this doesn't work anymore with this change.

What does everyone think of this change, and if it is good, what should I do about those tests?

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -89,8 +89,6 @@ Marionette.Region = Marionette.Object.extend({
       // we can not reuse it.
       view.once('destroy', this.empty, this);
 
-      this.renderView(view, options);
-
       view._parent = this;
 
       if (isChangingView) {
@@ -99,6 +97,8 @@ Marionette.Region = Marionette.Object.extend({
 
       this.triggerMethod('before:show', view, this, options);
       Marionette.triggerMethodOn(view, 'before:show', view, this, options);
+
+      this.renderView(view, options);
 
       if (isChangingView) {
         this.triggerMethod('swapOut', this.currentView, this, options);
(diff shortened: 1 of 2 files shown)


paulfalgout:
I think the before:show should be the first thing to happen in the show once the show is going to occur.  So rather than moving the view render down, the trigger should move up.  In my mind, this is about the event on the region as much as it is about the event on the view.

scottaj:
@paulfalgout I agree in principle. I was going for the simplest implementation that would pass the test case I set up. 

I have to admit, I was a little hesitant to mess with the order of the destroy and swap events getting fired. I still work in Marionette ~1.8 day to day and the implementation is different there. I wasn't sure what, if any, side effect or assumptions are expected when those events get fired, particularly the swap one.

I can look into it more though.

scottaj:
@paulfalgout The issue you linked is interesting. I agree with what was said in there. I think my perspective is just a little different because I use 1.8, so I don't have swap and attach events `onBeforeAttach` actually sounds like exactly what I needed to avoid the issue that made me open this PR originally (needing to do stuff in a layout when it is shown but before it is in the DOM).

paulfalgout:
This now contains an outdated commit.  The render/show event order is in `next` now.

scottaj:
@paulfalgout I'll rebase it tonight.

ianmstew:
Just pinging this thread to say we are in favor of this change for v3 and will revisit before long.

jasonLaster:
closing this as i don't think we need it anymore

ianmstew:
Even though this has been idle for months, it's one of the TODO items on #2560.

paulfalgout:
This is gonna get handled finally here: https://github.com/marionettejs/backbone.marionette/pull/2850

PR #2556, opened 2015-05-14
Allow `CompositeView` to attach to existing HTML with `template:false`

suprememoocow:
This PR allows `CompositeViews` to be attached to existing HTML by setting `template:false` in much the same way that `ItemView` can, [as documented](http://marionettejs.com/docs/marionette.itemview.html#template-less-itemview)


jasonLaster:
Looks good @suprememoocow. I like the tests and how simple the code is.

Mind opening this against the `next` branch. Also it'd be great to add some tests. 

p.s. I apologize for `jscs` it doesn't know any better.

suprememoocow:
Closed in favour of #2565

PR #2555, opened 2015-05-14
Update Bb / _ dependencies

paulfalgout:
We'd need to update radio, metal, and babysitter, but this is just a start to check out the failing tests and begin the conversation of what it might look like to have the latest dependencies for the v3 launch.

would resolve https://github.com/marionettejs/backbone.marionette/issues/2446

PR #2561, opened 2015-05-16
Update dependencies

### LINKED ISSUE #2446: Too old underscore.js bower dependency
The last version of underscore.js is `1.8.2` but Marionette.js restricts the version with `1.4.4 - 1.6.0`. Is it done intensionally? Can we increase the allowed version of underscore.js?

ahumphreys87:
We'd need to update ~~radio~~, ~~metal~~, and ~~babysitter~~, but this is just a start to check out the failing tests and begin the conversation of what it might look like to have the latest dependencies for the v3 launch.

would resolve #2446

Opening in favour of #2555


paulfalgout:
So these failing tests must be fixed here https://github.com/jashkenas/backbone/pull/3613  correct?

ahumphreys87:
I believe that is the real fix - however we can make our tests bypass this
code in backbone by enabling pushState on the router - do we wanna do that
in our tests? Would allow us to get this released - pretty confident it's
jsDom specific

paulfalgout:
I'll feel better when `1.2.1` comes out fixing:
https://github.com/jashkenas/backbone/pull/3603
https://github.com/jashkenas/backbone/pull/3615

But I don't know that it is worth waiting on either..

paulfalgout:
I think maybe we should go with `1.0.0 - ~1.2` for backbone.  perhaps even `1.4.4 - ~1.8` for underscore.

There will likely be some `1.2` patch fixes in the near future.  But there's more discussion about a Bb 2.0 than a 1.3, so rather than have to manage dependencies for the next few month, I'd bet it'd be safe to go this ~ route.

ahumphreys87:
If we do that we are relying on BB not releasing anything breaking in a minor or patch. Should we risk that?

paulfalgout:
And there we have it.  Bb `1.2.1` released.

ahumphreys87:
Awesome I'll update the deps. How do you feel about us updating wreqr to
support latest BB in 2.x @paulfalgout? I think we should tbh - 3.0 looks a
while away still and we don't want to block people

paulfalgout:
agreed.  those who can't upgrade to `v3` quickly may still want to upgrade to Bb `1.2.1` and I don't see any reason to prevent that.

ahumphreys87:
bumped backbone to 1.2.1 and rebased.

Once this is in ill cherry-pick onto 2.4.2

paulfalgout:
This is great once:
- [ ] https://github.com/marionettejs/backbone.wreqr/pull/67
- [ ] https://github.com/marionettejs/backbone.babysitter/pull/60
- [ ] https://github.com/marionettejs/backbone.radio/pull/225

And for `next` we'll really need to update metal.

ahumphreys87:
I checked metal and it seemed ok. Not specific like ours though.

ianmstew:
Agree this is much needed in `2.4.2`.  With all tests passing I don't see any reason not to merge.  :+1:

PR #2565, opened 2015-05-19
Allow `CompositeView` to attach to existing HTML with `template:false`

suprememoocow:
Same as #2556, except against `next`. This PR allows CompositeViews to be attached to existing HTML by setting template:false in much the same way that ItemView can, [as documented](http://marionettejs.com/docs/marionette.itemview.html#template-less-itemview).

This is useful when the contents of a `CompositeView`  has been prerendered on the server and the client-side needs to attach to the prerendered HTML on startup.


jasonLaster:
Legit

ahumphreys87:
I'm not sure what benefit using CompositeView over ItemView gives us in this instance. Unless I'm missing something???

ahumphreys87:
@marionettejs/marionette-core I'm on the fence with this one - it won't do any harm, but I cant really see the use case. Particularly with changes planned in 3.0

jasonLaster:
I think it's nice, it's also consistent so meets expectations.

samccone:
:+1:

samccone:
would love to get this into a 2.5 release thoughts @marionettejs/marionette-core ?

ahumphreys87:
I think I've been out voted :stuck_out_tongue:, code looks good and has test coverage.

@suprememoocow could you squash then this should be good to merge

samccone:
To me, part of the power with Mn and Backbone is the ability to bind to existing DOM nodes. This seems like a massive win!

One question @suprememoocow I notice that you are not setting up childViews instances to attach to each childNode of the compositeView, that seems like it might be useful... thoughts?

ahumphreys87:
Yeah thats what I was worried about - this implementation provides no functionality above `ItemView` the sorting/filtering wouldn't work so I couldn't see a use case for it.

if it would create child views based on the DOM then that would be great!

samccone:
yeah ^ perhaps a good first step would be to add this to collectionView and then go from there

ianmstew:
I'm pretty sure @suprememoocow IS attaching childView instances to each childNode in his Gitter app--he even presented in NYC [a trick to doing just that](http://www.slideshare.net/mjwbartlett/gitter-marionette-deck) (see slide 20).  I wonder if he can bring that experience to bear here, even it's it's something progressive/game-changing for `3.0`?  It _may_ depend on embedding `data-id`'s in the markup, which would take some discussion.

ahumphreys87:
angulette :boom:

samccone:
well keep in mind you do not need data-id's unless you want to connect it to a backing model. The default behavior would be to instantiate a childView matching the class that the collectionView and or compositeView specified. 

If you wanted to dynamically attach a model you would just have to override getChildView :dancer:

ianmstew:
Good point, @samccone.  I like it.  And if marionette takes one step closer to simple isomorphic javascript... huge win.

ianmstew:
Note self: test against strange child elements of `childViewContainer` el.

jasonLaster:
Manually Merged. thanks!

PR #2566, opened 2015-05-19
Fix ordering of arguments

acauamontiel:
* Fix ordering of arguments in "Module Classes"
* Add missing syntax highlight in some examples


paulfalgout:
Thanks!

PR #2571, opened 2015-05-27
Fixed reordering of collection/composite view bug.

abzainuddin:
**Why**
The collection/composite view have the option to `reorderOnSort`. If this is set to true and the comparator on the collection is changed to sort from asc/desc, the sort on the collection is not correctly rendered in the DOM.

**What**
The views have a `_index` attribute which is not updated during the sort if `reorderOnSort` is set to true. If this is updated after a sort, the sort on the collection is correctly rendered in the DOM.

**How**
`_sortViews()` uses the `view._index` to check if the order has changed. Only if this is true it will trigger `resortView()`. Depending on `reorderOnSort`, either `reorder()` or `render()` is called. `Reorder` will append each childViews el in the correct order in the DOM, but not update its `_index`. On subsequent sorts `_sortViews()` will use an outdated `_index` and fail. The expected behavior is described in the test `when changing the comparator multiple times`. The second added test `should reindex the views` makes sure the views index is updated.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -145,9 +145,12 @@ Marionette.CollectionView = Marionette.View.extend({
     if (modelsChanged) {
       this.render();
     } else {
-      // get the DOM nodes in the same order as the models
-      var els = _.map(models, function(model) {
-        return children.findByModel(model).el;
+      // Get the DOM nodes in the same order as the models and reindex the
+      // views.
+      var els = _.map(models, function(model, index) {
+        var childView = children.findByModel(model);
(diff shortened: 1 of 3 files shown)


ahumphreys87:
thanks for the PR @abzainuddin   this has already been fixed here: https://github.com/ahumphreys87/backbone.marionette/commit/e787839c9173dcd1cd13a5d1dc18e6b75ace1609

It is pending release of 2.4.2 which is very close. You can track its progress here: #2532

Issue #1078, opened 2014-03-31
Model's index in a collection doesn't match itemView index in .children collection


webapps4rim:
Say I have a collection of models like this one:

```
var itemsCollection = new Backbone.Collection([{name: 'item1'}]);
```

And a collection view:

```
var itemsView = new Marionette.CollectionView({
    collection: itemsCollection, 
    itemView: Marionette.View
});
```

Then I add a new item into the collection at index 0

```
itemsCollection.add({name: 'item0'}, {at:0});
```

Now `itemsView` is going to create a new item view with the new model `{name: 'item0'}` but the problem is that it will not get indexed with 0 in the `children` collection

So if I called `itemsView.children.findByIndex(0)` I will get back the view that has the model `{name: 'item1'}` instead of the one that has `{name: 'item0'}`

I can get around this issue if I do `itemsView.children.findByModel(itemsView.collection.at(0))` but I was hopping `itemsView.children.findByIndex(0)` could just work

Thanks

PR #2588, opened 2015-06-10
fix(collection-view): remove by index, not model

jiaweihli:
Collection models can differ from collection view models if
`buildChildView` is overridden.  Using `findByModel` will fail - use
`findByIndex` to avoid this.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -105,9 +105,9 @@ Marionette.CollectionView = Marionette.View.extend({
     }
   },
 
-  // get the child view by model it holds, and remove it
-  _onCollectionRemove: function(model) {
-    var view = this.children.findByModel(model);
+  // get the child view by index, and remove it
+  _onCollectionRemove: function(child, collection, opts) {
+    var view = this.children.findByIndex(opts.index);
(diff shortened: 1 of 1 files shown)


jiaweihli:
`:TODO:` add fail spec.

paulfalgout:
I'm not fully up to speed on this issue, but I know `findByIndex` was possibly going to be deprecated.  https://github.com/marionettejs/backbone.marionette/issues/1078 is the historical related conversation.

jiaweihli:
batman sign: @jmeas @jasonLaster @thejameskyle @samccone

samccone:
to be clear, __if__ you override a public method you can break things. This seems pretty reasonable to me.

Perhaps the real fix here is to improve the docs around what happens when you do change how the method works?

jiaweihli:
Hmm...I guess docs would work.  I rarely expect to break anything when I override a public method that is designated as 'override this for custom behavior'.

samccone:
very true and good point @jiaweihli that typically is the goal, hmmm perhaps this requires some more thinking

PR #2589, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


jasonLaster:
Closing as it's intended to be used in the future https://github.com/bower/bower.json-spec#version

kkirsche:
@jasonLaster please consider what the maintainer says (that it probably will never be used):
http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #2591, opened 2015-06-11
Add collectionView's setFilter() functionality

### LINKED ISSUE #2590: A more efficient CollectionView's Filter()
Hi everyone, I'm a Marionette fan who enjoys your nice framework. My question is about the new 'filter()' functionality (http://marionettejs.com/docs/v2.4.1/marionette.collectionview.html#collectionviews-filter).

In your docs, you encourage me to use `render()` function to 'render again' the `CollectionView` when I have modified the filter to display the changes. However, this behaviour will render the whole `CollectionView` again, when I only need to remove or add some `childrenView`. Maybe an alternative function like this could be better that a simple `render`(in pseudocode):

```javascript
/*
if the filter is the same than before, return
else
   for each model in collectionView
      if filter(model) is true
         if  not existRepresentation(model,view)
           this._onCollectionAdd(model)
      else
         if existRepresentation(model, view)
           this._onCollectionRemove(model)
*/
```

What are your thoughts? Thanks in advance!

frankcortes:
fixes #2590

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -128,6 +128,40 @@ Marionette.CollectionView = Marionette.View.extend({
     return this;
   },
 
+  // An efficient rendering used for filtering. Instead of modifying
+  // the whole DOM for the collection view, we are only adding or
+  // removing the related childrenViews.
+  setFilter: function(filter) {
+    if (!filter || this.filter === filter) {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


paulfalgout:
Pull against `next` please.  Sorry for the confusion.

frankcortes:
Ok, I close this PR. Thank you!

PR #2595, opened 2015-06-13
Added options parameter to showChildView()

### LINKED ISSUE #2572: showChildView() options
I have a situation where I have a region and its view can be one of two Views, which may be swapped later.

```
onRender: function() {
    var view = this.textView;
    if (model.get('someproperty') === true) {
        view = this.formView;
    }
    this.showChildView('myregion', view, {preventDestroy: true});
},
onDestroy: function() {
    this.view1.destroy();
    this.view2.destroy();
}
````

Turns out `showChildView` does not take an `options` hash. Is this by design?

### PR DESCRIPTION
**Pull request by jbarreiros**
Fixes #2572.

> ... `showChildView` does not take an `options` hash.

This PR changes `showChildView(view)` to `showChildView(view, options)` to provide consistency with `Marionette.Region.show(view, options)`.

I'm new to writing unit tests, so please let me know if the unit test needs adjusting.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -160,8 +160,9 @@ Marionette.View = Marionette.AbstractView.extend({
     return this.regionManager.removeRegion(name);
   },
 
-  showChildView: function(regionName, view) {
-    return this.getRegion(regionName).show(view);
+  showChildView: function(regionName, view, options) {
+    var region = this.getRegion(regionName);
+    return region.show.apply(region, _.rest(arguments));
   },
 
   getChildView: function(regionName) {
--- a/docs/marionette.layoutview.md
+++ b/docs/marionette.layoutview.md
@@ -75,17 +75,17 @@ Once you've rendered the layoutView, you now have direct access
 to all of the specified regions as region managers.
 
 ```js
-layoutView.getRegion('menu').show(new MenuView());
+layoutView.getRegion('menu').show(new MenuView(), options);
 
-layoutView.getRegion('content').show(new MainContentView());
+layoutView.getRegion('content').show(new MainContentView(), options);
 ```
 
 There are also helpful shortcuts for more concise syntax.
 
 ```js
-layoutView.showChildView('menu', new MenuView());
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by paulfalgout** [comment]
This likely needs a little bit of modification to documentation

File: src/view.js [hunk 1]
```diff
@@ -144,8 +144,8 @@ Marionette.View = Marionette.AbstractView.extend({
     return this.regionManager.removeRegion(name);
   },
 
-  showChildView: function(regionName, view) {
-    return this.getRegion(regionName).show(view);
+  showChildView: function(regionName, view, options) {
+    return this.getRegion(regionName).show(view, options);
```
samccone: lets do 

```
this.getRegion(regionName).show.apply(this, _.rest(arguments)
```

samccone:
@jbarreiros mind adding docs and we can :ship: this

src/view.js
jasonLaster: yep, @jbarreiros the benefit is that we can mantain the airity of the options

jbarreiros: Good idea. Updated.

jbarreiros:
@samccone Looking for a little direction on where to update the documentation. Do we just want to highlight that `options` can be passed to both `Region.show` and `Region.showChildView`?

samccone:
that would be perfect @jbarreiros

jasonLaster:
ping @jbarreiros - have a minute to update the docs to include `options`?

jbarreiros:
Added the updates to the docs. Sorry about the extreme delay.

In writing the docs and reading through the code, I noticed that `options` is [not passed to `empty()`](https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L74). This actually contradicts the docs: 
selection_002

I cleaned up the line mentioning "you also have access to the `options` ...".

ahumphreys87:
lgtm :+1: @samccone @jasonLaster  could we get another set of :eyes:

PR #2602, opened 2015-06-17
Fix underscore compatibility version

jthomerson:
In Marionette 2.4.2 the underscore dependency was bumped to be compatible with up to 1.8.3, but the readme file did not get updated.


jthomerson:
Please note: there's also a discrepency on the backbone version between readme and package.json, but due to the wording in the readme file I wasn't sure if the backbone compatibility should be changed there.

paulfalgout:
sure enough.  seems like backbone should be updated as well.

PR #2608, opened 2015-06-20
Added 'reverse' option for CollectionView and CompositeView that displays items in reverse order.

AaronNGray:
This patch adds a 'reverse' option for CollectionView's and CompositeView's that displays items in reverse order.

The reasoning behind this is that a reverse ordering is a special case and that filter 
sorting items in Backbone.Collection's into reverse order can take too much time to do leading to a noticeable delay in displaying views.

The patch has been tested.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -543,7 +543,7 @@ Marionette.CollectionView = Marionette.View.extend({
     var elBuffer = document.createDocumentFragment();
     if (this.getOption('reverse')) {
       _.each(this._bufferedChildren, function(b) {
-        elBuffer.prependChild(b.el);
+        elBuffer.insertBefore(b.el, elBuffer.firstChild);
       });
     } else {
       _.each(this._bufferedChildren, function(b) {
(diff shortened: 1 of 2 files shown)


jasonLaster:
I think this could be nice. Mind adding unit tests and docs

AaronNGray:
Ah right 'next' okay I will do that, where does it say about that. Regarding unit tests, how do I go about that I am not sure how to test such a thing ? And documentation also not totally sure about that too. Are there any guides for adding features ?

samccone:
Hmm I am still not sold, why could we not just use the `viewcomparator ` http://marionettejs.com/docs/v2.4.2/marionette.collectionview.html#collectionviews-viewcomparator 

If you want to change the order of your items in your view layer, this currently is the "semantic" way to do it. What do you think @jasonLaster @AaronNGray

AaronNGray:
I have tried 'viewcomparator' and its too slow on even just 256 items.
Reverse seems like a special case to me.

samccone:
Hey @AaronNGray mind creating a reduced test case for us to evaluate this, I am not sure I 100% follow your use, and if it is applicable to enough people to go in core.

Marionette is an **extension** lib, so worst case you can override a method given your needs.

AaronNGray:
Ah okay I will just override then. Its only a few extra lines of code to
the code base and for anything like a feed or a log its essential.

samccone:
thanks for the PR and idea @AaronNGray, keep them coming!

PR #2612, opened 2015-06-24
Fix region.empty({preventDestroy: false}) bug.

### LINKED ISSUE #2611: Region.empty fails to destroy view when called with any options

```
region.show(view);
region.empty({});
```

```
```

bradlord:
Fixes #2611. I wasn't sure which branch to request this to be merged into. CONTRIBUTING.md suggests making changes to the minor branch, but that doesn't seem to have been updated recently.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -207,7 +207,8 @@ Marionette.Region = Marionette.Object.extend({
   empty: function(options) {
     var view = this.currentView;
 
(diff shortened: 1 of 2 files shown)


bradlord:
After some more digging around, it probably makes more sense to request to have this merged into master. I did so in pull request #2613, so I'll close this one.

PR #2613, opened 2015-06-24
Fix region.empty({preventDestroy: false}) bug.

### LINKED ISSUE #2611: Region.empty fails to destroy view when called with any options

bradlord:
Fixes #2611.

I wasn't sure which branch to request this to be merged into. CONTRIBUTING.md suggests making changes to the minor branch, but that doesn't seem to have been updated recently. I originally requested (#2612) that this be merged into the 2.4.2 branch, but after looking more closely, I realized that didn't make much sense.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -207,7 +207,8 @@ Marionette.Region = Marionette.Object.extend({
   empty: function(options) {
     var view = this.currentView;
 
-    var preventDestroy = Marionette._getValue(options, 'preventDestroy', this);
+    var emptyOptions = options || {};
+    var preventDestroy  = !!emptyOptions.preventDestroy;
     // If there is no view in the region
     // we should not remove anything
     if (!view) { return; }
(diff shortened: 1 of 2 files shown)


jasonLaster:
sorry for the confusion @bradlord. @ianmstew thoughts?

samccone:
@bradlord can you add the test case initially outlined in your origin issue please?

ianmstew:
@bradlord We have the correct patch branch sorted out.  Move this to the new `2.4.3` branch and we will move forward with merging.  That should be seamless for you since `master` and `2.4.2` have the same HEAD at this moment.

ianmstew:
Update: PR against `patch` instead.  This should still be seamless.

bradlord:
I created a new PR (#2631) against `patch`.

PR #2614, opened 2015-06-24
Simplify Travis Matrix

### LINKED ISSUE #2609: Simplify travis-ci testing matrix
We should remove node 0.11.
Testing should likely be done mostly on 0.12 with a few on 0.10 and on io.js

Example:
https://github.com/marionettejs/backbone-metal/blob/master/.travis.yml#L7

megawac:
Fixes #2609 

There doesn't seem to be any reason to run travis on multiple versions of node as `Marionette` runs primarily on the browser. Currently we're running **36** different travis VMs per commit. This brings it down to **12**.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ sudo: false
 language: node_js
 node_js:
   - "0.12"
-  - "0.11"
-  - "0.10"
 env:
   global: MAINRUN=true
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,25 @@ sudo: false
 language: node_js
 node_js:
   - "0.12"
-env:
-  global: MAINRUN=true
-  matrix:
-    - UNDERSCORE=1.4.4 BACKBONE=1.0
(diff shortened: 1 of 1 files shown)


megawac:
/cc @ahumphreys87

If we want to test multiple versions of node we can change the matrix on a per test level (I think we do that in Radio)

jasonLaster:
yea, totally agree. It'll speed stuff up for sure

samccone:
lets keep node 10 around plz enough people are still on it, that makes it important to ensure those people can contribute to the source

paulfalgout:
Yep I think the reason to try node 10.. or _maybe_ io is specifically for dev environments.  For a while 0.12 wasn't working for us at all.. which I think is why the matrix was changed to begin with, but we don't need to test everything with all 3..  just 1

megawac:
In that case, do you prefer this?

megawac:
There we go, fixed it

samccone:
:+1:

paulfalgout:
:+1:

PR #2616, opened 2015-06-24
fixes Issue #2615 by making sure all arguments are passed through Mar…

### LINKED ISSUE #2615: Marionette.Application does not pass all arguments to .initialize()
I'm working on upgrading from 2.2 to 2.4, and ran into an issue based on the new inheritance structure released in 2.3. I've got a place where I'm instantiating a Marionette.Application, and am passing in two arguments which are referenced inside the Application's `initialize()`. This worked fine in 2.2, but broke in 2.3 - only the first argument gets passed through.

I've tracked the issue to this line inside Marionette.Application's `constructor()`:
```
Marionette.Object.call(this, options);
```
I think if this were changed to `Marionette.Object.apply(this, arguments);` it'd remain backwards-compatible with the earlier release, without hurting anything. I'll go ahead and get PR together with the change, but did want to make sure the change wasn't intentional for some reason.

kgoggin:
Fixes #2615 

Uses `apply` instead of `call` inside Marionette.Application's constructor to make sure all arguments are passed up the chain to Marionette.Object, and thus to the `initialize()` method.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -11,7 +11,7 @@ Marionette.Application = Marionette.Object.extend({
     this.submodules = {};
     _.extend(this, options);
     this._initChannel();
-    Marionette.Object.call(this, options);
+    Marionette.Object.apply(this, arguments);
   },
 
   // Command execution, facilitated by Backbone.Wreqr.Commands
--- a/test/unit/application.spec.js
+++ b/test/unit/application.spec.js
@@ -5,7 +5,7 @@ describe('marionette application', function() {
     beforeEach(function() {
       this.fooOptions = {foo: 'bar'};
       this.appOptions = {baz: 'tah'};
-      this.extraArg = 'qux'
+      this.extraArg = 'qux';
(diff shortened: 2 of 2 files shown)


kgoggin:
@megawac done. Hope it was okay to just update the existing one...

jasonLaster:
looks close @kgoggin - 

small travis bug
https://travis-ci.org/marionettejs/backbone.marionette/jobs/68210517

paulfalgout:
the travis-bug is already on the `minor` branch.

samccone:
this looks good to go yeah? seems like a good reason to do 2.4.3?

kgoggin:
Yeah, I noticed that code style linting error but it was definitely there on the branch. Thought about fixing it but wasn't sure about the preferred style. It seems like a false positive to me. That said, let me know if you want me to try to address it and I'd be happy to.

paulfalgout:
Can you pull and squash this against the `next` branch, please?  Then we'll cherry-pick this onto a patch branch later.

PR #2617, opened 2015-06-24
Allow a user to define trigger handlers in options.

samccone:
For example

```js
new myView({
  onRender: function(){}
})
```

As compared to prior to this a user having to define these methods on
the actual prototype chain such as

```js
var MyView = Mn.ItemView.extend({
  onRender: function(){}
})
```


ahumphreys87:
should we add docs to point out this is now possible?

ahumphreys87:
:+1:

ahumphreys87:
manually merged

PR #2618, opened 2015-06-24
📝 Fix backbone and underscore readme version

samccone:
via #2602


jasonLaster:
yup - looks good!

samccone:
:+1:

PR #2619, opened 2015-06-24
Region.empty doesn't fail to destroy view when called with any options

### LINKED ISSUE #2611: Region.empty fails to destroy view when called with any options

frankcortes:
This is a simple pull request to fix #2611


frankcortes:
This PR was duplicate

PR #2621, opened 2015-06-25
Fix bug in metal configuration

megawac:
I was looking over some of the changes and I'm pretty sure this is a mistake (/cc @thejameskyle) as `Marionnette.Class.prototype` will have no keys (as a result of calling `Metal.Class.extend()`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "backbone.babysitter": "^0.1.0",
     "backbone.radio": "^0.9.0",
     "backbone": "1.0.0 - 1.2.1",
-    "backbone-metal": "^0.5.0",
+    "backbone-metal": "^0.6.0",
     "underscore": "1.4.4 - 1.8.3"
   },
   "devDependencies": {
--- a/src/metal.js
+++ b/src/metal.js
@@ -4,11 +4,12 @@
 
 Marionette.Class = Metal.Class.extend({});
 
+var props = _.keys(Metal.Class.prototype);
 function classify(obj) {
   return Marionette.Class.extend(
     _.extend(
       {constructor: obj},
-      _.omit(obj.prototype, _.keys(Metal.Class.prototype))
+      _.omit(obj.prototype, props)
     )
   );
 }
--- a/src/metal.js
+++ b/src/metal.js
@@ -8,7 +8,7 @@ function classify(obj) {
   return Marionette.Class.extend(
     _.extend(
       {constructor: obj},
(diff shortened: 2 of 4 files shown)


thejameskyle:
Yup, would you mind taking a look at this as well: https://github.com/marionettejs/backbone.marionette/pull/2549#issuecomment-113992945

megawac:
Looks like some interesting build failures in old underscore xD

jasonLaster:
nice catch. heh - not sure why travis is sad...

megawac:
This is going to drive me mad. Why do you have to do this @thejameskyle?!!

Anyway there are so many issues and I'm not exactly sure what is causing them.

jasonLaster:
heh - @megawac why aren't you running the tests locally?

megawac:
I have been, it keeps failing on different versions of underscore and
passing on others

paulfalgout:
Sorry I'm a little late in checking this out, but isn't the point of `Marionette.Class` to add keys that then get `classify`ed?  Essentially so that base utility functions can be added everywhere within Marionette?

megawac:
@paulfalgout the `Marionettte.Class.prototype` has no immediate properties as it inherits from `Metal.Class`, thus `keys(Marionette.Class.prototype)` is empty and the classify does nothing currently

paulfalgout:
Very true, but I think that was intentional as part of an effort to make the initial commit small.  And keys would then be added in an additional commit to `Marionette.Class`.  @jasonLaster?

megawac:
That makes adding Metal kind of useless as its currently not doing anything. Anyway the bug seems to be in metal not supporting old as hell underscore

megawac:
Actually, it seems to be propogating from `Backbone.View` change where its no longer compatible with old underscore

megawac:
Alright I've tracked it down to the stupidest bug: `_.noop` doesn't exist in old underscore lol

paulfalgout:
Ah wait.. I think I'm tracking now.. Metal.extend, mixin, include, etc aren't getting to the `classify`ed things?  But this change will still add things added to `Marionette.Class` to all of the various things.

paulfalgout:
Not sure why travis isn't catching Metal `0.6.0` but you might as well up the version in package and bower as `0.5.0` won't pass tests.

ahumphreys87:
@megawac you free to bump version in package.json and bower to see if tests pass?

ahumphreys87:
@megawac any ideas on failures here?

jasonLaster:
closing as this is 3.x work

PR #2625, opened 2015-06-25
view can replace regions's el

ahumphreys87:
Thanks to @trezy for the original PR.

Ive added some tests and fixed an issue where `replaceView` could clash with `preventDestroy`

ping @marionettejs/marionette-core thoughts on getting this into 2.5 ready for CompositeView removal in 3.0?

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -49,6 +49,7 @@ Marionette.Region = Marionette.Object.extend({
     var showOptions     = options || {};
     var isDifferentView = view !== this.currentView;
     var forceShow       = !!showOptions.forceShow;
+    var replaceElement  = !!showOptions.replaceElement;
 
     // We are only changing the view if there is a current view to change to begin with
     var isChangingView = !!this.currentView;
@@ -61,6 +62,9 @@ Marionette.Region = Marionette.Object.extend({
     // the current view or if we want to re-show the view. Note that if
     // `_shouldDestroyView` is true, then `_shouldShowView` is also necessarily true.
     var _shouldShowView = isDifferentView || forceShow;
+
+    // only replace the region's element with the view's element if explicitly set
+    var _shouldReplaceElement = replaceElement;
 
     if (isChangingView) {
       this.triggerMethod('before:swapOut', this.currentView, this, options);
@@ -122,7 +126,7 @@ Marionette.Region = Marionette.Object.extend({
         this._triggerAttach(displayedViews, 'before:');
       }
 
-      this.attachHtml(view);
+      this.attachHtml(view, _shouldReplaceElement);
       this.currentView = view;
 
       if (attachedRegion && attachOptions.triggerAttach) {
@@ -204,12 +208,50 @@ Marionette.Region = Marionette.Object.extend({
     return Backbone.$(el, Marionette._getValue(this.options.parentEl, this));
   },
 
-  // Override this method to change how the new view is
-  // appended to the `$el` that the region is managing
-  attachHtml: function(view) {
+  // Replace the region's DOM element with the view's DOM element.
+  _replaceEl: function(view) {
+    // empty el so we don't save any non-destroyed views
     this.$el.contents().detach();
 
-    this.el.appendChild(view.el);
+    // always restore the el to ensure the regions el is
+    // present before replacing
+    this._restoreEl();
+
+    var parent = this.el.parentNode;
+
+    parent.replaceChild(view.el, this.el);
+    this.replaced = true;
+  },
+
+  // Restore the region's element in the DOM.
+  _restoreEl: function() {
+    if (!this.currentView) {
+      return;
+    }
+
+    var view = this.currentView;
+    var parent = view.el.parentNode;
+
+    if (!parent) {
+      return;
+    }
+
+    parent.replaceChild(this.el, view.el);
+    this.replaced = false;
+  },
+
+  // Override this method to change how the new view is
+  // appended to the `$el` that the region is managing
+  attachHtml: function(view, shouldReplace) {
+    if (shouldReplace) {
+      // replace the region's node with the view's node
+      this._replaceEl(view);
+    } else {
+      // empty the node and append new view
+      this.$el.contents().detach();
+
+      this.el.appendChild(view.el);
+    }
   },
 
   // Destroy the current view, if there is one. If there is no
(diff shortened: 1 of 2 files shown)


jasonLaster:
:nut_and_bolt: :wrench: :hammer:

jmeas:
Curious to hear @jridgewell's thoughts on this implementation.

ahumphreys87:
yup it's a worryingly simple implementation so would be good to get more opinions. I think the tests cover most cases though

src/region.js
```diff
@@ -204,12 +208,38 @@ Marionette.Region = Marionette.Object.extend({
     return Backbone.$(el, Marionette._getValue(this.options.parentEl, this));
   },
 
+  // Replace the region's DOM element with the view's DOM element.
+  replaceEl: function(view) {
+    // empty el so we don't save any non-destroyed views
+    this.$el.contents().detach();
```
samccone: worrying about leaks here

test/unit/region.spec.js
```diff
@@ -466,6 +497,65 @@ describe('region', function() {
...
+        });
+
+        it('should restore the region\'s "el" to the DOM', function() {
+          expect(this.$parentEl).to.contain.$html('<div id="region"></div>');
+        });
+      });
+
+      describe('and showing another view', function() {
+        beforeEach(function() {
+          this.MyView2 = Marionette.View.extend({
```
samccone: any reason why this is just a generic view?

src/region.js
jridgewell: I'm curious why this is needed. The comment seems to point to a subview inside this regions element being `#remove`d, but that'll just work naturally with `$#remove`.

trezy: I'm with @samccone and @jridgewell. If you look at [my original PR](https://github.com/marionettejs/backbone.marionette/commit/fe0d47fdc6162484bff2c97918ef2720e05ea4a0#diff-b8da5292cb07ec757ac64e78ff5af5c1R143) I didn't do anything fancy. I just replaced the old `el` with the new `el` so that we retain references to everything but we're not storing anything unnecessary.

src/region.js
```diff
@@ -204,12 +208,38 @@ Marionette.Region = Marionette.Object.extend({
...
   // appended to the `$el` that the region is managing
-  attachHtml: function(view) {
-    this.$el.contents().detach();
+  attachHtml: function(view, shouldReplace) {
+    if (shouldReplace) {
+      // replace the region's node with the view's node
+      this.replaceEl(view);
+    } else {
+      // empty the node and append new view
+      this.$el.contents().detach();
```
trezy: Same as @samccone's comment above. We should be rebinding events and such when we do this so `detach` could leave unwanted data in the nether whereas `remove` should cover our GC needs.

src/region.js
```diff
@@ -204,12 +208,38 @@ Marionette.Region = Marionette.Object.extend({
...
+  },
+
   // Override this method to change how the new view is
   // appended to the `$el` that the region is managing
-  attachHtml: function(view) {
-    this.$el.contents().detach();
+  attachHtml: function(view, shouldReplace) {
+    if (shouldReplace) {
+      // replace the region's node with the view's node
+      this.replaceEl(view);
```
jridgewell: I'm not sure this is can be called twice, which is a valid usecase. Seems `#replaceEl` will try replacing the `this.el` with the new view's `el` and, since `this.el.parentNode = null` the second time around, an error will be thrown.

src/region.js
ahumphreys87: Ok so this was added to handle the case where replaceEl was used with preventDestroy.

If we don't remove the old view in here when we restore it the old view will come back as we haven't destroyed it.

I just took this from attachHtml - I'll try with remove

src/region.js
ahumphreys87: As above - detach is used in the current implementation of attachHtml I've just moved it to the else case here

test/unit/region.spec.js
paulfalgout: This is `next` so `View === ItemView === LayoutView`?  and the old `View` is now `AbstractView`

trezy:
I'm glad to finally see this getting into core! I won't have to use my shim anymore. :-)

I can also say that I've been using my shim since I made the original PR which would have been long enough for a child to gestate and I've had zero issues with it. Performance is good and I haven't seen any view collisions.

jdaudier:
@trezy length of a child gestation... lol

src/region.js
trezy: Alternatively we could detach and place the results in storage, we just need to make sure that view gets destroyed at some point for proper garbage collection.

ahumphreys87: yeah thats the thing with `preventDestroy` if that is used Mn has no power over memory leaks, personally I hate the option (views should be recreated) but it is used quite a bit.

I think for 2.5 we'll have to stick with `detach` here. For 3.0 im open to removing the `preventDestroy` function but that needs some discussion. Ill open an issue

src/region.js
ahumphreys87: ah good catch - ill add a test case and fix!

src/region.js
trezy: A better option would probably be to prevent using both options together. `replaceEl` + `preventDestroy` could throw an error or, at least, should throw a `console.info` that warns the dev about it. I hate the idea of building in a feature that almost encourages poor performance.

trezy: After talking to @ahumphreys87 in Gitter I think we're fine here. `preventDestroy` is a tool specifically for creating memory leaks so I think it's all good. ;-)

ahumphreys87:
@samccone @jasonLaster bring it on!!!!!

src/region.js
ahumphreys87: @jridgewell added test and modifed `replaceEl` to cater for this case.

jasonLaster:
Looks good @ahumphreys87. :+1: Lets move forward into the future

ahumphreys87:
@marionettejs/marionette-core needs 1 more :smile:

paulfalgout:
I think `restoreEl` and `replaceEl` should probably be private.  Otherwise :+1:

ahumphreys87:
@paulfalgout methods now private

jasonLaster:
At long last :runner: :athletic_shoe:

trezy:
Woooooooooohooooooooo! No more shims for meeeeee!

howardroark:
Now I'm just waiting for the PR named "template can replace view's el" :) :)

PR #2626, opened 2015-06-25
adds support for Object.isDestroyed

### LINKED ISSUE #2620: Add isDestroyed flag to Mn.Object/Mn.Class
When an object is destroyed, it should be marked with `isDestroyed = true` in the same way that views are.

Use case: I have an object which calls one of its methods on a 5-second `_.delay` after being instantiated. If the object is destroyed in the mean time, that delayed method needs a way to detect that the object has been destroyed and return early.

Workaround: I've added an `onDestroy` method to my controller and applied the `isDestroyed` flag myself.

(I understand that Marionette.Object is getting replaced by Marionette.Class in a future version; the same issue applies.)

pascalpp:
[fixes #2620]

also adds 2 tests to object.spec.js

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -19,11 +19,20 @@ Marionette.Object.extend = Marionette.extend;
 _.extend(Marionette.Object.prototype, Backbone.Events, {
   cidPrefix: 'mno',
 
+  // for parity with Marionette.AbstractView lifecyle
+  isDestroyed: false,
+
   //this is a noop method intended to be overridden by classes that extend from this base
   initialize: function() {},
 
   destroy: function() {
+    if (this.isDestroyed) { return this; }
+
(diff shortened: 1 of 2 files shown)


pascalpp:
I'm not sure what to make of that build failure.

pascalpp:
I think it was just a blip with curl. Is there a way to retrigger Travis?

pascalpp:
@ahumphreys87 Can you review this PR and let me know if there's anything else I should do?

jasonLaster:
@pascalpp you can bump it in travis in the top right

jasonLaster:
looks good to me... thoughts crew?

ahumphreys87:
:+1:

ahumphreys87:
manually merged

PR #2607, opened 2015-06-19
Add ability to check if layout contains region

jisaacks:
fixes #2594

PR #2627, opened 2015-06-26
Add ability to check if layout contains region

### LINKED ISSUE #2594: RegionManager#hasRegion(regionName)
I was surprised to find that this doesn't exist!

`return Boolean(this.getRegion(regionName));`

Also, it goes without saying, but it should be added to the LayoutView prototype.

jisaacks:
fixes #2594
Also see #2607

### DIFF
--- a/api/layout-view.yaml
+++ b/api/layout-view.yaml
@@ -337,6 +337,25 @@ functions:
           layoutView.removeRegion("foo");
           ```
 
+  hasRegion:
+    description: |
+      Checks to see if the LayoutView contains a specified region.
+      @api public
+      @param {String} name - The name of the region being looked up.
+      @returns {Boolean} Whether or not the LayoutView has the region.
+    examples:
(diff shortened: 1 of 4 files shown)


megawac:
Looks good :+1:

ahumphreys87:
lgtm :+1:

ahumphreys87:
manually merged

jmeas:
If we don't plan to get rid of RegionManager in v3, then this method should prob. be added to the RegionManager instead and proxied like the other methods.

ahumphreys87:
@jmeas yeah we should keep this that in mind - Im quite keen on getting rid of it though. Especially now that _all_ views have `regions` I can't think of a major use case.

Also when #2625 lands it will clear the way to kill `CompositeView`  - we can replicate its functionality with a `region` in `View`. The dream of 2 views will become a reality!

PR #2635, opened 2015-06-27
fixes #2578

### LINKED ISSUE #2578: Reverse order in collection when items added with 'at' option
I think that I find a bug. Suppose you have a collection, associated with CollectionView instance. When you add a bunch of items (not one by one, but via array of models) to collection with 'at' option (to insert items at specified position) supplied, then child views added in reverse order. [Example on jsfiddle](https://jsfiddle.net/e76hak6g/5/). Code:

```javascript
var ChildView = Mn.ItemView.extend({
    tagName: 'li',
    template: function(data) { return data.name; }
});

var ContainerView = Mn.CollectionView.extend({
    tagName: 'ul',
    childView: ChildView
});

var list = new Backbone.Collection([
    new Backbone.Model({name: 'initial view 1'}),
    new Backbone.Model({name: 'initial view 2'}),
    new Backbone.Model({name: 'initial view 3'})
]);

var view = new ContainerView({collection: list});
view.render();
$('body').append(view.$el);

var additions = [
    new Backbone.Model({name: 'addition view 1'}),
    new Backbone.Model({name: 'addition view 2'}),
    new Backbone.Model({name: 'addition view 3'})
];

list.add(additions, {at: 2});
```

The result is:

```
initial view 1
initial view 2
addition view 3
addition view 2
addition view 1
initial view 3
```

But I'm expecting that it would be

```
initial view 1
initial view 2
addition view 1
addition view 2
addition view 3
initial view 3
```

Regards

cybernetlab:
actually, when `Backbone.Collection#add ` method called with a bunch of models, backbone triggers `add` event sequentially for each model, so where are no way to determine inside the `_onCollectionAdd` event handler whether one or many models are added. But backbone triggers `update` event after a number of `add` events, once per `Backbone.Collection#add` call. So I decide to add a "private" `_atIndex` flag to indicate this situation. I've add `_onCollectionUpdate` handler to reset this flag after one or more models have been added.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -90,6 +90,7 @@ Marionette.CollectionView = Marionette.View.extend({
       this.listenTo(this.collection, 'add', this._onCollectionAdd);
       this.listenTo(this.collection, 'remove', this._onCollectionRemove);
       this.listenTo(this.collection, 'reset', this.render);
+      this.listenTo(this.collection, 'update', this._onCollectionUpdate);
 
       if (this.getOption('sort')) {
         this.listenTo(this.collection, 'sort', this._sortViews);
@@ -97,11 +98,24 @@ Marionette.CollectionView = Marionette.View.extend({
     }
   },
 
+  // Handle a Backbone.Collection#add method finished (called once after a
+  // number of sequential _onCollectionAdd calls)
+  _onCollectionUpdate: function() {
+    this._atIndex = null;
+  },
+
   // Handle a child added to the collection
   _onCollectionAdd: function(child, collection, opts) {
     var index;
     if (opts.at !== undefined) {
-      index = opts.at;
+      if (this._atIndex) {
+        // correct child index if a number of models added to collection at once
+        this._atIndex += 1;
+        index = this._atIndex;
+      } else {
+        index = opts.at;
+        this._atIndex = index;
+      }
     } else {
       index = _.indexOf(this._filteredSortedModels(), child);
     }
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -100,8 +100,8 @@ Marionette.CollectionView = Marionette.View.extend({
   // Handle a child added to the collection
   _onCollectionAdd: function(child, collection, opts) {
     var index;
-    if (opts.index !== undefined && !this.getOption('filter')) {
-      index = opts.index;
+    if ((opts.index !== undefined || opts.at) && !this.getOption('filter')) {
+      index = opts.index || opts.at;
     } else {
       index = _.indexOf(this._filteredSortedModels(), child);
     }
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -101,7 +101,7 @@ Marionette.CollectionView = Marionette.View.extend({
   _onCollectionAdd: function(child, collection, opts) {
     var index;
     if (opts.at !== undefined && !this.getOption('filter')) {
-      index = opts.index || opts.at;
+      index = opts.index || _.indexOf(this._filteredSortedModels(), child);
     } else {
       index = _.indexOf(this._filteredSortedModels(), child);
     }
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -100,7 +100,7 @@ Marionette.CollectionView = Marionette.View.extend({
   // Handle a child added to the collection
   _onCollectionAdd: function(child, collection, opts) {
     var index;
-    if ((opts.index !== undefined || opts.at) && !this.getOption('filter')) {
+    if (opts.at !== undefined && !this.getOption('filter')) {
       index = opts.index || opts.at;
     } else {
       index = _.indexOf(this._filteredSortedModels(), child);
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -100,8 +100,8 @@ Marionette.CollectionView = Marionette.View.extend({
   // Handle a child added to the collection
   _onCollectionAdd: function(child, collection, opts) {
     var index;
-    if (opts.at !== undefined) {
-      index = opts.index !== undefined ? opts.index : collection.indexOf(child);
+    if (opts.index !== undefined && !this._isFilteredOrOrdered()) {
+      index = opts.index;
     } else {
       index = _.indexOf(this._filteredSortedModels(), child);
     }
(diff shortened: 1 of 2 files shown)


paulfalgout:
Hmm `update` is only available in Backbone `1.2.x`.
Can this be patched in way that supports Backbone `1.0.0`?

cybernetlab:
Yes, you right - here is last lines from `Backbone.Collection.set` (that is called by `Backbone.Collection.add`) for version `1.0.0`:

```js
      // Trigger `add` events.
      for (i = 0, l = toAdd.length; i < l; i++) {
        (model = toAdd[i]).trigger('add', model, this, options);
      }

      // Trigger `sort` if the collection was sorted.
      if (sort) this.trigger('sort', this, options);
      return this;
```

So, I have no idea how to handle situation of adding a list of model at once for `1.0.0` :(

cybernetlab:
Simpler is better! Thanks to @jridgewell for hint: 1.2 has `index` options in add event. So - just use it or `at`... As I say above it resolves bug only for backbone >= 1.2.0.

test/unit/collection-view.spec.js
```diff
@@ -467,6 +467,39 @@ describe('collection view', function() {
...
+      this.collection.add([this.model5, this.model6], {at: 2});
+    });
+
+    it('should add models and render views in right order', function() {
+      var order = _.pluck(this.collectionView.$el.find('span'), 'innerHTML');
+      var backboneVersion = _.map(Backbone.VERSION.split('.'), function(x) {
+        return parseInt(x);
+      });
+      if (backboneVersion[0] >= 1 && backboneVersion[1] >= 2) {
+        expect(order).to.deep.equal(['1', '3', '5', '6', '4', '2']);
```
jasonLaster: can we make these numbers mean something. My hope is that current backbone should have sequential numbers 1..6 and maybe older backbone has busted numbers... now they're both funky

jasonLaster:
I like this change! don't forget to squash when you're done

src/collection-view.js
```diff
@@ -101,7 +101,7 @@ Marionette.CollectionView = Marionette.View.extend({
   _onCollectionAdd: function(child, collection, opts) {
     var index;
     if (opts.at !== undefined) {
-      index = opts.at;
+      index = opts.index || opts.at;
```
jridgewell: This is still a bug when using earlier versions of Backbone (and, it'll mess up when `opts.index` is 0).

cybernetlab: As I wrote above:
> actually, when Backbone.Collection#add method called with a bunch of models, backbone triggers add event sequentially for each model, so where are no way to determine inside the _onCollectionAdd event handler whether one or many models are added

I think that there are no way to fix it for Backbone prior to 1.2.0
Replacing `at` with `index` will brake code for Backbone prior to 1.2.0
when `opts.index` is `0`, `opts.at` is `0` too...

jridgewell: > I think that there are no way to fix it for Backbone prior to 1.2.0

Falling back to the `_.indexOf` search when `opts.index === undefined` should work.

> when opts.index is 0, opts.at is 0 too...

You're right, ignore that part of my comment.

cybernetlab:
Where was an error in [travis](https://travis-ci.org/marionettejs/backbone.marionette/jobs/68989771) but I think that it's random and irrelevant to my changes

samccone:
I kicked a rebuild :volcano:

src/collection-view.js
cybernetlab: Nice trick, thanks!

src/collection-view.js
```diff
@@ -101,7 +101,7 @@ Marionette.CollectionView = Marionette.View.extend({
   _onCollectionAdd: function(child, collection, opts) {
     var index;
     if (opts.at !== undefined) {
-      index = opts.at;
+      index = opts.index !== undefined ? opts.index : collection.indexOf(child);
```
jasonLaster: why not `opts.index || collection.indexOf(child)`

samccone: index of 0 would be falsey :P

cybernetlab:
Thanks to @jridgewell fix is now applicable to all backbone versions

src/collection-view.js
jridgewell: This'll fail for filtered collections. Falling into the `else` path below should work.

```js
 if (opts.index !== undefined) {
   index = opts.index;
 } else {
   index = _.indexOf(this._filteredSortedModels(), child);
 }
```

samccone: The fact that our unit tests did not catch tells me we need to add a new test :)

cybernetlab: Strange, but spec for filtered collection passed with my current code.. May be I'm wrong in spec?

ahumphreys87:
nice fix! :+1:

ahumphreys87:
@cybernetlab could you squash your commits please? This looks good to go

jridgewell:
> This looks good to go

https://github.com/marionettejs/backbone.marionette/pull/2635/files#r33587359 is still an issue.

ahumphreys87:
Sorry @jridgewell I thought a test had been added for it since, doesn't this test case cover for the comment? https://github.com/marionettejs/backbone.marionette/pull/2635/files#diff-9c28bb5b0fc325301a45ddb7b591e97eR495

jridgewell:
I'm pretty sure https://github.com/cybernetlab/backbone.marionette/blob/patch/test/unit/collection-view.spec.js#L502 should be:

```js
      expect(order).to.deep.equal(['10', '5', '6']);
```

which makes me think the entire test is a false positive.

cybernetlab:
Sorry - I'm busy now. I'll review my PR at Saturday or Sunday.

cybernetlab:
Hi, guys! I'm going to finish this PR. Could you explain me one thing, As I understand, `opts.index` shouldn't be used at all if `this.getViewComparator()` return non-null (collection is sorted) or `this.getOption('filter')` return non-null (collection is filtered). Am I right? If so, I would to create helper function like `_isFilteredOrOrdered` and use it in my code. Thanks!

cybernetlab:
Sorry - I've pushed wrong commit :(

cybernetlab:
Ok, I think that I'm ready now. Can anybody kick a rebuild? (I've some random build fails on travis)

ahumphreys87:
lgtm @jridgewell nice catch on that test! :eyes: 

all comments updated, this gets my :+1:

jasonLaster:
Merged!  :boom: https://github.com/marionettejs/backbone.marionette/commit/9070bad2940564e7a975e4b0310d25768d986fea

cybernetlab:
Hi guys! I want to make some more for marionette. Have you any waiting issues?

ahumphreys87:
Hey thanks for the interest. Feel free to pick up any issues you like, we
have them tagged with version numbers so you can see what we are hoping to
get in next.

Give me a shout if you need a hand!

PR #2636, opened 2015-06-28
Fix region.empty({preventDestroy: false}) bug

### LINKED ISSUE #2611: Region.empty fails to destroy view when called with any options

bradlord:
Fixes #2611. See #2631 and #2613 for more comments.

Sorry for yet another pull request. I'm still getting the hang of these things, and I accidentally deleted my branch on github while trying to push the post-squashed commit.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -207,7 +207,8 @@ Marionette.Region = Marionette.Object.extend({
   empty: function(options) {
     var view = this.currentView;
 
-    var preventDestroy = Marionette._getValue(options, 'preventDestroy', this);
+    var emptyOptions = options || {};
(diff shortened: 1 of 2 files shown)


ianmstew:
:+1: 

@marionettejs/marionette-core, this PR removes a bad usage of `Marionette._getValue` from `Region#empty` that resulted in an empty options object `empty({})` being interpreted as `empty({preventDestroy: true});`.  @bradlord is now matching how `Region#show` parses its arguments.

ahumphreys87:
:+1:

ahumphreys87:
merged manually

PR #2639, opened 2015-06-29
Updating doc to suggest the use of Marionette Objects

### LINKED ISSUE #2638: Controllers usage message
In the docs the following is stated
>Warning: deprecated. The Controller object is deprecated. Instead of using the Controller
> class with the AppRouter, you should specify your callbacks on a plain Javascript object.

Should it be suggested to use a marionette object instead of/as well as a JavaScript object?

chrislaughlin:
Closes #2638

### DIFF
--- a/docs/marionette.controller.md
+++ b/docs/marionette.controller.md
@@ -3,7 +3,7 @@
 # Marionette.Controller
 
 > Warning: deprecated. The Controller object is deprecated. Instead of using the Controller
-> class with the AppRouter, you should specify your callbacks on a plain Javascript object.
+> class with the AppRouter, you should specify your callbacks on a plain Javascript object or a [Marionette Object](./marionette.object.md)
 
 A Controller is an object used in the Marionette Router. Controllers are where you store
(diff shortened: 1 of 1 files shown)


docs/marionette.controller.md
```diff
@@ -3,7 +3,7 @@
 # Marionette.Controller
 
 > Warning: deprecated. The Controller object is deprecated. Instead of using the Controller
-> class with the AppRouter, you should specify your callbacks on a plain Javascript object.
+> class with the AppRouter, you should specify your callbacks on a plain Javascript object or a [Marionette Object](./marionette.object.md)
```
samccone: why the `./` ? I am worried this might break on the www

chrislaughlin: This is following the same pattern as line 49 and 55

samccone: :+1:

samccone:
@chrislaughlin thanks so much for the PR :clap: keep them coming :heart:

Issue #2606, opened 2015-06-18
CollectionView does not respect false as a return value from isEmpty()


sdnetwork:
I think there is a little problem with the use of this function : 

in the doc the isempty is describe by  : 

    If you want to control when the empty view is rendered, you can override isEmpty

in my case i don't want display the empty view if my collection model query is empty.

so i override the isEmpty like that : 

    isEmpty: function() {
    		if (this.collection.length == 0) {
    			if (this.ui.query.val() == "") {
    				return false;
    			}
    			return true;
    		}
    		return false;
    	}

but the problem is in the _renderchildren

     // If we have shown children and none have passed the filter, show the empty view
      if (this.children.isEmpty()) {
        this.showEmptyView();
      }

to correct the problem i have done that but ...

    // If we have shown children and none have passed the filter, show the empty view
        if (this.children.isEmpty() && this.getOption('filter') && this.getOption('filter') != '') {
          this.showEmptyView();
        }

do you think that is a good workaround ?

vanessanutrio:
I am in the process of upgrading to 2.4 and noticed this issue as well. I am using  isEmpty as specified in the documentation to not show the empty view but now it is not working. To me, the workaround proposed makes sense based on the comment and the commit description when this was added.

https://github.com/marionettejs/backbone.marionette/commit/4c5ef018d813116e66fce421a874c44fdda13a1a

jasonLaster:
seems reasonable. @cmaher @ahumphreys87 ?

ahumphreys87:
Is the workaround proposed to make isEmpty look at children and the filtered set? Then change the check in _renderChildren to look at isEmpty?

I'd be happy to accept a PR for that change, I don't think changing isEmpty would cause issues elsewhere so let's do it.

@sdnetwork would you be interested in submitting a PR?

ianmstew:
@sdnetwork Any word on a PR?  Otherwise we can take a stab at it.

sdnetwork:
@ianmstew i have just done https://github.com/marionettejs/backbone.marionette/pull/2641

ianmstew:
Fixed by https://github.com/marionettejs/backbone.marionette/commit/a8771a4f5cd94d847d4e94db88dca26ea51a1acf.

PR #2641, opened 2015-06-29
Check if filter is set

sdnetwork:


samccone:
@sdnetwork thanks for the PR. mind adding a test case please

ianmstew:
This is intended to fix #2606.

ahumphreys87:
@sdnetwork this looks good, just needs a test that proves the bug fix and prevents a regression :smile:

jridgewell:
I'm not sure why [lines 244-248](https://github.com/sdnetwork/backbone.marionette/blob/master/src/collection-view.js#L244-L248) are ever needed? Shouldn't the [original `#isEmpty`](https://github.com/sdnetwork/backbone.marionette/blob/master/src/collection-view.js#L236-L237) check have returned `true`?

ahumphreys87:
@jridgewell i think its to catch the case where a filter "filters" all the children. The first is checking the collection which doesn't actually change with the filter

JSteunou:
from #2661 I was wondering if 

> Should the isEmpty check on children is a filter is set, and if not, fall-back on collection && collection.length?

jridgewell:
> Should the #isEmpty check on children is a filter is set, and if not, fall-back on collection && collection.length?

I think the issue is there isn't a performant way to determine if a filtered set is empty, without running the filter on all `n` models.

As it stands right now (and with this PR), `CollectionView#isEmpty` has lost its usefulness. It's no longer smart enough to handled a filtered collection, since we don't want that performance impact.

We could store the filtered models in an array and provide an API to access that filtered array. That'll take care of the performance impact, and allow `CollecitonView#isEmpty` to be useful again. But at that point, we're reinventing a `Collection`. :frowning: 

I think the real issue stems from where we've landed this functionality. It shouldn't be a view's responsibility to filter the collection's models, it should be the collection's. Using something like [VirtualCollection](https://github.com/p3drosola/Backbone.VirtualCollection) seems like an appropriate solution.

ahumphreys87:
@jridgewell I see where you are coming from and I know there was a big discussion about where to land the filter functionality back in the original PR.

However, as things stand this PR fixes a bug - I'm :+1: for getting this into a 2.x release. Maybe we can look at how we implement `filter` differently in a major version? I think your right - it probably needs a rethink.

ianmstew:
Manually merged via https://github.com/marionettejs/backbone.marionette/commit/a8771a4f5cd94d847d4e94db88dca26ea51a1acf.

JSteunou:
Does it fix #2661 too?

ianmstew:
@JSteunou good call!

Also, for future reference, the current solution is a nonbreaking stopgap.  The more comprehensive solution is already in `next` for 3.0: https://github.com/marionettejs/backbone.marionette/blob/2af55e5020da9d74c043bd5c121e1f95c45457d9/src/collection-view.js#L277-L285

Issue #2581, opened 2015-06-03
Get coverage up to 100% for next


paulfalgout:
We lost coverage recently https://coveralls.io/builds/2719284

Some additional tests should be written to maintain 100%.

ianmstew:
:cry:

paulfalgout:
This is mostly down to filter related issues in collectionView, but I think we should also add something for `if(this.isDestroyed)` in `Object.destroy` and the RegionManager `_store` adding the same named region.

Really it'd be great to get branch coverage up to 100%.  Often the branches can indicate unused code.

Some of that is https://github.com/marionettejs/backbone.marionette/pull/2733 and https://github.com/marionettejs/backbone.marionette/pull/2732

paulfalgout:
```md

paulfalgout:
This is very close to coverage already.. actually the more worrying stat is some of the branches not covered.. anyhow.. the tests need love anyhow.. but this shouldn't hold up v3.

rafde:
@paulfalgout this can be closed.

paulfalgout:
Thanks

PR #2642, opened 2015-06-30
Remove unused triggerMethod optimization

paulfalgout:
This optimization of `triggerMethod` was previously used in `Controller`, but are now gone.  Seeing as we have some custom overrides for `triggerMethod` in `AbstractView` it may be more reasonable to just not worry about it.  It could only be utilized internally if we do want to keep it.. none of the public API can benefit.  We'd need to call `Marionette._triggerMethod` directly with no event arguments.

This was discovered while researching https://github.com/marionettejs/backbone.marionette/issues/2581

### DIFF
--- a/src/trigger-method.js
+++ b/src/trigger-method.js
@@ -33,11 +33,7 @@ Marionette._triggerMethod = (function() {
 
     // trigger the event, if a trigger method exists
     if (_.isFunction(context.trigger)) {
-      if (noEventArg + args.length > 1) {
-        context.trigger.apply(context, noEventArg ? args : [event].concat(_.drop(args, 0)));
-      } else {
-        context.trigger(event);
-      }
+      context.trigger.apply(context, noEventArg ? args : [event].concat(_.drop(args, 0)));
     }
 
     return result;


src/trigger-method.js
```diff
@@ -33,11 +33,7 @@ Marionette._triggerMethod = (function() {
 
     // trigger the event, if a trigger method exists
     if (_.isFunction(context.trigger)) {
-      if (noEventArg + args.length > 1) {
-        context.trigger.apply(context, noEventArg ? args : [event].concat(_.drop(args, 0)));
-      } else {
-        context.trigger(event);
-      }
+      context.trigger.apply(context, noEventArg ? args : [event].concat(_.drop(args, 0)));
```
jridgewell: Why are you dropping zero?

paulfalgout: Hmm I don't actually know.  I didn't change this line at all.  
From here: https://github.com/marionettejs/backbone.marionette/commit/1316716980080173eccb18f47dfa5c80485645fb#diff-a20005534e3338e0b6f5c03de7ba9952R38
But the `,0` can be dropped (no pun intended) with no consequence with lodash right?

samccone: we can try and drop the 0 

ping @megawac

megawac: Nope, `drop(0)` is the only consistent way to do a fast slice on an arguments object in both lodash>=2 and underscore^1.4.2

@jridgewell its a faster way of doing `toArray` in underscore currently

jridgewell: But what do we need to slice it for? We're just passing it to concat...

megawac: Args is an `arguments` object, concat doesn't support them

ahumphreys87:
:+1:

PR #2643, opened 2015-06-30
Create Mn.Class - WIP

### LINKED ISSUE #2534: Metal - Create Mn.Class
We should add Mn.Class with Object feature parity:

```diff
+Marionette.Class = Metal.Class.extend({
+  // destroy: function() {
+  //   this.triggerMethod('before:destroy');
+  //   this.triggerMethod('destroy');
+  //   this.stopListening();
+  // },
+
+  // Import the `triggerMethod` to trigger events with corresponding
+  // methods if the method exists
+  triggerMethod: Marionette.triggerMethod,
+
+  // Proxy `getOption` to enable getting options from this or this.options by name.
+  getOption: Marionette.proxyGetOption,
+
+  // Proxy `bindEntityEvents` to enable binding view's events from another entity.
+  bindEntityEvents: Marionette.proxyBindEntityEvents,
+
+  // Proxy `unbindEntityEvents` to enable unbinding view's events from another entity.
+  unbindEntityEvents: Marionette.proxyUnbindEntityEvents
+
+});
```

I'm not sure if we should repurpose the tests here...

 Mn.mergeOptions
 Attach options to this.options in constructor
 proxyRadioHandlers in constructor

paulfalgout:
Resolves #2534 

I updated the WIP here and temporarily cherry-picked @megawac 's metal-fix.

### DIFF
--- a/src/metal.js
+++ b/src/metal.js
@@ -38,11 +38,12 @@ Marionette.Class = Metal.Class.extend({
   unbindEntityEvents: Marionette.proxyUnbindEntityEvents
 });
 
+var props = _.keys(Metal.Class.prototype);
 function classify(obj) {
   return Marionette.Class.extend(
     _.extend(
       {constructor: obj},
-      _.omit(obj.prototype, _.keys(Metal.Class.prototype))
+      _.omit(obj.prototype, props)
     )
   );
 }
--- a/src/metal.js
+++ b/src/metal.js
@@ -38,12 +38,12 @@ Marionette.Class = Metal.Class.extend({
   unbindEntityEvents: Marionette.proxyUnbindEntityEvents
 });
 
-var props = _.keys(Metal.Class.prototype);
 function classify(obj) {
   return Marionette.Class.extend(
-    _.extend(
+    _.defaults(
       {constructor: obj},
-      _.omit(obj.prototype, props)
+      Metal.Class.prototype,
+      obj.prototype
     )
   );
 }
--- a/src/metal.js
+++ b/src/metal.js
@@ -1,8 +1,42 @@
 // Missing Metal Items
-// 1. Marionette.Class parity
+// 1. Marionette.Error
 // 2. Marionette.Deprecate
 
-Marionette.Class = Metal.Class.extend({});
+Marionette.Class = Metal.Class.extend({
+  constructor: function(options) {
+    this.options = _.extend({}, _.result(this, 'options'), options);
+    Marionette.proxyRadioHandlers.apply(this);
+    this.cid = _.uniqueId(this.cidPrefix);
+    this.initialize.apply(this, arguments);
+  },
+
+  cidPrefix: 'mnc',
+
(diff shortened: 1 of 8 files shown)


src/metal.js
```diff
@@ -1,14 +1,49 @@
...
+  unbindEntityEvents: Marionette.proxyUnbindEntityEvents
+});
 
 function classify(obj) {
   return Marionette.Class.extend(
-    _.extend(
+    _.defaults(
       {constructor: obj},
-      _.omit(obj.prototype, _.keys(Marionette.Class.prototype))
+      Metal.Class.prototype,
```
megawac: BTW, this can be changed to `Marionette.Class` now as `default` iterates in keys

src/metal.js
```diff
@@ -1,14 +1,49 @@
 // Missing Metal Items
-// 1. Marionette.Class parity
+// 1. Marionette.Error
 // 2. Marionette.Deprecate
 
-Marionette.Class = Metal.Class.extend({});
+Marionette.Class = Metal.Class.extend({
+  constructor: function(options) {
+    this.options = _.extend({}, _.result(this, 'options'), options);
```
megawac: Are we keeping `getOption` around?

src/metal.js
```diff
@@ -1,14 +1,49 @@
...
+    this.cid = _.uniqueId(this.cidPrefix);
+    this.initialize.apply(this, arguments);
+  },
+
+  cidPrefix: 'mnc',
+
+  destroy: function() {
+    this.triggerMethod('before:destroy');
+    this.triggerMethod('destroy');
+    Marionette.unproxyRadioHandlers.apply(this);
```
megawac: Should this be on prototype too ?

paulfalgout: The radio handler stuff?  maybe.  I haven't really investigated it yet.  I don't know that the radio stuff belongs in the Class at all TBH.

paulfalgout:
It looks to me as if most of the issues here have to do with using `Backbone.View` where if there's no `destroy` it's cleaning up the view with `remove`, but now that we're classifying it _does_ have a destroy.

src/metal.js
paulfalgout: With `Metal` it'd be cool if the radio stuff was just a mixin.  I dunno that it needs to be available on every single Mn and Bb class there is. But that would require https://github.com/marionettejs/backbone-metal/issues/20  or you'd have to init the radio handler mixin yourself.

jasonLaster:
oh interesting... do we want to be clever here?

paulfalgout:
@marionettejs/marionette-core 
Well this is improving.  Still `classify` issues at various versions.
The tests that were failing for the `destroy` issues are not up to date and need some work in general.

The implementation needs tests of its own still, but it is getting closer.

jasonLaster:
@paulfalgout where are we at here?

jasonLaster:
closing as this is 3.x work

PR #2645, opened 2015-06-30
Prevent Object's Radio API from silently failing

paulfalgout:
If you give the Radio API a bad function name it will silently fail.

This is inconsistent with other hashes that accept functions.  If you mistype a function, javascript can handle the error.

I found this while investigating https://github.com/marionettejs/backbone.marionette/issues/2581  This line is currently not tested, but I think it is better it is simply removed.

### DIFF
--- a/src/radio-helpers.js
+++ b/src/radio-helpers.js
@@ -25,9 +25,6 @@
       }
       _.each(hash, function(handler, radioMessage) {
         handler = normalizeHandler.call(this, handler);
-        if (!handler) {
-          return;
-        }
         var messageComponents = radioMessage.split(' ');
         var channel = messageComponents[0];
         var messageName = messageComponents[1];


src/radio-helpers.js
```diff
@@ -25,9 +25,6 @@
       }
       _.each(hash, function(handler, radioMessage) {
         handler = normalizeHandler.call(this, handler);
-        if (!handler) {
-          return;
-        }
         var messageComponents = radioMessage.split(' ');
```
jasonLaster: do we want a descriptive error?

```js
throw new Error('hey you there!')
```

paulfalgout: I'm on the fence.  We do throw errors for `modelEvents` or anything using `bindEntityEvents` but you're not going to get a pretty error for a `listenTo` or a `reply`, but then you do have Radio's debug mode.

To some degree we have some overlap where this handling of errors might be best handled with on utility: https://github.com/marionettejs/backbone.marionette/blob/next/src/bind-entity-events.js#L23

However _technically_ a radio request can utilize undefined as the handler, and the use-case isn't necessarily a stretch:
```js
const ENV = 'PROD';
const DEBUG_MODE = 0;
someChannel.reply('env', ENV);
someChannel.reply('debug', DEBUG_MODE);

...
someChannel.request('env'); // => 'PROD'
someChannel.request('debug') // => 0
```

currently in this implementation anything falsy would return undefined.

So `radioRequests` / `radioCommands` are currently broken per Radio's documentation, but we could potentially only throw an error for `radioEvents`

jasonLaster:
:+1:

PR #2646, opened 2015-06-30
Make TemplateCache.loadTemplate() work for empty templates (fixes #2347)

### LINKED ISSUE #2347: TemplateCache.loadTemplate does not work for empty existing templates
Some templates may be intentionally left blank for various design reasons (e.g. some stuff can be unavailable for certain user languages). Current implementation of [TemplateCache.loadTemplate function](https://github.com/marionettejs/backbone.marionette/blob/2b900bba24d74c6c0cffad2740e07df69a15cc20/src/template-cache.js#L76) fails to consider empty templates as valid ones and throws an error. The reason for this is in relying on the length of the template's html:

```js
loadTemplate: function(templateId) {
    var template = Backbone.$(templateId).html();
    if (!template || template.length === 0) {
        throw new Marionette.Error({
            name: 'NoTemplateError',
            message: 'Could not find template: "' + templateId + '"'
        });
    }
    return template;
},
```

The problem can be avoided with this simple fix:

```js
loadTemplate: function(templateId) {
    var $template = Backbone.$(templateId);
    if (!$template.length) {
        throw new Marionette.Error({
            name : 'NoTemplateError',
            message : 'Could not find template: "' + templateId + '"'
        });
    }
    return $template.html();
},

```

kachkaev:


jasonLaster:
@kachkaev the change makes sense. Mind updating the test https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/template-cache.spec.js

samccone:
:+1: makes sense, @kachkaev mind adding a test for this?

ianmstew:
Fixes #2347

kachkaev:
Done (hopefully, the right way).

The new test fails on the current master branch and passes in the patched version.

jridgewell:
I think the tests would be more appropriate in `unit/template-cache.spec.js`.

src/template-cache.js
```diff
@@ -74,16 +74,15 @@ _.extend(Marionette.TemplateCache.prototype, {
   // using a template-loader plugin as described here:
   // https://github.com/marionettejs/backbone.marionette/wiki/Using-marionette-with-requirejs
   loadTemplate: function(templateId, options) {
-    var template = Backbone.$(templateId).html();
+    var $template = Backbone.$(templateId);
```
samccone: i know this is not directly related to this PR... but can anyone think of a reason we create a jquery object here only to call html on it?

paulfalgout: this:
```
<script type="text/template" id="some-template">
<p>Hello World</p>
</script>
```
```js
$('#some-template').html(); // <= '<p>Hello World</p>'
```

samccone: derp sorry totally fumbled reading this line ha :cactus:

ahumphreys87:
@kachkaev this is looking good. However, could you move the tests to `unit/template-cache.spec.js`? Then squash and open against the `patch` branch?

kachkaev:
Hi @ahumphreys87,
@ahumphreys87 I see what you mean and what the change is needed for. I'm not sure I can produce the fix quickly though. I understand that I have to somehow destroy the latest commit on the branch and replace it with another one, and this is something that I might stuck with (never amended branch history in the remote repo).

If you guys need to merge the fix ASAP to prepare the release, I guess it would be quicker for a an experienced person to grab a patch and move the test in another pull request. The changes in the code I've made are so tiny, so this can't be ever counted as stealing my right to become a contributor to Marionette :–)

I'll keep a reminder about this task in my inbox and will try to look at the solution myself when I find free time (if no one closes the issue before me, of course).

Hope to see the fix in the next release!

ahumphreys87:
squashed and opened at #2674

@kachkaev i cherry picked your commit so you will still get the contribution :)

kachkaev:
@ahumphreys87 great thanks!

PR #2658, opened 2015-07-03
Improve support for Backbone.View childViews

### LINKED ISSUE #2651: childView and emptyView no longer support `Backbone.View`s, now they require `Marionette.View`s in 2.4.2
This commit: https://github.com/marionettejs/backbone.marionette/commit/2cefc20d9ef2552a398b8b051b3151107f25a9fd

Has introduced a bug in which `childView` and `emptyView` in `CollectionView`s can no longer take a `Backbone.View` and require a `Marionette.View` instead. This is because of the added call to `view._getNestedViews()` in `addEmptyView` and `_addChildView`, which `Backbone.View` does not define.

### LINKED ISSUE #2657: Better Backbone.View childView support
As I began to write a failing test for issue #2651, I became aware of two issues.

1. There is a relatively wide gap in our specs concerning Backbone.View.
2. There are 2 instances in `master` where Marionette view lifecycle events are not applied to a Backbone view due to usage of `view.on('render', ...)`.  These are `onDomRefresh` and CollectionView childView `onBeforeAttach`.

I propose that, to start, we refactor `collection-view-spec.js` and `on-attach-spec.js` to test solely against Backbone.View childViews.  This exercise will result in failing tests.  Next, I propose we either remove instances of `view.on('render'`) from core or else externally trigger `render` events on Backbone views to address the failing tests.

I have already moved forward with experimentally refactoring tests and have some fixes in mind, but the main decision to be made is: Do we support `onDomRefresh` on Backbone views?  If so, do we support it by triggering `render` on Backbone views externally, or by not relying on `render` events from Backbone views at all?

ianmstew:
**This PR**

1. Cleans up collection-view and on-attach specs.
2. Significantly improves Backbone View child view test coverage (fixes #2657).  Moving child views to Backbone.View in `collection-view.spec.js` and `on-attach.spec.js` broke a handful of existing tests.
3. Fixes #2651.
4. Adds support for `before:render`, `render`, `before:destroy`, `destroy`, and `dom:refresh` view lifecycle events specifically for Backbone.View child views of Regions (and therefore LayoutViews) and CollectionViews.  This _completely fixed_ the broken tests in **(2)** above.

**Thoughts on better testing**

To improve on testing the Backbone View child view case, I converted `collection-view.spec.js` and `on-attach.spec.js` to use Backbone View child views by default.  I imagine this approach can be carried forward to future test cases and backward to any other existing child view test cases.

**Note on dom:refresh support for Backbone.View**

I had to add a flag `_isDomRefreshMonitored` to views within `Marionette.MonitorDOMRefresh` in order to gain `dom:refresh` from Backbone.Views shown by `Region#show` (see [Region#show commit diff](https://github.com/ianmstew/backbone.marionette/commit/730914c6087351f4214a4e7d5ec2957c135578b6#diff-b8da5292cb07ec757ac64e78ff5af5c1R204)).  Please let me know if anyone disapproves.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -358,7 +358,7 @@ Marionette.CollectionView = Marionette.View.extend({
     // Trigger `before:attach` following `render` to avoid adding logic and event triggers
     // to public method `renderChildView()`.
     if (canTriggerAttach && this._triggerBeforeAttach) {
-      nestedViews = [view].concat(view._getNestedViews());
+      nestedViews = this._getViewAndNested(view);
       view.once('render', function() {
         this._triggerMethodMany(nestedViews, this, 'before:attach');
       }, this);
@@ -369,7 +369,7 @@ Marionette.CollectionView = Marionette.View.extend({
 
     // Trigger `attach`
     if (canTriggerAttach && this._triggerAttach) {
-      nestedViews = [view].concat(view._getNestedViews());
+      nestedViews = this._getViewAndNested(view);
       this._triggerMethodMany(nestedViews, this, 'attach');
     }
     // call the 'show' method if the collection view has already been shown
@@ -461,7 +461,7 @@ Marionette.CollectionView = Marionette.View.extend({
     // Trigger `before:attach` following `render` to avoid adding logic and event triggers
     // to public method `renderChildView()`.
     if (canTriggerAttach && this._triggerBeforeAttach) {
-      nestedViews = [view].concat(view._getNestedViews());
+      nestedViews = this._getViewAndNested(view);
       view.once('render', function() {
         this._triggerMethodMany(nestedViews, this, 'before:attach');
       }, this);
@@ -471,7 +471,7 @@ Marionette.CollectionView = Marionette.View.extend({
 
     // Trigger `attach`
     if (canTriggerAttach && this._triggerAttach) {
-      nestedViews = [view].concat(view._getNestedViews());
+      nestedViews = this._getViewAndNested(view);
       this._triggerMethodMany(nestedViews, this, 'attach');
     }
     // Trigger `show`
@@ -482,7 +482,13 @@ Marionette.CollectionView = Marionette.View.extend({
 
   // render the child view
   renderChildView: function(view, index) {
+    if (!(view instanceof Marionette.View)) {
+      Marionette.triggerMethodOn(view, 'before:render', view);
+    }
     view.render();
+    if (!(view instanceof Marionette.View)) {
+      Marionette.triggerMethodOn(view, 'render', view);
+    }
     this.attachHtml(this, view, index);
     return view;
   },
@@ -490,7 +496,11 @@ Marionette.CollectionView = Marionette.View.extend({
   // Build a `childView` for a model in the collection.
   buildChildView: function(child, ChildViewClass, childViewOptions) {
     var options = _.extend({model: child}, childViewOptions);
-    return new ChildViewClass(options);
+    var childView = new ChildViewClass(options);
+    if (!(childView instanceof Marionette.View)) {
+      Marionette.MonitorDOMRefresh(childView);
+    }
+    return childView;
   },
 
   // Remove the child view and destroy it.
@@ -506,7 +516,9 @@ Marionette.CollectionView = Marionette.View.extend({
       if (view.destroy) {
         view.destroy();
       } else if (view.remove) {
+        Marionette.triggerMethodOn(view, 'before:destroy', view);
         view.remove();
+        Marionette.triggerMethodOn(view, 'destroy', view);
       }
 
       delete view._parent;
@@ -665,6 +677,11 @@ Marionette.CollectionView = Marionette.View.extend({
 
   _getImmediateChildren: function() {
     return _.values(this.children._views);
+  },
+
+  _getViewAndNested: function(view) {
+    // This will not fail on Backbone.View which does not have #_getNestedViews.
+    return [view].concat(_.result(view, '_getNestedViews') || []);
   },
 
   getViewComparator: function() {
--- a/src/dom-refresh.js
+++ b/src/dom-refresh.js
@@ -6,6 +6,7 @@
 // re-rendered.
 
 Marionette.MonitorDOMRefresh = function(view) {
+  view._isDomRefreshMonitored = true;
 
   // track when the view has been shown in the DOM,
   // using a Marionette.Region (or by other means of triggering "show")
@@ -23,9 +24,7 @@ Marionette.MonitorDOMRefresh = function(view) {
   // Trigger the "dom:refresh" event and corresponding "onDomRefresh" method
   function triggerDOMRefresh() {
     if (view._isShown && view._isRendered && Marionette.isNodeAttached(view.el)) {
-      if (_.isFunction(view.triggerMethod)) {
-        view.triggerMethod('dom:refresh');
-      }
+      Marionette.triggerMethodOn(view, 'dom:refresh', view);
     }
   }
 
--- a/test/unit/on-attach.spec.js
+++ b/test/unit/on-attach.spec.js
@@ -2,6 +2,7 @@ describe('onAttach', function() {
   'use strict';
 
   beforeEach(function() {
+    var spec = this;
 
     // A Region to show our LayoutView within
     this.setFixtures('<div id="region"></div>');
@@ -11,11 +12,31 @@ describe('onAttach', function() {
     // A view we can use as nested child views
     this.BasicView = Marionette.ItemView.extend({
       template: false,
+      constructor: function(options) {
+        Marionette.ItemView.prototype.constructor.call(this, options);
+        spec.sinon.spy(this, 'onAttach');
+        spec.sinon.spy(this, 'onBeforeAttach');
+      },
       onAttach: function() {},
       onBeforeAttach: function() {}
     });
 
-    var spec = this;
+    this.BasicLayoutView = Marionette.LayoutView.extend({
+      template: _.template('<header></header><main></main><footer></footer>'),
+      regions: {
+        header: 'header',
+        main: 'main',
+        footer: 'footer'
+      },
+      constructor: function() {
+        Marionette.LayoutView.prototype.constructor.apply(this, arguments);
+        spec.sinon.spy(this, 'onBeforeAttach');
+        spec.sinon.spy(this, 'onAttach');
+      },
+      onAttach: function() {},
+      onBeforeAttach: function() {}
+    });
+
     this.EmptyView = Marionette.ItemView.extend({
       template: false,
       constructor: function(options) {
@@ -231,29 +252,9 @@ describe('onAttach', function() {
   });
 
   describe('when the parent view is initially detached', function() {
-    beforeEach(function() {
-
-      // A LayoutView class that we can use for all of our tests
-      this.LayoutView = Marionette.LayoutView.extend({
-        template: _.template('<main></main><footer></footer>'),
-        regions: {
-          main: 'main',
-          footer: 'footer'
-        },
-        onBeforeAttach: function() {},
-        onAttach: function() {}
-      });
-    });
-
     describe('When showing a View in a Region', function() {
       beforeEach(function() {
-        this.MyView = Marionette.ItemView.extend({
-          template: _.template(''),
-          onBeforeAttach: this.sinon.stub(),
-          onAttach: this.sinon.stub()
-        });
-
-        this.myView = new this.MyView();
+        this.myView = new this.BasicView();
         this.region.show(this.myView);
       });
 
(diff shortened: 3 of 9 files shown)


ianmstew:
Note to self: Add in @megawac's suggestion to supplement Backbone.View with `render` and `destroy` events and see if tests pass.

test/unit/on-attach.spec.js
```diff
@@ -986,18 +908,22 @@ describe('onAttach', function() {
...
+          .and.to.not.have.been.calledOn(this.childView1)
+          .and.to.not.have.been.calledWith(this.childView1);
+        expect(this.childView2.onBeforeAttach)
+          .and.to.not.have.been.calledOnce
+          .and.to.not.have.been.calledOn(this.childView2)
+          .and.to.not.have.been.calledWith(this.childView2);
+        expect(this.childView2.onAttach)
+          .and.to.not.have.been.calledOnce
+          .and.to.not.have.been.calledOn(this.childView2)
+          .and.to.not.have.been.calledWith(this.childView2);
```
jasonLaster: Looks so nice

jasonLaster:
Looks good Ian

src/collection-view.js
```diff
@@ -369,7 +369,7 @@ Marionette.CollectionView = Marionette.View.extend({
 
     // Trigger `attach`
     if (canTriggerAttach && this._triggerAttach) {
-      nestedViews = [view].concat(view._getNestedViews());
+      nestedViews = [view].concat(_.result(view, '_getNestedViews') || []);
```
jasonLaster: Not sure what the advantage of _.result is. 

Also could the [] default go in getNestedViewd

ianmstew: Ahh, I actually copied that from [Region#show](https://github.com/marionettejs/backbone.marionette/blob/4c00dabc3f41a95edac2773824ebecaf6c130986/src/region.js#L155).  It's a not-so-obvious fix for Backbone Views (#2651) that don't have `_getNestedViews`--the `result` call won't fail if that method doesn't exist, and in the non-exist case we get the `[]`.  I do believe `_getNestedViews`, when it does exist, already returns the `[]` default.

ianmstew:
@jasonLaster @samccone @megawac @paulfalgout See my updated description.  This PR adds `triggerMethodOn`-style support for the full view event lifecycle on Backbone.View when shown by either a Region or CollectionView.

src/collection-view.js
```diff
@@ -482,7 +487,13 @@ Marionette.CollectionView = Marionette.View.extend({
 
   // render the child view
   renderChildView: function(view, index) {
+    if (!(view instanceof Marionette.View)) {
+      Marionette.triggerMethodOn(view, 'before:render', view);
+    }
     view.render();
+    if (!(view instanceof Marionette.View)) {
```
megawac: Whats the justification of the order of this. Surely they can be added ad the same time?

ianmstew: You mean the order of fire `before:render` -> call `render()` -> fire `render`?  I am trying to maintain the natural order, such as in [ItemView#render](https://github.com/marionettejs/backbone.marionette/blob/4c00dabc3f41a95edac2773824ebecaf6c130986/src/item-view.js#L54-L60).

jasonLaster: What is this `view instanceof Marionette.View` guarding against?

sorry should be obvious

test/unit/on-attach.spec.js
```diff
@@ -900,18 +808,22 @@ describe('onAttach', function() {
...
+        .and.to.have.been.calledOn(this.childView1)
+        .and.to.have.been.calledWith(this.childView1);
+      expect(this.childView2.onBeforeAttach)
+        .and.to.have.been.calledOnce
+        .and.to.have.been.calledOn(this.childView2)
+        .and.to.have.been.calledWith(this.childView2);
+      expect(this.childView2.onAttach)
+        .and.to.have.been.calledOnce
+        .and.to.have.been.calledOn(this.childView2)
+        .and.to.have.been.calledWith(this.childView2);
```
jasonLaster: I love this style of cleanup. Tons of improvements here @thejameskyle would be proud

test/unit/on-dom-refresh.spec.js
```diff
@@ -2,33 +2,93 @@ describe('onDomRefresh', function() {
...
+      this.detachedRegion.show(this.bbView);
     });
 
     it('should never trigger onDomRefresh', function() {
-      expect(this.onDomRefreshStub).not.to.have.been.called;
+      expect(this.bbView.onDomRefresh).not.to.have.been.calledOnce;
+    });
+  });
+
+  describe('when a Marionette view is shown detached from the DOM', function() {
```
jasonLaster: I like these describes. Much more precision

test/unit/on-attach.spec.js
ianmstew: Inspired by the one and only @jmeas :)

src/collection-view.js
ianmstew: It's an unfortunate guard, but the goal here is to target Backbone views for external `render` event triggering-on.

src/collection-view.js
```diff
@@ -482,7 +487,13 @@ Marionette.CollectionView = Marionette.View.extend({
 
   // render the child view
   renderChildView: function(view, index) {
+    if (!(view instanceof Marionette.View)) {
```
jasonLaster: Oh, I get it. We're manually firing the event on B.V ha. okay, I like this. sorry took me awhile

src/dom-refresh.js
```diff
@@ -24,7 +25,9 @@ Marionette.MonitorDOMRefresh = function(view) {
   function triggerDOMRefresh() {
     if (view._isShown && view._isRendered && Marionette.isNodeAttached(view.el)) {
       if (_.isFunction(view.triggerMethod)) {
-        view.triggerMethod('dom:refresh');
+        view.triggerMethod('dom:refresh', view);
+      } else {
+        Marionette.triggerMethodOn(view, 'dom:refresh', view);
```
jasonLaster: could this just be 

```js
Marionette.triggerMethodOn(view, 'dom:refresh', view);
```
without ` if (_.isFunction(view.triggerMethod)) {`

jasonLaster:
This is awesome. :+1:

src/dom-refresh.js
ianmstew: Great point.  I'll kill the `if`.

src/collection-view.js
megawac: @ianmstew what happens if the view is rendered twice?

jmeas: Can `!(view instanceof Mn.View)` be cached and reuzed?

jmeas: Actually, looks like it's used multiple times throughout the app. This makes me wanna vom a little but `isMarionetteView` could be a helper we introduce lol

src/collection-view.js
```diff
@@ -419,6 +419,11 @@ Marionette.CollectionView = Marionette.View.extend({
     return view;
   },
 
+  _getViewAndNested: function(view) {
+    // This will not fail on Backbone.View which does not have #_getNestedViews.
+    return [view].concat(_.result(view, '_getNestedViews') || []);
```
jmeas: :+1: nice catch

src/collection-view.js
megawac: Also is `view instanceof Mn.View` the best way to do this or should we check for the presence of individual methods/hooks

jasonLaster:
Can we add a section in the M.View docs for supporting full lifecyle in B.Views

```md
#### How do i support the full view lifecyle in my Backbone.View 

yada yada yada
```

jmeas:
> Can we add a section in the M.View docs for supporting full lifecyle in B.Views

I think we should break from our usual "associate useful content with a specific Marionette classes" pattern here; it's a symptom of our history of conflating guides with API docs, I think. I could instead see there being an entire guide dedicated to "Interoperability with Backbone.Views" (or whatever), separate from any specific Marionette view.

One day in the future when we group some guides under 'Working with Views" (or whatever), this guide could go there.

src/collection-view.js
ianmstew: The only other hook we test for in core currently is the absence of a `destroy` method.  The only thing I don't like about the destroy method check is that it's not obvious what we're asking.  Is `_isMarionetteView` more performant than an instanceof check?  I _could_ add something like that into this commit.

ianmstew:
@jmeas @jasonLaster I can see adding a note about what additional we provide to BB Views within the Region and CollectionView docs.  This could be a halfway point between the two proposals until we release the guidebook.

ianmstew:
@megawac voiced a concern about duplicated events.  I can see one case where they would be duplicated: if a Backbone View handles its own `render` and `destroy` events internally:

```js
var FooView = Backbone.View.extend({
  render: function () {
    Marionette.triggerMethodOn(this, 'before:render', this);
    this.$el.append(this.template(this.model.attributes));
    Marionette.triggerMethodOn(this, 'render', this);
  },
  remove: function () {
    Marionette.triggerMethodOn(this, 'before:destroy', this);
    Backbone.View.remove.apply(this, arguments);
    Marionette.triggerMethodOn(this, 'destroy', this);
  }
});
```

Is this a concern?

jmeas:
> Is this a concern?

I think so. I'm not really a fan of us trying to 'recreate' Marionette's API on the BB Views.

ianmstew:
Okay, so the only two events in question are `render` and `destroy`.  The rest of the events are native to Region, and are therefore appropriate to fire on Mn and Bb views alike.  There is one alternative: Refactor both core code (`MonitorDomRefresh`) and unit tests to not depend on `render` or `destroy` except when explicitly expecting a Mn view.  Otherwise, all core and tests should be Bb/Mn-agnostic when it comes to child views.  In that end, that's the end goal--the implementation is flexible.

ianmstew:
@jmeas What do you think of that alternative?

jasonLaster:
@jmeas @ianmstew easiest thing for me would be to change the check from `instanceof` to `_supportsRenderLifecycle` `_supportsDestroyLifecycle` that way we add those fields and if your view supports it you add it too

ianmstew:
@jasonLaster So we could in good faith trigger `render` and `destroy` on views that don't include those while giving the developer the option to opt out of our API standardization.  That's pretty slick.  @jmeas?

jmeas:
I would opt for the simpler API and never trigger the events on Bb Views.

It might just be me, but I would rather manually trigger the events on my BB.View than add a `_supportsRenderLifecycle` flag on it. There are three reasons why I feel this way:

1. The former is more expressive; it's much clearer that by triggering the events, I'm, well, getting triggered events. The property doesn't communicate *exactly* what's going on, you know?

2. It also seems a bit silly to burden developers with requiring knowledge of an obscure, "internal" Mn property, I think, for something as simple as firing some events. If it was something more complex, then maybe I could justify it being hidden behind a flag. But triggering events is so simple!

3. @jasonLaster picked what I think is a pretty good name for this concept, I think: "render lifecycle." But this is a brand new concept in Marionette that developers haven't yet had to directly think about afaik. "What is a render lifecycle?" is the first question I would have if I read that code.

The code in this PR that fixes errors when using BB Views is really solid, but I'm not sold on adding new features to Backbone Views when they're used in conjunction w. Mn.

ianmstew:
@jmeas I think the point of `_supportsRenderLifecycle` is that Bb Views get `render` by default _unless_ the `_supportsRenderLifecycle` flag is present.  That means when a Bb View fires `render` manually the developer should add that flag to be sure Mn doesn't cause a double `render` trigger.  At the point a Bb View is firing its own `render` events, the developer is at least a _little_ more aware of the "render lifecycle".

That being said, these points of yours still stand:

- We would be _adding_ to the Bb View API.  This goes against simplicity and isn't necessarily expected.
- If a developer wants to fire `render` manually, the developer must be educated about `_supportsRenderLifecycle`.  I can see it being not necessarily ideal that a developer has to know more than "just trigger `render`" to get the job done.

After talking this through, my main concern is that `render/onRender`, `destroy/onDestroy`, and `dom:refresh/onDomRefresh` are invoked during `Region#show` and CollectionView lifecycle but not when calling `this.render()` or `this.remove()`.  It seems that the only way to reliably fire those events in all circumstances is if the developer takes responsibility to fire them in `render()` or `remove()` overrides, thus requiring education about `_supportsRenderLifecycle`.

I do like us "just taking care of" those events transparently for internal use, but the second a developer learns about and starts relying on those events, there could be a logical gap that opens up.  It seems that, no matter which direction we go, the developer will need some education on full lifecycle support for BB Views.  Considering education is necessarily either way, perhaps simpler is better.

All of this conversation has been extremely valuable to help me understand our vision for supplementing Bb better.  I am going to try an alternative PR that addresses the test fails without patching `render` or `destroy`.  I think we will be better able to then finalize the best approach.

jasonLaster:
Discussed as a group in the weekly meeting and decided to move forward.

@ianmstew great job cleaning up the tests.

Issue #2601, opened 2015-06-17
Remove component.json support?


paulfalgout:
Does anyone use this?  Babysitter's had been a version outdated for a while and no one complained.

If it is in use, is it important that the dependency's component.json files are maintained / available

For instance:

https://github.com/marionettejs/backbone-metal/pull/23

Is it necessary?

thejameskyle:
Component seems pretty dead to me: http://npm-stat.com/charts.html?package=component

ianmstew:
I would :+1: its removal in `next`.  Since Metal is behind a feature flag in `2.5` I think it's acceptable to keep Marionette supporting it while Metal does not until the breaking release when Metal is fully integrated by default.

PR #2664, opened 2015-07-09
Remove component.json

cmisenas:
Fixes https://github.com/marionettejs/backbone.marionette/issues/2601


pascalpp:
Does marionette also need to be 'unpublished' from component?

cmisenas:
Good question. Looks like they use a [wiki page](https://github.com/componentjs/component/wiki/Components)

samccone:
I think doing this would be a breaking change right?

pascalpp:
No match for marionette on http://component.github.io. So maybe it was never published to component?

cmisenas:
@pascalpp It is on the old registry (their wiki page) which I believe is still being used cause I can run `component install marionette` successfully. 

@samccone yes. From my understanding though reading #2601 is that not many people are using it nowadays.

ahumphreys87:
yup this would be breaking, its going into `next` though so that's fine - we just dont push till 3.0.

:+1:

jasonLaster:
Merged. Thanks @cmisenas! We can always re-add if this registry takes off

PR #2675, opened 2015-07-16
`attachView` triggers events on parent view

### LINKED ISSUE #2568: childEvents do not get triggered on LayoutViews for child views added with attachView
I am attaching a child view to a `LayoutView` by calling `attachView` on one of the layout's regions and defining a `childEvents` attribute for the layout to handle events from the child view. I expected the `childEvents` to work the same whether I call `region.show(view)` or `region.attachView(view)`, but it turns out that `childEvents` handlers are never called when the view is added via `attachView`.

It looks like the `childEvents` handlers are not called because `attachView` does not set the `_parent` attribute on the view. Manually setting `_parent` on the view makes `childEvents` work as expected. Is this the intended behavior?

I am seeing this issue in Marionette 2.4.1.

Here is some example code to demonstrate the issue:

https://jsfiddle.net/72vxrufh/1/

```
var ButtonView = Marionette.ItemView.extend({
    triggers: {
        'click': 'clicked'
    }
});

var Layout = Marionette.LayoutView.extend({
    el: '#layout',
    
    template: false,
    
    regions: {
        buttonRegion: '#button-region'
    },
    
    childEvents: {
        clicked: function() {
            this.$el.append('<div>clicked</div>')
        }
    },
    
    onRender: function() {
        var buttonView = new ButtonView({el: $('#button')});
        // Uncommenting the following makes child events work as expected
        // buttonView._parent = this.buttonRegion;
        this.buttonRegion.attachView(buttonView);
    }
});

new Layout().render();
```

cmisenas:
Fixes #2568

This was a bit tricky for me to understand so I kept old commits in case anybody also wants to compare `show` and `attachView` behavior. Will squash after :eyes:

### DIFF
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -887,39 +887,57 @@ describe('region', function() {
 
   describe('when attaching an existing view to a region', function() {
     beforeEach(function() {
-      this.setFixtures('<div id="foo">bar</div>');
+      this.setFixtures('<div id="foo">bar<div id="bar"><div id="baz"></div></div></div>');
 
-      this.View = Backbone.View.extend({
-        onShow: function() {}
+      this.ItemView = Backbone.Marionette.ItemView.extend({
+        onShow: function() {},
+        triggers: {
+          'click': 'attachViewClicked'
+        }
       });
 
-      this.view = new this.View();
+      this.viewAttached = new this.ItemView({el: '#baz'});
 
-      this.sinon.spy(this.view, 'render');
-      this.sinon.spy(this.view, 'onShow');
+      this.sinon.spy(this.viewAttached, 'render');
+      this.sinon.spy(this.viewAttached, 'onShow');
 
-      this.region = new Backbone.Marionette.Region({
-        el: '#foo'
+      this.LayoutView = Backbone.Marionette.LayoutView.extend({
+        el: '#foo',
+        regions: {
+          barRegion: '#bar'
+        },
+        childEvents: {
+          attachViewClicked: function() {},
+        }
       });
 
-      this.sinon.spy(this.region, 'attachView');
-      this.region.attachView(this.view);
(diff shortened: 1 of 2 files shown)


megawac:
:+1:

cmisenas:
Hmm..gh now completely removes comments on old commits. 
Anyways, @samccone changed overriding `render` with `template: _.template('')` to make sure first tests were true.

jasonLaster:
:+1:

test/unit/region.spec.js
```diff
@@ -887,39 +887,57 @@ describe('region', function() {
...
     });
 
     it('should return the region', function() {
-      expect(this.region.attachView).to.have.returned(this.region);
+      expect(this.barRegion.attachView).to.have.returned(this.fooView.barRegion);
+    });
+
+    it('calls the child events defined on parent view', function() {
+      this.viewAttached.$el.click();
+      expect(this.fooView.childEvents.attachViewClicked).to.have.been.called;
```
jasonLaster: is this the test that would have failed earlier?

cmisenas: yep

ahumphreys87:
great work @cmisenas :+1:

ahumphreys87:
could you squash this and then we can merge :smile:

cmisenas:
done :)

jasonLaster:
@cmisenas looks like you've got a random commit in your PR that's not on patch. Mind killiing that guy

https://github.com/cmisenas/backbone.marionette/commits/patch

cmisenas:
woops. fixed

ianmstew:
:+1:

jasonLaster:
Nice work @cmisenas! Keep em coming

cmisenas:
Thanks. I will. :smile_cat:

PR #2676, opened 2015-07-16
Allow over-ridable renderer for Item and Composite views at an instance level.

jshthornton:


jasonLaster:
We've discussed this in the past, but I think this makes sense.

jasonLaster:
At a minimum we can add a getRenderer that wraps Mn.renderer

ahumphreys87:
yeah, I'd prefer `getRenderer` for consistency.

@jshthornton could you remove the changes in the built versions please?

paulfalgout:
`getRenderer` seems like it could just be added to the base `View` as well

jshthornton:
I agree with this. I might be able to find the time to contribute the changes.

jshthornton:
@ahumphreys87 I need to use this fork as production until this PR is applied, therefore I need the build files to accompany them.

jshthornton:
Changes have been applied. Awaiting Travis.

src/view.js
```diff
@@ -349,6 +349,11 @@ Marionette.View = Backbone.View.extend({
     });
   },
 
+  // Returns the view's renderer.
+  getRenderer: function() {
```
samccone: we need to add docs about this

src/item-view.js
```diff
@@ -85,7 +84,7 @@ Marionette.ItemView = Marionette.View.extend({
     var data = this.mixinTemplateHelpers(this.serializeData());
 
     // Render and add to el
-    var html = Marionette.Renderer.render(template, data, this);
+    var html = (this.getRenderer()).render(template, data, this);
```
samccone: We should user our getOption helper here to get this value

src/composite-view.js
```diff
@@ -92,7 +91,7 @@ Marionette.CompositeView = Marionette.CollectionView.extend({
     this.triggerMethod('before:render:template');
 
     var template = this.getTemplate();
-    var html = Marionette.Renderer.render(template, data, this);
+    var html = (this.getRenderer()).render(template, data, this);
```
samccone: We should user our getOption helper here to get this value

src/item-view.js
```diff
@@ -5,7 +5,6 @@
 // with underscore.js templates, serializing the view's model or collection,
 // and calling several methods on extended views, such as `onRender`.
 Marionette.ItemView = Marionette.View.extend({
-
```
samccone: :x: whitespace change

src/composite-view.js
```diff
@@ -7,7 +7,6 @@
 // Extends directly from CollectionView and also renders an
 // a child view as `modelView`, for the top leaf
 Marionette.CompositeView = Marionette.CollectionView.extend({
-
```
samccone: :x: whitespace change

jasonLaster:
@jshthornton mind re-opening PR against `next` also you don't need to do a build or anything, just your patch :)

ianmstew:
@jshthornton Can you remove the build assets from the commit?  `lib/*`

ahumphreys87:
Closed in favour of #2682

PR #2682, opened 2015-07-29
Get renderer functionality for views

jshthornton:
This will allow views to specify the renderer they want to use at a class or instance level.
Therefore some views can use mustache, some can use lodash, and others can use Virtual DOM.


ianmstew:
I'm sure you're on top of this, but looks like this need a rebase.  Also, once you're satisfied this supersedes #2676 can you close it?

paulfalgout:
Closing in favor of #2700

PR #2687, opened 2015-07-30
Update behaviors.js

wvankuipers:
Fixed a small typo

### DIFF
--- a/src/behaviors.js
+++ b/src/behaviors.js
@@ -47,7 +47,7 @@ Marionette.Behaviors = (function(Marionette, _) {
         behaviorEvents = Marionette.normalizeUIKeys(behaviorEvents, getBehaviorsUI(b));
 
         var j = 0;
-        _.each(behaviorEvents, function(behaviour, key) {
+        _.each(behaviorEvents, function(behaviorHandler, key) {
           var match     = key.match(delegateEventSplitter);
 
           // Set event name to be namespaced using the view cid,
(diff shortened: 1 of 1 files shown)


src/behaviors.js
```diff
@@ -47,7 +47,7 @@ Marionette.Behaviors = (function(Marionette, _) {
         behaviorEvents = Marionette.normalizeUIKeys(behaviorEvents, getBehaviorsUI(b));
 
         var j = 0;
-        _.each(behaviorEvents, function(behaviour, key) {
+        _.each(behaviorEvents, function(behaviorEvent, key) {
```
ahumphreys87: I'm not sure `behaviorEvent` is the correct term here either. Maybe `handler`?

wvankuipers: It's a item from the behaviorEvents collection, so behaviorEvent seems to be the correct phrase. Handler could also be working, but I find it a bit more abstract.

ahumphreys87: I think the actual event is the 'key' though. In this case 'behaviorEvent'
is what actually responds to the event.

wvankuipers: Noted, maybe `eventHandler` is the right therm then?

ahumphreys87: Yeah sounds good to me :)

wvankuipers:
@ahumphreys87 Done!

jasonLaster:
that's fine, Andrew is being nice because he's british. `behaviour` is a perfectly valid spelling in my book :)

wvankuipers:
@jasonLaster Yes it is, but it's the only time it's being used in the entire codebase! I noticed because I couldn't find it using a search ;)

blurrah:
+1 for Marionette's independence day!

ahumphreys87:
Hehe I hadn't even realised it was the british spelling! I'm a convert now - too many PR comments making me change spelling :stuck_out_tongue:

paulfalgout:
Let's get these commits squashed and I'm :+1:

wvankuipers:
Here you go

jasonLaster:
Thanks @wvankuipers. 

@marionettejs/marionette-core note the PR is against master and not `patch`

jasonLaster:
Thanks! it's in

PR #2690, opened 2015-08-04
Expose `originalEvent` to trigger handler args

### LINKED ISSUE #2677: Wouldn't it be wonderful if triggers would also pass original event down?
Now you're passing only this:
```javascript
        var args = {
          view: this,
          model: this.model,
          collection: this.collection
        };
```
Example: i have collection view that listens to its children to open modal window to play video, but since i can't have access to original event, i can't get video id to find it in model to get link to it.

cvisco:
Fixes #2677

This is a minor change to `AbstractView#_buildViewTrigger`, which allows
access to the DOM event that was triggered on the view. This is useful
when you need to access properties of the event that would not otherwise
be available on just the `view` instance.

### DIFF
--- a/docs/marionette.abstractview.md
+++ b/docs/marionette.abstractview.md
@@ -282,8 +282,11 @@ includes the following:
 * view
 * model
 * collection
+* originalEvent
 
-These properties match the `view`, `model`, and `collection` properties of the view that triggered the event.
+The first three properties match the `view`, `model`, and `collection`
+properties of the view that triggered the event, whereas `originalEvent`
+is a reference to the DOM event that was initially triggered on the view.
 
 ```js
 var MyView = Marionette.View.extend({
@@ -300,6 +303,7 @@ view.on("some:event", function(args){
   args.view; // => the view instance that triggered the event
   args.model; // => the view.model, if one was set on the view
   args.collection; // => the view.collection, if one was set on the view
+  args.originalEvent; // => the click event that was triggered on the view
 });
 ```
 
--- a/src/abstract-view.js
+++ b/src/abstract-view.js
@@ -254,7 +254,15 @@ Marionette.AbstractView = Backbone.View.extend({
     var eventName = hasOptions ? options.event : triggerDef;
 
     return function(e) {
+      var args = {
+        view: this,
+        model: this.model,
+        collection: this.collection
+      };
+
       if (e) {
+        args.originalEvent = e;
+
         if (e.preventDefault && options.preventDefault) {
           e.preventDefault();
         }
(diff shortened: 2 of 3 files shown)


jmeas:
The existing (arguably weird) method signature comes from Derick's desire to **not** expose the event, which `triggerMethod` does by emitting an event. One might argue that Backbone nudges us toward this, as well, since the `events` hash doesn't convert a DOM event into a BB one (which would effectively 'expose' the DOM event), but, rather, gives you a hook to a method on the view itself.

I can understand the reasoning behind this: it seems reasonable to expect that each view should manage its own DOM events. I mean, that's kinda what views are for, you know?

@cvisco , are you using this because you want the view *itself* to have access to the event? Or because you want some other view to access it?

If it's the former, I encourage you to use `events`, then manually fire an event afterward. If it's the latter, I'm curious to hear more about the use case.

cvisco:
@jmeas My use case(s) primarily deal with behaviors. In general I don't want to include an events hash in my behavior code because the event signature might change slightly from view to view. Instead, I'd prefer to only ever use triggers to convert the unique event in the DOM to a standard method in the behavior. Now, this works as-is most of the time, but there are situations where it does not. 

For example, let's say I have created

cvisco:
@jmeas My use case(s) primarily deal with behaviors. In general I don't want to include an `events` hash in my behavior code because the event signature might change slightly from view to view. Instead, I'd prefer to only use `triggers` to convert the unique event in the DOM to a standard method in the behavior. This  keeps the behavior code simple and easier to test. 

Now, this works as-is most of the time, but there are situations where it does not. For example, let's say I have created a `ContextMenuBehavior` that uses a jQueryUI plugin to display a custom context menu. Ideally I would set this up so the `View` simply triggers a method:

```js
triggers: {
    'contextmenu': 'show:menu'
}
``` 

The behavior in this case only needs an `onShowMenu` method. But, the jQueryUI context menu plugin wants the actual event to inspect, which the `onShowMenu` method doesn't currently get access to. In this case I have to find a workaround. I can include an events hash in the behavior, or as you mention above, I could use an `events` hash on the view to call a view method that can then re-trigger a `show:menu` event but pass along the original event. 

In the first case, this means the event signature must always be the same in every view the behavior is mixed in to (e.g. I can't have `'contextmenu': 'show:menu'` in one view and `'contextmenu .some-element': 'show:menu'` in another). In the latter case, I'm adding extra boilerplate code to each view to work around a limitation (intentional or not) of the framework. 

Passing along a reference to the original event allows a behavior access to certain information it might need—"what was the mouse x and y for this event so I can position a tooltip", for example—while keeping them decoupled from the view's unique DOM and maintaining a concise, declarative style in the views.

This is my take on it. I'm curious if @nezaidu (who filed the original issue) or @ahumphreys87 has other thoughts.

jasonLaster:
@cvisco thanks for the suggestion. I'm closing the PR as we want DOM events to be encapsulated to the view. If you'd like to share action specific data you should do

```js
this.triggerMethod('show:menu' { activeMenu: 'poop' } );
```

Issue #2689, opened 2015-08-03
underdash/Lodash v4 support


megawac:
There are some major changes taking place in lodash v4/underdash.

First of I'd like to suggest that v3 update its backbone and underscore dependencies to something like `>= 1.1.2` and either `>= 1.6` or `>= 1.8`, respectively (which would have other advantages besides making it less of a pain to support cross version compatibility). Next, we if we want to be future compatible with lodash v4 (should be coming soon) and underdash there are some refactoring to be done. Below is a non exhaustive list of changes (feel free to edit this issue directly):

- [ ] Remove any uses of alias methods
- [ ] drop all uses of context binding (`_.each(items, function, context)` be replaced with `_.each(items, _.bind(function, context))` For this one in particular, I think using babel for arrow functions would be a great win
- [ ] Try to use only core functions (see underdash)

I haven't gone over the codebase for specifics, this list is only from memory. For specific changes see https://github.com/lodash/lodash/wiki/Roadmap

megawac:
This may be a good reason to reopen/fork https://github.com/marionettejs/backbone.marionette/pull/2440

paulfalgout:
I'd be :+1: for updating the Bb and _ dependencies.  I'm not sure if there's any strong feelings against that for v3?  `1.1.2` and `1.6.0` are pretty old now, but I'm sure widely used..  I'd doubt that anyone pre-that would be upgrading to Mn v3.  But we should do it particularly if that opens the path easier for future proofing and less maintenance headaches.

Everything else though seems doable and non-breaking outside of a major release?  I'm just nervous about anything else holding up v3.

jasonLaster:
Thanks @megawac. I think we should feel safe upgrading in a major release. I agree with Paul with respect to starting another Major project right now. If a PR was opened that did the upgrade and the tests passed I'd merge, or if there was a way to do a small breaking PR and then do incremental non-breaking improvements that would be fine.

This might be one of the rare cases where an AST search could help: http://www.graspjs.com/ identify how much work there is

ahumphreys87:
yup im pro for doing this as well

anthony-redFox:
Hello, I try create report using underscore/lodash in MarionetteJs (only folder src).
https://docs.google.com/spreadsheets/d/1auq_MbgMFqbhx7u__ky4XhnU_j_Z16LEdQCzVThQAp0/edit?usp=sharing

In core Underdash
_.invert
_.keys
_.each
_.extend
[...]

PR #2693, opened 2015-08-07
Remove _.isUndefined usage

anthony-redFox:
#2689

typeof more clear then void(0) and the same by performance.
https://jsperf.com/typeof-vs-undefined-check/12

### DIFF
--- a/src/abstract-view.js
+++ b/src/abstract-view.js
@@ -96,7 +96,7 @@ Marionette.AbstractView = Backbone.View.extend({
 
     // normalize ui keys
     events = this.normalizeUIKeys(events);
-    if (_.isUndefined(eventsArg)) {this.events = events;}
+    if (typeof eventsArg === 'undefined') {this.events = events;}
 
     var combinedEvents = {};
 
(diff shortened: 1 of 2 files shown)


paulfalgout:
@megawac care to weigh in?  underscore switched to `=== void 0` from typeof and lodash is simply `=== undefined`

anthony-redFox:
The undefined is not a reserved word, it can be used as an variable name in any scope other than the global scope. So `=== undefined` should be slow than `=== void 0` or `typeof` with scopes.

Underscore swhiched to `=== void 0` because of check ActiveX Objects from Internet Explorer.
We don't use ActiveX Objects.

paulfalgout:
Ah, yeah in lodash I believe undefined is referring to a local `var undefined;`

megawac:
I would suggest == null

anthony-redFox:
@megawac  - my opinion it is not good idea, it is conflict with jshint with option 'eqeqeq',  and many misunderstanding contains '=='

samccone:
I am ok with this

jasonLaster:
@samccone which way do you propose
@anthony-redFox mind rebaing off of next. You'll get passing tests :)

anthony-redFox:
@jasonLaster done

PR #2694, opened 2015-08-08
Modify how views execute the View constructor to better allow monkey patching

timsnadden:
In my app I wanted to add some custom code to all views as follows:

```javascript
Marionette.View.prototype.constructor = (function (original) {
	return function () {
		// Do some custom thing to all views before executing original constructor
		// ...
		original.apply(this, arguments);
	};
})(Marionette.View.prototype.constructor);
```

Without the changes in this pull request the monkey patched function doesn't get executed.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -25,7 +25,7 @@ Marionette.CollectionView = Marionette.View.extend({
     this.once('render', this._initialEvents);
     this._initChildViewStorage();
 
-    Marionette.View.apply(this, arguments);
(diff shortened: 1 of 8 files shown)


jasonLaster:
I like the change. I've thought for awhile we should let debs have the ability to hook in here. It's of course dangerous, but if you communicate it appropriately j think it's fine. 

Please drop the guild commit :)

timsnadden:
I've dropped the build commit. D'oh!

jasonLaster:
Nice

ahumphreys87:
closing in favour of #2725

PR #2698, opened 2015-08-11
drop iojs from matrix

jasonLaster:


paulfalgout:
:+1:  BTW this is related to some issue with old jsdom and io.js.  Not particularly with io.js and Mn

PR #2700, opened 2015-08-12
Utilize getRenderer to allow for custom renderers

paulfalgout:
Opening in favor of https://github.com/marionettejs/backbone.marionette/pull/2682

Still needs:
- [ ] Documentation
- [ ] API Docs
- [ ] A test?
- [ ] Sanity check thumbs

### DIFF
--- a/src/abstract-view.js
+++ b/src/abstract-view.js
@@ -366,6 +366,11 @@ Marionette.AbstractView = Backbone.View.extend({
     });
   },
 
+  // Returns the view's renderer.
+  getRenderer: function() {
+    return Marionette.Renderer;
+  },
+
   // Imports the "normalizeMethods" to transform hashes of
   // events=>function references/names to a hash of events=>function references
   normalizeMethods: Marionette.normalizeMethods,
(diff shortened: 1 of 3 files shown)


paulfalgout:
Alright, in the interest of not expanding the API, I'm now  :-1: on this implementation.

We already provide a way for the [renderer to be aware of the view](https://github.com/marionettejs/backbone.marionette/blob/master/docs/marionette.itemview.md#itemview-render) and we provide a method for [customizing the renderer](https://github.com/marionettejs/backbone.marionette/blob/master/docs/marionette.renderer.md#custom-template-selection-and-rendering).

To solve the OP's the original concern you could certainly designate a view type (or any hundred of other ways) but your rendered could simply be:
```js
Marionette.Renderer.render = function(template, data, view){
  if(view.getOption('viewType' === 'IDOM'){
    return awesomeSauceRender(template)(data);
  }
  return $(template).tmpl(data);
};
```

I'm closing this for now.  In the very least it should likely be a 3.x conversation at this point.  Certainly reopen if necessary

Issue #2464, opened 2015-03-30
View childViewEvents should support trigger


paulfalgout:
`LayoutView` only supports `triggerMethod`. This inconsistency with `CollectionView` `childEvents`  has come up more than once now in gitter.

jasonLaster:
Agreed

jasonLaster:
I'm not sure if there's a ticket for this, but I wonder if the "fix" could be to make trigger triggerMethod in V3. In pseudo code, backbone.Events.trigger = triggerMethod. 

The advantage is that we'd be supercharging B.E not competing. If this is a distraction we'll just override trigger for M.AV to support bubbling. Thoughts?

paulfalgout:
So perhaps this is bad pattern, but `triggerMethod` is not async safe, where-as `trigger` is.  Arguably you can use `triggerMethod` safely if you setup a different listener, or you can constantly be watching for situations not to trigger, but for instance:
This will trigger safely.  onSomeEvent will never be called.
```js
$.when(promise).done(_.bind(this.trigger, this, 'some:event'));
this.on('some:event', this.onSomeEvent);
this.destroy();
promise.resolve();
```
This will still call the method onSomeEvent:
```js
$.when(promise).done(_.bind(this.triggerMethod, this, 'some:event'));
this.destroy();
promise.resolve();
```

jasonLaster:
@paulfalgout - this is a really interesting point, mind creating a separate issue for this discussion. I'd like to see why that is w/ the implementation

ianmstew:
Hey guys, can we bump this to 2.5?  I am super anxious to have this is in, especially after making it clear in 2.4.3 docs that LayoutView does not support `childEvents` triggered with, er, `trigger`.

paulfalgout:
FWIW if we fix this it will make bubbling work everywhere https://github.com/marionettejs/backbone.marionette/issues/2667

Previously I would have considered this a major issue.  With `childViewEventPrefix: false`  (hopefully the default for v4) I think this would actually be ok.  It would be easy enough to avoid bubbling or utilizing it when you needed to, and I think this is already the expected behavior.

PR #2705, opened 2015-08-14
Improved childEvents docs

### LINKED ISSUE #2531: Users keep getting confused by the fact that 'childEvents' relies on 'triggers' and not 'events'
All,

Simply reporting this because I have heard it come up twice in 24 hours in the Gitter channel.

When a user sets up a childEvents object like so:

    childEvents: {
        'item:clicked': _.noop
    }

they initially believe that the correct way to run this code from a child is by defining:

    events: {
       'click': 'item:clicked' 
    }

rather than: 

    triggers: {
        'click': 'item:clicked'
    }

I understand both arguments. Events are being triggered at the lower level and then the event is handled at an upper level... but, in my experience, most newcomers do not understand the point of 'triggers' after becoming familiar with 'events' and generally don't think to use them unless forced. So, when something says "childEvents" rather than "childTriggers", it leads them to believe they should go through 'events'.

I'm not sure if this warrants a change at all, but thought it was worth mentioning.

ianmstew:
Aims to fix #2531 through clearer examples rather than a "hey, don't do this" warning (unless someone prefers to have that warning).  I also made it clearer the first argument of a childView event is the child view and took note that LayoutView only supports child view events triggered with `triggerMethod`.

Side note: this very much highlights the need for #2464 resolution.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -213,16 +213,16 @@ in the constructor function call, to get a view instance.
 
 ### CollectionView's `childEvents`
 
-You can specify a `childEvents` hash or method which allows you to capture all bubbling childEvents without having to manually set bindings. The keys of the hash can either be a function or a string that is the name of a method on the collection view.
+A `childEvents` hash or method permits handling of child view events without manually setting bindings. The values of the hash can either be a function or a string method name on the collection view.
 
 ```js
 // childEvents can be specified as a hash...
 var MyCollectionView = Marionette.CollectionView.extend({
 
-  // This callback will be called whenever a child is rendered or emits a `render` event
   childEvents: {
+    // This callback will be called whenever a child is rendered or emits a `render` event
     render: function() {
-      console.log("a childView has been rendered");
+      console.log('A child view has been rendered.');
     }
   }
 });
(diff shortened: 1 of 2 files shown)


ianmstew:
@paulfalgout @jasonLaster this will knock out the last issue left in `2.4.3`.

ahumphreys87:
:+1: lgtm to me

docs/marionette.collectionview.md
```diff
@@ -234,33 +234,53 @@ var MyCollectionView = Marionette.CollectionView.extend({
...
 var ParentView = Marionette.CollectionView.extend({
+
   childView: ChildView,
 
   childEvents: {
-    'show:message': function () {
-      console.log('The show:message event bubbled up to the parent.');
-    }
+    'show:message': 'onChildShowMessage',
+    'submit:form': 'onChildSubmit'
```
paulfalgout: this should be `onChildSubmitForm`

docs/marionette.layoutview.md
```diff
@@ -105,73 +105,71 @@ new Marionette.LayoutView({
...
-
-The function is called in the context of the view. The first parameter is
-the child view, which emitted the event, the remainder are the arguments
-associated with the event.
+A `childEvents` hash or method permits handling of child view events without manually setting bindings. The values of the hash can either be a function or a string method name on the collection view.
 
 ```js
 // childEvents can be specified as a hash...
-var MyLayoutView = Marionette.LayoutView.extend({
+var MyCollectionView = Marionette.CollectionView.extend({
```
paulfalgout: This and lines 123 and 155 should be `LayoutView`

docs/marionette.layoutview.md
```diff
@@ -105,73 +105,71 @@ new Marionette.LayoutView({
...
   }
 });
 
 // ...or as a function that returns a hash.
-var MyLayoutView = Marionette.LayoutView.extend({
+var MyCollectionView = Marionette.CollectionView.extend({
 
   childEvents: function() {
     return {
-      render: this.onChildRender
```
paulfalgout: I think I actually prefer not adding the "ed", but that could be a matter of opinion..  I suppose it was already this way in the CollectionView docs, but I try and stick to before prefixes and all of my after events are present tense to match how the events are written.

docs/marionette.layoutview.md
```diff
@@ -105,73 +105,71 @@ new Marionette.LayoutView({
...
 
-  onChildRender: function(childView) {
+  onChildRendered: function () {
+    console.log('A child view has been rendered.');
   }
 });
 ```
 
-This also works for custom events that you might fire on your child views.
+`childEvents` also catches custom events fired by a child view.  Take note that the first argument to a `childEvents` handler is the child view itself.  Caution: Events triggered on the child view through the `triggers` hash or `this.trigger` are not yet supported for LayoutView `childEvents`.  Use strictly `triggerMethod` within the child view.
```
paulfalgout: this isn't true.  The `triggers` hash will work right?  `triggers` calls triggerMethod internally so I think it should be fine.  It's just the `Backbone.Event.trigger` function that isn't going to work.

docs/marionette.layoutview.md
```diff
@@ -105,73 +105,71 @@ new Marionette.LayoutView({
...
-  var ParentView = new Marionette.LayoutView.extend({
-    childEvents: {
-      'show:message': function (childView, msg) {
-        console.log('The show:message event bubbled up to the parent.');
-      }
-    },
+  // Events hash defines local event handlers that in turn may call `triggerMethod`.
+  events: {
+    'click .button': 'onClickButton',
+    'submit form': function () { this.triggerMethod('submit:form'); }
```
paulfalgout: I think the triggers hash will work, so if so, we should move this to match the `CollectionView` example

docs/marionette.layoutview.md
```diff
@@ -105,73 +105,71 @@ new Marionette.LayoutView({
...
-```
+  onClickButton: function () {
+    this.triggerMethod('show:message', 'foo');
+  }
+});
+
+// The parent uses childEvents to catch that custom event on the child view
+var ParentView = Marionette.CollectionView.extend({
+
+  childView: ChildView,
```
paulfalgout: this won't be true for a layout

docs/marionette.layoutview.md
```diff
@@ -105,73 +105,71 @@ new Marionette.LayoutView({
...
+
+  childEvents: {
+    'show:message': 'onChildShowMessage',
+    'submit:form': 'onChildSubmit'
```
paulfalgout: dupe of above, `onChildSubmitForm`

ahumphreys87:
closing in favour of #2710

ianmstew:
I forgot to say thanks for the review, @paulfalgout, and @ahumphreys87 for running with the changes.

PR #2440, opened 2015-03-20
Transition to ES6 modules.

trusktr:
WIP

PR #2706, opened 2015-08-14
Use ES6 modules

cmisenas:
**WIP**
This is in continuation of PR #2440.
~~For now, I simply picked up the old branch and rebased off of `next`.~~
* [x] Run babel-boilerplate/Create gulpfile
* [ ] Create `Mn` index file
* [ ] Move existing tasks from grunt to gulp/Remove grunt
* [ ] Babel transpiling support
* [ ] Convert files to es6 modules


.gitignore
```diff
@@ -1,16 +1,31 @@
...
+# Dependency directory
+# Commenting this out is preferred by some people, see
+# https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git-
+node_modules
+bower_components
+coverage
+tmp
+
+# Users Environment Variables
+.lock-wscript
```
jasonLaster: is this babel boilerplate? what's the rationale for all these changes?

.travis.yml
```diff
@@ -1,29 +1,9 @@
...
-    - env: UNDERSCORE=1.6 BACKBONE=1.0
-    - env: UNDERSCORE=1.4.4 BACKBONE=1.1.0
-    - env: UNDERSCORE=1.5 BACKBONE=1.1
-    - env: UNDERSCORE=1.7 BACKBONE=1.2
-    - node_js: "0.10"
-      env: UNDERSCORE=1.8 BACKBONE=1.2
-    - env: LODASH=2.4 BACKBONE=1.1
-    - env: LODASH=3.0 BACKBONE=1.1
-    - node_js: "io.js"
-      env: LODASH=3.1 BACKBONE=1.0
```
jasonLaster: why remove the matrix?

.travis.yml
```diff
@@ -1,29 +1,9 @@
...
-env: MAINRUN=true
-before_install:
-  - npm config set ca ""
-install:
-  - curl -L https://rawgit.com/marionettejs/travis-ci-setup/master/ci.sh | bash
```
jasonLaster: why is install gone? is this replaced w/ some boilerplate stuff?

Gruntfile.js
```diff
@@ -122,7 +122,10 @@ module.exports = function(grunt) {
...
       tests: {
         options: {
-          require: 'test/unit/setup/node.js',
+          require: [
+            'babel/register',
+            'test/unit/setup/node.js'
+          ],
           reporter: grunt.option('mocha-reporter') || 'nyan',
           clearRequireCache: true,
           mocha: require('mocha')
```
jasonLaster: is the gruntfile used now?

gulpfile.js
```diff
@@ -0,0 +1,181 @@
...
+});
+
+// Set up a livereload environment for our spec runner
+gulp.task('test-browser', ['build-in-sequence'], function() {
+  $.livereload.listen({port: 35729, host: 'localhost', start: true});
+  return gulp.watch(otherWatchFiles, ['build-in-sequence']);
+});
+
+// An alias of test
+gulp.task('default', ['test']);
```
jasonLaster: how much is similar to boilerplate?

package.json
```diff
@@ -23,8 +23,10 @@
     }
   ],
   "scripts": {
-    "test": "grunt test --mocha-reporter=dot",
-    "coverage": "grunt coverage --mocha-reporter=dot"
+    "test": "gulp",
+    "test-browser": "gulp test-browser",
+    "build": "gulp build",
+    "coverage": "gulp coverage"
```
jasonLaster: w00t

package.json
```diff
@@ -47,37 +49,57 @@
...
-    "grunt-coveralls": "0.3.0",
-    "grunt-env": "0.4.1",
-    "grunt-istanbul": "0.3.0",
-    "grunt-jscs": "^1.5.0",
-    "grunt-lintspaces": "0.7.0",
-    "grunt-mocha-test": "0.10.2",
-    "grunt-plato": "0.1.4",
-    "grunt-preprocess": "4.0.0",
-    "grunt-template": "0.2.3",
+    "grunt-babel": "^5.0.1",
```
jasonLaster: yay

src/abstract-view.js
```diff
@@ -7,6 +7,18 @@
...
+import normalizeMethods         from './utils/normalizeMethods';
+import normalizeUIKeys          from './utils/normalizeUIKeys';
+import normalizeUIValues        from './utils/normalizeUIValues';
+import mergeOptions             from './utils/mergeOptions';
+import proxyGetOption           from './utils/proxyGetOption';
+import MonitorDOMRefresh        from './dom-refresh';
+import MarionetteError          from './error';
+import { proxyBindEntityEvents, proxyUnbindEntityEvents } from './bind-entity-events';
+import { _triggerMethod }       from './trigger-method';
+
```
jasonLaster: so wonderful

src/abstract-view.js
```diff
@@ -92,7 +104,7 @@ Marionette.AbstractView = Backbone.View.extend({
   // Overriding Backbone.View's `delegateEvents` to handle
   // `events` and `triggers`
   delegateEvents: function(eventsArg) {
-    var events = Marionette._getValue(eventsArg || this.events, this);
+    var events = _getValue(eventsArg || this.events, this);
```
jasonLaster: i can't comment everywhere, but these changes make me :smile: :smile_cat: :smiley_cat:

src/backbone-marionette.js
```diff
@@ -0,0 +1,65 @@
+import
```
jasonLaster: is there a corressponding red file?

src/bind-entity-events.js
```diff
@@ -15,92 +15,99 @@
...
+
+// Proxy `bindEntityEvents`
+function proxyBindEntityEvents(entity, bindings) {
+  return bindEntityEvents(this, entity, bindings);
+}
+
+// Proxy `unbindEntityEvents`
+function proxyUnbindEntityEvents(entity, bindings) {
+  return unbindEntityEvents(this, entity, bindings);
+}
```
jasonLaster: so great

gulpfile.js
cmisenas: Haven't changed any of them yet except for the test setup script path Mn uses.

jasonLaster:
I think I speak for everyone when I say that this is work is epic.

.travis.yml
cmisenas: Yeah, basically @jasonLaster . It was a big commit so missed this but will change it back again tomorrow.

Gruntfile.js
cmisenas: Nah, old commit. Will start removing grunt next.

package.json
cmisenas: Darn. Thought I got 'em all. Had to put some of the old deps back cause boilerplate overwrote them. Will change.

src/backbone-marionette.js
cmisenas: Not sure I understand?

.gitignore
paulfalgout: This is the boilerplate.  We'll want that DS_Store back at least...  jsdoc,,  not sure what else might be necessary

gulpfile.js
paulfalgout: whenever you get to adding a banner to the build files, I've done this on an older version of the boilerplate https://github.com/RoundingWellOS/marionette.toolkit/blob/master/gulpfile.js#L67 if it is useful.

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "Make your Backbone.js apps dance!",
   "version": "2.4.1",
   "homepage": "https://github.com/marionettejs/backbone.marionette",
-  "main": "lib/core/backbone.marionette.js",
+  "main": "dist/backbone.marionette.js",
```
paulfalgout: we may want to keep the dist locations the same as before?  particularly if we're keeping the core build?  @jasonLaster

package.json
```diff
@@ -47,37 +49,57 @@
...
+    "esperanto": "^0.7.4",
     "grunt": "0.4.5",
-    "grunt-cli": "0.1.13",
-    "grunt-contrib-clean": "0.5.0",
-    "grunt-contrib-concat": "0.1.2",
-    "grunt-contrib-connect": "0.1.2",
-    "grunt-contrib-jshint": "0.10.0",
-    "grunt-contrib-uglify": "0.2.0",
-    "grunt-contrib-watch": "0.6.1",
-    "grunt-coveralls": "0.3.0",
```
paulfalgout: are we dumping coveralls for code climate?  In general I find that's ok, but coveralls does sometimes warn in github on lost coverage which is nice?  Similarly we're losing plato.. which I honestly never used, but I'm sure it serves a purpose to someone?  @jasonLaster

src/backbone-marionette.js
```diff
@@ -0,0 +1,65 @@
...
+  } from './trigger-method';
+
+import { FEATURES, isEnabled } from './features';
+
+import MarionetteError from './error';
+
+Marionette = Marionette || {};
+
+// extend Marionette in the meantime
+Object.assign(Marionette, {
```
paulfalgout: I wonder if everything still needs to be attached to `Mn`?  Seems like a few things were attached to the root simply because we weren't using module loaders and couldn't import them.  Like specifically if we're not calling `Marionette._getValue()` anywhere we probably don't need to attach it..  I suspect maybe the proxy functions as well?

test/setup/node.js
```diff
@@ -0,0 +1,60 @@
...
+global.$ = global.jQuery = require('jquery');
+global._ = require('underscore');
+global.Backbone = require('backbone');
+global.Backbone.$ = global.$;
+global.Metal = require('backbone-metal');
+global.Marionette = Backbone.Marionette = {};
+require('backbone.babysitter');
+require('backbone.radio');
+global.slice = Array.prototype.slice;
+// requireHelper('bind-entity-events');
```
paulfalgout: So something the boilerplate does quite well is keep you from having to maintain what to require in..  I _think_ we could lose the `requireHelper` stuff and just setup jsDOM and Chai.. and then `require('babel/register');  require('./setup')();`  Then in the setup file, require in the dependencies, and just the main marionette src file.

test/unit/setup/node.js
```diff
@@ -14,10 +14,10 @@ if (!global.document || !global.window) {
   var jsdom = require('jsdom').jsdom;
```
paulfalgout: Is this file in use?  It seems like you only want `test/setup/node.js` and not this one.

paulfalgout:
Pretty exciting.. also daunting.

Seems like we need to come up with a list of what the end goal is here.  I'm sure there will be some additional stuff that could be worked on later.. ie: the bundled build, using coveralls over code climate maybe.. 

@megawac I know you were keen on converting to ES6.  Thoughts?

anthony-redFox:
PR to fix many tasks and large commits difficult to understanding.
The main goals in PR
1. Replace grunt on gulp
2. Added babeljs transpiring support
3. Convert to Es6 modules.

So its should be separate PR or Commits

gulpfile.js
cmisenas: Thanks @paulfalgout . Was trying to look around for an example.

src/backbone-marionette.js
cmisenas: Good catch. Will change.

test/unit/setup/node.js
cmisenas: @paulfalgout the old setup file used was this. I was thinking of keeping this and getting rid of the `test/setup/node.js` instead so not to break the existing setup. Would one be better than the other?

paulfalgout: I do think one of the strengths of the boilerplate is that you don't have to maintain the require list in a setup and a runner.html file.   And multiple times a new test will get added but we'll neglect the runner.html.  That said, I don't think that needs to be a requirement of this PR..

package.json
cmisenas: I'd suggest that be done in a new PR to keep this from blowing up? My next task is to push the old Grunt tasks over to gulp.

package.json
jasonLaster: yea, true. We should probably slot that in going from grunt to gulp.

jasonLaster:
Thanks @cmisenas I see your point about size and clarity. Maybe what I heard you saying is that we could do a small PR for adding the bare minimum gulp solution. 

Then, we can
 a) replace grunt w/ gulp
 b) add ES6 modules
 c) go nuts w/ babel

---

Thanks @anthony-redFox for weighing in!

gulpfile.js
```diff
@@ -0,0 +1,181 @@
+// Load Gulp and all of our Gulp plugins
```
thejameskyle: `gulpfile.babel.js` & use `import`'s

thejameskyle:
Just to note, changes like this are API changes as well since they can no longer be overwritten:

```diff
-    Marionette.MonitorDOMRefresh(this);
+    MonitorDOMRefresh(this);
```

thejameskyle:
> So its should be separate PR or Commits

Meh, just do it at once and get it over with.

jasonLaster:
@cmisenas closing this PR as we can create a new one after the gulp work lands

PR #2720, opened 2015-08-30
Fix pathname for tests in gruntfile

dougalg:
This is a submission towards dealing with https://github.com/marionettejs/backbone.marionette/issues/2715. I noticed that `grunt coverage` still fails with:

>Running "coveralls:default" (coveralls) task
>[warn] "2015-08-30T22:00:20.051Z"  'Repo token could not be determined.  Continuing without it.'
>\>\> Failed to submit coverage results to coveralls
>Warning: Task "coveralls:default" failed. Use --force to continue.
>
>Aborted due to warnings.

But I think this might be because I am not running grunt as travisci.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -112,7 +112,7 @@ module.exports = function(grunt) {
     },
 
     instrument: {
-      files: 'src/*.js',
+      files: 'src/**/*.js',
       options: {
         lazy: true,
         basePath: 'test'


dougalg:
I guess I should look into these tests failing in io.js

paulfalgout:
For io.js we actually just need to cherry-pick https://github.com/marionettejs/backbone.marionette/commit/ef405e0fb6bab2e2c59f3cdaea51b873c2e15a49 onto `next`

This PR LGTM :+1: 

Thanks so much!

jasonLaster:
@paulfalgout just pushed #ef405e0

jasonLaster:
Thanks @dougalg

dougalg:
My pleasure!

PR #2724, opened 2015-09-02
Improve handing of `at`-added models in filtered + sorted CollectionViews

### LINKED ISSUE #2714: Loss of code coverage for 2.4.3
https://github.com/marionettejs/backbone.marionette/commit/9070bad2940564e7a975e4b0310d25768d986fea

and

https://github.com/marionettejs/backbone.marionette/commit/454d2cbdd37e4c3f27abd3f655b9cbf84f04ea27

have taken code coverage below 100%

ianmstew:
- [x] Fixed edge case when adding models to filtered collectionView with `at` (#2726)
- [x] 100% coverage for collection-view.js (#2714)
- [x] 100% coverage for region.js (#2714)

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -99,17 +99,12 @@ Marionette.CollectionView = Marionette.View.extend({
 
   // Handle a child added to the collection
   _onCollectionAdd: function(child, collection, opts) {
-    var index;
-    if (opts.at !== undefined && !this.getOption('filter')) {
-      index = opts.index;
-      if (!index) {
-        // ignore sorting with at options specified
-        index = this.getViewComparator() ?
-                collection.indexOf(child) :
-                _.indexOf(this._filteredSortedModels(), child);
-      }
-    } else {
-      index = _.indexOf(this._filteredSortedModels(), child);
+    // `index` is present when adding with `at` since BB 1.2; indexOf fallback for < 1.2
+    var index = opts.at !== undefined && (opts.index || collection.indexOf(child));
+
+    // When filtered or when there is no initial index, calculate index.
+    if (this.getOption('filter') || index === false) {
+      index = _.indexOf(this._filteredSortedModels(index), child);
     }
 
     if (this._shouldAddChild(child, index)) {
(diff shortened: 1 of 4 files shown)


src/collection-view.js
```diff
@@ -101,8 +101,9 @@ Marionette.CollectionView = Marionette.View.extend({
   _onCollectionAdd: function(child, collection, opts) {
     var index;
     if (opts.at !== undefined && !this.getOption('filter')) {
-      index = opts.index;
-      if (!index) {
+      if (typeof opts.index === 'number') {
+        index = opts.index;
```
ianmstew: This line still needs coverage--have to look deeper into Backbone and where `index` is coming from.

paulfalgout: This was in gitter, but for documentation's sake.. opts.index is a Bb >= 1.2 feature so this check is essentially `if(Bb < 1.2){`  and the coverage report from travis is run with Bb 1.2.1.  Thus the coverage loss.

jasonLaster:
:+1:

ianmstew:
Hey, I added some extra code that took me hours to debug.  It seems the new logic introduced in https://github.com/marionettejs/backbone.marionette/commit/9070bad2940564e7a975e4b0310d25768d986fea led to a scenario where, when the view has a comparator AND a filter, the view is not added at the `at` index in the CollectionView.  Because the `at` index only makes sense _before_ filtering, the order of operations in `_filteredSortedModels()` is now: 1. Sort, 2. Optionally override `at` location, 3. Filter.  The override `at` location code isn't terribly efficient, but it being called at all is an edge case.

For the record, I don't like any of this, but status quo is that the `at` index in `collection.add(view, {at: index})` _also_ be the index in the CollectionView, _even when the CollectionView is explicitly overriding collection order by supplying a `viewComparator`_.  I think it would be MUCH more logical if the `at` index on the collection level did not override the CollectionView `viewComparator`.

ianmstew:
I don't know if any of this belong in a separate ticket/PR, because this is essentially a preemptive bugfix while addressing code coverage.

ianmstew:
Thanks @paulfalgout for the nudge.  This version is way more elegant.  I actually ended up adding a test case for BB < 1.2 because to write a workaround in the code compromised elegance :D.

Since this became such an unexpected deep-dive, I wrote up an issue with `2.4.2` that this solves (#2726) and re-cast the PR in terms of the CollectionView logic fixes, with test coverage being a side effect instead of the other way around.

ahumphreys87:
lgtm :+1: 

@jasonLaster @paulfalgout 1 more

paulfalgout:
so one small tweak I _think_ then I think it's :+1: merge ready

ianmstew:
Okay @paulfalgout, you should be able to merge it now :D

paulfalgout:
:+1:

PR #2727, opened 2015-09-04
Bump to Backbone 1.2.3

ianmstew:
Regressions causing test failures in `1.2.2` have been resolved in `1.2.3`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ matrix:
     - env: UNDERSCORE=1.6 BACKBONE=1.0
     - env: UNDERSCORE=1.4.4 BACKBONE=1.1.0
     - env: UNDERSCORE=1.5 BACKBONE=1.1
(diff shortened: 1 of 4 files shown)


ianmstew:
@paulfalgout [This line](https://github.com/ianmstew/backbone.marionette/blob/82defadd14e40bb819adfb308e6a0a69ffc56f2e/test/unit/sorted-views.spec.js#L61) is killing the build all of a sudden, even though previous builds containing this code are passing.  That line is attempting to array splice past the array's length, which I thought was valid because it works in the Chrome console.  Any ideas why this is suddenly an issue?

ianmstew:
Okay, after some investigation on my local, I think I should wait until @ahumphreys87 updates babysitter, wreqr, and radio before debugging this further.  It could be related to library node_modules version consolidation issues because of mismatched version ranges for backbone.

ahumphreys87:
:+1:

ianmstew:
@paulfalgout I think we're good!

PR #2728, opened 2015-09-04
Bump and build 2.4.3

ianmstew:
* [x] Changelog
* [x] Bumped version numbers
* [x] Rebase after all other PRs are merged
* [x] Build


ianmstew:
@ahumphreys87 @paulfalgout looks like we have a winner... agree?

lib/backbone.marionette.js
```diff
@@ -38,7 +38,7 @@
   /* istanbul ignore next */
   // Backbone.BabySitter
   // -------------------
-  // v0.1.7
+  // v0.1.8
```
ianmstew: Hmm, why is babysitter not 0.1.10?  Ahhh.  I think I know.

paulfalgout:
:+1:  finally :-)

jasonLaster:
@ianmstew - mind changing this PR to be against MASTER. That way the branch is truly representative of what will land. We should see all of the commits for example, including the merge commits.

ianmstew:
Moved to #2740

PR #2729, opened 2015-09-06
Simplify _renderTemplate

paulfalgout:
While researching the renderer I found _renderTemplate to be nearly identical between View and CompositeView with a couple of exceptions.
CompositeView's triggered events and called `bindUIElements` within it instead of in the render to ensure it was called before the children were rendered.

In this PR:
- I removed the events CompositeView was triggering.  I can't think of a use case where this event cannot be handled by either `before:render` or `before:render:collection` so I don't think there's reason to muddy up the event space.  _If_ someone is against their removal they can be added around a `Marionette.CollectionView.prototype._renderRemplate.call(this);`

- Then I moved `bindUIElements` up into the render between `_renderTemplate` and `_renderChildren`  It is called in the same place, resembles the regular view.render more and is enforced by tests.

- Additionally since the `_renderTemplate` was now the same I moved it into the `AbstractView`.  This will mean that `CollectionView` will now get this unused function, but it does make some things cleaner and might be moot if/when CompositeView is part of the `CollectionView`.

### DIFF
--- a/api/abstract-view.yaml
+++ b/api/abstract-view.yaml
@@ -16,6 +16,9 @@ functions:
   getTemplate: |
     @api public
     
+  _renderTemplate: |
+    @api private
+
   serializeModel: |
     @api public
     @param {} model
--- a/api/composite-view.yaml
+++ b/api/composite-view.yaml
@@ -54,17 +54,8 @@ functions:
   _renderChildren:
     description: |
       ...
-      
-      @api private
 
-    examples:
-      
-  _renderTemplate:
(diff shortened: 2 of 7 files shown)


ianmstew:
:+1: This is a beautiful thing.  Let's merge this @ahumphreys87 @jasonLaster.

ianmstew:
Also fixes #2752

jasonLaster:
:thumbsup: 

I like the render template, bind ui, render children change. Very clean

Issue #2695, opened 2015-08-09
rendering only childviews within a composite view


kp25:
I'm using composite view to render a table, with some filter like options on top of the table. I'm wrapping the table rows (collection) with table header and filter options. On applying the filter option i don't want the entire table to be re-rendered, i just want the collection, childviews to be re-rendered, so that template wrapping the collection could be rendered only once at the begining. 
Is there any way to do that like rendering only childviews within a composite view.? I'm looking for something like 
```javascript
//childView: xyz
childView.render(); // renders the childView 
```

jmeas:
:+1:

There should be a public method, say, `renderChildren()`, that does just this.

ianmstew:
Without a CompositeView at all in 3.0 (#2268), I don't see a need to separate `render()` and `renderChildren()` or maintain separate events (#1309).

PR #2730, opened 2015-09-06
Rename `render:collection` => `render:children

### LINKED ISSUE #1309: Rename `render:collection` => `render:children`
Because you're rendering the children not the collection

paulfalgout:
Resolves #1309

I think this is the correct naming strategy.  It is what the function is called, which will likely be made [public](https://github.com/marionettejs/backbone.marionette/issues/2695)

Addressing previous concerns..  the difference with the collectionview's children vs children views of a view, would be that views are not responsible for rendering their children in the same way that collectionview's are, so I don't think there will be confusion anymore than our current situation, being that a simple view doesn't currently throw `render:collection` even though it can certainly render a collection.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -49,7 +49,7 @@ will provide features such as `onShow` callbacks, etc. Please see
   * ["before:apply:filter" / "apply:filter" event](#beforeapplyfilter--applyfilter-event)
   * ["before:reorder" / "reorder" event](#beforereorder--reorder-event)
   * ["before:destroy" event](#beforedestroy-event)
-  * ["destroy" / "destroy:collection" event](#destroy--destroycollection-event)
+  * ["destroy" / "destroy:childen" event](#destroy--destroychildren-event)
   * ["before:add:child" / "add:child" event](#beforeaddchild--addchild-event)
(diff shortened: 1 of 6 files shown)


paulfalgout:
I added an additional commit for updating `destroy:collection` to match `destroy:children`

Really it seems like when these are triggered should be addressed.  If you were to call the public `destroyChildren` function these events will not get triggered, but that is likely an [issue for another day](https://github.com/marionettejs/backbone.marionette/issues/2731).

Particularly if we do work to reduce the CollectionView's API fixing this behavior may become trivial.

ianmstew:
Why are these events even necessary/useful in the first place?  Rendering a CollectionView implies rendering its children.  Same with destroying a CollectionView.

ianmstew:
My question applies more obviously to `destroy:children` since there isn't actually a separate `destroyChildren()` action.  `render:children` only makes sense to me if `renderChildren()` remains a separate action from `render()`, but without CompositeView there's less of a need for that separation.

paulfalgout:
so as `CompositeView` may be sticking around for v3, it is probably important for these to stay.  In which case we should likely just make `renderChildren` public.. `destroyChildren` should also have it's events fixed which is super easy.. and we can remove them for v4

Issue #2515, opened 2015-04-20
Object should support a default radio channel


paulfalgout:
Radio Events currently handle a hash containing `'channel eventName' : 'handler'`
```js
radioCommands: {
  'app doFoo': 'executeFoo',
},

radioRequests: {
  'resources bar': 'getBar',
},
```

As radio commands and requests can only be handled once per channel, I find most of my Objects are only utilizing a single channel for commands and requests... 

```js
radioRequests: {
  'resources foo': 'getFoo',
  'resources bar': 'getBar',
  'resources baz': 'getBaz',
  'other baz': 'getOtherBaz'
},
```

Therefore I think a `defaultRadioChannel` would be useful:

```js
defaultRadioChannel: 'resources',
radioRequests: {
  'foo': 'getFoo',
  'bar': 'getBar',
  'baz': 'getBaz',
  'other baz': 'getOtherBaz'
},
```

jasonLaster:
Hmm, I could see going either way on this one. Ultimately, I feel like keeping the channel in the key is useful documentation and consistent. 

Two usecases:
1. new developer sees `radioRequests: { 'foo': 'getFoo'}` and is not sure what's happening.
2. seasoned Mn developer sees  `radioRequests: { 'foo': 'getFoo', 'pizza foo': 'getFoo'}` and thinks what's that key for obj channel... and has to look for it

One additional consideration:
For requests and commands, it makes sense that shop model handles requests/commands on that channel. For events, it's less clear that the shop model cares about events throughout the app...

paulfalgout:
One particular use case where this feature would shine is when your channel name depends on the data.  for instance:
```js
defaultRadioChannel: function(){
  return 'user' + this.model.id;
},
radioCommands: {
  'doFoo': 'executeFoo'
},
executeFoo: function(){
  // do stuff you wouldn't do on the model directly
}
```

you can of course do this without but you'd have to:
```js
radioCommands: function(){
  var commands= {};
  commands[ 'user' + this.model.id + ' doFoo'] = 'executeFoo';
  return commands;
}
```

ianmstew:
I rather think that the most common pattern if for a single object to listen on a single channel, and that an object listening on multiple channels is the exception.  In my own code, I prefer doing:

```js
channelName: // channel name
channel: // channel instance
```

[...]

PR #2733, opened 2015-09-06
Simplify triggerMethod

paulfalgout:
Now due to how we use `_triggerMethod` internally `noEventArg` is always true (according to our test suite and as far as I can tell) So removing that condition simplifies things quite a bit.

Additionally we're checking to see if `trigger` exists.  I don't know of a use case in which `triggerMethod` should ever be called on something that doesn't also have `Backbone.Events` but considering we're are looking into merging this into `trigger` for `Marionette.Events` it seems like an unnecessary check on a fairly hot function.

Additionally, I haven't removed it, but I think returning the result of the onMethod is an anti-pattern and we should remove it.  Events are by nature one to many, and we have that situation here as well where behaviors might listen to the same event, but cannot return a value.
~~If we want to do something like this it should follow more of the request pattern.  Perhaps piggybacking off of https://github.com/marionettejs/backbone.marionette/issues/2515~~  Except since you already have the obj you're triggering on, you can just call the function.

### DIFF
--- a/src/abstract-view.js
+++ b/src/abstract-view.js
@@ -345,7 +345,7 @@ Marionette.AbstractView = Backbone.View.extend({
     var prefixedEventName = eventPrefix + ':' + eventName;
     var callArgs = [this].concat(args);
 
-    Marionette._triggerMethod(layoutView, prefixedEventName, callArgs);
+    Marionette._triggerMethod(layoutView, [prefixedEventName].concat(callArgs));
 
     // call the parent view's childEvents handler
     var childEvents = Marionette.getOption(layoutView, 'childEvents');
--- a/src/trigger-method.js
+++ b/src/trigger-method.js
@@ -13,12 +13,8 @@ Marionette._triggerMethod = (function() {
     return eventName.toUpperCase();
   }
 
-  return function(context, event, args) {
-    var noEventArg = arguments.length < 3;
-    if (noEventArg) {
-      args = event;
-      event = args[0];
-    }
+  return function(context, args) {
+    var event = args[0];
 
     // get the method name from the event name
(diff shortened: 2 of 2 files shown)


ianmstew:
I just did a sweep for `_triggerMethod` in the `next` sources and you're right, `noEventArg` is always true.  But I'm wondering why we even have `_triggerMethod` at all anymore?  This seems like needless indirection: https://github.com/marionettejs/backbone.marionette/blob/4695ae1f32e581440f789ab430a9e1ef246001d2/src/trigger-method.js#L50-L52.

paulfalgout:
Oh interesting.  I think eventually we will want the private,  where the public would be overridden for bubbling and/or private onMethods

ianmstew:
Ohhh, I see that now.  You're right.  Let's merge this @ahumphreys87 @jasonLaster :+1:

src/abstract-view.js
```diff
@@ -345,7 +345,7 @@ Marionette.AbstractView = Backbone.View.extend({
     var prefixedEventName = eventPrefix + ':' + eventName;
     var callArgs = [this].concat(args);
 
-    Marionette._triggerMethod(layoutView, prefixedEventName, callArgs);
+    Marionette._triggerMethod(layoutView, [prefixedEventName].concat(callArgs));
```
ianmstew: @megawac We added a concat back in here because Paul's win of converting `_triggerMethod()` to two arguments seemed worth it.  Plus we will be moving `_triggerEventOnParentLayout` function to `Marionette.Events` in the near future.

ianmstew:
Travis is rate limited; tests passed locally; merging.

PR #2735, opened 2015-09-07
Fix currentView might be undefined on swapOut #2055

### LINKED ISSUE #2055: currentView might be undefined on swapOut
https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.region.js#L181

Because `this.empty` is called above which `delete this.currentView`

paulfalgout:
Resolves #2055

Ignoring that this function is a beast, the fix was relatively easy.  
Maybe too easy?  Am I missing something?  

All tests pass.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -52,7 +52,8 @@ Marionette.Region = Marionette.Object.extend({
     var replaceElement  = !!showOptions.replaceElement;
 
     // We are only changing the view if there is a current view to change to begin with
-    var isChangingView = !!this.currentView;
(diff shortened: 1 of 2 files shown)


paulfalgout:
Next will be to rename `swap` to `swapIn` and to cherry-pick and update https://github.com/marionettejs/backbone.marionette/pull/1868

jasonLaster:
:thumbsup: looks like a clean fix. I think we should rename Mn to region.show. Nice work

PR #2736, opened 2015-09-07
View mixin proposal

paulfalgout:
~~This is mainly a POC.~~

I think we should get rid of `AbstractView` and make it a simple mixin.
- we're not calling the prototype except for on constructor and destroy.
- in both of those cases improvements may be made by moving them to their respective views
- it keeps things simple
- it keeps the inheritance shallow
- constructors and destroys similar functionality may be made into methods on the mixin, (think `_initBehaviors`)
- we can breakup the mixins into smaller mixins (utils, ui, events, etc)

What are the cons?


paulfalgout:
By the way.. there are two tests I couldn't fix right away.. some magic was happening somewhere setting the prototype that isn't working now.  less clever tests I think will pass.

jasonLaster:
I'm a big fan of this. In addition to the small mixins, we could make it easy for ppl to override small mixins to get their preferred behavior

jasonLaster:
I'd like to dry up the constructors before they land, but yea this is nice.

paulfalgout:
I totally agree about drying up both the constructors and the destroys to a point.

I'm thinking along with the Lifecycle Monitor the two constructors would look like this:
```js
Marionette.View = Backbone.View.extend({
  constructors: function(options) {
    _.bindAll(this, 'render');

    // should be able to handle _firstRender
    Marionette.MonitorLifecycle(this);    

    // this function will get much simpler
    this._initializeRegion();
    this._initializeBehaviors();

    Backbone.View.call(this, this.options);
    this.delegateEntityEvents();
  }
});

Marionette.CollectionView = Backbone.View.extend({
  constructor: function(options) {
    _.bindAll(this, 'render');

    // should be able to the this.on for _onBeforeShowCalled etc
    Marionette.MonitorLifecycle(this);    

    this.once('render', this._initialEvents);
    this._initChildViewStorage();
    this._initRenderBuffer();
    this._initializeBehaviors();

    Backbone.View.call(this, this.options);
    this.delegateEntityEvents();
  }
});
```

So there's definitely still some repetition, but it's very readable, I think much further abstraction would make the constructor harder to understand.

Slightly different conversation, but similarly I think destroy should stop calling `Backbone.View.remove`  It is making it difficult to accurate order the events that need to occur in `destroy` and we are creating an additional difficult to override dependency on jQuery with our `destroyImmediately` where we should be overriding and utilizing [_removeElement](http://backbonejs.org/docs/backbone.html#section-159) and then just calling `stopListening` as the very last call in the function.

paulfalgout:
@ahumphreys87 @ianmstew @jasonLaster 

I may have gone off the deep-end here a bit, but in trying to foresee what might be, it was easier to just start refactoring wholesale to see what an end result might look like.  But I don't want to muddy up the PRs here with a ridiculous PR, so I pulled it on my origin for review: https://github.com/paulfalgout/backbone.marionette/pull/3

Essentially it attempts to clean up `View` and make it solely responsible for building and managing regions by changing the RegionManager into a mixin.

ianmstew:
I LOVE the RegionManager mixin here https://github.com/paulfalgout/backbone.marionette/pull/3.  That's an extremely solid direction.  The ViewMixin is still seems too large, but you already mentioned you want to split it up further.  (For example, maybe a ViewEventsMixin, ViewUIMixin, ViewDataMixin, etc.)

ianmstew:
Then again, maybe ViewMixin being large isn't an issue, because everything there is core to all views.  Dividing it further doesn't improve logic the way RegionManagerMixin does, it's merely organizational... which probably isn't necessary.  I'm wondering if "ChildrenMixin" could apply to the CollectionView in the same way.

paulfalgout:
I think I'd like to attempt a few separate mixins.  It's be cool, but possibly unrealistic to do one for behaviors since it's pretty intermixed with other functionality.

But mainly for organization and readability if we can pull it off..   for instance `_buildViewTrigger` is only used once in the mixin by `configureTriggers` and then again in the behaviors, but those functions are over 100 lines apart and some of the functions in between are fairly unrelated.  It could be that we should just move stuff around a little, but a couple of separation points might make things easier to follow.

paulfalgout:
after talking through the `destroyImmediate` stuff on gitter I think I can fix this PR up so that it's cleaner and ready to go.  The tests that are failing would get quite a bit simpler and should pass.

jasonLaster:
Sounds good Paul. I like this direction.

jasonLaster:
@paulfalgout - looking forward to revisiting this now that ES6 has landed and imports will be that much clearer

src/view-mixin.js
```diff
@@ -138,6 +118,9 @@ var AbstractView = Backbone.View.extend({
   // Overriding Backbone.View's `delegateEvents` to handle
   // `events` and `triggers`
   delegateEvents: function(eventsArg) {
+    // proxy behavior $el to the view's $el.
+    _.invoke(this._behaviors, 'proxyViewProperties', this);
```
paulfalgout: Moved this from `setElement` as it removed the necessity for another protocol call, but will still occur in the same order.

src/view-mixin.js
```diff
@@ -208,26 +188,29 @@ var AbstractView = Backbone.View.extend({
...
+    this._isRendered = false;
 
     // unbind UI elements
     this.unbindUIElements();
 
-    this._isRendered = false;
-
     // remove the view from the DOM
-    this.remove();
+    https://github.com/jashkenas/backbone/blob/1.2.3/backbone.js#L1235
```
paulfalgout: destroy event was moved to the end of this call, which in general makes more sense, but this required breaking up `Backbone#remove` to move `stopListening` until after the trigger.  Although `_removeElement` is _'d it is documented as overrideable, and Bb isn't semver anyways, so this seems ok.

src/view-mixin.js
```diff
@@ -208,26 +188,29 @@ var AbstractView = Backbone.View.extend({
...
+    this._removeElement();
 
+    // remove children after the remove to prevent extra paints
+    this._removeChildren();
```
paulfalgout: removeChildren was made as a generic call that collection view could use to destroy its children and view could use to remove its regions.  This allowed for `destroy` to be shared where it is otherwise identical.  Alternatively I could see just setting up a listener for destroy inside of `_initRegions` and `_initBehaviors` which might simplify things a bit more.  Thoughts?

test/unit/collection-view.spec.js
```diff
@@ -1399,16 +1399,6 @@ describe('collection view', function() {
     });
   });
 
-  describe('has a valid inheritance chain back to Backbone.View', function() {
```
paulfalgout: I'm not sure what this was attempting to prove.. I think it was just a poorly maintained test.  `this.ChildView` is a `Backbone.View`

test/unit/trigger-method.spec.js
```diff
@@ -141,8 +141,8 @@ describe('trigger event and method name', function() {
       this.childView.$el.click();
     });
 
-    it('should fire the event method', function() {
-      expect(this.onChildviewFooClickStub).to.have.been.called;
+    it('should fire the event method once', function() {
+      expect(this.onChildviewFooClickStub).to.have.been.calledOnce;
```
paulfalgout: this test was modified in the ES6 module switch, but somewhere along the way whatever was causing the problem was fixed. :tada:

test/unit/view.child-views.spec.js
```diff
@@ -142,12 +142,6 @@ describe('layoutView', function() {
       this.layoutView.render();
     });
 
-    it('should supply the layoutView.options to the function when calling it', function() {
```
paulfalgout: destroyImmediate is now the default action

test/unit/view.spec.js
```diff
@@ -295,12 +295,16 @@ describe('item view', function() {
       expect(this.onBeforeDestroyStub.lastCall.returnValue.isDestroyed).not.to.be.ok;
     });
 
+    it('should be rendered when "onBeforeDestroy" is called', function() {
```
paulfalgout: we would expect the view to be rendered before destroy, but not after.

paulfalgout:
Ok @marionettejs/marionette-core  this needs :eyes:   I think the most controversial thing will be the `_removeElement`  It also causes some problems.  That function was introduced in Bb 1.2.0.  So we'd either need to up our dependencies or define our own `_removeElement`  I actually think either is completely reasonable and much better than having a fairly useless and confusing `onDestroy`

jasonLaster:
Looks good. Will review thoroughly

jasonLaster:
Looks like there's a merge conflict syntax error.

jasonLaster:
:eyes: let's get this merged soon

paulfalgout:
> Looks like there's a merge conflict syntax error.

I don't follow.. where?

megawac:
+1 for upping dependencies of lodash/underscore & backbone in v3

jasonLaster:
Merging. This is a big step forward. Any improvements can easily be made w follow up PRs

PR #2742, opened 2015-09-11
Clean up application

paulfalgout:
Marionette.Application is the only Object where the passed in options are extended directly on the Object.  This makes the currently documents [Application.mergeOptions](https://github.com/marionettejs/backbone.marionette/blob/master/docs/marionette.application.md#applicationmergeoptions) particularly useless and is a pattern that all other Marionette Objects have moved away from previously.
It should be removed and anything that needs to be attached should be done explicitly in the `initializer` with `mergeOptions`

Additionally there was quite a bit of clean up necessary on the docs.  This was a rough pass to remove any glaring issues.  I'm sure most of the docs could use a once over for vibe and clarity.

### DIFF
--- a/api/application.yaml
+++ b/api/application.yaml
@@ -3,327 +3,73 @@ name: Application
 class: Marionette.Application
 
 description: |
-  Applications are the entry point into most Marionette Applications. 
+  Applications are the entry point into most Marionette Applications.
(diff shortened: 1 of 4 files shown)


jasonLaster:
looks good :+1:

Issue #1866, opened 2014-09-11
Consider merging the events swapOut and swap (Marionette.Region)


paulovieira:
After reading the documentation (see #1859) and inspecting the code in the `show` method (`Marionette.Region`), I don't see the advantage of triggering 2 different events on the region for the act of swapping a view: `swap` and `swapOut`. 

What if we had only a single `swap` event?

That is:

`before:swap`

Triggered before the **new view** has replaced the **old view** within the Region. The new view is passed as the first argument. The old view is passed as the second argument. 

Note: at the moment this event is triggered, `region.currentView` is still the **old view**.

``` javascript
    if (isChangingView) {
        this.triggerMethod('before:swap', view, this.currentView);
    }
```

`swap`

Triggered when the **new view** has replaced the  **old view** within the Region. The new view is passed as the first argument. The old view is passed as the second argument. 

Note: at the moment this event is triggered, `region.currentView` is already the **new view**.

``` javascript

    this.attachHtml(view);
    var oldView = this.currentView;  // <- new code
    this.currentView = view;

    if (isChangingView) {
        this.triggerMethod('before:swap', view, oldView);
    }
```

This would make things more simple, no?

PR #2744, opened 2015-09-13
Rename region swap event to swapIn

paulfalgout:
Ref https://github.com/marionettejs/backbone.marionette/issues/1866#issuecomment-55342998


jasonLaster:
:thumbsup:

Issue #1872, opened 2014-09-12
swapOut triggered at the wrong time


jmeas:
It should be triggered after the swapping has occurred.

[Ref](https://github.com/marionettejs/backbone.marionette/issues/1866#issuecomment-55354093)

Potentially resolved by #1868

PR #2747, opened 2015-09-15
Correct the placement of the swapOut trigger to be in accordance with the documentation.

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/1872

The event order isn't being tested.  That should probably be a todo somewhere.


ianmstew:
:+1:

Issue #2279, opened 2015-02-05
Marionette Application needs a "root"


samccone:
With the deprecation of the application regionManager on the Application object, we need a way for people to "kick off" their view stack, scoped to an application.

```js

app = new Mn.Application({rootEl: "#my-app"})

app.onStart () =>
  app.root.show(new MyViewClass)
```

This is something that we can add to minor and allow people to start migrating away from a region manager today, it will also provide a better answer to people when they ask ... what do i do now since region managers were deprecated.

jmeas:
:+1:

Two things:

1. It should be called `rootElement` or `rootEl`.
2. It should be `body` by default.

jfairbank:
`rootEl`

jmeas:
Also, I like this best if the `rootElement` becomes something more like Ember's outlet, vs. a region.

In Marionette, there's no easy way to get access to `app.root` from within a route. One option is accessing the app directly. Another option is Radio. A third option is the way that Ember does it. Consider this URL (from within a Route)

```js
show() {
  this.showIn('root', new View());
}
```

`'root'` would use the [resolver](https://github.com/marionettejs/backbone.marionette/issues/2207) pattern to find the region in the DOM tree hierarchy. The Router would keep track of these nested regions.

samccone:
:+1: Priority bump just based on the number of questions that we have been getting regarding this

jasonLaster:
:dancers:

ahumphreys87:
fixed via #2763

PR #2757, opened 2015-09-20
Make Application extend View

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2279

Application in v3 does essentially [nothing](https://github.com/marionettejs/backbone.marionette/blob/next/src/application.js).

It is essentially [what people will do](https://github.com/samccone/todomvc/commit/e37df9935e7068d46316d449ba8d50925f37d20a#diff-4bf7cf8088b9cae0e4bfc8a6539148f7R8) anyhow.

While it pretty much brings back Application.Regions, it's not the same thing as adding a RegionManager to an Object.  But this might make the upgrade path easier for people still using them.

The failing test here is related to `destroyImmediate` being part of the `Marionette.View` `options` hash instead of on the view.  We can test around it, but I think it'd be better to make `destroyImmediate` the regular behavior.


jasonLaster:
Not sure bout this one. Let's talk about it on Tuesday.

jasonLaster:
closing this for the time being.

PR #2758, opened 2015-09-20
Add radio mixin

paulfalgout:
Proposal API for https://github.com/marionettejs/backbone.marionette/issues/2515 resolution

To use extend into object and call `_initRadio` in the constructor


jasonLaster:
I like it. What would we do w the other radio setup?

paulfalgout:
So making use of the `bindEntityEvents` utility we need a name for the other, but it would reduce this to essentially: 
```js
var RadioMixin = {
  _initRadio: function() {
    var channelName = Marionette._getValue(this.getOption('channelName'), this);

    if (!channelName) {
      return;
    }

    this._channel = Backbone.Radio.channel(channelName);

    // could be this.bindEntityEvents using the proxyBindEntityEvents
    Marionette.bindEntityEvents(this, this._channel, this.getOption('radioEvents'));
    
    Marionette.bindEntityRequests(this, this._channel, this.getOption('radioEvents'));

    this.on('destroy', this._destroyRadio);
  },

  // destroy anything on this channel that this object is replying to
  _destroyRadio: function() {
    this._channel.stopReplying(null, null, this);
  },

  getChannel: function() {
    return this._channel;
  }
};
```

Thoughts on `bindEntityRequests`?  Also we'll probably want to modify bind-entity-events.js to reuse most of the logic of figuring out the strings / functions for requests.

So would we want `bindEntityRequests` `unbindEntityRequests` `proxyBindEntityRequests` and `proxyUnbindEntityRequests` to match Events?

jasonLaster:
got it, i needed to see this line to understand what you're doing
```js
   Marionette.bindEntityEvents(this, this._channel, this.getOption('radioEvents'));
```

I could see the public API being, `bindRadioEvents` `bindRadioRequests` so what if `bindRadioEvents` and `bindEntityEvents` is like the same thing - radio is a special entity.

paulfalgout:
true.  suppose it could check to ensure it's a channel

paulfalgout:
Related issue.  would we do a `proxyBindRadioEvents`?  like: https://github.com/marionettejs/backbone.marionette/blob/master/src/bind-entity-events.js#L98

jasonLaster:
ahh, that's where the proxyProxyPeroxyPeroxide business came from. I suppose so.

paulfalgout:
:-) lol

jasonLaster:
@paulfalgout does this API become cleaner w/ ES6 modules?

Also is this ready to land crew?

paulfalgout:
I still have to make a couple revisions I think..  however pulling out the other radio stuff would certainly relieve some of the breaking tests on ES6 :-P   It might get easier with ES6.  I've kind of been leaving it until last cause it's easier.

jasonLaster:
Interesting idea. Wouldn't be opposed to doing that.

paulfalgout:
Clearly this is lacking any kind of test or documentation.  But otherwise I think it is very close.

jasonLaster:
Agreed.

paulfalgout:
This is still lacking the tests which I'll get to ASAP, but it has been updated to match our new patterns..

And @jasonLaster I changed one related thing which I might need to revert..  `bindEntityEvents` and so `bindRadioEvents` were taking an POJO or a function that returns a POJO via a `getValue` within the bind functions.
I actually removed this and now prior to calling any of these bind functions there's a now very common `this.getValue(this.getOption('modelEvents'));` etc.  I think maybe we _should_ reintroduce that `getOptionValue` you were working on..  There is only a couple of times we use `getValue` where it is not this pattern exactly.

I think doing this makes the patterns simpler and less complicated to learn, test, and document.

Thoughts?

jasonLaster:
Yea. Why not. Let's do it

paulfalgout:
@marionettejs/marionette-core this has tests now and should be ready.

ahumphreys87:
Lgtm :+1: I say lets get this merged asap

jasonLaster:
Great work @paulfalgout!

PR #2759, opened 2015-09-20
deprecate CompositeView

ahumphreys87:


jasonLaster:
So much salmon. 

Let's review and see if there are tests or docs we should bring over. 

NOTE: Community, we're not arbitrarily removing a swath of functionality. With @ahumphreys87 prior work, it's possible to do everything from CompositeView could do from View.

paulfalgout:
So it seems if we're ready for this, we should theoretically be able to build the CompositeView API out of a View and CollectionView that would pass all of the tests?  I suspect some of the tests are too tied to the inner-workings and would need some adjustments, but I think it makes sense to attempt to do this.

jasonLaster:
@ianmstew brought up a point in channel - we should probably deprecate and feature flag CompositeView. That way, users can upgrade to 3 and then begin converting from composite view to the base view. 

I'd argue that we should have dynamic deprecation notices in the code so that in DEV users see a warning if they're using CompositeView, although setting up deprecation notices is a separate ticket.

ianmstew:
@paulfalgout If we do a deprecate plan alongside a well written upgrade guide, I don't think we'll have to keep the entire CompositeView API in v4, but you make a good point and there are probably some spec ports to CollectionView that we could make.  @jasonLaster I like the DEV deprecation notice idea.  What's the `isEnabled()` check about?

jasonLaster:
@ianmstew it's the [feature flag]( https://github.com/marionettejs/backbone.marionette/blob/next/src/features.js#L1-L7). I don't know if it's needed, but the idea is that once it's on then the class is not available.

ianmstew:
Ahh, got it.  Thanks.

paulfalgout:
I do wonder if when we remove the `CompositeView` that we could make an implementation of it that passed it's current tests in a separate repo or something.  Just something that proves that says you can still do all of the things CompositeView was providing.

ianmstew:
That would be excellent, I didn't mean to discourage that, even in the core spec library.  I'm really only wondering about CompositeView-specific APIs that will be no longer be necessary:  `getChildViewContainer()`, `resetChildViewContainer()`, `renderChildren()`, `destroyChildren()`.

JSteunou:
Along side a Module polyfill, a CompositeView polyfill made of actual View + CollectionView would be great!

I'm afraid to not be able to go on the v3 train without it. Having a lot  of `div > ul > li` with `ul` as `childViewContainer` in a CompositeView in order to avoid another useless `div`.

jasonLaster:
@JSteunou - I hear you. We'll probably end up deprecating in 3, which will give people time to upgrade to our newly improved Region class which avoids useless `div`s

ahumphreys87:
@marionettejs/marionette-core ok so I've put CompositeView back and just deprecated for 3.0.0.

Once we have a pre-release for 3.0.0 we should create a guide for implementing a `CompositeView` with `View` and `regions`

ahumphreys87:
@JSteunou for reference - using a `replaceElement` option when showing a view in a region will provide the same functionality as `childViewContainer`. This means you can create a `CompositeView` using `LayoutView` (`View` in v3) and `CollectionView`

https://github.com/marionettejs/backbone.marionette/commit/83aea13b23574c5b5b06a1c500cc1e6d88ec42bb

paulfalgout:
:+1:

JSteunou:
This is very nice, great work.

AndrewHenderson:
> . . . it's possible to do everything from CompositeView could do from View.

@ahumphreys87 @jasonLaster How would you recreate a nested folder structure using a `LayoutView` and `CollectionView`?

My understanding of `CompositeView` is that it was meant to recursively use a View class on the view's children to a depth of n.

I suppose I'm not clear on the reasons for deprecation. It seems most devs miss the intended use for `CompositeView` by using it for the simple case where a view needs both a `model` and a `collection`. The title of this issue (https://github.com/marionettejs/backbone.marionette/issues/2800) is evidence of that.

ahumphreys87:
@AndrewHenderson so in 3.0.0 `ItemView` and `LayoutView` have been merged. So the children of the  `CollectionView` will also have regions that you can nest views into

ahumphreys87:
The reasons for deprecation are to simplify the views we have. In 3.0.0 you only need 2 views: `View` (LayoutView and ItemView merged) and `CollectionView`

paulfalgout:
This?
```js
var TreeView = Marionette.CollectionView.extend({
  buildChildView: function(child, ChildViewClass, childViewOptions){
    if(child.nodes.length){
      return new TreeView(_.extend({collection: child.nodes}, childViewOptions));
    }
    return new ChildViewClass(_.extend({model: child}, childViewOptions));
  },
  childView: MyLeafView
});
```

jmeas:
> @ahumphreys87 @jasonLaster How would you recreate a nested folder structure using a LayoutView and CollectionView?

This should be as simple as setting the `childView` of the `CollectionView` to be the `CollectionView` itself. Or, if you want the template, then it should be set to the `ItemView` that then renders the CollectionView.

> My understanding of CompositeView is that it was meant to recursively use a View class on the view's children to a depth of n.

This is one use case for it, but very few people use it like this, we found. Our research showed that most people use it to get around the fact that Regions create a wrapper Node. Because Regions in v3 won't require a wrapper node, CompositeViews main benefit no longer requires a separate concept for it.

> I suppose I'm not clear on the reasons for deprecation. 

Marionette is absolutely enormous relative to the features that it provides. It's enormous in three ways:

1. The number of individual "concepts" one must learn to use the framework
2. The size of the library in bytes
3. The surface area of the API

This is an effort to reduce all three of those things, while retaining 100% of the existing functionality. And, actually, adding more functionality by making Regions more powerful.

also hey @AndrewHenderson :wave: :stuck_out_tongue:

AndrewHenderson:
Given Marionette moving to a subset of the current Views, is there a recommended use case so that upgrading is much easier. For instance, should we avoid use of CompositeView and instead use a combo of X. Should we avoid CollectionView?

paulfalgout:
So hopefully we can make an upgrade tool to make it a pretty seamless transition.  You could start using only layoutviews instead of itemviews, but there _might_ be a minor perf hit doing that in v2 that won't be in v3.. With CompositeView it is likely best to avoid them, but we don't yet have all of the recipes for how.  If we can find a use-case that can't be replaced with the v3 view's we'll probably need to fix the views to accommodate.  That said, CompositeView will be around for a little bit.  We're just going to deprecate them until there is a *super* clear upgrade path.

Issue #2326, opened 2015-02-13
Introduce a "rootEl" property on the application object


samccone:
With the deprecation of the application regionManager on the Application object, we need a way for people to "kick off" their view stack, scoped to an application.

I propose a solution of allowing a users to specify a `roolEl` property on their application. 

```js
var myApp = new Application({
  rootEl: "#my-app"
}); 

myApp.on("start", () =>
   this.rootEl.show(new FooLayout)
)
```

This is something that we can add to minor and allow people to start migrating away from a region manager today, it will also provide a better answer to people when they ask ... what do i do now since region managers were deprecated.

jmeas:
Closing as a duplicate of #2279

samccone:
derp I thought I had written this before

jmeas:
:stuck_out_tongue:

PR #2763, opened 2015-09-22
Introduce a "rootEl" property on the application object

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2279

Also https://github.com/marionettejs/backbone.marionette/issues/2326

Alternate to https://github.com/marionettejs/backbone.marionette/pull/2757

I think we want the rootEl to be a `Marionette.Region`.. otherwise we could be duplicating much of the region related code relating to a view's lifecycle and introducing new areas for regression.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -1,19 +1,58 @@
 // Application
 // -----------
-
-import MNObject from './object';
+import _                from 'underscore';
+import MarionetteObject from './object';
+import Region           from './region';
 
 // A container for a Marionette application.
-var Application = MNObject.extend({
+var Application = MarionetteObject.extend({
   cidPrefix: 'mna',
 
+  constructor: function(options) {
+    options = options || {};
+
+    this._initRegion(options);
+
+    MarionetteObject.prototype.constructor.apply(this, arguments);
+  },
+
+  regionClass: Region,
+
+  _initRegion: function(options) {
+    var region = options.region || this.region;
+    var RegionClass = options.regionClass || this.regionClass;
+
+    // if the region is a string expect an el or selector
+    // and instantiate a region
+    if (_.isString(region)) {
+      this._region = new RegionClass({
+        el: region
+      });
+      return;
+    }
+
+    this._region = region;
+  },
+
+  getRegion: function() {
+    return this._region;
+  },
+
+  showView: function(view, options) {
+    var region = this.getRegion();
+    return region.show.apply(region, arguments);
+  },
+
+  getView: function() {
+    return this.getRegion().currentView;
+  },
+
   // kick off all of the application's processes.
-  // initializes all of the regions that have been added
-  // to the app, and runs all of the initializer functions
   start: function(options) {
     this.triggerMethod('before:start', options);
     this.triggerMethod('start', options);
   }
+
 });
 
 export default Application;


src/application.js
```diff
@@ -5,9 +5,33 @@
...
+  constructor: function(options) {
+    options = options || {};
+
+    this._initRootRegion(options);
+
+    Marionette.Object.prototype.constructor.apply(this, arguments);
+  },
+
+  _initRootRegion: function(options) {
+    var rootEl = options.rootEl || this.rootEl;
```
jasonLaster:
```js
var rootEl = this.getOption('rootEl');
```

src/application.js
```diff
@@ -5,9 +5,33 @@
...
+    rootEl = Marionette._getValue(rootEl, this, options);
+
+    // rootEl is optional
+    if (rootEl) {
+
+      var RootRegion = Marionette.Region.extend({
+        el: rootEl
+      });
+
+      this.rootEl = new RootRegion();
```
jasonLaster: maybe `this._rootEl` so that we maintain separation.

src/application.js
paulfalgout: `getOption` isn't available yet :-(  not until after the Object constructor

src/application.js
paulfalgout: Yeah I was following this:
```
var myApp = new Application({
  rootEl: "#my-app"
}); 

myApp.on("start", () =>
   this.rootEl.show(new FooLayout)
)
```
I agree with you about the separation, but what we set here will be what you call show on.. so I suppose two options..
we could store the original `rootEl` selector string on `this._rootEl` kind of like uiBindings..  or we could make the region name different than the selector, but I'd suggest something without the leading _?  maybe just root? so:  `this.root.show(new FooLayout)`

jasonLaster:
Nice work. This improves the getting started story quite a bit, while keeping the separation between architecture and view tree.

src/application.js
paulfalgout: ninja edit.  Yeah I like that.  `showRootView` abstracts the region even more here.  :+1: :+1:   change coming soon

paulfalgout: one other thought.. there's not any region events we'll want to proxy here?  it kinda doesn't seem like it.  But we could add a `before:show:rootView` and `show:rootView`  any value?

RobertCZ:
Why not call it just 'region'? Name 'rootEl' doesnt work very well when you want to pass previously created region. With 'region' name, it can still be initialized as el, selector or region, but will give us a much nicer pattern for child apps - basically every app would have a region to use sent from outside (eg parent app or bootstrap):

    // main
    let rootApp = new RootApp({region: document.body}) // can be default value

    // RootApp#ctor
    this.view = new View()
    this.region.show(this.view)

    // RootApp#startChildApp
    let region = this.view.getRegion('child')
    let childApp = new ChildApp({region})
    childApp.start()

Also, regions' show and maybe getRegion(s) can be aliased directly on app (from this.region and this.region.currentView)

    this.show(new View())
    let region = this.getRegion('child')
    (new ChildApp({region})).start()

JSteunou:
Hooooo this is nice @paulfalgout 

Definitely a must have. 2nd top personal need after subapp :p

paulfalgout:
@marionettejs/marionette-core trying out @RobertCZ's `region` idea.  As implied this removes the expectation that `Application` is the very top item... so I went with it.  I added an `isRunning` and a `stop`  This might be the least opinionated way to replace Modules and provide a bare bones solution for subapps.

Thoughts?  The more I sit with it, the more I like it.

JSteunou:
I do not like it. The same way regions was removed from application to avoid mixing responsibility, this change add it again but the other way. Something related to view / region mixing with app / subapp. Name that region "application" and it will mirror the old application object.

Having a rootEl / rootRegion / rootLayout is something that is needed but not as a module or subapp.

RobertCZ:
I'd still prefer region as a param name, but I'm hesitating now what should the default be. While for main app automatically created region on document.body makes sense, some child apps might be regionless (eg services) and that would require passing explicitly {region: false} or some such. Not sure which one is more elegant. Perhaps passing false is more in the spirit of template: false for views so it'd be less confusing for beginners:

    // (1) default is no region
    let rootApp = new RootApp({region: document.body})
    let serviceApp = new ServiceApp()

    // (2) default is body
    let rootApp = new RootApp()
    let serviceApp = new ServiceApp({region: false})

src/application.js
```diff
@@ -5,11 +5,53 @@
...
+    return this._region;
+  },
+
+  showView: function(view) {
+    this.getRegion().show(view);
+  },
+
+  isRunning: function() {
+    return !!this._isRunning;
+  },
```
jmeas: Why have this? We should get rid of this as well as `start` and `stop`, unless we want to hide the router configuration behind `start`, like...

```js
MyApp.start({
  html5: true
});
```

paulfalgout:
@RobertCZ I don't think apps should have a default region if they're reusable.  And they don't really need to be required which alleviates the need for a `region: false`

The start/stop stuff I feel was mainly useful in regards to a stop-gap for those using modules as lifecycle items.

I am now of the opinion that we don't need to add `stop` or a running state and can suggest if someone needs a module type start/stop to either build it out of an Object (insert recipe), use another library, or extend application and add a stop.  And we just keep Application as very bare bone.

So other than an app region, the opinion this would be adding that we don't currently have, is that if you want, multiple Applications is ok.  I think that's a pretty reasonable opinion.

I don't feel super gung-ho about any of this.. I just don't want to get overly opinionated.. But I want to provide a starting place.  That said, I'm still ok with #2757 as well.

But right now my todo list on this looks like:
- [ ] Remove stop and running states
- [ ] Make sure showView / getView are similar API to `showChildView`
- [ ] Tests
- [ ] Some form of documentation (Although I still think docs are an issue all their own ATM)

RobertCZ:
If I understand correctly: without default region (agreed) and without stop/running (agreed) we'll have just something like:

    class Application {
        constructor(options = {}) {
            this.region = options.region
        }
        start() {
            // trigger 2x
        }
    }

This almost looks like a candidate for recipe in a guide not something that should be in Mn. Maybe it's time to go back to the Marionette view-managing roots and deprecate Application object completely. Controller and Module are already deprecated, right? Would be cool if Mn were down to just View and Region :-)

paulfalgout:
So this is updated to be the least opinionated solution that involves having a single region on an application.  It leaves `start` where it was.  The particular merits of `Application` are certainly still in consideration for the future, but in lieu of [replacing it entirely with just a plain view](https://github.com/marionettejs/backbone.marionette/pull/2757), this seems to be the smallest compromise that still provides some sort of clear starting point.

JSteunou:
I'm still not confortable with Application getting `showView` `getView` but this is maybe the simplest & accessible solution.

RobertCZ:
I like this version but I agree show/getView sugar is not strictly necessary

ahumphreys87:
:+1: @marionettejs/marionette-core lets get this in - we can iterate in `next`

jmeas:
Just cross-referencing [this comment](https://github.com/marionettejs/backbone.marionette/issues/2279#issuecomment-72970656), which describes how this simple change could be super useful with an updated router.

paulfalgout:
The show/getView is essentially consistency.

The resolver pattern is super interesting.

JSteunou:
resolver pattern + app / sub app + multi router = :fireworks:

PR #2765, opened 2015-09-23
Fix issue with serializeCollection

paulfalgout:
Calling clone on each model in the collection is actually re-initializing the models.  Fine I suppose in many cases, but if you were doing something expensive in initialize, or using a library like `backbone.uniquemodel` suddenly you're creating new models in memory you don't need.  And that could lead to all sorts of problems.

`Collection.map` and `_.clone` are going to be cheaper and it's what [backbone is doing](https://github.com/jashkenas/backbone/blob/1.2.3/backbone.js#L791)

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -51,7 +51,7 @@ Marionette.View = Marionette.AbstractView.extend({
   // its model's attributes
   serializeCollection: function() {
     if (!this.collection) { return {}; }
-    return _.pluck(this.collection.invoke('clone'), 'attributes');
+    return this.collection.map(function(model) { return _.clone(model.attributes); });
   },
 
   // Render the view, defaulting to underscore.js templates.


jasonLaster:
:+1: 

How'd you find this bug?

TerrenceLJones:
:+1:

paulfalgout:
we're backporting this change as we needed to over-write `toJSON` for the server.
It worked great except that we just recently setup a particular model that listens to a radio channel on initialize.  We passed that collection to an itemview in one instance.. and rendering that view suddenly caused duplicate events to occur.  Biiiiggg rabbit trail.

JSteunou:
Woo nice catch @paulfalgout! Will this fix be backported to a 2.4.4?

paulfalgout:
no.  `serializeCollection` works quite differently in `2.x` backporting this would be breaking.

JSteunou:
Ok, I though it was the same.

ahumphreys87:
good catch @paulfalgout :+1:

ahumphreys87:
manually merged into `next`

Issue #2296, opened 2015-02-05
Marionette should be built w/ babel (6to5) boilerplate principles


jasonLaster:
Ember and Lodash are built w/ these principles. We have two experts on the matter here @jmeas, @thejameskyle. There would be several benefits:
+ module-based build strategy (think imports on the top, export on the bottom)
+ module resolution `import CollectionView from 'backbone.marionette/source/collection-view'`
+ es6 features

There would be one significant downside:
+ we would be es5 compliant, which means that those who want to support es3 for ie8 need a shim

I'll leave it to the pros to outline an implementation strategy:
+ refactor
+ build tools
+ package management

What do y'all think?

jmeas:
https://github.com/marionettejs/backbone.marionette/issues/2250

samccone:
@jasonLaster I don't want to drop ie8 support because we use ES6. If we are going to use things that need polyfills then we need to bake them into the core lib.

jasonLaster:
Reasonable

samccone:
that said, we can use es6, @thejameskyle alluded to the fact that we can simply avoid certain es6 things that need polyfills

paulfalgout:
I am unfortunately tied to IE8 for quite a while   :skull:

jmeas:
I think that it's fine if we support IE9 out of the box, with IE8 support if you include an ES5 shim. It's 2015, and more and more people make applications that don't need to support legacy browsers.

jasonLaster:
Reopening Bc this is where the convo is. 

I don't think ie8 should be a blocker. I'm tied to ie8 :poop:s.com is optimized for it.

samccone:
Pretty much every project we do here needs ie8 support, and marionette just working in IE8 is a really great thing. 

That said, I agree @jmeas it should not hold us back, we just need to make sure that we give people the polyfill that they need.

instanceofpro:
We can use autopolyfiller https://github.com/azproduction/autopolyfiller. It scans code and applies only required polyfills.

jmeas:
> marionette just working in IE8 is a really great thing.

Yup. This won't change.

> That said, I agree @jmeas it should not hold us back, we just need to make sure that we give people the polyfill that they need.

It isn't our job to supply the polyfill – [it already exists](https://github.com/es-shims/es5-shim).

[...]

PR #2769, opened 2015-09-24
Add ES6 modules

jasonLaster:
This PR adds support for ES6 Modules. It's a huge team effort @trusktr and @cmisenas and @marionettejs/marionette-core have all been involved.


jasonLaster:
just updated it to support feature flags

Running into two problems:
1. tests have lots of undefined functions (requires some checking)
2. The specRunner can't just be opened (marionette probably has to be bundled)

src/abstract-view.js
```diff
@@ -6,8 +6,23 @@
...
+import Backbone                 from 'backbone';
+import _                        from 'underscore';
+import Behaviors                from './behaviors';
+import _getValue                from './utils/_getValue';
+import getOption                from './utils/getOption';
+import normalizeMethods         from './utils/normalizeMethods';
+import normalizeUIKeys          from './utils/normalizeUIKeys';
+import normalizeUIValues        from './utils/normalizeUIValues';
+import mergeOptions             from './utils/mergeOptions';
+import proxyGetOption           from './utils/proxyGetOption';
```
davesnx: Why not

```javasctipt
import {
    _getValue,
    getOption,
    normalizeMethods,
    normalizeUIKeys,
    normalizeUIValues,
    mergeOptions,
    proxyGetOption
} from  './utils
```

blikblum:
Wondering if there's an impact in built script file size. Smaller? Bigger?

src/abstract-view.js
jasonLaster: nice idea.

package.json
```diff
@@ -63,7 +63,6 @@
     "gulp-sourcemaps": "^1.6.0",
     "gulp-uglify": "^1.4.2",
     "gulp-unwrapper": "^0.1.0",
-    "gulp-util": "^3.0.7",
```
jasonLaster: oops, that should be there

jasonLaster:
Hey @marionettejs/marionette-core! Just rebased off of next to get all of the new work @paulfalgout did!

We're now at 
```
  1049 passing (43s)
  33 failing
```

I think if we remove Metal it could go to 0. Also, if somebody were too do a pass, we could really clean up some of this stuff. Some todos

+ get all tests passing
+ investigate why the Error tests had to change
+ remove backbone-marionette `assign`
+ change Marionette.MarionetteError to Marionette.Error

src/backbone-marionette.js
```diff
@@ -0,0 +1,111 @@
...
+  triggerMethodOn,
+  triggerMethodMany,
+  MarionetteError,
+  FEATURES,
+  isEnabled,
+  setEnabled,
+  Application,
+  AppRouter,
+  MonitorDOMRefresh,
+  MNObject,
```
thejameskyle: MNObject?

cmisenas: @jasonLaster we forgot to remove this, in favor of [line 101](https://github.com/marionettejs/backbone.marionette/pull/2769/files#diff-505d5912b9c98bf87e95e162f9390b4fR101).

jasonLaster:
just pushed a version that comments out all of the failing tests (42) pending now

Couple of things that are happening
1. jQuery is not in scope, but $ is...
2. Router is not referencing ref, is this a BB version thing?
3. Object is not emitting an event, is that something to do w/ Radio versions...

paulfalgout:
https://github.com/jasonLaster/backbone.marionette/pull/3

jasonLaster:
@paulfalgout, just merged your awesome build work for browser tests and fixes to all of the `xit` tests. 

I also did a rebase from next, so everything is uptodate. This includes all of the awesome work to kill region-manager and other goodies.

@marionettejs/marionette-core this has been a project since August that has involved @trusktr, @cmisenas, myself and @paulfalgout. It's the most important thing we're working on.

### Progress: 13 failing tests out of 1033 (98.7% done)

Here are the failing tests to motivate you to go in and fix them. If you want to work on it, I recommend checking out the branch, running `gulp test-browser`, and then filing a PR on my branch. There are two gotchas now (you need to re-run `npm i`, and you need to be on node 4 or 5). No biggie.

errors 1

errors 2


----

### It's time

trusktr:
@blikblum 

> Wondering if there's an impact in built script file size. Smaller? Bigger?

This actually helps file size. In your project, import whatever you need using ES6 modules. Then (thanks the the ability for ES6 modules to be statically analyzed by build tools like Webpack+Babel, Browserify+Babel, etc) your resulting code bundle will only have the things you use compiled into it.

If you're not using a module build system and just adding Marionette to your page via `<script>` tag as a to make it available as a global variable, then the code will be the same size, and it will include everything, and you won't have the benefit of the compiler not compiling unused code.

src/abstract-view.js
jasonLaster: Hmm, looks like that's not yet possible with a directory. We could add an index.js file in the folder, but i don't think that would help in the long run

jasonLaster:
@trusktr are you saying if somebody is building an app like this:

```js
import Marionette from 'marionette'
import Backbone from 'backbone'
import _ from 'underscore'

// do crazy stuff
```

Their bundle will only have one Backbone and one Underscore? (Just trying to clarify for everyone)

jasonLaster:
@cmisenas @paulfalgout just fixed the tests. Fine, i confess I also deleted one test and modified another (it's probably fine). Will explain in comments. Big picture, we're green! Lets critique the code and merge when we're ready. The long march is coming to an end.

test/unit/trigger-method.spec.js
```diff
@@ -186,12 +184,6 @@ describe('trigger event and method name', function() {
       it('should call a method named on{Event}', function() {
         expect(this.methodHandler).to.have.been.calledOnce;
       });
-
-      it('should return the value returned by the on{Event} method', function() {
```
jasonLaster: I deleted this test because it required spying on `Marionette.triggerMethod`. In the new world of modules, we'd have to split triggerMethod and triggerMethodOn's modules and then stub on the `import` level which is possible but more overhead. Not an essential test and I went w/ killing it

test/unit/trigger-method.spec.js
```diff
@@ -141,8 +141,8 @@ describe('trigger event and method name', function() {
       this.childView.$el.click();
     });
 
-    it('should fire the event method once', function() {
-      expect(this.onChildviewFooClickStub).to.have.been.calledOnce;
+    it('should fire the event method', function() {
+      expect(this.onChildviewFooClickStub).to.have.been.called;
```
jasonLaster: This is truly bazaar, for some reason the triggerMethod is called twice for what looks like a Mocha reason, but I have no idea why. Should be safe as it's the only instance of this

src/application.js
```diff
@@ -1,8 +1,10 @@
 // Application
 // -----------
 
+import MNObject from './object';
```
jasonLaster: This should be `MarionetteObject` not a big deal though

src/backbone.marionette.js
```diff
@@ -1,6 +1,100 @@
...
+import {
+  triggerMethod,
+  triggerMethodOn,
+  triggerMethodMany
+} from './trigger-method';
+
+Marionette = Marionette || {};
+
+// extend Marionette in the meantime
+Object.assign(Marionette, {
```
jasonLaster: We shouldn't use `Object.assign` for IE reasons i believe

src/behavior.js
```diff
@@ -6,7 +6,10 @@
 // Behaviors allow you to blackbox View specific interactions
 // into portable logical chunks, keeping your views simple and your code DRY.
 
-Marionette.Behavior = Marionette.Object.extend({
+import _        from 'underscore';
+import MNObject from './object';
```
jasonLaster: should be `MarionetteObject`

test/unit/error.spec.js
```diff
@@ -1,11 +1,11 @@
 describe('Marionette.Error', function() {
   it('should be subclass of native Error', function() {
-    expect(new Marionette.Error()).to.be.instanceOf(Error);
+    expect(new Marionette.MarionetteError()).to.be.instanceOf(Error);
```
jasonLaster: We should export `Marionette.Error`

test/unit/region.build-region.spec.js
```diff
@@ -258,10 +258,7 @@ describe('Region', function() {
       });
 
       it('throws an error', function() {
-        expect(this.buildRegion).to.throw(Marionette.Error, new Marionette.Error({
-          message: 'Improper region configuration type.',
-          url: 'marionette.region.html#region-configuration-types'
```
jasonLaster: these errors are strange, probably a side-effect of whether we're using Metal or Marionette Error and how it's exported. the two things to look for are are URL and version, which were issues before but should be resolved

jasonLaster:
Did a pass and left lots of comments for my later self.

test/unit/trigger-method.spec.js
paulfalgout: I think we should prove this isn't an issue.  log a trace or something.. I do think there are child event issues being introduced by combining itemview and layoutview, but I am surprised we hadn't seen a failing test.

trusktr:
@jasonLaster 

> @trusktr are you saying if somebody is building an app like this:
> 
> ```
> import Marionette from 'marionette'
> import Backbone from 'backbone'
> import _ from 'underscore'
> ```
> 
> // do crazy stuff
> Their bundle will only have one Backbone and one Underscore? (Just trying to clarify for everyone)

That correct, we can `import Backbone from 'backbone'` in as many files as we want, and a single reference will be shared to all places where Backbone is imported. But, the thing to consider when importing a whole library like that is that most build tools right now don't do [tree shaking](http://blog.sethladd.com/2013/01/minification-is-not-enough-you-need.html) out-of-the-box, so, the `Backbone` reference contains the entire Backbone library, and same with `_` and `Marionette`.

What I was talking about was a little different. In order to prevent from loading an entire library (so that a compiled bundle can be as small as possible), we can easily import just the things we need directly, assuming a library is structured properly (`underscore` isn't, and I'll explain why).

So, for example, `lodash` *is* structure properly, and we can do this in our project files:

```js
import debounce from 'lodash/function/debounce'

let debouncedFunction = debounce(function() { /* ... */ })
```

in order to import only the `debounce` function from `lodash`. When we compile an app bundle, only the `debounce` function will be compiled into the bundle's source, and so our clients will have a smaller file to download. 

If you install `underscore` locally, from NPM, you'll notice inside `node_modules/underscore` that all of the code lives in a single file, so the only option we have with `underscore` is to import the whole library without the possibility to import specific things. The good news is, if we use ES6 (ES2015) modules which are statically analyzable, then build tools can have an easier time tree shaking, so doing `import _ from 'underscore'` won't necessarily mean we'll get all of underscore built into our app bundle.

To be safe, the best practice would be to structure the library the way it is already being done in this PR, keeping in mind to keep import dependencies as lean as possible, and then for users to import only what they need instead of the whole library.

In the near past, before module bundlers and compile tools (namely, before Node.js proliferated) it was common to use script tags, so f.e.

```html
<script src="/path/to/marionette.js"></script>
```

Doing things that way would obviously load the whole library, and be less efficient, so I was just pointing out that with ES6 we could be more efficient (by importing only what is specifically needed).

Looks like Webpack [now has](https://github.com/webpack/webpack/tree/master/examples/harmony-unused) some [desired tree-shaking functionality](https://github.com/webpack/webpack/issues/1433).

We can't predict if the user's build system will have tree-shaking though, so best practice is to do what is already being done here, and hope the app developer imports specific things. Perhaps the examples in the documentation can be done that way.

trusktr:
Note: lodash is peculiar. They have a build step to convert from their source format to the properly structured modular format seen in the NPM module.

paulfalgout:
Yep related https://github.com/rollup/rollup does static analysis bundling, but that guy thinks lodash _shouldn't_ be built into the `nested/module/format`  And when static analysis or somehow doing that over HTTP2 is a thing, that'll probably be very true.

src/backbone.marionette.js
trusktr: @jasonLaster If we're compiling ES6 modules (which aren't natively supported in any environment yet) we may as well compile Object.assign and any other `>ES5` features in the process. The global version of the library (a bundle compiled from an index file that imports everything, generating a `marionette.js` file that can be loaded via `<script>` tag) can contain the compiled code, so Object.assign will work anywhere. Some ES6+ features compile into nice clean ES5 code, while some others compile into complex, less-performant ES5 code (generator functions compiled for environments without native generator support incur a performance cost, for example (see [Facebook's regenerator](https://github.com/facebook/regenerator))). So, it would be a matter of choosing wisely what features to use and to compile at the same time as the ES6 modules for the global build.

For browserify users, the good news is that configuring all browserify transforms needed for the user's browserify project is in our hands, and the user need not worry about it.

In the case of Webpack, the app developer (not us) has to configure Webpack. The best we can do is clearly document which Webpack configurations are needed when and if the end user of the Marionette library wants to import ES6 code directly.

A solution to battle these things is to compile ES6 modules into a CommonJS format on `prepublish` so that the shipped library is consumable in all common build tools and environments, with all non ES5 things already pre-compiled.

To see an example of that, check this out (though the project is not complete): https://github.com/infamous/motor/blob/joe/package.json#L14

See this thread on details of how this package.json configuration is widely consumable in many build systems: http://forums.infamous.io/t/api-build-system-goals/46

There are still some things to do, which I've put off until the library itself becomes more mature (it's not at all ready for consumption yet): include the babel-runtime so that Babel's helper functions aren't duplicated in every built file, but are instead imported from `babel-runtime` as needed. Babel will place the proper require calls into the resulting CommonJS modules during compilation.

trusktr:
@paulfalgout I like the concept, but this depends on using rollup or another build tool that has the same ability. If we write all code in the `nested/module/format` from the get go, we'll have the best coverage of bundles-that-are-as-small-as-possible across all the common build tools of today. I think `nested/module/format` works over HTTP/2, as seen with [JSPM](http://jspm.io). Nested or not, all dependencies can still be sent a one through HTTP/2, with the logic of what gets sent (tree-shaken or not) on the server side. If we add tree-shaking to the mix *on top of* the nested/module/format, I think that would result in the most possible good outcomes for reduced file size across all build tools.

jasonLaster:
Alright, couple points from @paulfalgout 

+ remove version.md in favor of package.json

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2296

jasonLaster:
Thanks @trusktr. Sounds like this is something we can continue investigating, but we have not closed the door on improvements that would make thing like `import CollectionView from 'marionette/collection-view;` impossible.

jasonLaster:
# Marionette has ES6 Modules

Thanks @paulfalgout, @trusktr, @cmisenas, @paulfalgout, @ianmstew for all of your hard work. 

---

This is a big PR, with lots of small inconsequential changes. If you see anything related in next or the PR you'd like to change, file an issue and a PR. This PR has gotten tons of eyes over the past four months and I'm confident that it is well reviewed.

---

Next steps. With this PR, we have a full ES6 build system based off of the babel boilerplate. I encourage you to take a fresh look at Marionette's src. Also, if you have a favorite ES6 feature like arrows, default values, or splats, now is your chance to get your line count changes in Marionette. Open hunting season all, regex away.

PR #2774, opened 2015-09-30
Replace _delegateDOMEvents with an override of backbone's delegateEvents

kworth:
Migrate model, collection, and behavior event delegation to initialize, and allow backbone setElement to call override of deleteEvents

See https://github.com/jashkenas/backbone/issues/3802

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -22,7 +22,7 @@ Marionette.View = Backbone.View.extend({
     this._behaviors = Marionette.Behaviors(this);
 
     Backbone.View.call(this, this.options);
-
+    this._delegateCollectionModelEvents();
     Marionette.MonitorDOMRefresh(this);
   },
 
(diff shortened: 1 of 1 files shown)


paulfalgout:
Thanks for the work here.  However this has [already been done](https://github.com/marionettejs/backbone.marionette/blob/next/src/abstract-view.js#L40) for the `next` branch, and since this is a breaking change, that's where it needs to go.

jridgewell:
@paulfalgout: This is a backport to fix a breaking change in Backbone.

paulfalgout:
It is almost identical to this commit https://github.com/marionettejs/backbone.marionette/commit/2196ee9b82bcf03c859f6a834a14105b2d3a24a9 already in `next`
I don't think currently the plan was to backport this into 2.x, but we certainly could.  I think the backport would have less downsides than not having it.
But here is the original discussion around this issue: https://github.com/marionettejs/backbone.marionette/pull/1569

JSteunou:
As we are told about v3 since a year now, and it still has no deadline, I'm in favour of a backport too.

kworth:
I'm new to the project, so forgive my ignorance. It looks like next is dramatically disconnected from master. Have the changes made on master been cherry-picked onto next? Looks like a merge of master and next yields a fair amount of conflicts. Has there been discussion around who will resolve these and when? If everything from master has been cherry-picked onto next, does that mean with the next merge of next onto master we would just overwrite master and take the changes from next lock, stock, and barrel?

PR #2777, opened 2015-10-01
Make test for LayoutView#when_destroying selfcontained

eth0lo:
There were a series of tests in the LayoutView tests, that setup information for the next upcoming test, this makes the test dependant on each other, which then makes difficult to move tests around or removing any of those.

On top of that they are harder to understand from where the setup information is coming.

### DIFF
--- a/test/unit/view.child-views.spec.js
+++ b/test/unit/view.child-views.spec.js
@@ -213,77 +213,87 @@ describe('layoutView', function() {
   });
 
   describe('when destroying', function() {
-    beforeEach(function() {
-      this.layoutViewManager = new this.View(this.viewOptions);
-      $('<span id="parent">').append(this.layoutViewManager.el);
-      this.layoutViewManager.render();
 
-      this.regionOne = this.layoutViewManager.regionOne;
-      this.regionTwo = this.layoutViewManager.regionTwo;
+    before(function() {
+      this.setup = function(viewOptions) {
+        this.layoutViewManager = new this.View(viewOptions);
+        $('<span id="parent">').append(this.layoutViewManager.el);
+        this.layoutViewManager.render();
 
-      var View = Marionette.View.extend({
-        template: false,
-        destroy: function() {
-          this.hadParent = this.$el.closest('#parent').length > 0;
-          return View.__super__.destroy.call(this);
-        }
-      });
+        this.regionOne = this.layoutViewManager.regionOne;
+        this.regionTwo = this.layoutViewManager.regionTwo;
 
-      this.regionOneView = new View();
-      this.regionOne.show(this.regionOneView);
+        var View = Marionette.View.extend({
+          template: false,
+          destroy: function() {
+            this.hadParent = this.$el.closest('#parent').length > 0;
+            return View.__super__.destroy.call(this);
+          }
+        });
 
-      this.sinon.spy(this.regionOne, 'empty');
-      this.sinon.spy(this.regionTwo, 'empty');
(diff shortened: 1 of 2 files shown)


paulfalgout:
:+1:  @marionettejs/marionette-core let's get this in

test/unit/view.child-views.spec.js
```diff
@@ -213,77 +213,88 @@ describe('layoutView', function() {
   });
 
   describe('when destroying', function() {
-    beforeEach(function() {
-      this.layoutViewManager = new this.View(this.viewOptions);
-      $('<span id="parent">').append(this.layoutViewManager.el);
-      this.layoutViewManager.render();
+    var oldLayoutView;
```
samccone: lets use this.oldLayoutView instead of making a new var in this "void" area

test/unit/view.child-views.spec.js
```diff
@@ -213,77 +213,88 @@ describe('layoutView', function() {
...
-      this.layoutViewManager.destroy();
-      this.layoutViewManager.destroy();
+        this.sinon.spy(this.layoutViewManager, 'destroy');
+        this.layoutViewManager.destroy();
+        this.layoutViewManager.destroy();
+      };
     });
 
     it('should empty the region managers', function() {
+      this.setup();
```
samccone: instead of setup, can we use a beforeEach ?

eth0lo: The main reason to called setup is because it needs to be configurable, beforeEach does not have that ability, the previous solution, which used beforeEach, relied on the order of the tests to set up the viewOptions for the beforeEach block.

The function setup was introduced to make the beforeEach block configurable, so now each test can configure the viewOptions for the test.

I don't know if I'm making my self clear here

samccone: >  make the beforeEach block configurable,

Yep and that is awesome, however it looks like we are not doing that anywhere in this PR so this seems somewhat like a premature optimization?

paulfalgout: I think he's referring to line 270.

FWIW this was a super brittle tests that was making development difficult.  I do think this whole file is up for a re-write, but this fix does make these tests a non-blocker for other development.

samccone: ah ah i missed that one ;)

samccone:
@eth0lo mind squashing into a single commit and then :ship: 

thanks a bunch!

test/unit/view.child-views.spec.js
eth0lo: Not only line 270, if you look at split view, instead of the consolidated one, it's more clear that all setup of test have been moved from the end of the previous test to the beginning of the test where the actual test happened.

I agree with @paulfalgout but I did not want to deviate a lot from the current structure of the file

eth0lo:
@samccone sure np

test/unit/view.child-views.spec.js
eth0lo: To make it more clear:

Lines 259-261(old version) were the setup for `should remove itself from the DOM before destroying child regions if flag set via options` but instead was found in `should not remove itself from the DOM before destroying child regions by default`

Lines 266-267 (old version) were the setup for `should remove itself from the DOM before destroying child regions if flag set on proto options` but instead was found in `should remove itself from the DOM before destroying child regions if flag set via options`

And so on hope this makes more clear what was the issue

samccone:
A+ work

PR #2790, opened 2015-10-23
Warn that events using triggers hash do not have access to DOM evt

Yahkob:
I recently came across this specific part of the docs while setting up a layoutView and thought it would be helpful to have a comment above the example trigger hash's method. I have to agree that all logic that uses DOM events should handle said logic in its own view but I think for a lot of developers it seems intuitive that one of the parameters in the function would be the DOM event object.  I didn't go into much detail about why this logic should be encapsulated since it seems implicit.


Yahkob:
Just read the contributing guide going to reopen as a PR to patch :smile:

PR #2791, opened 2015-10-23
Triggers hash details

Yahkob:
Originally opened a PR to merge to master, read contributing guide and opened this.

I recently came across this specific part of the docs while setting up a layoutView and thought it would be helpful to have a comment above the example trigger hash's method. I have to agree that all logic that uses DOM events should handle said logic in its own view but I think for a lot of developers it seems intuitive that one of the parameters in the function would be the DOM event object. I didn't go into much detail about why this logic should be encapsulated since it seems implicit.


Yahkob:
Woops commit history is messed up on this one. Let me know if the changes in the docs for layoutView looks good and you want to merge this and I'll clean up the PR.

samccone:
:+1: :LGTM

Yahkob:
Had sketchy commit history for some reason heres the cleaned up pr https://github.com/marionettejs/backbone.marionette/pull/2794

Issue #2797, opened 2015-10-28
CollectionView.reorder() filter not working when filter function produces subset of existing children.


msiconolfi:
This is my first issue, so maybe I am using the library wrong, but when I run the `reorder()` function on a `CollectionView`, the filter function doesn't seem to behave the way I would expect. If my filter function reduces the items in the collection to a subset of the currently rendered models then all the models are still rendered even if they should be filtered out by the filter function on the `CollectionView`. Below is a jasmine test which illustrates my point.

```javascript
describe('Marionette CollectionView', function() {
	var UserModel = Backbone.Model.extend({
		defaults: {
			first_name: '',
			last_name: '',
			username: ''
		}
	});

	var UserCollection = Backbone.Collection.extend({
		model: UserModel
	});

	var users,
		usersView;

	var UserView = Marionette.ItemView.extend({
		className: 'user',
		template: _.template("<label><%= username %>:</label><span><%= first_name %>  <%= last_name %></span>")
	});

	var UsersView = Marionette.CollectionView.extend({
		childView: UserView,
		sortDirection: 1,
		reverseDirection: function () {
			this.sortDirection *= -1;
		},
		viewComparator: function(model1, model2) {
			var name1 = model1.get('first_name');
			var name2 = model2.get('first_name');
			if (name1 < name2) {
				return 1 * this.sortDirection;
			}
			else if (name1 > name2) {
				return 1 * this.sortDirection;
			}
			return 0;
		},
	});

	beforeEach(function(){
		users = new UserCollection([
			{first_name: 'John', last_name: 'Doe', username: 'jdoe'},
			{first_name: 'Jane', last_name: 'Doe', username: 'jane.doe'},
			{first_name: 'Mike', last_name: 'Doe', username: 'mike.doe'},
		]);

		setFixtures('<div class="container"></div');

		usersView = new UsersView({
			el: '.container',
			collection: users,
		});

		usersView.render();
	});

	it('Filtering ', function() {
		expect(usersView.$el.find('.user').length).toBe(3);

		usersView.filter = function(model) {
			return model.get('first_name')[0] == 'J';
		};

		usersView.reorder();
		expect(usersView.$el.find('.user').length).toBe(2);
	});

});
```

[...]

PR #2798, opened 2015-10-28
Adding fix for filtering out views in CollectionView.reorder()

msiconolfi:
Based on my finding in this [Issue](https://github.com/marionettejs/backbone.marionette/issues/2797) I have created a pull request to address the issue.

The previous `CollectionView.reorder()` did not remove any previous views that might have been filtered out after calling `._filteredSortedModels()`. This patch will identify the models that have been removed and manually remove them using the `.removeChildView()` method. 

I added some test cases to cover this logic.


msiconolfi:
Closing Pull Request as it's against master instead of patch.

PR #2799, opened 2015-10-28
Adding fix for filtering out views in CollectionView.reorder()

msiconolfi:
Based on my finding in this [Issue](https://github.com/marionettejs/backbone.marionette/issues/2797) I have created a pull request to address the issue.

The previous `CollectionView.reorder()` did not remove any previous views that might have been filtered out after calling `._filteredSortedModels()`. This patch will identify the models that have been filtered out and manually remove their views using the `.removeChildView()` method. 

I added some test cases to cover this logic.

*Note* This is my second PR as the first one I had made against `master` as opposed to `patch`

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -165,27 +165,38 @@ Marionette.CollectionView = Marionette.View.extend({
   reorder: function() {
     var children = this.children;
     var models = this._filteredSortedModels();
-    var modelsChanged = _.find(models, function(model) {
+    var anyModelsAdded = _.some(models, function(model) {
       return !children.findByModel(model);
     });
 
-    // If the models we're displaying have changed due to filtering
-    // We need to add and/or remove child views
+    // If there are any new models added due to filtering
+    // We need to add child views
     // So render as normal
-    if (modelsChanged) {
+    if (anyModelsAdded) {
       this.render();
     } else {
       // get the DOM nodes in the same order as the models
-      var els = _.map(models, function(model, index) {
+      var elsToReorder = _.map(models, function(model, index) {
         var view = children.findByModel(model);
         view._index = index;
         return view.el;
       });
 
+      // find the views that were children before but arent in this new ordering
+      var filteredOutViews = children.filter(function(view) {
+        return !_.contains(elsToReorder, view.el);
+      });
+
+      this.triggerMethod('before:reorder');
+
       // since append moves elements that are already in the DOM,
       // appending the elements will effectively reorder them
-      this.triggerMethod('before:reorder');
-      this._appendReorderedChildren(els);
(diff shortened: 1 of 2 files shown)


jasonLaster:
Looks good

ahumphreys87:
nice catch @msiconolfi would you mind squashing these commits then I think we can get this in.

msiconolfi:
Ok, had a bit of trouble squashing, but I think we're good now. Just waiting for the Travis Checks to complete.

src/collection-view.js
```diff
@@ -165,27 +165,38 @@ Marionette.CollectionView = Marionette.View.extend({
   reorder: function() {
     var children = this.children;
     var models = this._filteredSortedModels();
-    var modelsChanged = _.find(models, function(model) {
+    var modelsAdded = _.find(models, function(model) {
```
megawac: should this be find or filter. seems like it should be a filter case (otherwise rename var please)

src/collection-view.js
```diff
@@ -165,27 +165,38 @@ Marionette.CollectionView = Marionette.View.extend({
...
       // get the DOM nodes in the same order as the models
-      var els = _.map(models, function(model, index) {
+      var elsToReorder = _.map(models, function(model, index) {
         var view = children.findByModel(model);
         view._index = index;
         return view.el;
       });
 
+      // find the views that were children before but arent in this new ordering
+      var filteredOutViews = children.filter(function(view) {
```
megawac: use _.filter

src/collection-view.js
ahumphreys87: we're only using this `var` to check if a model has been added so `find` is fine - but perhaps `_.some` would be a better option as that actually returns a `boolean`.

msiconolfi: :+1: I think the intent was to short circuit when an item was found. `_.some` might be more semantic given that we check it's truthiness below. I can make the change if you guys think it makes it more clear.

megawac: Definitely. Please use some

msiconolfi: Might mess up the git history. But adding `_.some` calling the `var` `anyModelsAdded` as I fell it reads better than `someModelsAdded` let me know if you want me to change it to `someModelsAdded`

ahumphreys87:
Lgtm :+1:

paulfalgout:
:+1:

src/collection-view.js
```diff
@@ -165,27 +165,38 @@ Marionette.CollectionView = Marionette.View.extend({
...
+        return !_.contains(elsToReorder, view.el);
+      });
```
megawac: wish underscore had `differenceBy`

Issue #2767, opened 2015-09-23
Improve gulp runner


jasonLaster:
Here's a checklist of things to look into for the new gulp runner

- [ ] jscs is not being run
- [ ] takes longer than grunt to run.
- [ ] locally the output is a mess.. 
- [ ] I think it’s linting the instrumented files or something.

paulfalgout:
I actually think eslint should be able to replace jscs.  It might take a little bit of tweaking to get it to line up.

jasonLaster:
@paulfalgout perhaps - JSCS was running last night when I was working on the ES6 modules ticket.

paulfalgout:
I was seeing linting in the output oh those crazy files.  I do wonder if most of the issues with the gulp stuff have to do with preexisting temp stuff left over by grunt.  I haven't tried from a clean dir.
Like locally try
```
grunt coverage
gulp
```

jasonLaster:
yea, that's interesting.

PR #2801, opened 2015-10-31
Improve gulp

paulfalgout:
This is related to https://github.com/marionettejs/backbone.marionette/pull/2769

Moving to gulp was :+1: 
But keeping grunt around and some of the legacy things were subtly breaking tests or making things seem ok when they weren't.  And things have changed a bit for the better in the node world since we started.  This should be more future friendly.

**Also this halves the time of the tests/build on Travis.**

I spent a bit of time, ripping all of the build tools out and putting the back carefully and minimally.

- jshint is now completely gone and eslint should match the old hint settings.
- grunt is gone and the duplicate test setup files were removed.  They were causing plenty of issues
- jsdom is upgraded to ^7.0 which is good for our windows devs.  This also means that node 4+ is a requirement
- coveralls and coverage are working
- the gulp tasks are broken up by build, test, and linter in an effort to make things more readable.
- I made `gulp-unwrapper` to make `unwrap` a gulp plugin and make it a cleaner implementation
- I moved the includes list into it's own file so that both bundler.js and core.js didn't have to be maintained.
- Dependencies have been updated and removed if not in use (other than dox)

*Issues*
- This does include a strange fix for Metal for the instance in which Travis loads a different backbone version.  This might indicate we have a bigger problem though.
- ~~Lodash isn't loading for the two 3.x builds and so this won't pass initially.  @megawac @samccone any ideas here in Travis?~~
- ~~There was also a strange failing test that was checking for undefined being passed in a very strange way that was modified to work.~~

*ES6* and https://github.com/marionettejs/backbone.marionette/pull/2769
This does not further the ES6 stuff at all.  In fact, some of that has been removed in the PR.  Hopefully it can be added easier following this new style.  Plus with babel6 and updates to the boilerplate for that v6 including webpack, I believe there are better ways to handle the ES6 stuff that we should get on now.  Some of the fixes on that branch are handled here as well, but differently.

There will be three separate efforts involved there.
1. The most important will be removing the `requireHelper` from the tests and having babel/webpack package the file for testing.  All ES6 dev can be done from here.  Most of it is done in some form on another branch and just needs to be added cleanly.
2. The build.  I think this can be left until last.
3. Eventually we'll want to implement boilerplates test-browser with the live-reload


test/unit/collection-view.empty-view.spec.js
```diff
@@ -279,7 +279,7 @@ describe('collectionview - emptyView', function() {
       });
 
       it('renders other empty view instance', function() {
-        expect(this.getEmptyViewStub).to.have.been.calledWith(this.collectionView.children.first());
+        expect(this.getEmptyViewStub).to.have.been.called;
```
ahumphreys87: what is the reason for this change?

paulfalgout: So this happened after some dependency updates.  I am not 100%, but I expect (pun intended)  that this is a later version of chai failing the test.   `getEmptyView` is not called with _any_ arguments.

paulfalgout: Ok I can verify.. I pulled a fresh `next` and updated the chai's and mocha's and sinon's.. I'm not sure which one it is for certain, but updating them with no other changes will fail this test:
```
AssertionError: expected getEmptyView to have been called with arguments undefined
      getEmptyView() at [object Object].Marionette.CollectionView.Marionette.AbstractView.extend.showEmptyView (src/collection-view.js:348:26)
      at Context.<anonymous> (test/unit/collection-view.empty-view.spec.js:282:52)
```

paulfalgout:
Specifically regarding ES6 we should likely do something very similar to:
For the build: https://github.com/babel/generator-babel-boilerplate/blob/v6.0.0/app/templates/gulpfile.babel.js#L63 
For the test-browser: https://github.com/babel/generator-babel-boilerplate/blob/v6.0.0/app/templates/gulpfile.babel.js#L125

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2767

I will say that local output isn't perfect when running everything at once..  This is because sometimes the linters bleed into the tests.. so you get the dot reporter intermixed with the linter output at first.  It looks fine later if there are issues.  Fixing this would mean serializing and extending the time it takes to run.

src/build/mn.js
```diff
@@ -0,0 +1,53 @@
...
+// @include ../abstract-view.js
+// @include ../view.js
+// @include ../collection-view.js
+// @include ../composite-view.js
+
+// @include ../behavior.js
+// @include ../behaviors.js
+
+// @include ../app-router.js
+// @include ../application.js
```
jasonLaster: This file is lovely. Once we add modules it will go, but it's such a nice refactor for core and bundled

jasonLaster:
@paulfalgout this PR is awesome on many levels. I feel much better about our infrastructure with this landing and I cannot wait to pull it down and play.

@marionettejs/marionette-core if you have other comments on the infrastructure, comment here, file issues, make PRs. I'm merging this in relatively quickly for us (2 days) because we iterate on infrastructure and can always improve more. This is decidedly different from code our users see.

ES6 Modules branch. Lets make landing this a priority, there's lots of good `src` work in there to make it all happen. We should do it.

PR #2803, opened 2015-11-02
Consistently return instance for Region#empty

duclet:
The empty method for Region is supposed to return itself when done. It works as expected when there is a view but returns undefined when there isn't. This breaks the ability to chain the method.


paulfalgout:
Thanks!  @duclet can you add a test for this?

duclet:
Test added.

jasonLaster:
Thanks @duclet!

PR #2804, opened 2015-11-02
Added support for passing down arguments to Marionette.Object.destroy

olivierbeaulieu:
When calling a Marionette.View's destroy method, you can pass options, which in turn are passed down to the view's `onBeforeDestroy` and `onDestroy` handlers.

I was quite surprised to see that this behaviour is not consistent with Marionette.Object's destroy method. Since this is a pretty useful pattern, I added that feature to Marionette.Object. This way we have consistency in behaviour across the board when destroying objects. The fix is pretty straight-forward.

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -20,9 +20,11 @@ _.extend(Marionette.Object.prototype, Backbone.Events, {
   //this is a noop method intended to be overridden by classes that extend from this base
   initialize: function() {},
 
-  destroy: function() {
-    this.triggerMethod('before:destroy');
-    this.triggerMethod('destroy');
+  destroy: function(options) {
+    options = options || {};
+
+    this.triggerMethod('before:destroy', options);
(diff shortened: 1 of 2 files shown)


paulfalgout:
Thanks!  This LGTM.  However can you add a line to the docs similar to the one on the view?
> The onDestroy method will be passed any arguments that destroy was invoked with.

olivierbeaulieu:
@paulfalgout The docs actually already say it does! But the behaviour wasn't actually implemented.

"*These calls will be passed any arguments destroy was invoked with.*"

https://github.com/olivierbeaulieu/backbone.marionette/blob/master/docs/marionette.object.md#destroying-a-object

paulfalgout:
Ah! ouch.  ok :+1:

ahumphreys87:
:+1:

Issue #2743, opened 2015-09-13
Reconsider the addition of region management code to Marionette.View


blikblum:
In next  branch (future 3.0), LayoutView was removed and the region management code was added to Marionette.View. 
I understand the reasoning behind this move (simplify the API) but there's a potential performance hit considering that majority of Marionette.View (formerly ItemView) instances in one typical app will not use the region code (think of list items, table rows). These are views that generally are the most rendered and rerendered. 

I think of two options:
- keep LayoutView and View classes as is in current version
- add first class mixin (functional mixin plus advice?) support that would allow to declare something like:
```
var MyViewWithRegions = Mn.View.extend({
   mixins: [Mn.RegionsMixin]
});
```
The later would open other possibilities to be used in other features

paulfalgout:
Well I think in v3 the simplification will go further than just making everything a layout.  Essentially I do not think there will be any perf decrease than if you were to add `_.isEmpty({})` to your `ItemView`'s initialize.  And region management will only occur when needed.

ianmstew:
What about lazily instantiating the RegionManager?

blikblum:
> there will be any perf decrease than if you were to add _.isEmpty({}) to your ItemView's initialize

This is the region management code called when a View is instantiate:
````
_initializeRegions
  _initRegionManager
    getRegionManager
      new Marionette.RegionManager
    4x listenTo(this.regionManager)
  _addRegions
    Marionette._getValue
    normalizeUIValues
    _buildRegions
      regionManager.addRegions
  _addRegions
    Marionette._getValue
    normalizeUIValues
    _buildRegions
      regionManager.addRegions
```

When re-rendered:

    _reInitializeRegions

AFAICS is more expensive then _.isEmpty({}) and also increase memory usage (a RegionManager instance thus the events addition)

ianmstew:
@blikblum Great point, and I think we are actually all in agreement.  @paulfalgout is working on a simplification to Marionette.View that will allow it to support regions _without_ the burden of RegionManager.  An early proposal is here: https://github.com/paulfalgout/backbone.marionette/pull/3/files#diff-54628cdbb43b8b10d359b3bfa2a80387R6.

blikblum:
Fine. I'll keep looking.

paulfalgout:
@blikblum the latest merge should help alleviate this issue

blikblum:
Thanks. Nice work

Issue #2477, opened 2015-04-09
Merge layoutView docs into ItemView


jasonLaster:
https://github.com/marionettejs/backbone.marionette/issues/2432#issuecomment-91217622

They were dropped in a merge.

paulfalgout:
Also need to check `SpecRunner.html`

paulfalgout:
Update: I suspect `SpecRunner.html` is updated in this missing PR https://github.com/marionettejs/backbone.marionette/commit/c63b5847e213b17be2cb7a5387b02c16ff6bd058

paulfalgout:
closing in favor of #2920

PR #2806, opened 2015-11-04
Remove RegionManager

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2743

This vastly simplifies the relationship between view and region.

It should lessen the concern about any perf-hit from all Views having regions.


JSteunou:
:+1: this sounds very nice!

src/view.js
```diff
@@ -180,64 +103,11 @@ Marionette.View = Marionette.AbstractView.extend({
...
-  _getImmediateChildren: function() {
-    return _.chain(this.regionManager.getRegions())
-      .pluck('currentView')
-      .compact()
-      .value();
+    return Marionette.AbstractView.prototype.destroy.apply(this, arguments);
   }
 });
+
+_.extend(Marionette.View.prototype, Marionette.RegionsMixin);
```
ahumphreys87: do we have Metal mixins yet?

jmeas:
:heart_eyes:

src/view.js
jasonLaster: nope, i'm going to remove metal too. It's currently flagged off, but still a headache

jasonLaster:
I'd like to keep `actsAsCollection` it's useful and ppl might use it :)

paulfalgout:
We should make `actsAsCollection` a separate utility maybe?  Possibly use it in babysitter as well.  Babysitter has it's own version of this, and they both get out of date and out of sync too easily.  But the utility most likely won't be used in the library at all unless babysitter is brought in.
Plus would we just attach it to the parent `Marionette`?  Seems like in ES6 world it makes more sense just to `import actAsCollection from 'backbone-actascollection';`

paulfalgout:
@jmeas yeah I can delete the doc I suppose.. although there's lingering stuff in various places of the docs on this branch.  They're essentially unmaintainable at the moment and need a lot of love.  related https://github.com/marionettejs/backbone.marionette/issues/2477

jmeas:
> although there's lingering stuff in various places of the docs on this branch. They're essentially unmaintainable at the moment and need a lot of love.

Gotcha. Np then; it shouldn't hope up this good work.

paulfalgout:
@jasonLaster the previously mentioned collection function on babysitter is here: https://github.com/marionettejs/backbone.babysitter/blob/master/src/childviewcontainer.js#L135

jasonLaster:
`import actAsCollection from 'backbone-actascollection'` sounds good to me. Lets kill it

jasonLaster:
:+1: 

To be clear, we should merge this ASAP. PRs this good, and this big should not be allowed to go stale :bread: :egg: , else they look like :skull: which nobody wants :hear_no_evil:

paulfalgout:
To be fair a lot of the concepts here are from various @jfairbank @jmeas discussions

ahumphreys87:
:+1:

PR #2809, opened 2015-11-04
(WIP) Re-adds a bowser test runner

jasonLaster:
This brings back our live-reloading, source maps powered, test runner.

Well, kinda... needs one or two teaks to get over the finish line. 

@cmisenas @jmeas @paulfalgout mind lending your expertise here. A little stuck

### DIFF
--- a/gulp/browser_test.js
+++ b/gulp/browser_test.js
@@ -1,0 +1,84 @@
+import gulp from 'gulp';
+const plumber = require('gulp-plumber');
+const glob = require('glob');
+const babelify = require('babelify');
+const watchify = require('watchify');
+const browserify = require('browserify');
+const runSequence = require('run-sequence');
+const liveReload = require('gulp-livereload');
+const source = require('vinyl-source-stream');
+const buffer = require('vinyl-buffer');
+
+function getBundler() {
+  // Our browserify bundle is made up of our unit tests, which
+  // should individually load up pieces of our application.
+  // We also include the browserify setup file.
+  var testFiles = glob.sync('./test/unit/**/*');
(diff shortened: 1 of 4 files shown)


gulp/browser_test.js
```diff
@@ -0,0 +1,84 @@
...
+gulp.task('browserify', function() {
+  return bundle();
+});
+
+// Ensure that linting occurs before browserify runs. This prevents
+// the build from breaking due to poorly formatted code.
+gulp.task('build-in-sequence', function(callback) {
+  runSequence([
+    // 'lint-src',
+    'lint-test'
```
jasonLaster: oops, was trying to speed it up by commenting it out

paulfalgout:
Yeah I'd like to take a look at this.  In fact I really think we need to do what the boilerplate is doing in an upcoming version: https://github.com/babel/generator-babel-boilerplate/blob/v6.0.0/app/templates/gulpfile.babel.js#L125

It uses webpack instead of browserify.  I don't know that there is any _big_ advantage to doing that for the live browser, but specifically it is useful for using webpack instead of esperanto for the build..

Also.. and maybe why it isn't working.. the library still has to be built somehow.. like `specRunner.html` is making `Backbone.Marionette = {};` and each hardcoded, ordered file is building out the object..  but the boilerplate builds a js file to test with using the dependency tree which we don't have yet.  I'm not 100% sure what the intermediate step looks like.

paulfalgout:
Yeah.. the more I look into this the more trouble I think it is.  We should just do this along side https://github.com/marionettejs/backbone.marionette/pull/2769

The solution for our current preprocess build system is sooooo different than what it is when the tests can just import in their dependencies and get bundled.

Issue #2793, opened 2015-10-27
Bug: `childEvents` along with a function which returns hash doesn't work


andrei-placinta-sociomantic:
The problem:
you are not be able to use `childEvents` along with function which returns hash (only simple hash can be used).
But according to doc, you should able to:
http://marionettejs.com/docs/v2.4.3/marionette.layoutview.html#layoutview-childevents
E.g. `onSomeEventMethod` will never be called: 
`childEvents: function () {  return { 'someEvent': 'onSomeEventMethod' };  }`

The minimal working example:
```
var LV = Marionette.LayoutView.extend({
    template: _.template('<div class="js-region"></div>'),
    regions: 
    {
        testRegion: '.js-region'
    },
    childEvents: function () // <<< childEvents returning a function does NOT work
    {
        return {
            'test': function () 
            {
                console.log('function test')
            }
        };
    }
})

var V = Marionette.ItemView.extend({
    template: _.template('<div>view</div>')
});

var lv = new LV();
var v = new V();

lv.render();
lv.$el.appendTo('body');
lv.testRegion.show( v );

v.triggerMethod('test'); // <<< nothing happens

// but if you change `childEvents` to simple hash - it will work
lv.childEvents = {
    'test': function () 
    {
        console.log('hash test');
    }
};

v.triggerMethod('test'); // <<< `hash test` will be printed

```

PS:
I investigated a bit, and found that inside marionette sources after this line:
`var childEvents = Marionette.getOption(layoutView, 'childEvents');`
missed `_getValue` call, something like:
`childEvents = Marionette._getValue( childEvents, layoutView );`

paulfalgout:
Confirmed

paulfalgout:
https://github.com/marionettejs/backbone.marionette/blob/master/src/view.js#L306 does not get the result of the function.

msiconolfi:
Saw the *help wanted* tag, definitely be interested into working on a fix for this if it's still open.

paulfalgout:
@msiconolfi sure thing!

msiconolfi:
Thanks, started working on this. It looks like @andrei-placinta-sociomantic was right with the fix: `childEvents = Marionette._getValue( childEvents, layoutView );` so it should be pretty straight forward just working on some test cases and should have a PR up in a couple of days.

Another question I had why do we have a helper method called `_getValue` instead of `getValue` is it intended to be private?

PR #2815, opened 2015-11-09
Fixing View.childEvents to support using a function which returns a hash (Issue #2793)

msiconolfi:
This it the fix for [Issue 2793](https://github.com/marionettejs/backbone.marionette/issues/2793). The implementation change was trivial (thanks @andrei-placinta-sociomantic for making it easy). I also added a single test case to test this new logic. Let me know if you think I should add more tests.


samccone:
:+1: to this landing

jasonLaster:
Great fix!

paulfalgout:
:clap: Thanks @msiconolfi !

samccone:
:bow:

Issue #2786, opened 2015-10-16
LayoutView childEvents: Does not respond to child triggering event from within onRender


MeoMix:
http://codepen.io/anon/pen/MaOyKz

	var ChildView = Marionette.LayoutView.extend({
	  template: false,
	  onRender: function(){
		console.log('onRender');
		this.triggerMethod('content:updated');

		_.defer(function(){
		  console.log('_defer onRender');
		  this.triggerMethod('content:updated');
		}.bind(this));
	  }
	});

	var ParentView = Marionette.LayoutView.extend({
	  template: _.template('<div data-region=content></div>'),
	  regions: {
		content: '[data-region=content]'
	  },
	  childEvents:{
		'content:updated': function(){
		  console.log('content:updated has fired');
		}
	  },
	  onRender: function(){
		this.showChildView('content', new ChildView());
	  }
	});

	var parentView = new ParentView();
	parentView.render();

Feel like this should work fine, but the `content:updated` event handler only runs once.

paulfalgout:
I _think_ if https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L94 was moved up above the render it would solve this problem with no side effects.

MeoMix:
Surely this is a breaking change? If someone had code which was dependent on childEvents not running during `onRender` that it would cause their code to work incorrectly? Maybe that's not considering a breaking change?

paulfalgout:
Hmm that'd be code that worked around what I think is clearly a bug.  I don't think there's anything to suggest that the expected behavior is that triggerMethod wouldn't trigger a childEvent inside the first render.  I don't think this is a breaking change, I think it's fixing an expected behavior.  @marionettejs/marionette-core ?

paulfalgout:
Also it's not that it _doesn't_ work in `onRender` it's that it doesn't work in the _first_ `onRender` from a `region.show`  For instance you can pre-render the view (which won't trigger it because it doesn't have a parent..) then `attachView` it to the region instead of show.  Then render the child, and it will trigger.

PR #2817, opened 2015-11-11
Moving parent binding to before child rendering in Region.show() (#Issue 2786)

msiconolfi:
This is a potential fix for [Issue 2786](https://github.com/marionettejs/backbone.marionette/issues/2786). I modified an existing test in `layout-view.spec.js` to explicitly mention it can work with basic `Backbone.View`'s. I also added a similar test case to the one in the issue using a `Marionette.ItemView` since it supports the 'onRender` and `triggerMethod` methods.

Let me know if the tests look ok or you need something done differently.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -89,9 +89,12 @@ Marionette.Region = Marionette.Object.extend({
       // we can not reuse it.
       view.once('destroy', this.empty, this);
 
-      this._renderView(view);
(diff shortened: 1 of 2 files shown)


paulfalgout:
:+1:

jasonLaster:
nice.

PR #2818, opened 2015-11-11
Implement childTriggers

### LINKED ISSUE #2341: Why do View.triggers not bubble up through views like DOM events?
Hey,

I'm just wondering if there's a specific reason for triggers not bubbling up. They do for CollectionView -> Children, but not for LayoutView -> Region -> ItemView.

Is this just a huge performance concern? Are there other implications I haven't considered? I think that the best way to approach it currently is to use a channel and pass an ID through the channel to ensure that if there's multiple views subscribed to the event that they know which one it is intended for.

For example:

    Grandparenta
        Parent1a
            Child1a
        Parent2a

    Grandparentb
        Parent1b
            Child1b
        Parent2b

Child1a has an event on it and that information needs to get propagated to Parent2a, but not propagate to Parent2b. This requires either a chain of event listeners, or a channel with unique IDs.

Thoughts?

### LINKED ISSUE #2404: childTriggers or childProxy
On a view, `events` has a sister hash `triggers` for conveniently converting DOM events into view events.  Thanks to @jasonLaster's #2316, we are now uniformly dealing with `childEvents` across all views with children.  I propose a sibling hash to _childEvents_ called `childTriggers`.  _childTriggers_ would be a logical parallel to _triggers_ in that it converts child events to "my own" view events.

The logical parallel is nice, but the real value of _childTriggers_ would be the ability to declaratively proxy child events, in other words give the ability to bubble, or deeply link, events through views more than 2 levels deep.  This solution would avoid any unintended or complicated side effects of "auto proxying" the way DOM nodes do, and would serve as a more elegant solution to #2341 than we have today while prioritizing zero magic over automation.  I imagine the PR would apply childTriggers to _all_ views implementing _childEvents_.

Example:

Instead of
```js
childEvents: {
  activate: function () { this.triggerMethod(‘activate’); }
}
```

We would have
```js
childTriggers: {
  activate: 'activate'
}
```

Or even
```js
childProxy: {
  activate: true
}
```

### LINKED ISSUE #2667: triggerMethod bubble fail?
the are two cases: 

1) layoutView -> region -> collectionView -> layoutView -> itemView
2) layoutView -> region -> collectionView -> itemView

in the first case, i triggerMethod in the itemView, i can't catch the event in the first layoutView.
in the second case, i do the same thing, the layoutView catch the event

it seemed that layoutView can not bubble the event to the parent view.
is that right?

i also want to know is there a method to stop the event bubble? something like:

``` javascript
tirggerMethod('event', {bubbleOne: true})
```

sometime i just want the parent view catch the event, other than the parent parent view can also catch the event.

megawac:
See #2404 for details

/cc @ianmstew 

Fixes #2404, resolves #2341, resolves #2667 

Currently ignores `{'all': ??????}` cases. Probably should error.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -5,6 +5,8 @@
 // The core view class that other Marionette views extend from.
 Marionette.View = Backbone.View.extend({
   isDestroyed: false,
+  supportsRenderLifecycle: true,
+  supportsDestroyLifecycle: true,
 
   constructor: function(options) {
     this.render = _.bind(this.render, this);
--- a/src/view.js
+++ b/src/view.js
@@ -304,6 +304,10 @@ Marionette.View = Backbone.View.extend({
 
     // call the parent view's childEvents handler
     var childEvents = Marionette.getOption(layoutView, 'childEvents');
+
+    // since childEvents can be an object or a function use Marionette._getValue
+    // to handle the abstaction for us.
+    childEvents = Marionette._getValue(childEvents, layoutView);
     var normalizedChildEvents = layoutView.normalizeMethods(childEvents);
 
     if (normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
--- a/src/view.js
+++ b/src/view.js
@@ -7,7 +7,7 @@ Marionette.View = Backbone.View.extend({
   isDestroyed: false,
 
   constructor: function(options) {
-    _.bindAll(this, 'render');
+    this.render = _.bind(this.render, this);
 
     options = Marionette._getValue(options, this);
 
--- a/src/view.js
+++ b/src/view.js
@@ -313,6 +313,15 @@ Marionette.View = Backbone.View.extend({
     if (normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
       normalizedChildEvents[eventName].apply(layoutView, callArgs);
     }
+
+    // call the parent view's proxyEvent handlers
+    var childTriggers = Marionette.getOption(layoutView, 'childTriggers');
+    childTriggers = Marionette._getValue(childTriggers, layoutView);
+
+    // Call the event with the proxy name on the parent layout
+    if (childTriggers && _.isString(childTriggers[eventName])) {
+      layoutView.triggerMethod.apply(layoutView, [childTriggers[eventName]].concat(args));
+    }
   },
 
   // This method returns any views that are immediate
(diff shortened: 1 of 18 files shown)


src/view.js
```diff
@@ -313,6 +313,15 @@ Marionette.View = Backbone.View.extend({
     if (normalizedChildEvents && _.isFunction(normalizedChildEvents[eventName])) {
       normalizedChildEvents[eventName].apply(layoutView, callArgs);
     }
+
+    // call the parent view's proxyEvent handlers
+    var childTriggers = Marionette.getOption(layoutView, 'childTriggers');
+    childTriggers = Marionette._getValue(childTriggers, layoutView);
```
jasonLaster: we see this all over. wish it were done by default

jasonLaster:
:+1:

src/view.js
megawac: I'd be = on adding a helper composing getOption and getValue

megawac: By the way @jasonLaster, I think all this stuff should be precomputed rather than done during an event trigger (same for `childEvents`). Thoughts?

paulfalgout: ha yes!  I [totally agree](https://github.com/marionettejs/backbone.marionette/pull/2819#issuecomment-156258456)

I also wonder if we should allow `_getValue` to accept a string like:
```js
Marionette._getValue = function(value, context, params) {
    if(_.isString(value)){
      value = Marionette.getOption(context, value);
    }
    if (_.isFunction(value)) {
      value = params ? value.apply(context, params) : value.call(context);
    }
    return value;
  };
```

We don't currently have a use-case where the we'd want the value to sometimes actually be a string value..  but I suppose we could.. in which case this wouldn't work.

ahumphreys87:
@megawac this looks good, mind opening against `next` I think this was slated for 3.x

megawac:
I was aiming this for 2.5 if that is on the road map; I think a
re-evalutation/removal of childEvents would be more good discussion for 3.x

ahumphreys87:
We weren't planning on pushing a 2.5 out - we're trying to make 3.0 our
next release (except bug fixes). How do you feel about putting this into
3.0 as it is and opening a discussion around childEvents?

paulfalgout:
It isn't hard to roll a 2.5.  While I certainly don't want to be distracted, it could help bridge the 3.x gap. Might deserve a convo

JSteunou:
My 2 cents: you should focus on 3.0, it's been mention for 18 month or so now, some Marionette fans (myself included) just cannot wait to play with all the new toys, and wont stay on 2.5 if the 3.0 just go out a couple months later.

ahumphreys87:
moved to #2845

Issue #2783, opened 2015-10-14
LayoutView childEvents: Allow for filtering based on region.


MeoMix:
Consider the following:

```
var ConfirmDialogView = Marionette.LayoutView.extend({
    template: false,
    onShow: function(){
        this.triggerMethod('dialog:confirm', { result: 'Ok' });
    }
});

var FooView = Marionette.LayoutView.extend({
    template: _.template('<div data-region="dialog1"></div><div data-region="dialog2"></div>'),

    regions: {
        dialog1: '[data-region=dialog1]',
        dialog2: '[data-region=dialog2]'
    },

    childEvents: {
        'dialog:confirm': '_onChildDialogConfirm',
    },

    onRender: function(){
        this.showChildView('dialog1', new ConfirmDialogView());
        this.showChildView('dialog2', new ConfirmDialogView());
    }

    _onChildDialogConfirm: function(options){
        console.log('Confirm result:', options.result);
    }
});
```

It would be nice to be able to filter whether an event comes from the dialog1 or dialog2 region. Similar to in 'events' where the selector filters how/when event handlers run.

This would allow for:

```
childEvents: {
    '@region.dialog1 dialog:confirm': '_onRegion1DialogConfirm',
    '@region.dialog2 dialog:confirm': '_onRegion2DialogConfirm'
}
```

russplaysguitar:
:+1:

Tranfu:
:+1:

msiconolfi:
So I have a preliminary PR for this feature. Should I branch off of next for this? I don't see a `layout-view.js`. Did this get moved into something?

msiconolfi:
It looks like on `next` `Marionette.AbstractView` is what  `Marionette.View` was and `Marionette.View`is now what `Marionette.LayoutView` was?

jasonLaster:
I would prefer dropping the `@region.dialog1` in favor of `@dialog1`. 

In general I'm afraid of this pattern slightly because I would prefer `dialog:confirm` would be replaced with something action specific like `form:saved` `form:updated` where there's context and verb... but I'm not sure I'm necessarily opposed to this idea either.

paulfalgout:
Now that proxying view events is normalized between View and CollectionView I think this would be abnormally difficult to implement in any sort of performant fashion on a very very hot function.

We should not do this. :-1:

However I think a moderately simple behavior can handle the use-case

https://jsfiddle.net/g9znf174/

[...]

PR #2819, opened 2015-11-12
Adding region aliases into childEvents hash

msiconolfi:
Preliminary enhancement for supporting region aliases in the `View.childEvents` hash, as requested in [Issue 2763](https://github.com/marionettejs/backbone.marionette/issues/2783).
- Moved layout specific logic out of `AbstractView` and into `View` object, and simply calling `layoutView.triggerChildEvents(eventName, this, args);` from inside `AbstractView._triggerEventOnParentLayout`; to me this seems to separate the logic a little better, and also provides a hook for subclasses to override if needed.
- Added region alias detection to determine if the current view is the one specified in the `childEvents` hash, non-region aliases should behave the same as in earlier releases.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -234,6 +234,41 @@ Marionette.View = Marionette.AbstractView.extend({
     });
   },
 
+  // given an event name, childView and args determine if the layout view's
+  // we need to trigger any matching handlers defined in the childEvents options
+  triggerChildEvents: function(eventName, childView, args) {
+    // invoke triggerMethod on parent view
+    var eventPrefix = this.getOption('childViewEventPrefix');
+    var prefixedEventName = eventPrefix + ':' + eventName;
+    var callArgs = [childView].concat(args);
+
+    // trigger the prefixed event
+    Marionette._triggerMethod(this, [prefixedEventName].concat(callArgs));
+
+    var childEvents = this.getOption('childEvents');
+    var normalizedChildEvents = this.normalizeMethods(childEvents);
+
+    // iterate over all normalized events looking for matching childEventName
+    // determine if there is also a region alias that needs to be bound correctly
+    _.each(normalizedChildEvents, function(eventHandler, childEventName) {
+        var regionEventSplitter = /^(@region\.(.*?)\s+)?\s*(.*)$/;
+        var match = childEventName.match(regionEventSplitter);
+        var regionName = match[2];
+        var regionEventName = match[3];
+
+        // is there a valid region alias event bound to this childView
+        var validRegionAliasEvent = !!regionName &&
+                                    this.getChildView(regionName) === childView &&
+                                    regionEventName === eventName;
+
+        if (_.isFunction(eventHandler) && (validRegionAliasEvent || childEventName === eventName)) {
+          eventHandler.apply(this, callArgs);
+        }
+      }, this);
+
+    return childView;
+  },
+
   _getImmediateChildren: function() {
     return _.chain(this.regionManager.getRegions())
       .pluck('currentView')
--- a/src/view.js
+++ b/src/view.js
@@ -177,7 +177,7 @@ Marionette.View = Marionette.AbstractView.extend({
   },
   showChildView: function(regionName, view, options) {
     var region = this.getRegion(regionName);
-    // this._unbindRegionEvents(regionName, view);
+    this._unbindRegionEvents(regionName, view);
     this._bindRegionEvents(regionName, view);
 
     return region.show.apply(region, _.rest(arguments));
--- a/src/view.js
+++ b/src/view.js
@@ -126,18 +126,60 @@ Marionette.View = Marionette.AbstractView.extend({
 
     // Add the regions definitions to the regions property
     this.regions = _.extend({}, this.regions, regions);
+    this._regionEvents = {};
 
     return this._buildRegions(regions);
   },
 
   // Remove a single region from the LayoutView, by name
   removeRegion: function(name) {
+    this._unbindRegionEvents(name);
     delete this.regions[name];
     return this.regionManager.removeRegion(name);
   },
+  _bindRegionEvents: function(regionName, childView) {
+    var childEvents = this.getOption('childEvents');
+    var normalizedChildEvents = this.normalizeMethods(childEvents);
 
+    var regionEvents = {};
+
+    // determine if there is a region alias that needs to be bound
+    _.each(normalizedChildEvents, function(eventHandler, childEventName) {
+        var regionEventSplitter = /^(@region\.(.*?)\s+)?\s*(.*)$/;
+        var match = childEventName.match(regionEventSplitter);
+        var regionAlias = match[2];
+        var regionEventName = match[3];
+
+        // is there a valid region alias event bound to this childView
+        var validRegionEvent = _.isString(regionAlias) && regionAlias === regionName;
+
+        if (_.isFunction(eventHandler) && validRegionEvent) {
+          regionEvents[regionEventName] = function() {
+            var args = Array.prototype.slice.call(arguments);
+            eventHandler.apply(this, [childView].concat(args));
+          };
+
+          this.listenTo(childView, childEventName, regionEvents[regionEventName]);
+        }
+      }, this);
+
+    this._regionEvents[regionName] = regionEvents;
+    return regionEvents;
+  },
+  _unbindRegionEvents: function(regionName, childView) {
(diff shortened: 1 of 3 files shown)


src/view.js
```diff
@@ -234,6 +234,37 @@ Marionette.View = Marionette.AbstractView.extend({
     });
   },
 
+  // given an event name, childView and args determine if the layout view's
+  // we need to trigger any matching handlers defined in the childEvents options
+  triggerChildEvents: function(eventName, childView, args) {
+    // invoke triggerMethod on parent view
+    var eventPrefix = this.getOption('childViewEventPrefix');
+    var prefixedEventName = eventPrefix + ':' + eventName;
```
msiconolfi: Need to handle this `prefixedEventName` if it includes `@region.regionName`

jasonLaster:
nice @msiconolfi! 

I left a comment in #2783. But, i wonder if `@region.regionName` could be replaced with `@regionName`.

Also, not sold on the value but I'd defer to the group here.

Lastly, if we used the generic `@alias` maybe we could re-use the normalize methods in utils even more?

paulfalgout:
I think I'm :-1: on this.

While it follows the established pattern, I'm not confident in the current implementation.

My concern in general is that we are doing a lot of work for each event a child throws.. and those child events are already pretty noisy.  Seems like the @region childEvents could be analyzed before the view is shown one time and then those event hashes are managed separately?

I'm not sure exactly the solution, but in general I think we need to find ways to reduce trigger weight and to reduce event noise.

jasonLaster:
@paulfalgout i think you're right. We do that for the other aliases

msiconolfi:
@paulfalgout I agree the `_.each` is needlessly ran for every childEvent. I could do as you said and analyze the before the view is shown the first time. However, it seems like this feature may not ultimately desired.

If you think that this might be a viable feature to eventually release, then I can work on a better approach. Could others weigh in on whether adding @region aliases for childEvents is something we would want in the library one day?

paulfalgout:
Hmmm.. it may very well be a viable feature.. I personally think `childEvents` and [`childTriggers`](https://github.com/marionettejs/backbone.marionette/pull/2818) are an improvement to and should replace the auto-proxying of all of the `'onChildviewBeforeRender': 'magic'`, but I'd like to see a cheap lookup for what events should be proxied.

jasonLaster:
ping @msiconolfi hope you have a chance to tackle this :)

msiconolfi:
Still working on this. I wanted to review the library as whole to try and see how to best fit it in and provide the cheap lookup. I should have a second implementation in a couple of days.

jasonLaster:
Nice. What was your opinion of the library as a whole? Everything is busted? Not so bad :)

Also, did you see we landed the new ES6 build and modules yesterday? AKA everything is different :P

msiconolfi:
Regarding @paulfalgout comment "Seems like the @region childEvents could be analyzed before the view is shown one time and then those event hashes are managed separately?". This would work fine for when `childEvents` is defined as a hash but when it is defined as a function returning a hash, this would not work. Since the function returning a hash would be dynamic we wouldn't know until it was executed and wouldn't be able to take advantage of any pre-`show` lookup generation. 

Any ideas of how we could still make this a cheap lookup, at this point it may be too expensive for it to be a useful feature.

We could possibly add a `regionEvents` attribute which only supports a hash, but that might not fit the pattern of `childEvents`, `collectionEvents`, and `modelEvents` of allowing either a hash or function returning hash.

jasonLaster:
Should be fine. The childEvents function would be executed on view instantiation and saved to an indtance prop we can access.

jasonLaster:
:phone:  @msiconolfi!  How goes it?

msiconolfi:
Still working on it. I'm not in love with my current implementation, but would appreciate any feedback.

Here is the basic flow:
- Upon calling `View.showChildView()` regionEvents will be setup using the newly added `View._bindRegionEvents(regionName)` method.
- This will iterate over the normalized childEvents and parse out any region aliases that match the regionName.
- Any matching events will be bound directly to the childView using the `listenTo()` method.
- The events will then be stored in a private variable `this._regionEvents` so that they can eventually be unbound upon: new region show, or a region destroy.

I still need to add `View._unbindRegionEvents()` to a couple more places but not sure I am totally in love with this approach yet.

A couple things I don't like.
- I had originally wanted to do the lookup in my `View.triggerChildEvents()` method, however there was no easy way to determine in which region(s) the childView belonged so I could do the lookup on `View._regionEvents`. I could still continue this way but then I would also have to create a mapping between childViews and the region(s) they belonged. The difficult part was supporting the ability for one child view to possibly be bound to more than region, which may not be a reasonable expectation. If we lax that requirement it probably wouldn't be too bad to create a lookup that will map views to region.
- As an alternative the problem above, I decided that I could use `listenTo` to bind the events directly when the region was showed, since at that time I know the region name. For most cases the events will be unbound on the child view's `destroy()`, however if the view has the `preventDestroy` flag set then we need to manually clean up the regionEvents upon `showChildView` / `destroy()` of the `View`. This was the reason I had created the `View._unbindRegionEvents()` method.

jasonLaster:
Definitely going in the right direction. I like bind and unbind w/ listenTo. 

Mind rebasing. We've been busy :)

JSteunou:
Is this going v3? no milestone on original issue #2783

paulfalgout:
v3.x :-)

msiconolfi:
I can pick this up this week, sorry this fell of my plate. I just need to rebase and then re-work the bind/unbind. I should be able to get something up and running this week to review.

Thanks for the nudge.

paulfalgout:
@msiconolfi no pressure here at all.  Thanks for being up to working on it!  We might be in the weeds a bit soon with the v3 release candidate.  This could certainly land at any point, but take all the time you need :+1:

src/view.js
```diff
@@ -126,18 +126,60 @@ Marionette.View = Marionette.AbstractView.extend({
...
+        var match = childEventName.match(regionEventSplitter);
+        var regionAlias = match[2];
+        var regionEventName = match[3];
+
+        // is there a valid region alias event bound to this childView
+        var validRegionEvent = _.isString(regionAlias) && regionAlias === regionName;
+
+        if (_.isFunction(eventHandler) && validRegionEvent) {
+          regionEvents[regionEventName] = function() {
+            var args = Array.prototype.slice.call(arguments);
```
rafde: would `_.toArray` help shorten this line?

src/view.js
```diff
@@ -126,18 +126,60 @@ Marionette.View = Marionette.AbstractView.extend({
...
   },
-
+  _bindRegionEvents: function(regionName, childView) {
+    var childEvents = this.getOption('childEvents');
+    var normalizedChildEvents = this.normalizeMethods(childEvents);
+
+    var regionEvents = {};
+
+    // determine if there is a region alias that needs to be bound
+    _.each(normalizedChildEvents, function(eventHandler, childEventName) {
```
rafde:
```
_.each(normalizedChildEvents, _.bind(function(eventHandler, childEventName) {
...
}, this);
```
for lodash v4

src/view.js
maniator: Actually since we are using es6 just use spread would be better.

src/view.js
```diff
@@ -126,18 +126,60 @@ Marionette.View = Marionette.AbstractView.extend({
...
+        var regionEventName = match[3];
+
+        // is there a valid region alias event bound to this childView
+        var validRegionEvent = _.isString(regionAlias) && regionAlias === regionName;
+
+        if (_.isFunction(eventHandler) && validRegionEvent) {
+          regionEvents[regionEventName] = function() {
+            var args = Array.prototype.slice.call(arguments);
+            eventHandler.apply(this, [childView].concat(args));
+          };
```
maniator: regionEvents[regionEventName] = function(...args) {
        eventHandler.apply(this, [childView].concat(args));
    }

maniator: or can even do:

    regionEvents[regionEventName] = function(...args) {
        eventHandler.apply(this, [childView, ...args]);
    } 

Which would work too :-)

src/view.js
maniator: No need for that if you use the fat arrow in es6

```
_.each(normalizedChildEvents, (eventHandler, childEventName) => {
...
}));
```

scott-w:
Is this being subsumed into #3157?

paulfalgout:
This is pretty outdated now.  For now it seems like the best course is to close this and refer to it if anyone picks up the open issue.

PR #2820, opened 2015-11-13
Optimize lookup for parent layout

megawac:
This should be a quick :+1: and merge

/cc @jasonLaster


paulfalgout:
:+1:  great

jasonLaster:
Nice win. 

I still like getAncestors for debugging, but I don't think that's enough of a reason to keep it.

PR #2823, opened 2015-11-16
(WIP) Add getOptionValue

jasonLaster:
Based on our discussion, here's a basic implementation for a helper getOptionValue which composes two of our most used helpers.

I'll add docs and tests if we like it

### DIFF
--- a/src/utils/_getOptionValue.js
+++ b/src/utils/_getOptionValue.js
@@ -1,0 +1,8 @@
+// Composes getOption with the intent of
+// evaluating a target's property or option.
+// Useful when class property like `events` or `regions`
+// could either be an object or function.
+export default function(target, optionName, params) {
+  var option = target.getOption(optionName);
+  return target.getValue(option, params);
+}
--- a/src/mixins/behaviors.js
+++ b/src/mixins/behaviors.js
@@ -1,7 +1,8 @@
-import _                        from 'underscore';
-import getValue                 from '../utils/getValue';
-import { triggerMethod }        from '../trigger-method';
-import Marionette               from '../backbone.marionette';
+import _ from 'underscore';
+import getValue from '../utils/getValue';
+import _getOptionValue from '../utils/_getOptionValue';
+import { triggerMethod } from '../trigger-method';
+import Marionette from '../backbone.marionette';
 
 // Takes care of getting the behavior class
 // given options and a key.
@@ -38,7 +39,7 @@ function parseBehaviors(view, behaviors) {
 
 export default {
   _initBehaviors: function() {
-    var behaviors = this.getValue(this.getOption('behaviors'));
+    var behaviors = _getOptionValue(this, 'behaviors');
 
     // Behaviors defined on a view can be a flat object literal
     // or it can be a function that returns an object.


paulfalgout:
Also I suppose there's a reason getValue can't handle a string as an option?

jasonLaster:
So, I'm not convinced that there are a lot of uses, the motivation is that our DSL, which evaluates properties like `modelEvents` or `behaviors` for instance properties or options properties, object properties or function properties, is prone for error. Here's a [gist](https://gist.github.com/jasonLaster/fd83341e87665ea41a59) of the places where `getOption` is called. Unfortunately, we can't assume that every `getOption` call intends to immediately evaluate a function


##### Types:
1. (instance, object)  `View.extend({modelEvents: {}});`
2. (instance, fnc)  `View.extend({modelEvents: function() {return {}}});`
3. (options, object) `new View({modelEvents: {}})`
4. (options, fnc)   `new View({modelEvents: function() {return {}}})`

paulfalgout:
> Unfortunately, we can't assume that every getOption call intends to immediately evaluate a function

My 2nd questions is essentially the opposite.. Can we assume that all `getValue`s could handle a string as a `getOption`?

paulfalgout:
FWIW I think this pattern happens a couple other times but in separate steps.. For instance I ran into it kind of on the radio mixin
`getOption` here https://github.com/marionettejs/backbone.marionette/pull/2758/files#diff-951bb4cc49ca2d8446383594f46fe145R26
passed to something that `getValue`s https://github.com/marionettejs/backbone.marionette/pull/2758/files#diff-73c6cec277d5cc65ad42a85db80263adR31

paulfalgout:
Hmm I don't think `getValue` can take a string and assume `getOption`.. sometimes the field we're analyzing is either a string or a function returning a string, and we want the string to just pass through... so yeah.. :+1: Just needs tests, docs I guess.. though we have a general v3 doc issue.. and any other implementations in the src.

jasonLaster:
I kinda want to punt on this actually. Doesn't seem worth it...

paulfalgout:
cool.  yeah.. on the fence myself

paulfalgout:
With all of the nice refactors, I think this would clean some things up.  Time to reconsider?

jasonLaster:
lets do it. will rebase

jasonLaster:
@paulfalgout after thinking about it, I don't think this warrants tests or documentation as I'd like to keep the function internal and not exposed on `Marionette`. It'll be well tested as it will be used in lots of places.

paulfalgout:
Sounds good.  Can the util filename be _'d and the getValue is now public

jasonLaster:
sure

paulfalgout:
oh sorry.. actually the API for both getOption and getValue is different in this context.  neither accept the target argument and should be called with context.. and I think getValue now accepts n args instead of an array

jasonLaster:
sorry, tried and failed again i believe. looking at getOption again, it became clear that it expected a context `this`, which getOptionValue wouldn't have... Anyways, once again, not sure about the value here... any suggestions.

src/utils/_getOptionValue.js
```diff
@@ -0,0 +1,8 @@
+// Composes getOption with the intent of
+// evaluating a target's property or option.
+// Useful when class property like `events` or `regions`
+// could either be an object or function.
+export default function(target, optionName, params) {
```
paulfalgout: maybe
```js
export default function(target, optionName, ...args) {
  var options = getOption.call(target, optionName);
  return getValue.apply(target, [option].concat(args));
}
```

jasonLaster: yea, honestly, not feeling it... I think `this.getValue(this.getOption('foo'))` is fine in most cases

paulfalgout: :+1:

PR #2825, opened 2015-11-17
Add Rest & Spread ES6 syntax

jasonLaster:
Fun little PR that takes advantage of one of the nice ES6 features (rest and spread) to make the code a little bit more readable.


src/abstract-view.js
```diff
@@ -201,11 +201,9 @@ var AbstractView = Backbone.View.extend({
   // DOM and unbinding it. Regions will call this method
   // for you. You can specify an `onDestroy` method in your view to
   // add custom code that is called after the view is destroyed.
-  destroy: function() {
```
paulfalgout: we should apply this exact same thing to `Object#destroy` which explicitly passes `options` instead of the `args`

paulfalgout:
Love this.  It'll pass tests after #2834, but we need to dump the whitelist.  It ain't workin'.

paulfalgout:
I'd love to do this same thing to `Object#destroy` and `Application#showView`

src/collection-view.js
```diff
@@ -701,20 +701,17 @@ var CollectionView = Backbone.View.extend({
...
+      var childEventName = prefix + ':' + rootEvent;
 
       // call collectionView childViewEvent if defined
       if (typeof childViewEvents !== 'undefined' && _.isFunction(childViewEvents[rootEvent])) {
-        childViewEvents[rootEvent].apply(this, args.slice(1));
+        childViewEvents[rootEvent].apply(this, [view, ...args]);
       }
 
-      this.triggerMethod.apply(this, args);
+      this.triggerMethod.apply(this, [childEventName, view, ...args]);
```
ianmstew: I think this could get even smaller :D
```js
this.triggerMethod(childEventName, view, ...args)
```

paulfalgout: oh yeah.. no need to put it back into an array.  so sweet.

ianmstew: Does `call` work that way?  I.e., what's better, `call(context, arg, ...args)` or `apply(context, [arg, ...args])`?

paulfalgout: well in many of these situations the context doesn't change, so as you mentioned there's no need to do a call or apply.

ianmstew: True.

src/collection-view.js
```diff
@@ -701,20 +701,17 @@ var CollectionView = Backbone.View.extend({
...
 
-      args[0] = prefix + ':' + rootEvent;
-      args.splice(1, 0, view);
+      var childViewEvents = this.normalizeMethods(_.result(this, 'childViewEvents'));
+      var childEventName = prefix + ':' + rootEvent;
 
       // call collectionView childViewEvent if defined
       if (typeof childViewEvents !== 'undefined' && _.isFunction(childViewEvents[rootEvent])) {
-        childViewEvents[rootEvent].apply(this, args.slice(1));
+        childViewEvents[rootEvent].apply(this, [view, ...args]);
```
ianmstew: @paulfalgout Is this confusing?  `childViewEvents[rootEvent].call(this, view, ...args)`

jasonLaster: i think that's a little nicer @ianmstew

ianmstew: Cool, TIL

test/unit/view-mixin.spec.js
```diff
@@ -1,4 +1,4 @@
-describe('base view', function() {
+describe('view mixin', function() {
```
jasonLaster: small cleanup from earlier. changing the file name and top-level describe

test/unit/object.spec.js
```diff
@@ -82,12 +82,12 @@ describe('marionette object', function() {
...
-    it('should pass an empty object options down to the onBeforeDestroy method', function() {
-      expect(this.beforeDestroyHandler).to.have.been.calledWith({});
+    it('should pass an empty array of options options down to the onBeforeDestroy method', function() {
+      expect(this.beforeDestroyHandler).to.have.been.calledWith();
     });
 
-    it('should pass an empty object options down to the onDestroy method', function() {
-      expect(this.onDestroyHandler).to.have.been.calledWith({});
+    it('should pass an empty array of options down to the onDestroy method', function() {
+      expect(this.onDestroyHandler).to.have.been.calledWith();
```
jasonLaster: I think the semantics of `triggerMethod('foo')` are that the handler is called w/ no params. In this one case we were calling it w/ an empty obj.

src/collection-view.js
jasonLaster: Damn... this is cool

src/object.js
```diff
@@ -43,17 +43,15 @@ _.extend(MarionetteObject.prototype, Backbone.Events, {
...
   initialize: function() {},
 
-  destroy: function(options) {
+  destroy: function(...args) {
     if (this._isDestroyed) { return this; }
 
-    options = options || {};
-
-    this.triggerMethod('before:destroy', options);
+    this.triggerMethod.apply(this, ['before:destroy', ...args]);
```
paulfalgout: I don't know why we'd do the `options = options || {}` thing in this case when it is not that way on view destroy..  however.. you could still support this with ['before:destroy', options, ...args]`  _or_ I _think_ we could be doing `destroy: function(options = {}) {` WHAAAAA?

test/unit/object.spec.js
```diff
@@ -82,12 +82,12 @@ describe('marionette object', function() {
       expect(this.object.destroy).to.have.returned(this.object);
     });
 
-    it('should pass an empty object options down to the onBeforeDestroy method', function() {
-      expect(this.beforeDestroyHandler).to.have.been.calledWith({});
+    it('should pass an empty array of options options down to the onBeforeDestroy method', function() {
```
paulfalgout: so this isn't true... it's just passing `undefined` instead of `{}`

test/unit/view-mixin.spec.js
paulfalgout: :+1:

src/object.js
```diff
@@ -43,17 +43,15 @@ _.extend(MarionetteObject.prototype, Backbone.Events, {
...
   initialize: function() {},
 
-  destroy: function(options) {
+  destroy: function(...args) {
     if (this._isDestroyed) { return this; }
 
-    options = options || {};
-
-    this.triggerMethod('before:destroy', options);
+    this.triggerMethod('before:destroy', ...args);
```
paulfalgout: gak.. comment got outdated.. was saying we could still support this via 
```
options = options || {};
this.triggerMethod('before:destroy', options, ...args);
```
or perhaps better.. I think this should work now `destroy: function(options={}) {`

But I'm ok with the change too.. dunno why we are doing it here and not on `View#destroy`

test/unit/object.spec.js
jasonLaster: gotcha, will change it to be `shouldn't pass any options`

ianmstew:
:+1:

PR #2826, opened 2015-11-17
Fix _parent link in view

ftdebugger:
Do not remove view link to parent region, when it pass twice to the same region


jasonLaster:
:+1:

paulfalgout:
Thanks!  Manually merged into patch

PR #2830, opened 2015-11-24
Patch

aqqaluk:


paulfalgout:
Patch was cherry-picked onto master  For instance https://github.com/marionettejs/backbone.marionette/commit/11dc5511bbe812c54a0f25f40ea4b8af378bcbcf

If there anything in particular that appears to be missing?

PR #2832, opened 2015-11-25
childEvents -> childViewEvents

### LINKED ISSUE #2812: childEvents and childViewEvents
As we're trying to wrap up the v3 stuff, the last hard thing is related to what needs to be finalized around view events and what can be pushed till a v4.  

There are certainly some unfortunate quirks with the attach events..  but a lot of that won't be any worse.

The biggest issue I can see if we have both `childEvents` and `childViewEvents` and our simplified view structure I think will make them clash a lot more if we rename the View's to `childViewEvents` (as we should)

paulfalgout:
Resolves #2812


jasonLaster:
:thumbsup:

jasonLaster:
Merging as it's a very simple `regex`. If others have comments, questions we can make a new PR very easily.

PR #2834, opened 2015-11-29
Gulp build es6 libs

paulfalgout:
@marionettejs/marionette-core this is a big one and needs :eyes:

I'd also suggest that we look for anything glaringly wrong, and be less concerned with whether or not we should be using a different tool.  I tried many options, but this one was the only one that both worked and worked well in my opinion.  I would suggest as long as this isn't incorrect, that we merge it and any opinions as to other bundlers or versions of babel be addressed at a later date so as not to hold up v3 any longer.

So there were definitely a few things wrong with the ES6 stuff in general in ways that the tests are forgiving.  At some point it might be worth having the tests import what they need and not the full library so that it is easier to catch these issues.

In any case I attempted babel 6 a bunch of different ways with various builds and I couldn't get a build that works or that produced a sane output.

Rollup for it's flaws does a good job of making a readable lib file, but it doesn't play nicely with babel 6 and gulp.

I also attempted to load the tests via rollup and remove the browserify dependency..  I do think this might be possible except that all of the test files would HAVE to switch to import and ditch require.

Anyhow, I've also included a commit for review of what the build produces.  When this is reviewed and ready to go, I need to pull out that commit.

This will also fix the tests for #2825


jasonLaster:
First of all, This is awesome work. I agree, this should be merged ASAP because it gets us a built lib. 

@paulfalgout mind removing the build commit and putting the patch in a gist for others. 

I think we should review this work and merge it, but as a philosophical point I am okay walking away from a readable/beautiful lib file and single annotated source. Hopefully we encourage two patterns for reading our source: src files, in devtools w/ source maps.

gulp/build.js
```diff
@@ -1,41 +1,108 @@
-import gulp from 'gulp';
-import header from 'gulp-header';
-import preprocess from 'gulp-preprocess';
-import rename from 'gulp-rename';
+import gulp       from 'gulp';
+import header     from 'gulp-header';
+import plumber    from 'gulp-plumber';
+import file       from 'gulp-file';
+import filter     from 'gulp-filter';
+import rename     from 'gulp-rename';
```
ahumphreys87: why we doing this?

ahumphreys87:
@paulfalgout this is great! 1 slight nit - but I suspect this could be the eslint rules. Agree that we should get this merged asap once build commit is removed.

gulp/build.js
paulfalgout: We don't.  It was done for the imports on the es6 stuff,  so this made that more consistent.   I do think it looks nicer, but it does make it a pain to maintain and blame.

ahumphreys87: Im not a fan of lining up for those reasons. That said - its a separate
conversation and shouldn't hold up this PR

paulfalgout:
My biggest issue with the less-readable variety build is that you really then have to fully trust the tool since it's hard to understand the output, plus it typically adds weight and helpers that'd be duplicates for people repackaging the library.  If there's a clear benefit to a tool that produces less-readable code, I'm all for it.. but at the moment, most of these things are in their infancy and don't play well with others.  And I know we've talked about removing the bundled versions in the past, but keeping them is non-trivial for some build situations.

I fully expect the tools will improve and there will be a clearer path, and I'm certainly not married to the solution, but it is a _moderately_ simple setup that changes as little as possible from the state of things right now.  ie: we don't _have_ to remove the annotated source page.. we don't _have_ to remove the bundled build..

paulfalgout:
https://gist.github.com/paulfalgout/27ca9265f81c7e62390a

jasonLaster:
@paulfalgout you're about as reasonable as they come. I'd like to research some of these tools for my own edification, but you make a lot of sense.

paulfalgout:
FWIW rollup has a lot of potential, but some flaws... specifically when looking at babel v6 and gulp https://github.com/rollup/rollup-plugin-babel/issues/12#issuecomment-157265775  But because you can use babel during the build it's pretty good.

But rollup has no concept of require() so we can't package our browser tests in their current state with it.  

Even babel itself is close to being able to solve the issue of UMD https://www.npmjs.com/package/babel-plugin-transform-es2015-modules-umd but no real sense of packaging.. plus it has very little options, but it is a super simple implementation.. and can be customized.  

Then browserify/webpack and even jspm could be used for packaging the library, but none of them are purposefully built to build libraries and so to me everything feels a bit hacky to me and the output is moderately impossible to control.  I am perhaps most intrigued by jspm, but it feels like the wild west.  But without tree shaking at least browserify and webpack both fall short as the _entire_ package.json would be loaded into the lib.  I _think_ jspm could do this.. and webpack is working on it.

In any case, I find it fun but pretty challenging.  The worst parts of it is that you can do a bunch of things that _seem_ right only to find out that you're importing 6 different versions of _ or you've redeclared something that you shouldn't have.. And the tools will adapt to these issues and you'll end up `_$1` `_$2` `_$3`.. which isn't at all what you want.  But for instance.. all of our tests passed because browserify and node worked fine with 6 different underscores..  but that is particularly bad for our build.

But a good way to compare is to look at the js outputted by the browserify build of the tests.  Mn starts around line 13200. and the start looks like:
```js
Object.defineProperty(exports, '__esModule', {
  value: true
});

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }

var _backbone = require('backbone');

var _backbone2 = _interopRequireDefault(_backbone);

var _underscore = require('underscore');

var _underscore2 = _interopRequireDefault(_underscore);

var _behaviors = require('./behaviors');

var _behaviors2 = _interopRequireDefault(_behaviors);
```

Issue #2292, opened 2015-02-05
Simplify bindEntityEvents


jmeas:
[All of this bloat](https://github.com/marionettejs/backbone.marionette/blob/master/src/bind-entity-events.js#L18-L107) could maybe become as small [as this](https://github.com/jmeas/puppets/blob/v0.0.1/src/utilities.js#L37-L54)? The Puppets version doesn't support space-separated names, so I suppose I'd have to add that in, but it shouldn't be too hard!

thejameskyle:
:+1:

PR #2835, opened 2015-11-29
Simplify bindEntityEvents

paulfalgout:
While looking for patterns for https://github.com/marionettejs/backbone.marionette/pull/2758 it seemed like some clean-up would be good.

Turns out as usual @jmeas had already figured that out.

Resolves https://github.com/marionettejs/backbone.marionette/issues/2292

This keeps the current functionality.  I suppose it adds the possibility of an error on unbinding from a string where it wouldn't have been before.

### DIFF
--- a/src/bind-entity-events.js
+++ b/src/bind-entity-events.js
@@ -17,46 +17,23 @@ import _               from 'underscore';
 import _getValue       from './utils/_getValue';
 import MarionetteError from './error';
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
:thumbsup:

jasonLaster:
Nice improvement, if others have revisions they'd like to make we can open a new PR or re-open this one.

PR #2838, opened 2015-11-30
Add Rest & Spread ES6 syntax

jasonLaster:
followup on #2825. hopefully will trigger travis


jasonLaster:
oh. looks like the force push did

PR #2839, opened 2015-11-30
Refactor: Add default ES6 options

jasonLaster:
This is mostly, just another syntatic cleanup. I added ES6 default
options and _.result default options.

It's worth pointing out that in the process I found a couple places
where we were not cloning the prototype property, which left the
opportunity open to mutate it. This is *not* a desirable behavior.

Also, I believe `|| {}` is mostly removed from the codebase, except
places where we need to do some local property accessing for passed in
options.

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -27,8 +27,8 @@ import { triggerMethod }       from './trigger-method';
 
 var AppRouter = Backbone.Router.extend({
 
-  constructor: function(options) {
-    this.options = options || {};
+  constructor: function(options = {}) {
+    this.options = options;
 
     Backbone.Router.apply(this, arguments);
 
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -18,7 +18,7 @@ var Behavior = MNObject.extend({
     // wants to directly talk up the chain
     // to the view.
     this.view = view;
-    this.defaults = _.result(this, 'defaults') || {};
+    this.defaults = _.clone(_.result(this, 'defaults', {}));
     this.options  = _.extend({}, this.defaults, options);
     // Construct an internal UI hash using
     // the behaviors UI hash and then the view UI hash.
--- a/src/region.js
+++ b/src/region.js
@@ -16,11 +16,11 @@ import { triggerMethodOn, triggerMethodMany } from './trigger-method';
 var Region = MNObject.extend({
   cidPrefix: 'mnr',
 
-  constructor: function(options) {
+  constructor: function(options = {}) {
 
     // set options temporarily so that we can get `el`.
     // options will be overriden by Object.constructor
-    this.options = options || {};
+    this.options = options;
     this.el = this.getOption('el');
 
     // Handle when this.el is passed in as a $ wrapped element.
--- a/src/regions-mixin.js
+++ b/src/regions-mixin.js
@@ -8,7 +8,7 @@ export default {
 
   // Internal method to initialize the regions that have been defined in a
   // `regions` attribute on this View.
-  _initRegions: function(options) {
+  _initRegions: function() {
 
     // init regions hash
     this.regions =  this.regions || {};
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ node_js:
   - "stable"
 matrix:
   include:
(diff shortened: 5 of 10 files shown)


src/behavior.js
```diff
@@ -18,7 +18,7 @@ var Behavior = MNObject.extend({
     // wants to directly talk up the chain
     // to the view.
     this.view = view;
-    this.defaults = _.result(this, 'defaults') || {};
+    this.defaults = _.clone(_.result(this, 'defaults', {}));
```
paulfalgout: Well.. so do we want to update the underscore dep to 1.8?  1.7 doesn't have a default value for `_.result`  Not sure about when lodash added it either.

src/region.js
```diff
@@ -16,11 +16,11 @@ import { triggerMethodOn, triggerMethodMany } from './trigger-method';
 var Region = MNObject.extend({
   cidPrefix: 'mnr',
 
-  constructor: function(options) {
+  constructor: function(options = {}) {
```
paulfalgout: this would work without the default too.. `getOption` will handle the undefined

src/regions-mixin.js
```diff
@@ -11,7 +11,7 @@ export default {
   _initRegions: function(options) {
 
     // init regions hash
-    this.regions =  this.regions || {};
+    this.regions =  _.clone(_.result(this, 'regions', {}));
```
paulfalgout: I think this is actually just ensuring the hash exists.. we shouldn't do the result here.  `addRegions` will do that

src/regions-mixin.js
```diff
@@ -11,7 +11,7 @@ export default {
   _initRegions: function(options) {
```
paulfalgout: hmm this options is unused after the viewmixin change.. it can be removed..

src/application.js
```diff
@@ -8,11 +8,8 @@ import Region           from './region';
 var Application = MarionetteObject.extend({
   cidPrefix: 'mna',
 
-  constructor: function(options) {
-    options = options || {};
-
+  constructor: function(options = {}) {
     this._initRegion(options);
```
paulfalgout: hmm this is actually broken.
this shouldn't pass options to `_initRegion`
It should
```
  // set options temporarily so `_initRegion` can use `getOption`
  this.options = options;
```

Concurrent branch issues here.. also Application isn't tested super well.. (but it will be)

src/app-router.js
```diff
@@ -27,8 +27,8 @@ import { triggerMethod }       from './trigger-method';
 
 var AppRouter = Backbone.Router.extend({
 
-  constructor: function(options) {
-    this.options = options || {};
+  constructor: function(options = {}) {
```
paulfalgout: I'm not 100% sure we need the default here at all.. this is just attaching options for `getOption` I assume.. and `getOption` will actually work better if it's undefined (one less check)

paulfalgout:
Really glad you explored this.. pointed out some issues..  also.. I mentioned this above but to fully work this'll need underscore `1.8.0`+  for the default result..  I'm actually fine with that.. `1.7.x` has some compatibility issues anyhow reverted in `1.8.0` so we'll potentially have less issues pop up with not support `1.7.x`

src/regions-mixin.js
jasonLaster: cool

src/app-router.js
jasonLaster: hmm, I think some people use `this.options` so it'd be nice to keep the default behavior

jasonLaster:
thanks @paulfalgout. I'll leave a note that we're supporting Underscore 1.8 and up in the doc.

src/application.js
jasonLaster: Hmm... made some application.js Changes

paulfalgout:
cool... change bower.json, package.json and travis to reflect dropping 1.7 too.. then it's done and done.

src/app-router.js
paulfalgout: oh right true.. good point

paulfalgout:
Ok this looks good.  We can PR changes if need be later.  Too much going on right now to leave blockers.

PR #2840, opened 2015-11-30
Refactor Behavior and UI Mixin out of View Mixin

jasonLaster:
This PR, extracts the behavior and UI logic out of Views Mixin so that
it is relatively-self contained.

NOTE: there are three instance props maintained in Behaviors Mixin
(_behaviors, behaviorEvents, behaviorTriggers)

----

*Patterns*: I tried to think through some patterns that might be useful
in future extractions.

1. any expression that touched behavior state should be a function
defined in the behaviors mixin

2. the behaviors mixin should be mixed into both views (collection,
view) that way in the future some views can get one mixin and not
another.

3. all view mixins should live inside of a `mixins` folder. I move the region mixin into it.

----

It also, kills the behaviors.js file and moves that logic to some util files and the mixin.

This involves a couple things:

1. extract parseBehaviors, BehaviorsTriggerBuilder, and behaviorsLookup
out of the Behaviors src file

2. move the Behaviors constructor logic to initBehaviors

3. finally kill the `Behaviors.wrap` function which was setting the
behaviorEvents function on the prototype. It's my most hated function.

---

Is this a big PR? Yes. Is it terribly scary? No, because we're just rejiggering things a bit and moving them around.


src/mixins/view.js
```diff
@@ -119,24 +116,23 @@ export default {
...
-    var behaviorEvents = _.result(this, 'behaviorEvents') || {};
-    var triggers = this.configureTriggers();
-    var behaviorTriggers = _.result(this, 'behaviorTriggers') || {};
+    if (typeof eventsArg === 'undefined') {
+      this.events = events;
+    }
 
-    // behavior events will be overriden by view events and or triggers
-    _.extend(combinedEvents, behaviorEvents, events, triggers, behaviorTriggers);
+    var combinedEvents = _.extend({},
```
paulfalgout: I know this isn't necessarily part of this PR, but I do think this._getBehaviorTriggers() should come before this.configureTriggers(), to match events..  I think behavior triggers were added later and the order wasn't considered.

paulfalgout:
should we do anything about mixin dependencies? Like the UI mixin will not work without the behaviors mixin.  I wish there was a clean way to pull out the behaviors mixin that did not create mixin dependencies.

src/mixins/view.js
```diff
@@ -215,58 +187,7 @@ export default {
 
   bindUIElements: function() {
```
paulfalgout: OHHH this is perfect.. composes both of the mixin behaviors together.  This just means that `unbindUIElements` should be pulled out of the ui mixin and put here.

jasonLaster: Oops. Nice find.

src/mixins/view.js
jasonLaster: Cool

src/backbone.marionette.js
```diff
@@ -85,6 +85,9 @@ Marionette.triggerMethodMany = triggerMethodMany;
 Marionette.isEnabled = isEnabled;
 Marionette.setEnabled = setEnabled;
 
+Marionette.Behaviors = {};
```
jasonLaster: This used to be in behaviors.js

src/backbone.marionette.js
```diff
@@ -95,7 +98,6 @@ Marionette.View = View;
 Marionette.CollectionView = CollectionView;
 Marionette.CompositeView = CompositeView;
 Marionette.Behavior = Behavior;
-Marionette.Behaviors = Behaviors;
```
jasonLaster: killing this thing

src/behaviors.js
```diff
@@ -1,179 +0,0 @@
...
-  }
-
-  // Behaviors defined on a view can be a flat object literal
-  // or it can be a function that returns an object.
-  behaviors = Behaviors.parseBehaviors(view, behaviors);
-
-  // Wraps several of the view's methods
-  // calling the methods first on each behavior
-  // and then eventually calling the method on the view.
-  Behaviors.wrap(view, behaviors, _.keys(methods));
```
jasonLaster: So... happy to see this go

src/utils/parseBehaviors.js
```diff
@@ -0,0 +1,37 @@
...
+function getBehaviorClass(options, key) {
+  if (options.behaviorClass) {
+    return options.behaviorClass;
+    //treat functions as a Behavior constructor
+  } else if (_.isFunction(options)) {
+    return options;
+  }
+
+  // behaviorsLookup can be either a flat object or a method
+  return _getValue(Marionette.Behaviors.behaviorsLookup, this, [options, key])[key];
```
paulfalgout: This shouldn't import Behaviors here, not Marionette.Behaviors

src/mixins/ui.js
```diff
@@ -0,0 +1,64 @@
...
+
+    // reset the ui element to the original bindings configuration
+    this.ui = this._uiBindings;
+    delete this._uiBindings;
+    delete this._ui;
+  },
+
+  getUI: function(name) {
+    this._ensureViewIsIntact();
+
```
jasonLaster: probably should delete this line

src/utils/parseBehaviors.js
jasonLaster: sadly, that'd break everything. We need to use `Marionette.Behaviors` so that it can be overidden in userland.

src/utils/behavior-trigger-builder.js
```diff
@@ -0,0 +1,44 @@
...
+_.extend(BehaviorTriggersBuilder.prototype, {
+  // Main method to build the triggers hash with event keys and handlers
+  buildBehaviorTriggers: function() {
+    _.each(this._behaviors, this._buildTriggerHandlersForBehavior, this);
+    return this._triggers;
+  },
+
+  // Internal method to build all trigger handlers for a given behavior
+  _buildTriggerHandlersForBehavior: function(behavior, i) {
+    var triggersHash = _.clone(_.result(behavior, 'triggers')) || {};
```
paulfalgout: _.result default should be {}

src/utils/parseBehaviors.js
paulfalgout: Ah well that's ok..  in that case this still needs to have an import Marionette from 'backbone.marionette';

paulfalgout: You may as well add the `Marionette.Behaviors = {};` to the top of this file instead of inside `backbone.marionette` so it's in context..

src/backbone.marionette.js
paulfalgout: :boom:

src/utils/parseBehaviors.js
paulfalgout: errr.. and the `Marionette.Behaviors.behaviorsLookup = behaviorsLookup;`

src/backbone.marionette.js
```diff
@@ -20,11 +20,11 @@ import View               from './view';
 import CollectionView     from './collection-view';
 import CompositeView      from './composite-view';
 import Behavior           from './behavior';
-import Behaviors          from './behaviors';
 import Region             from './region';
 import Application        from './application';
 import AppRouter          from './app-router';
 import MarionetteError    from './error';
+import behaviorsLookup    from './behaviors-lookup';
```
paulfalgout: any thoughts on putting this and features into like a config directory?  Not sure if there's really anything else that would qualify as config related things?

src/mixins/behaviors.js
```diff
@@ -0,0 +1,121 @@
...
+var delegateEventSplitter = /^(\S+)\s*(.*)$/;
+
+function getBehaviorsUI(behavior) {
+  return behavior._uiBindings || behavior.ui;
+}
+
+export default {
+  _initBehaviors: function() {
+    var behaviors = _getValue(this.getOption('behaviors'), this);
+    var instanciatedBehaviors = {};
```
paulfalgout: `s/instanciatedBehaviors/instantiatedBehaviors/g`

src/utils/parseBehaviors.js
jasonLaster: Hmm. Okay.

src/utils/behavior-trigger-builder.js
jasonLaster: Yup

src/backbone.marionette.js
jasonLaster: No. That's smart.

src/mixins/behaviors.js
jasonLaster: Damn

src/mixins/behaviors.js
```diff
@@ -0,0 +1,121 @@
...
+  _getBehaviorTriggers: function() {
+    return _.result(this, 'behaviorTriggers', {});
+  },
+
+  behaviorTriggers: function(unusedBehaviorTriggers) {
+    var triggerBuilder = new BehaviorTriggersBuilder(this, this._behaviors);
+    return triggerBuilder.buildBehaviorTriggers();
+  },
+
+  behaviorEvents: function(unusedBehaviorEvents) {
```
paulfalgout: I think you could just rename this function `_getBehaviorEvents` also fortunately this `unusedBehaviorEvents` and the "unused" above this can go away

jasonLaster: True. And true. Actually meant to delete unused but got distracted.

paulfalgout:
This is arguably getting out of hand :-D but https://github.com/jasonLaster/backbone.marionette/pull/4  @jasonLaster check out this PR before taking this further..

Essentially I had a thought before bed that `Behavior` could just use the `UIMixin` directly.. and it all went down hill from there.

src/mixins/ui.js
```diff
@@ -0,0 +1,106 @@
+// allows for the use of the @ui. syntax within
```
jasonLaster: This mixin is a really nice example of modularity in my mind

gulp/browser-test.js
```diff
@@ -40,6 +40,6 @@ function browserWatch() {
   gulp.watch(['src/**/*.js', 'test/**/*.js'], ['browser-bundle']);
 }
 
-gulp.task('browser-bundle', ['lint-src', 'lint-test'],   bundle);
+gulp.task('browser-bundle', [],   bundle);
```
paulfalgout: don't like the linting?  We could probably make these run async to the tests, but their sync because often a linting error can break the build.

src/backbone.marionette.js
paulfalgout: Yeah.. so that turned out to be more difficult than I expected with the final build.  importing Marionette works for DEV_MODE for instance because it's used inside a function, but declaring things in the global scope of the module doesn't jive well with `backbone.marionette` later.  There might be a cleaner solution, but I'm not sure atm what it is.

src/mixins/behaviors.js
```diff
@@ -0,0 +1,98 @@
...
+  },
+
+  _getBehaviorTriggers: function() {
+    var triggers = _.invoke(this._behaviors, 'getTriggers');
+    return _.extend({}, ...triggers);
+  },
+
+  _getBehaviorEvents: function() {
+    var events = _.invoke(this._behaviors, 'getEvents');
+    return _.extend({}, ...events);
```
paulfalgout: woah.. that works eh?  I guess it would.. that is amazing.

src/backbone.marionette.js
jasonLaster: Cool  we can come back to it

jasonLaster:
Alright, I think this PR is in a good spot. Don't be afraid of the size. We're just re-shuffling, here are some stats.

**Files Killed**
```
src/utils/normalizeUIKeys.js
src/utils/normalizeUIString.js
src/utils/normalizeUIValues.js
src/behaviors.js
````

**Files Added**
```
src/mixins/behaviors.js
src/mixins/ui.js
src/utils/getUniqueEventName.js
 src/behaviors-lookup.js
```

**Ugliness Removed**
```
1. Behaviors.wrap
2. Behaviors.method and basically all of this code was weird where it was
3. View.delegateEvents is much cleaner
4. Behavior is much cleaner
5. UImixin has 4 public methods that are much easier to follow
6. View Mixin lost 80 lines of code and a lot of complexity
```

**Best Thing about this work?** There are lots of ways, this is just the beginning. Things are going to get much nicer.

File: gulp/browser-test.js (hunk 13 above)
Comment (jasonLaster) [inline gulp/browser-test.js]: async would be good. My behavior is to run both the browser and terminal tests so that i can quickly see if there's something for me to investigate in the browser... Could be weird...

File: src/view.js [hunk 15]
```diff
@@ -105,7 +107,8 @@ var View = Backbone.View.extend({
 });
 
 _.extend(View.prototype, ViewMixin);
-
 _.extend(View.prototype, RegionsMixin);
+_.extend(View.prototype, BehaviorsMixin);
```
paulfalgout: So I initially was all for listing every mixin.. but I think I'm swinging back the other way.  ViewMixin will not work without BehaviorsMixin and UIMixin.. so I think they should be mixed into it, and they don't need to be here.  ViewMixin may have more stuff taken out of it, but it is a mixin that composes others.  And that doesn't keep us from using the mixins by themselves like we're doing with UIMixin on behavior.

gulp/browser-test.js
paulfalgout: cool worth investigating.. should revert this back though for this PR.

src/view.js
jasonLaster: I'm fine w that. Was debating it

PR #2841, opened 2015-12-01
Move serializeModel out of viewmixin

paulfalgout:
Now that @jasonLaster started cleaning up the `ViewMixin`, I noticed `serializeModel` felt out of place there.  it makes more sense in context of `View`.. and the only reason it was in the base was to share it with `CompositeView` since that is deprecated, it is cleaner to go ahead and move it to the two use-cases.


jasonLaster:
:+1:

PR #2842, opened 2015-12-01
Move View specific functionality to view

paulfalgout:
Related to https://github.com/marionettejs/backbone.marionette/pull/2841

Seems like there are some very View specific things that are in the shared to support `CompositeView`  This moves all of that to the `View` and then to avoid further duplication, mixes in the methods from `View` into `CompositeView`

### DIFF
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -8,6 +8,7 @@ import getValue        from './utils/getValue';
 import getOption       from './utils/getOption';
 import MarionetteError from './error';
 import CollectionView  from './collection-view';
+import View            from './view';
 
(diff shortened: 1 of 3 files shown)


jasonLaster:
:+1: I like two things about this:
(1) `ViewMixin` becomes simpler
(2) `View` becomes easier to reason about

I actually expect, that slowly going through some of these remaining shared methods will help us learn something about Marionette. As a sidenote, I was going to submit the massive PR w/ mixins, but I prefer this approach where we pluck a couple researched methods at a time.

jasonLaster:
obvious win.

PR #2844, opened 2015-12-01
Make normalizeMethods util more consistent

paulfalgout:
normalizeMethods is currently attached to Marionette, but as it uses `this` internally it isn't super useful.  Most of our utilities that are attached to Marionette get the context passed in, and for the local use cases we proxy the method.
In this case I believe we've attached normalizeMethods to things mostly for internal use, and that it isn't super useful day to day.  If you do need it, it's a very simple utility to employ with the target.
This reduces the overall API and I believe makes the utility more useful globally.


jasonLaster:
:+1:

paulfalgout:
I may have something better... TBD

PR #2845, opened 2015-12-01
Implement childTriggers (#2404)

ahumphreys87:


megawac:
As #2818, I would personally prefer `childTriggers` get implemented in 2.5 as a transitional step to 3.0 in order to remove `childEvents` for performance reasons. I'm fine with 2.5 being skipped but I would really like to see childEvents (at least their current implementation) dropped or changed

paulfalgout:
I pretty much agree with removing all implementations of childEvents and using childTriggers as an explicit way to proxy, but I don't think we can drop it for v3.  We're going to need time for childTriggers to be around as a viable alternative.  However maybe we could put childEvents in it's current state on a feature flag that is turned on by default?  Also.. I haven't looked into this thoroughly yet, but this will work for both view and collectionview?

paulfalgout:
@ahumphreys87 this is only partially implemented.  There were 2 commits on the other PR. https://github.com/megawac/backbone.marionette/commit/be9cc161caf77e35c2d525b39411d7409e35ca2c

ahumphreys87:
@paulfalgout I dont think the 2nd commit is needed now as `.rest` was removed with the ES6 work

paulfalgout:
The previous implementation on the 2nd commit appended the childview to the first element of args.  That is not occurring here.  But that 2nd commit was also reducing the concats which could also be done here.
```diff
-var callArgs = [this].concat(args);

+_triggerMethod(layoutView, prefixedEventName, this, ...args);
-_triggerMethod(layoutView, [prefixedEventName].concat(callArgs));

-normalizedChildEvents[eventName].apply(layoutView, callArgs);
+normalizedChildEvents[eventName].call(layoutView, this, ...args);

-layoutView.triggerMethod.apply(layoutView, [childTriggers[eventName]].concat(args));
+_triggerMethod(layoutView, childTriggers[eventName], this, ...args);
```

ahumphreys87:
@paulfalgout @jasonLaster :dancer: :dancer:

ahumphreys87:
urgh sorry @paulfalgout my force wiped out your comments :(

I have removed `triggerMethodOn`.
Regarding the context, the test checks that the method was called on `this.superView` so we can use `this` within that. The method is also passed the `layoutView` and the `childView` so we have all 3 view objects accessible in the method.

jasonLaster:
:+1:

paulfalgout:
the issue is that the triggerMethod in the test is passing the childView.  I think it should automatically append this as it does with `childViewEvents`  Otherwise you have to design every view as if it could potentially be a child of something that needs it's events in the future

ahumphreys87:
I think in this case the test is slightly confusing as its starting 2 levels deep. `this.layoutView` is actually the child in this case. `this.childView` is a child of that one - so without passing it up there aint a way to access it.

In the code https://github.com/ahumphreys87/backbone.marionette/blob/childTriggers/src/mixins/view.js#L243-L245 `layoutView` actually maps to `this.superView` in the tests.

paulfalgout:
I guess what I do not understand is why we aren't passing the childView as part of the `childViewTriggers` implementation.  Is there a reason we wouldn't do this all the time?

We do it for prefixed child events:
`layoutView.triggerMethod(prefixedEventName, `**this**`, ...args);`

We do it for `childViewEvents`:
`normalizedChildEvents[eventName].call(layoutView, `**this**`, ...args);`

For instance let's say you have a layout with 2 regions both containing `CollectionView`s.

If you, on your layout add:
```js
childViewTriggers: {
  'remove:empty': 'child:remove:empty'
},
onChildRemoveEmpty: function(){
  //There is absolutely no way for me to have any idea which collection view this occurred in.
}

```

megawac:
@paulfalgout I considered that originally, however it didn't seem useful. Do you really want to have all the views the event gets proxied through on the arguments?

Right now it the source view is the first argument and the view the event listener is on is the `this` context

paulfalgout:
It seems inconsistent that the handlers for `events` and `triggers` would have the same arguments, but `childViewEvents` would contain the child and `childViewTriggers` would not.  I agree that it gets messier when you're proxying multiple layers down the line.

Perhaps a better name for this functionality is `childViewProxy`?

paulfalgout:
One possible alternative I could see here for v3 would be to change our use internally of `triggerMethod` such that we always pass the triggering View `this.triggerMethod('event', this, 'foo', 'bar');`

megawac:
Aren't we planning to remove `childViewEvents` in any case

megawac:
I would be strongly against adding the current view to the arguments passed
through the triggers. It would make this feature very annoying to use when
proxying multiple levels

paulfalgout:
I'm personally more ok with `childViewEvents` than with all of the unlistened `childview:before:render` `childview:render`s being thrown.

I think this works for me if we stop passing the childview as the 1st argument elsewhere.. proxied events and `childViewEvents` then as a convention anytime we `triggerMethod` internally the 1st arg is itself.  As it is the args now for _most_ of our internal events are
`onChildviewRender: function(childView, childView){}`

ahumphreys87:
The thing is the view is passed in args, not sure how! Hehe

If we added 'this' to the call there would be 2 passed. The test looks for
both the child and the child of that child (this.childView)

paulfalgout:
It really doesn't though.
`childViewEvents` does pass the childView.  Here's the relevant parts of the `childViewEvents` test:
```js
this.layoutView.showChildView('child', this.childView);
this.childView.triggerMethod('boom', 'foo', 'bar');

.to.have.been.calledWith(this.childView, 'foo', 'bar')
```
And here is the `childViewTriggers` test:
```js
this.layoutView.showChildView('child', this.childView);
this.childView.triggerMethod('whack', this.childView, 'foo', 'bar');

.to.have.been.calledWith(this.layoutView, this.childView, 'foo', 'bar')
```
And you can see the difference is that in the case of `childViewTriggers` inside the _test_ we're passing the `childView` it isn't part of the implementation

paulfalgout:
So `this.layoutView.onRattle`'s arguments would be `this.childView, 'foo', 'bar'` but _only_ because the triggerMethod that was originally called passed `this.childView`
This might be best, but it should likely be consistent everywhere.
If we made it consistent, and then consistently made sure all view triggers passed the view as the 1st argument we could avoid weird stuff like https://github.com/marionettejs/backbone.marionette/blob/next/test/unit/collection-view.spec.js#L1027

ahumphreys87:
But in that test layoutView is the child, childView is a child of that one.
The test shows an event triggered on childView, LayoutView then triggers an
event which is caught by the childViewTriggers on superView

ahumphreys87:
childView is just a param, it is a view further down the tree that doesn't
really need to be passed

paulfalgout:
I think I need to talk about a simpler bit of code to get on the same page here.
```js
this.LayoutView = Marionette.View.extend({
  template: _.template('<div class="child"></div><div class="child2"></div>'),
  regions: {
    'child': '.child'
  },
  childViewEventPrefix: 'autoproxy',
  childViewEvents: {
    'whack': 'onWhackEvents'
  },
  childViewTriggers: {
    'whack': 'whack:triggered'
  }
});
this.childView = new Marionette.ItemView({ template: false });
this.layoutView.render();
this.layoutView.showChildView('child', this.childView);

this.layoutView.onWhackEvents = function(){
  console.log(arguments); // [childView, 'foo', 'bar']
};
this.layoutView.onAutoproxyWhack = function(){
  console.log(arguments); // [childView, 'foo', 'bar']
};
this.layoutView.onWhackTriggered = function(){
  console.log(arguments); // ['foo', 'bar']
};

this.childView.triggerMethod('whack', 'foo', 'bar');
```
This is the current implementation.

paulfalgout:
So as is the case in the current tests... when you need to pass down `this.childView` so that the "super" knows about the eventing view, now the intermediate view's auto proxied event is going to look like
`layoutView.onChildviewWhack = function(childView, childView, foo, bar){};`
So leaving it off in one implementation to make it cleaner, makes a different feature worse.

megawac:
Passing the event origin source/initiator is common in backbone and in the
DOM. In a Backbone Collection for example, one can listen to events on its
children; the source model initiating an event will be passed as the first
argument. The DOM has event bubbling which lets one listen to events on
some element and if the event occurs on any nested children the event will
bubble up; this event will have `this` set to the target element.

Anyway, I think being able to access the view initiating the event may be
important and should be left in. Getting the views the event is proxied
through is not useful in my opinion and shouldn't be included

paulfalgout:
@megawac I agree that we should be attaching the initiator, I am just suggesting that we should not be overriding the triggerMethod to auto attach it in 2 out of 3 situations.  We should just make sure to attach it via any `triggerMethod` we call within the views.  This is actually something we do pretty consistently already.. So _many_ of the events people use get the childView twice as it is.
```js
this.LayoutView = Marionette.View.extend({
  // auto-proxy
  onChildViewRender: function(childView, childView){},

  childViewEvents: { 'render': 'onChildRender' },
  onChildRender: function(childView, childView){},

  childViewTriggers: { 'render': 'child:rendered' },
  onChildRendered: function(childView){}
});
```
So I don't know why we need to append the childView inside of triggerMethod.

paulfalgout:
I'm going to :+1:  this and we can continue the initiator attaching over on https://github.com/marionettejs/backbone.marionette/issues/2855

Issue #2814, opened 2015-11-08
Make _getValue public.


paulfalgout:
On #2793  @msiconolfi asked why `_getValue` was private.  I don't really have an answer for that.
It was requested here: https://github.com/marionettejs/backbone.marionette/pull/2136#issuecomment-67424178

@samccone @marionettejs/marionette-core what was the reason this was privatized?  Seems to me like it could be useful publicly.

msiconolfi:
Also, I noticed the [comment](https://github.com/marionettejs/backbone.marionette/blob/master/src/helpers.js#L47-L48) before the `Marionette._getValue` definition should probably remove the line `if the value is undefined return a default value`.

Based on the conversation in  [#2136](https://github.com/marionettejs/backbone.marionette/pull/2136#issuecomment-67424178) made by @megawac *"I couldn't decide whether there should be a defaultVal or an args param, I went with the args param as it seemed more useful internally."* 

It looks like this [comment](https://github.com/marionettejs/backbone.marionette/blob/master/src/helpers.js#L47-L48) never got updated after the decision was made to have an args param.

jasonLaster:
nice. lets do this

samccone:
I think the initial reason why we did this was because we were not 100% sure of the API and keeping this private let us iterate on the API... it also kept the overall api namespace down, but if this is something that we want to expose to others then :+1: :ship:

msiconolfi:
I can take this on. Are we simply just removing the underscore and calling it `getValue` and replacing all invocations with the new name?

samccone:
yep @msiconolfi :) you will also want to document the method

jasonLaster:
Yep. 

The other part too make it public would be to add it to the Marionette namespace in backbone.marioette.js and of course the docs.

paulfalgout:
Also look at doing this off of the `next` branch

PR #2847, opened 2015-12-03
Standardize utils and remove old proxies.

paulfalgout:
This is a lot.. not sure how to break it up yet, but this PR is overkill.. and there's some tests failing I need to check into..

But essentially this was looking at how all of the various utils and mixed in functions could be made more similar.

There's a couple of rabbit trails related to similar consistency concerns.. but I just wanted to open a PR to see if this was a reasonable direction before I fix it.

### DIFF
--- a/src/mixins/behaviors.js
+++ b/src/mixins/behaviors.js
@@ -1,7 +1,6 @@
 import _                        from 'underscore';
-import _getValue                from '../utils/_getValue';
-import getOption                from '../utils/getOption';
-import { _triggerMethod }       from '../trigger-method';
+import getValue                 from '../utils/getValue';
+import { triggerMethod }        from '../trigger-method';
 import Marionette               from '../backbone.marionette';
 
 // Takes care of getting the behavior class
@@ -19,7 +18,7 @@ function getBehaviorClass(options, key) {
   }
 
   // behaviorsLookup can be either a flat object or a method
-  return _getValue(Marionette.Behaviors.behaviorsLookup, this, [options, key])[key];
+  return getValue(Marionette.Behaviors.behaviorsLookup, options, key)[key];
 }
 
 // Iterate over the behaviors object, for each behavior
@@ -39,7 +38,7 @@ function parseBehaviors(view, behaviors) {
 
 export default {
   _initBehaviors: function() {
-    var behaviors = _getValue(this.getOption('behaviors'), this);
+    var behaviors = this.getValue(this.getOption('behaviors'));
 
     // Behaviors defined on a view can be a flat object literal
     // or it can be a function that returns an object.
@@ -87,12 +86,11 @@ export default {
     _.invoke(this._behaviors, 'unbindUIElements');
   },
 
-  _triggerEventOnBehaviors: function(args) {
-    var triggerMethod = _triggerMethod;
+  _triggerEventOnBehaviors: function(...args) {
     var behaviors = this._behaviors;
     // Use good ol' for as this is a very hot function
     for (var i = 0, length = behaviors && behaviors.length; i < length; i++) {
-      triggerMethod(behaviors[i], args);
+      triggerMethod.call(behaviors[i], ...args);
     }
   }
 };
--- a/src/mixins/view.js
+++ b/src/mixins/view.js
@@ -3,20 +3,12 @@
 
 import Backbone                 from 'backbone';
 import _                        from 'underscore';
-import _getValue                from '../utils/_getValue';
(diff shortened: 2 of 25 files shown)


paulfalgout:
Ok.  I think this at a good state.

It introduces 2 new utils for internal use 
- `_setOptions` which is for the constructors and merges and attaches options to the object.. in a couple of cases we were just `this.options = options` so that we could use a `this.getOption` prior to a `Marionette.Object.call`, but that actually breaks the ability to predefine options on the definition of those objects.
- `proxy` which essentially relieves the need for the various proxy functions but still allows for the generic util attached to `Marionette` whose first argument is the target.

Additionally I've made `getValue` public and have changed the API so that it works like the others and expects to be added to a prototype or `proxy`'d on to `Marionette`.  Also it passes any addition ...args to the method passed to it, where-as it used to take an array of the arguments to pass to the method.

It also adds a CommonMixin for the utils common to every Mn class.

And it simplifies `triggerMethod` in ways doomed to conflict with https://github.com/marionettejs/backbone.marionette/pull/2845

@marionettejs/marionette-core :eyes:   This one probably shouldn't hang around too long.. Lemme know if adjustments need to be made.

paulfalgout:
This'll resolve https://github.com/marionettejs/backbone.marionette/issues/2814

jasonLaster:
Great deep dive. Utils look 10x better. Every class is so much easier on the eyes. 

:thumbsup:

I'll merge today if there are no objections. This is big, but there are only a couple key ideas applied at scale.

paulfalgout:
@jasonLaster FWIW it might be more interesting now to reopen the idea of `getOptionValue` some of the  consolidating of styles made it such that there are more places where it'd be useful.. 
Even `triggerMethod` could become:
```js
function triggerMethod(event, ...args) {
  // get the method name from the event name
  var methodName = 'on' + event.replace(splitter, getEventName);
  var result = getOptionValue.call(this, methodName, ...args);
  
  // trigger the event
  this.trigger(event, ...args);

  return result;
}
```
That _might_ change the return value of `triggerMethod`, in some strange cases if you were to have an `onEvent` defined that isn't a function, but otherwise functionally the same.

Also I haven't analyzed the output at all as far as what babel transpiles to.. how concerned, or with what functions should we be concerned about the output here?

paulfalgout:
Checking into how some of the spreads transpile.. and hmmm :disappointed:   

https://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=var%20method%20%3D%20function()%7B%7D%3B%0Afunction%20triggerMethod(...args)%7B%0Amethod.call(this%2C%20...args)%3B%0A%7D

megawac:
Yeah, if we're bumping the version range of _ to 1.8/3.* we could also use
restParam

jasonLaster:
@paulfalgout, this is a great win for the readability of all of our common classes. Also, it'll make it much easier to do `getOptionValue` or other small utils.

Not so worried about the babel splat code. The way I see it, if we cared enough we could probably override that function with an `_.toArray(arguments)` call `babel-plugin-arguments-to-array-ftw`

megawac:
@jasonlaster if we decide we care we should just replace splats with
https://lodash.com/docs#restParam

src/mixins/behaviors.js
```diff
@@ -87,12 +86,11 @@ export default {
...
   },
 
-  _triggerEventOnBehaviors: function(args) {
-    var triggerMethod = _triggerMethod;
+  _triggerEventOnBehaviors: function(...args) {
     var behaviors = this._behaviors;
     // Use good ol' for as this is a very hot function
     for (var i = 0, length = behaviors && behaviors.length; i < length; i++) {
-      triggerMethod(behaviors[i], args);
+      triggerMethod.call(behaviors[i], ...args);
```
megawac: This seems very, very weird. Should this not be `triggerMethod.apply(behaviours[i], args)`

src/mixins/view.js
```diff
@@ -236,38 +228,37 @@ export default {
 
   // import the `triggerMethod` to trigger events with corresponding
   // methods if the method exists
-  triggerMethod: function(eventName, ...args) {
-    var ret = _triggerMethod(this, arguments);
+  triggerMethod: function(...args) {
+    var ret = triggerMethod.call(this, ...args);
```
megawac: Same here, you want `triggerMethod.apply(this, arguments)`

megawac:
Looks good, just remove uses of applied splats with `.call`

Issue #2849, opened 2015-12-07
Remove spread calls in next


paulfalgout:
https://github.com/marionettejs/backbone.marionette/pull/2847#issuecomment-162008534

jasonLaster:
all spreads, or just the one's w/ apply?

paulfalgout:
just the one's w/ `call`

jasonLaster:
yup. cool

paulfalgout:
https://github.com/marionettejs/backbone.marionette/blob/next/src/utils/getValue.js#L9

ahumphreys87:
fixed via #2864

Issue #2846, opened 2015-12-03
Region creation too slow due to jQuery


jiaweihli:
We're using Marionette for a medium-size codebase (~100 views) with heavy nesting of `LayoutViews`.  One perf issue we noticed was how expensive region creation was - nearly 5ms went to element lookup.  When a single action can create 10 regions, this already costs 50ms.  We're trying to keep app latency per action below 100ms to avoid a poor user experience, and this was killing our budget.

We reduced element lookup time to 1ms by using [qwery](https://github.com/ded/qwery) v3.4.2:

``` typescript
class QweryLookupRegion extends Marionette.Region {
  getEl(el: string): JQuery {
    // :UPDATE: (marionette) We patched the implementation to use `qwery` instead of `$.find`.
    // This reduces lookup time from ~5ms to 1ms.
    const jquery: JQuery =  Backbone.$(qwery(el, Marionette['_getValue'](this['options'].parentEl, this)));
    // :KLUDGE: Since we're using a `Node` and not a string in constructing the JQuery object, patch
    // the object to make it seem like we used a string.  (Marionette expects this.)
    jquery.selector = el;

    return jquery;
  }
}

class QweryLayoutView<T extends Backbone.Model> extends Marionette.LayoutView<T> {
  get regionClass(): { new(...args: Array<any>): Marionette.Region } { return QweryLookupRegion; }
}

export default QweryLayoutView;
```

qwery v4 removed the documentation describing which selectors it doesn't support.  Might be worthwhile to ask @ded if it's fully compatible / if perf is similar to v3.  

Food for thought, in case anyone else runs into the same issue.

paulfalgout:
oh super interesting  cc: @ianmstew 

There's has been some discussion of what it would look like to reduce/remove dependency on jQuery, but it won't make v3.

That said, have you seen Backbone.NativeView?  Particularly: https://github.com/akre54/Backbone.NativeView/blob/master/backbone.nativeview.js#L76 

that's a fairly similar functionality with no external dependencies.. not sure speed-wise how that might compare with qwery.

jiaweihli:
I hadn't heard of this until now.  It should give perf similar to qwery v4, which is also built on top of `querySelectorAll`.

paulfalgout:
I think this can be considered closed since the DOM mixin will allow you to swap it out

PR #2850, opened 2015-12-08
View events refactor / Region and CollectionView cleanup

ianmstew:
- [x] Fix broken tests (113 of 113 fixed)
- [x] Remove `triggerAttach/Detach` options from Region
- [x] Add [before:]detach to View destroy()
- [x] Add [before:]detach to CollectionView
- [x] [before:]detach tests
- [x] Tests to enforce that _isRendered is set before onRender and _isAttached is set before onAttach
- [x] Squash and rebase against `next`

Region#show
1. before:show moved to beginning of method
2. before:show/show no longer triggered on view
3. Added triggerDetach option and [before:]detach events
4. Attach/detach events are triggered on children, not on all descendants
5. triggerBeforeAttach option removed, triggerAttach option now affects both (same for detach)
6. Merged shouldDestroyView() into show()
7. Merged renderView() into show()
9. Attach events no longer triggered on view with region as the second argument
10. [before:]show events have arguments {region, view, options} instead of {view, region, options}
11. Removed triggerAttach option; it is now only a class option
12. Removed swap events
13. Removed forceShow option; it isn ow only a class option

Region
1. Removed attachView() (no longer necessary because of conditional rendering in onShow())
2. [before:]empty events have arguments {region, view, options} instead of {view, region, options}
3. onRender is the only officially supported callback for rendering children (this could be a discussion, but after some deep research, supporting onAttach for child view rendering was not practical)

MonitorDOMRefresh => MonitorViewEvents
1. Converted onShow handler to onAttach and _isShown flag to _isAttached
2. dom:refresh is now render + attach
3. Added onDetach check
4. Renamed to MonitorViewEvents
5. Added propagation for [before:]attach/detach events
6. Removed "is view element attached" check now that we're tracking _isAttached

CollectionView
1. Removed before:show/show triggering on view
2. Attach events only triggered on children, not on all descendants
3. Added [before:]detach events
4. Relying on MonitorViewEvents to handle propagation of events to child views
5. Made initRenderBuffer() private
6. Made startBuffering() private
7. Made endBuffering() private
8. Made showCollection() private
9. Made showEmptyView() private
10. Made destroyEmptyView() private
11. Made buildChildView() private
12. Made checkEmpty() private
13. Made destroyChildren() private
14. Made proxyChildEvents() private
15. Made removeChildView() private
16. Made addChild() private
17. Made isBuffering private
18. Events are no longer triggered on children with the CollectionView as the second argument
19. Attach-triggering behavior does not change when shown by a Region with triggerAttach set to false. Attach-triggering is independently set with its own class option triggerAttach.

Other
1. Removed `triggerMethodMany()`

/cc @paulfalgout @jasonLaster


src/region.js
```diff
@@ -42,151 +41,145 @@ var Region = MarionetteObject.extend({
   // the old view being destroyed on show.
   // The `forceShow` option can be used to force a view to be
   // re-rendered if it's already shown in the region.
-  show: function(view, options) {
+  show: function(view, options={}) {
```
paulfalgout: this is going to always pass a `{}` to your `triggerMethod`s

src/region.js
```diff
@@ -42,151 +41,145 @@ var Region = MarionetteObject.extend({
...
     if (!this._ensureElement()) {
       return;
     }
 
-    this._ensureViewIsIntact(view);
```
paulfalgout: any particular reason to inline this method in show?

src/region.js
```diff
@@ -42,151 +41,145 @@ var Region = MarionetteObject.extend({
...
+    }
+    view.render();
+    if (!view.supportsRenderLifecycle) {
+      triggerMethodOn(view, 'render', view);
+    }
 
-      this.triggerMethod('show', view, this, options);
-      triggerMethodOn(view, 'show', view, this, options);
+    // Attach the view
+    if (isAttachedRegion && shouldTriggerAttach) {
```
paulfalgout: I wonder if it'd be possible to move this through 120 to an `attachView`?  This way you could attach an already rendered view to a region, and still get the attach events.

src/region.js
```diff
@@ -200,30 +139,6 @@ var Region = MarionetteObject.extend({
...
-        name: 'ViewDestroyedError',
-        message: 'View (cid: "' + view.cid + '") has already been destroyed and cannot be used.'
-      });
-    }
-  },
-
-  // Override this method to change how the region finds the DOM
-  // element that it manages. Return a jQuery selector object scoped
-  // to a provided parent el or the document if none exists.
-  getEl: function(el) {
```
paulfalgout: we might want to keep this for this use-case: https://github.com/marionettejs/backbone.marionette/issues/2846

Until we can provide a better jQuery alternative

src/region.js
ianmstew: Right!  I was part of that discussion, thanks for catching that.

src/region.js
ianmstew: Only that show() got a good bit shorter, so I unrolled some methods back into show() for ease of following the logic.  However just throwing exceptions isn't core method logic, so I agree with keeping `_ensureViewIsIntact()`.

src/region.js
paulfalgout: Just an additional though.  I don't know that `preventDestroy` without an `attachView` is very useful if using the view in a region again is going to force a render

ianmstew: Now _that_ would actually be a useful case for `attachView`... I'll see what that looks like.

paulfalgout:
I think `checkEmpty` may need to be public.  I think it for overriding?
I'm not sure about swapIn/swapOut getting removed.  We need a solid upgrade plan for people who utilize that

src/region.js
paulfalgout: Also it'd fix like half of the breaking tests ;-P

paulfalgout:
Looks like most of the breaking tests are just looking for privatized functions.. so that's not too bad.

src/collection-view.js
```diff
@@ -249,11 +209,11 @@ var CollectionView = Backbone.View.extend({
 
       // since append moves elements that are already in the DOM,
       // appending the elements will effectively reorder them
-      this._appendReorderedChildren(elsToReorder);
```
paulfalgout: this guy does something different in CompositeView

ianmstew:
At first I was thinking swapIn/swapOut may be redundant, but I agree the added logic of "if changing view" is nice--I'll add back.

A lot more decision-making is delegated to `isEmpty()` in `3.0`, so `isEmpty()` might be all the overriding needed.  Can you see another case for `checkEmpty`?

Awesome about the breaking tests.  I was hoping they weren't too bad :fearful:

src/collection-view.js
ianmstew: My mistake, I actually thought "CompositeView is gone so we don't need it", but it's only _deprecated_.

src/region.js
ianmstew: Great reference.

paulfalgout:
oh yep.. sure enough.. I think I dislike the name "checkEmpty" but I dunno what else it'd be.  all good there. isEmpty was what I was thinking about.

src/region.js
```diff
@@ -321,46 +247,13 @@ var Region = MarionetteObject.extend({
...
-  // no view is present.
-  hasView: function() {
-    return !!this.currentView;
-  },
-
-  // Reset the region by destroying any existing view and
-  // clearing out the cached `$el`. The next time a view
-  // is shown via this region, the region will re-query the
-  // DOM for the region's `el`.
-  reset: function() {
```
paulfalgout: This'll need to change https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/regions.js#L21
Somehow...

ianmstew: Hmm, good find.  Do you think that this is a case for a privatized `reset()`?

ianmstew: Actually, that would be treating leading underscores as "package-private" when most of the time we mean "class private".  We may be calling leading underscore methods from across classes in other cases, but I'm not sure I like the potential for confusion.  Maybe just keep it public.

src/monitor-view-events.js
```diff
@@ -0,0 +1,53 @@
+// DOM Refresh
```
jasonLaster: hmm, maybe this could be moved to `src/utils/monitor-view-events`... it's not a primary class

src/region.js
jasonLaster: hmm, see your thinking but I'd vote on keeping it out

src/region.js
```diff
@@ -42,151 +41,91 @@ var Region = MarionetteObject.extend({
...
-      this.triggerMethod('before:swapOut', changingView, this, options);
-    }
+    const isChangingView = !!this.currentView;
+    const isDifferentView = view !== this.currentView;
+    const isAttachedRegion = isNodeAttached(this.el);
+    const shouldForceShow = !!options.forceShow;
+    const shouldReplaceEl = !!options.replaceElement;
+    const shouldDestroyView = view !== this.currentView && options.preventDestroy;
+    const shouldShowView = isDifferentView || shouldForceShow;
+    const shouldTriggerAttach = options.triggerAttach !== false && this.triggerAttach;
```
jasonLaster: :+1: on const

src/region.js
```diff
@@ -297,7 +212,13 @@ var Region = MarionetteObject.extend({
     delete this.currentView;
 
     if (preventDestroy) {
+      if (shouldTriggerDetach) {
+        triggerMethodOn(view, 'before:detach', view);
+      }
       this.$el.contents().detach();
+      if (shouldTriggerDetach) {
+        triggerMethodOn(view, 'detach', view);
+      }
```
jasonLaster: :+1:

src/region.js
```diff
@@ -321,46 +247,13 @@ var Region = MarionetteObject.extend({
...
-    this.currentView = view;
-    return this;
-  },
-
-  // Checks whether a view is currently present within
-  // the region. Returns `true` if there is and `false` if
-  // no view is present.
-  hasView: function() {
-    return !!this.currentView;
-  },
```
jasonLaster: refresh my memory, what is the reason for deleting `attachView` and `hasView`?

jasonLaster:
@ianmstew this is awesome! Cleaning up our events will be a huge win for the community. I also love how much simpler and more consistent CollectionView is becoming.

Great work, is there any way that I can help?

src/region.js
ianmstew: :+1:

src/monitor-view-events.js
ianmstew: Good point, I'll move it.

paulfalgout: So I'm not sure this is a util either..  Is bind-events and triggerMethod a util?

src/region.js
ianmstew: Well, I don't know of a recommended pattern that uses `attachView`, I wonder if it's more of an anti-pattern to get that low level with a Region.  I already added back `hasView` locally this morning, though--will push soon.  But basically I had a feeling I would err on the side of slightly over-aggressive API reduction, so I wanted your thoughts.  I'm fine adding it back in.

src/monitor-view-events.js
paulfalgout: I think Renderer and TemplateCache could also potentially go into whatever bucket this belongs in

ianmstew:
Thanks, @jasonLaster!  I have some code to push soon, and also will check out your PR.

paulfalgout:
New related thought..  I think `attachView` can go away completely if we did this:
```js
if (!view.isRendered()) {
  if (!view.supportsRenderLifecycle) {
    triggerMethodOn(view, 'before:render', view);
  }
  view.render();
  if (!view.supportsRenderLifecycle) {
    triggerMethodOn(view, 'render', view);
  }
}
```

src/region.js
```diff
@@ -33,155 +33,108 @@ var Region = MarionetteObject.extend({
...
     }
   },
 
-  renderView: function(view, options) {
-    view.render();
-  },
-
-  triggerBeforeAttach: true,
-  triggerAttach: true,
+  _attachView: function(view, options={}) {
```
paulfalgout: I think a public `attachView` would simply be:
```
attachView(view, options){
  view._parent = this;
  this._attachView(view, options);
}
```
The other possible use-case for `attachView` is when you have a view that isn't rendered that you attach to an existing `el`.

src/region.js
```diff
@@ -33,155 +33,108 @@ var Region = MarionetteObject.extend({
...
+    // We need to listen for if a view is destroyed in a way other than through the region.
+    // If this happens we need to remove the reference to the currentView since once a view
+    // has been destroyed we can not reuse it.
+    view.once('destroy', this.empty, this);
 
-      if (attachedRegion && attachOptions.triggerBeforeAttach) {
-        displayedViews = this._displayedViews(view);
-        this._triggerAttach(displayedViews, 'before:');
-      }
+    // Make this region the view's parent.
```
paulfalgout: @ianmstew is there some strange use-case mentioned here with triggering during a render I broke by not re-rendering an already rendered view?

src/region.js
```diff
@@ -33,155 +33,108 @@ var Region = MarionetteObject.extend({
...
-
-    // only replace the region's element with the view's element if explicitly set
-    var _shouldReplaceElement = replaceElement;
+    const { forceShow, replaceElement, preventDestroy, triggerAttach } = options || {};
+    const changingView = this.currentView;
+    const isChangingView = !!changingView;
+    const isDifferentView = view !== this.currentView;
+    const shouldForceShow = !!forceShow;
+    const shouldDestroyView = !preventDestroy;
+    const shouldEmptyRegion = isDifferentView && shouldDestroyView;
```
paulfalgout: hmm this should actually be `isChangingView && isDifferentView && shouldDestroyView`
If you don't have a view to begin with, there's no reason to empty

src/region.js
ianmstew: A view that isn't rendered that you attach to an existing `el`...  Is that a legitimate/common enough need to warrant a public API method?  IMO the _ideal_ `Region#show()` should "do the right thing" when you show any view--including avoiding to render a view that is attached to an existing `el`.  I wonder if there's a way to accomplish the latter?

paulfalgout: Perhaps we need a way to detect or define that a view is rendered when
instantiated.

src/region.js
ianmstew: Fixed locally, will be in next commit.

src/region.js
ianmstew: Hmm, I don't think so... I think the only concern here is that a parent receives the child's render events triggered during a show, but if there is no render there is no need for those events.  I can not think of a case where we are internally tracking a child's render events.

src/region.js
ianmstew: That's a great idea.  `isRendered()` has holes besides this one if it doesn't reflect what intuitively I would use it for--when a view 1. is bound to DOM, and 2. doesn't need to be rendered.

src/region.js
ianmstew: I'll make a new issue, because given that `isRendered()` means the above, Region code should be fine as it stands.

src/region.js
ianmstew: I'll make a new issue, because given that `isRendered()` means the above, Region code should be fine as it stands.

src/region.js
ianmstew: Hmm, so I just noticed that `empty()` has special logic for when a `preventDestroy` is true.  This means we should probably drop the `shouldDestroyView` check from `shouldEmptyRegion`.

src/region.js
```diff
@@ -223,7 +176,6 @@ var Region = MarionetteObject.extend({
     return Backbone.$(el, this.getValue(this.getOption('parentEl')));
   },
 
-  // Replace the region's DOM element with the view's DOM element.
   _replaceEl: function(view) {
     // empty el so we don't save any non-destroyed views
     this.$el.contents().detach();
```
ianmstew: I can't come up with a scenario where there would be an un-destroyed view here.  `empty()` takes care of detaching views when `preventDestroy` is true.  @ahumphreys87 @paulfalgout any ideas why this is necessary?

src/region.js
paulfalgout: Oh interesting..  I wonder if we should just pass the options to the `empty` and not worry about any of the logic about `preventDestroy` or whether or not we should empty..  and just setup a guard in empty..

src/region.js
```diff
@@ -1,19 +1,19 @@
...
-import MarionetteError   from './error';
-import MonitorDOMRefresh from './dom-refresh';
+import _ from 'underscore';
+import Backbone from 'backbone';
+import isNodeAttached from './utils/isNodeAttached';
+import MarionetteObject from './object';
+import MarionetteError from './error';
+import MonitorViewEvents from './monitor-view-events';
+import getValue from './utils/getValue';
 import { triggerMethodOn, triggerMethodMany } from './trigger-method';
```
paulfalgout: looks like you no longer need `triggerMethodMany`

src/region.js
```diff
@@ -288,16 +240,22 @@ var Region = MarionetteObject.extend({
...
+    if (!shouldPreventDestroy) {
+      this._destroyView(options);
     }
     this.triggerMethod('empty', view);
 
     // Remove region pointer to the currentView
     delete this.currentView;
 
-    if (preventDestroy) {
+    if (shouldPreventDestroy) {
```
paulfalgout: I think this, and maybe even the delete above this should go before the `triggerMethod('empty')`
FWIW, this could match show more and look like:
```
if(shouldPreventDestroy){
  this._detachView(view, options);
} else {
  this._destroyView(view, options);
}
```

src/region.js
```diff
@@ -232,7 +184,7 @@ var Region = MarionetteObject.extend({
     // present before replacing
     this._restoreEl();
 
-    var parent = this.el.parentNode;
+    const parent = this.el.parentNode;
 
     parent.replaceChild(view.el, this.el);
     this.replaced = true;
```
paulfalgout: We should probably go ahead and match our styles here and go with `_isReplaced` or something similar..

src/region.js
```diff
@@ -33,155 +33,108 @@ var Region = MarionetteObject.extend({
...
-
-    var showOptions     = options || {};
-    var isDifferentView = view !== this.currentView;
-    var forceShow       = !!showOptions.forceShow;
-    var replaceElement  = !!showOptions.replaceElement;
-
-    // We are only changing the view if there is a current view to change to begin with
-    var changingView = this.currentView;
-    var isChangingView = !!changingView;
+    MonitorViewEvents(view);
```
paulfalgout: I think maybe this should occur either right before or right after the `before:show` gets triggered.  Particularly if the show gets prevented.

src/region.js
paulfalgout: so, if we don't make `attachView` public, this isn't necessary.  Otherwise `attachView` becomes leaky if there's a view in there, right?  But honestly I think we needed `attachView` because show always re-rendered a view.  so I don't think there is a use-case anymore for `attachView`

src/region.js
```diff
@@ -33,155 +33,108 @@ var Region = MarionetteObject.extend({
...
+    const shouldDestroyView = !preventDestroy;
+    const shouldEmptyRegion = isDifferentView && shouldDestroyView;
+    const shouldShowView = isDifferentView || shouldForceShow;
 
     if (isChangingView) {
       this.triggerMethod('before:swapOut', changingView, this, options);
     }
 
-    if (this.currentView && isDifferentView) {
+    if (isChangingView) {
```
paulfalgout: I'm not sure why this is in the show..  it should probably be in the `empty`

src/region.js
```diff
@@ -33,155 +33,108 @@ var Region = MarionetteObject.extend({
...
+    const { forceShow, replaceElement, preventDestroy, triggerAttach } = options || {};
+    const changingView = this.currentView;
+    const isChangingView = !!changingView;
+    const isDifferentView = view !== this.currentView;
+    const shouldForceShow = !!forceShow;
+    const shouldDestroyView = !preventDestroy;
+    const shouldEmptyRegion = isDifferentView && shouldDestroyView;
+    const shouldShowView = isDifferentView || shouldForceShow;
 
     if (isChangingView) {
```
paulfalgout: Is it strange that it is possible for this to be triggered but `before:show` and every event after it not?

However, for the life of me, I can't come up with a use-case for the swap stuff.  Tracing it back it seemed to be a first attempt at `preventDestroy`

src/region.js
ianmstew: Weird, I wasn't getting a linter warning...

ianmstew: I added a linter rule for this to `.eslintrc`

src/region.js
ianmstew: Updated locally.

src/region.js
ianmstew: Fixed locally.

src/region.js
ianmstew: `swap` events have been removed after reasoning that exact conditions for a `swap` can be reproduced with show/empty events and condition on `this.currentView`.

src/region.js
ianmstew: Moved to `empty()`.

src/region.js
ianmstew: Done.  I exposed with `isReplaced()` in keeping with view state methods, but I'm not sure that's essential public information.

ianmstew:
I made several updates after discussion with @p, including:

- Cut logic from `show()`
- Cut `swap` events
-

blikblum:
Hi, i've been following the work on v3 and you are doing a good job.

Just allow me doing an observation about `triggerMethodOnCond`

First, i did not understand what was supposed to do at first sight. After i looked what replaced i got it. triggerMethodOnCondition would make clearer its usage

Second, there's the generated code as seen below. There's extra code to pass the proper arguments.

Personally i think the saved code  is not worth the extra function and the added code for arguments manipulation

```
function triggerMethodOnCond(condition) {
  if (condition) {
    for (var _len = arguments.length, args = Array(_len > 1 ? _len - 1 : 0), _key = 1; _key < _len; _key++) {
      args[_key - 1] = arguments[_key];
    }

    triggerMethodOn.apply(undefined, args);
  }
}
```

src/monitor-view-events.js
jasonLaster: i see renderer and templateCache as more of an object singeleton thing, but yea... shades of gray

src/trigger-method.js
```diff
@@ -44,12 +44,16 @@ function triggerMethod(event, ...args) {
 //
 // e.g. `Marionette.triggerMethodOn(view, 'show')`
 // will trigger a "show" event or invoke onShow the view.
-function triggerMethodOn(context, ...args) {
-  var fnc = _.isFunction(context.triggerMethod) ?
-                context.triggerMethod :
-                triggerMethod;
+export function triggerMethodOn(context, ...args) {
+  var fnc = _.isFunction(context.triggerMethod) ? context.triggerMethod : triggerMethod;
+  return fnc.call(context, ...args);
```
paulfalgout: This should be `fnc.apply(context, args);` per https://github.com/marionettejs/backbone.marionette/issues/2849

ianmstew:
@blikblum

> triggerMethodOnCondition would make clearer its usage

Good point, thanks for that feedback.

> Personally i think the saved code is not worth the extra function and the added code for arguments manipulation

I'm hesitant to make decisions based on compiled code when the point of Babel is to transparently depend on new language features that improve the readability of the code.  However, for methods that called frequently, I'm not opposed to making an exception given a performance profile like jsPerf.  Care to create one?

ianmstew:
@paulfalgout I added `_isRendered` to the prototype.

As far as location, I think setting up all prototype properties at the top of a class definition is clearer than locating them near methods that depend on them.  If a class is getting so large that separating them becomes a problem, then there could be another way to trim things down, such as with mixins, for example.

src/trigger-method.js
ianmstew: Fixed

paulfalgout:
Ah sorry..  Crossed pronouns.   I meant the isReplaced function could move instead of being at the bottom.

ianmstew:
Good call, will do!  I have now fixed 62 of 113 broken tests, and while I was at it I ran into a few more considerations to discuss.  The following breaking changes would make the API much more consistent:

- No longer trigger attach events with region as second argument, to make these more consistent with other view lifecycle events; `onAttach(view)`, not `onAttach(view, region)`.
- As mentioned in the issue description, no more `triggerBeforeAttach` option--triggerAttach affects both `attach` and `before:attach`.  Removing a second tree traversal does not impact the algorithmic runtime--O(2 lg n) === O(lg n), and if you really need to get that granular with options, you're either in the 0.1% (guesstimate) of users or you're doing something wrong.  Removing the extra option cleans up our logic a good bit.
- Swap `before:show/show` arguments: `this, view, options` instead of `view, this, options`.  To follow our pattern elsewhere, the trigger-er should be the first argument, not the second.
- Do no attempt to pass `{triggerAttach: false}` behavior from a Region to a CollectionView.  Both should rely on their own triggerAttach options independently.  I had attempted to influence children's attach-triggering behavior in 2.x, but I no longer think this makes sense and permits hard to follow side effects.  Either a CollectionView or Region triggers attach or it doesn't!
- Remove `triggerAttach` method option from Region#show entirely, for two reasons:
  1. CollectionView does not have a show() method parallel where that option makes sense, so CollectionView only has the class option.
  2. I argue that altering side effects of a method through options is potentially confusing, and altering side effects through _both_ class options and method-level options is certainly confusing.  A method should ideally behave one way every time it's called, and perhaps alter its side effects based on characteristics of the class OR its options should be truly method options, not both.  I want to eliminate the odd overlap between "this is a region that does/does not trigger attach" vs. "this is a single method invocation that does/does not attach, regardless of the behavior set on the class."  I personally see attach-triggering as a characteristic as a region (and more powerful there), so I argue for removing the method option.
- I would also like to move `replaceElement` from a method option to a class option for similar reasons to `triggerAttach`.  "This is a region that replaces elements (characteristic)," _not_ "This is a singular `show()` call that will cause the region to behave fundamentally differently."  Additionally, moving the option to a class option fits our model of configuring a class via constructor options (making a class's behavior more explicit), and allowing users to switch their app to `replaceElement` wholesale by assigning to the Region prototype, `Region.prototype.replaceElement = true`.

I realize this is a lot, so if you all skim this for now perhaps we can discuss at length on our team call.

src/collection-view.js
```diff
@@ -35,67 +36,43 @@ var CollectionView = Backbone.View.extend({
...
     this.delegateEntityEvents();
   },
 
-  // Instead of inserting elements one by one into the page,
-  // it's much more performant to insert elements into a document
-  // fragment and then insert that document fragment into the page
-  initRenderBuffer: function() {
+  // Instead of inserting elements one by one into the page, it's much more performant to insert
+  // elements into a document fragment and then insert that document fragment into the page
+  _resetRenderBuffer: function() {
```
paulfalgout: So this is a better name, although I'm now wondering why we don't just change line 58 to `this._bufferedChildren = [];`

I don't think it needs to be initialized, and everything that needs it is called after `_startBuffering`

src/collection-view.js
paulfalgout: I think putting this back will fix a few tests

jasonLaster:
@ianmstew that was a long post, but i agreed with your arguments. Thanks for the writeup and you have the greenlight from me for those changes.

src/collection-view.js
ianmstew: That's a great point, we're not gaining anything from "resetting" the render buffer in the constructor, or even technically in `_endBuffering()`.

ianmstew:
@jasonLaster Thanks for the quick response, I'll keep moving forward.  I'm sure we'll get more comments on the final code.

src/collection-view.js
ianmstew: My mistake--I added back in the method but not the call.  Fixed.

ianmstew:
Tests fixed.  Moving on to wrap up tasks.

jasonLaster:
:dancer: awesome Ian!

ianmstew:
:D  It's almost there!  Just added `detach` events (pending tests) and removed `triggerAttach/Detach` flags app-wide after discussing with Paul; this made implementing `detach` way easier and without the need for weird hacks.

ianmstew:
@jasonLaster @paulfalgout A couple more changes for consistency:

- We are no longer passing the "source" (trigger-er) as the second argument to any events in Region or CollectionView.
- `triggerMethodMany` no longer passes as the second argument the "source" (trigger-er).
- I actually had to move away from `triggerMethodMany` in CollectionView in order to set `_isAttached` on each child within a loop, guaranteeing that during `onAttach` `isAttached()` will return `true`, and now `triggerMethodMany` is no longer used anywhere in the codebase.  I'm thinking we just remove it at this point, what do you say?
- I killed off my `triggerMethodOnCondition` after realizing bloating our API and adding to the thought cycle necessary to interpret the codebase wasn't worth extrapolating an IF statement.

src/region.js
```diff
@@ -315,42 +236,39 @@ var Region = MarionetteObject.extend({
     if (view.destroy) {
       view.destroy();
     } else {
+      const shouldTriggerDetach = !!view._isAttached;
```
paulfalgout: I think it might be nice stylistically to move everything in this else to a method that helps document the functionality.. like `_destroyBackboneView` or something

src/region.js
```diff
@@ -33,158 +32,76 @@ var Region = MarionetteObject.extend({
...
       triggerMethodOn(view, 'render', view);
     }
   },
 
-  renderView: function(view, options) {
-    view.render();
-  },
+  _attachView(view) {
+    const shouldTriggerAttach = !view._isAttached && isNodeAttached(this.el);
+    const shouldReplaceEl = !!this.replaceElement;
```
paulfalgout: should this be a `getOption`?

That would allow for 
```js
regions: {
  foo: {
    replaceElement: true,
    el: '.some-region'
  }
}
```

src/collection-view.js
```diff
@@ -30,73 +26,54 @@ var CollectionView = Backbone.View.extend({
...
+    this._isBuffering = false;
 
-    if (this._isShown) {
-      triggerMethodMany(this._bufferedChildren, this, 'before:show');
-    }
-    if (canTriggerAttach && this._triggerBeforeAttach) {
-      nestedViews = this._getNestedViews();
-      triggerMethodMany(nestedViews, this, 'before:attach');
+    if (shouldTriggerAttach) {
+      _.each(this._getImmediateChildren(), child => {
```
paulfalgout: any value in caching `_getImmeditateChildren` for both attach triggers?

src/collection-view.js
```diff
@@ -169,19 +120,21 @@ var CollectionView = Backbone.View.extend({
...
 
-  // An efficient rendering used for filtering. Instead of modifying
-  // the whole DOM for the collection view, we are only adding or
-  // removing the related childrenViews.
-  setFilter: function(filter, options = {}) {
-    var viewCanBeRendered = this._isRendered && !this._isDestroyed;
+  // An efficient rendering used for filtering. Instead of modifying the whole DOM for the
+  // collection view, we are only adding or removing the related childrenViews.
+  setFilter(filter, {preventRender} = {}) {
+    const viewCanBeRendered = this._isRendered && !this._isDestroyed;
```
paulfalgout: I know this is pre-existing behavior, but does it seem strange that if the view isn't destroyed and hasn't been rendered yet, that calling `setFilter` will not change `view.filter`?

src/collection-view.js
```diff
@@ -553,12 +482,27 @@ var CollectionView = Backbone.View.extend({
     if (!view.supportsDestroyLifecycle) {
       triggerMethodOn(view, 'before:destroy', view);
     }
+
     // call 'destroy' or 'remove', depending on which is found
     if (view.destroy) {
       view.destroy();
     } else {
+      const shouldTriggerDetach = !!view._isAttached;
```
paulfalgout: I mentioned on region in a very similar spot that it might be nice to move the else logic into a method named something like `_destroyBackboneView` but now that there are 2 places we're doing this, I think a tiny module might be good so that we don't accidentally change the behavior in one place and not the other.

src/mixins/behaviors.js
```diff
@@ -70,12 +70,12 @@ export default {
...
   },
 
-  _destroyBehaviors: function(args) {
+  _destroyBehaviors: function(options) {
     // Call destroy on each behavior after
     // destroying the view.
     // This unbinds event listeners
     // that behaviors have registered for.
-    _.invoke(this._behaviors, 'destroy', args);
+    _.invoke(this._behaviors, 'destroy', options);
```
paulfalgout: Actually I think `args` is more correct than options and this should be `...args`  options is most likely `args[0]`
This is getting called from within `destroy: function(...args) {` and passed args, so those arguments would get spread for the destroy events on the behavior event.

src/mixins/view.js
```diff
@@ -163,11 +167,18 @@ export default {
...
+      this._isAttached = false;
+      this.triggerMethod('detach', this);
+    }
+
     // remove children after the remove to prevent extra paints
     this._removeChildren();
 
     this._destroyBehaviors(args);
 
+    this._isDestroyed = true;
```
paulfalgout: possible we want to make sure the view is destroyed prior to the behaviors?  Not 100% sure, but the behaviors know about the view... don't know if there's ever a use case for the behavior understanding if it is being destroyed on its own or as part of the view being destroyed.  Actually it might be reasonable to put this before `_removeChildren` as well.  if a childView triggers an event during a destroy that the parent is listening to, it may want to know if it is being destroyed or not.

paulfalgout:
Nice!  I am on the side of the fence for removing `triggerMethodMany` if we aren't using it.. at this point if it was going to get used it woulda, and this probably just indicates that the code is easier to understand.  Agree about the conditional triggerMethod as well.

JSteunou:
You rock guys, cheers!

src/region.js
ianmstew: I like the idea of making this self-documenting.  `_destroyView()` is almost 100% `_destroyBackboneView()` anyway, so I renamed `_destroyView()` -> `_destroyBackboneView()` and pulled the decision to call it up to `empty()`.  (Next commit)

src/region.js
ianmstew: Sure, that makes sense.

src/collection-view.js
ianmstew: Yes, definitely.  I don't see any public method calls between the two, so there's no sane reason the immediate children could change.

src/collection-view.js
ianmstew: Great observation.  `filter` should indeed be set no matter what.  I have improved the logic like so:

```js
    const canBeRendered = this._isRendered && !this._isDestroyed;
    const filterChanged = this.filter !== filter;
    const shouldRender = canBeRendered && filterChanged && !preventRender;
    // if shouldRender, render and set this.filter, otherwise set this.filter
```

src/region.js
ianmstew: I actually did one better and moved all "removal" logic to `_removeView`, which is only called if view isn't yet destroyed.

```js
  _removeView(view, {preventDestroy} = {}) {
    const shouldPreventDestroy = !!preventDestroy;

    if (shouldPreventDestroy) {
      this._detachView(view);
      return;
    }

    if (view.destroy) {
      view.destroy();
    } else {
      destroyBackboneView(view);
    }
  },
```

src/collection-view.js
ianmstew: Very nice.  I implemented as such.

src/mixins/behaviors.js
ianmstew: Reverted.  I'm curious, though, why does `destroy()` support variable arguments?

src/mixins/view.js
paulfalgout: related to this.. something I recently noticed.  view triggerMethods get proxied directly on the behavior like `onRender`... this works for everything now except for `onDestroy` `onDestroy` is called for the view and the behavior itself.  Granted what triggered it will be the view or the behavior as the 1st arg.. but it might make a case for knowing that the `onDestroy` of the behavior is occurring during the `onDestroy` of the view..  though I don't know the use case for destroying a behavior apart from the view

ianmstew: Hmm, I will have to think more about this.  Maybe we can discuss on the call.  My first impression is that a view is not fully destroyed until its children and behaviors are gone as well.  If knowledge of "event in progress" is desired, setting a `this.isEventHappening = true` in user code within `onBeforeEvent()` could become a generalized recommendation to our users.

src/region.js
paulfalgout: :+1:

src/mixins/view.js
paulfalgout: I actually this the flag was initially added to prevent looping destroys.. where... something like this would happen:
```js
myLayout.listenTo(myChild, 'destroy', myLayout.destroy);
myLayout.destroy();
// stack overflow
```

ianmstew:
@paulfalgout Removed `triggerMethodMany` locally.  I am pushing the above mentioned changes along with `detach` tests later tonight.

src/mixins/view.js
ianmstew: I see what you're saying... that would be a problem.  I wasn't thinking in terms of side effects during children/behavior destruction, I was thinking in terms of our pattern of triggering events like attach, render, and destroy.  Especially after this PR, we are setting a flag _immediately_ before triggering a corresponding event in every case.  In other words, at the instant we change the state of an object (rendered, destroyed, attached) we also announce that fact with an event.

The question as I see it is, when exactly the view is "destroyed" as a semantic concept?  That could be before or after destroying either children, behaviors, or both.  Attempting to answer that question myself, I would have a hard time considered a view "destroyed" while pieces of it (children, behaviors) are still "alive."  I'd rather leave both flag and fire at the end method because I think the pattern is more important than protecting against side effects of a behavior or child destroy, which is probably a symptom of bad app design.

paulfalgout: I don't disagree.

But we should remove the comments such as:
https://github.com/marionettejs/backbone.marionette/blob/next/src/object.js#L44
which indicate the opposite.

ianmstew: :anguished: Great reference.  I agree, that type of comment indicates we _are_ accounting for destroy side effects.  Maybe we can finalize the call on `_isDestroyed` on the call this week, I just want to make sure we're in agreement.

ianmstew:
The last major hurdle is done--`detach` tests added to `on-attach.spec.js` :smile:  What is left at this point is mostly cleanup.  @paulfalgout thanks for practically pair-programming on this one.  Your :eyes: have been essential.  The task list is updated with what remains.

paulfalgout:
good time. :tada: pretty excited about getting this one in.  It already has my :+1:   I'm good with merging it whenever and dealing with any issue in an RC.

ianmstew:
Cool, I can move the cleanup items over to a separate ticket.  In a PR this large, I would almost say keeping multiple commits is probably better than one massive squashed commit.  What do you think?  There is some tension between digestable diffs and pushing only non-breaking commits.

ianmstew:
@ahumphreys87 @paulfalgout This is as squashed as I want to go right now.  :+1: ?

ahumphreys87:
:+1: :dancer: :dancer: :dancer:

JSteunou:
kudos guys, this is freaking exciting!

ianmstew:
@JSteunou :smile: :tada:

paulovieira:
Magnific! Thank you all for the great work! :+1:

PR #2853, opened 2015-12-08
View events

paulfalgout:
I _think_ the show logic can be broken up to be more clear.  But I probably broke something


paulfalgout:
I keep stomping on my commit comments:   Here's some of my thinking behind the logic changes:
https://github.com/paulfalgout/backbone.marionette/commit/591d35022a2595ef121e746978532a572a653ef7

The top part I think is more important than the bottom part.   I _think_ we weren't cleaning up the destroy listener correctly surrounding `preventDestroy`

ianmstew:
Manually merged into #2850

PR #2855, opened 2015-12-10
child event proxy consistency

paulfalgout:
This is POC following https://github.com/marionettejs/backbone.marionette/pull/2845

* Added `childViewTriggers` to `CollectionView`
* Removed the proxied `childView` from proxied events
* made sure internal `triggerMethod`s passed the view
* moved the `childViewEvents` and `childViewTrigger` hash builds into a cache triggered by `delegateEvents`
* fixed a couple of `.call(...args)`

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -245,7 +245,7 @@ var CollectionView = Backbone.View.extend({
         return !_.contains(elsToReorder, view.el);
       });
 
-      this.triggerMethod('before:reorder');
+      this.triggerMethod('before:reorder', this);
 
       // since append moves elements that are already in the DOM,
       // appending the elements will effectively reorder them
@@ -255,7 +255,7 @@ var CollectionView = Backbone.View.extend({
       _.each(filteredOutViews, this.removeChildView, this);
       this.checkEmpty();
 
-      this.triggerMethod('reorder');
+      this.triggerMethod('reorder', this);
     }
   },
 
@@ -387,9 +387,9 @@ var CollectionView = Backbone.View.extend({
 
       var view = this.buildChildView(model, EmptyView, emptyViewOptions);
 
-      this.triggerMethod('before:render:empty', view);
+      this.triggerMethod('before:render:empty', this, view);
       this._addChildView(view, 0);
-      this.triggerMethod('render:empty', view);
+      this.triggerMethod('render:empty', this, view);
 
       view._parent = this;
     }
@@ -400,12 +400,12 @@ var CollectionView = Backbone.View.extend({
   // rendered empty, and then a child is added to the collection.
   destroyEmptyView: function() {
     if (this._showingEmptyView) {
-      this.triggerMethod('before:remove:empty');
+      this.triggerMethod('before:remove:empty', this);
 
       this.destroyChildren();
       delete this._showingEmptyView;
 
-      this.triggerMethod('remove:empty');
+      this.triggerMethod('remove:empty', this);
     }
   },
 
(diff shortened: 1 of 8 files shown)


paulfalgout:
@ahumphreys87 @megawac @jasonLaster I know there's been a lot of back and forth over on #2845  I don't entirely disagree that `childViewTriggers` shouldn't pass the relevant `childView`, but perhaps we should make them all consistent.  In most places for the internal view events the childView is redundant anyhow.  I think appending that first argument, though sometimes convenient, is just [complicating things](https://github.com/marionettejs/backbone.marionette/blob/next/test/unit/collection-view.spec.js#L1027)

paulfalgout:
:-/ the travis matrix lodash hack is breaking all of the sudden :crying_cat_face: :sob: :pear: 
Normal tests pass

src/mixins/view.js
```diff
@@ -211,6 +212,16 @@ export default {
...
   },
 
+  // Cache `childViewEvents` and `childViewTriggers`
+  _buildEventProxies: function() {
+
+    var childViewEvents = this.getOption('childViewEvents');
+    this._childViewEvents = this.getValue(childViewEvents);
+
+    var childViewTriggers = this.getOption('childViewTriggers');
+    this._childViewTriggers = this.getValue(childViewTriggers);
```
jasonLaster: small knit, but I'd prefer to see `this.getValue(this.getOption('childViewTriggers'))` I think this is nicer, even w/o the util

src/mixins/view.js
```diff
@@ -221,18 +232,21 @@ export default {
...
-    layoutView.triggerMethod(prefixedEventName, this, ...args);
+    layoutView.triggerMethod(prefixedEventName, ...args);
+
+    // use the parent view's childViewEvents handler
+    var childViewEvents = layoutView.normalizeMethods(layoutView._childViewEvents);
 
-    // call the parent view's childViewEvents handler
-    var childViewEvents = layoutView.getOption('childViewEvents');
+    if (!!childViewEvents && _.isFunction(childViewEvents[eventName])) {
+      childViewEvents[eventName].call(layoutView, ...args);
```
jasonLaster: does apply work here?

jasonLaster:
some comments, but :+1:

src/mixins/view.js
paulfalgout: :+1:

src/mixins/view.js
paulfalgout: sure would.

paulfalgout:
The tests should pass after https://github.com/marionettejs/backbone.marionette/pull/2857 is in.

paulfalgout:
Along the same lines as this pr, `view#triggers` should probably build a `triggerMethod` where the argument is simply the initiating view instead of the `args` https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/view.js#L114

The args is hard to explain and doesn't really do anything for us than actually just passing the view along.. if the model or collection is useful, it is always attached to the view.  

Then all view triggerMethods are the same.

jasonLaster:
:+1: things are definitely getting simpler.

src/collection-view.js
```diff
@@ -704,17 +704,26 @@ var CollectionView = Backbone.View.extend({
...
 
-      var childViewEvents = this.normalizeMethods(_.result(this, 'childViewEvents'));
-      var childEventName = prefix + ':' + rootEvent;
+      var childEventName = prefix + ':' + eventName;
+
+      var childViewEvents = this.normalizeMethods(this._childViewEvents);
 
       // call collectionView childViewEvent if defined
-      if (typeof childViewEvents !== 'undefined' && _.isFunction(childViewEvents[rootEvent])) {
-        childViewEvents[rootEvent].call(this, view, ...args);
```
ianmstew: Massive win.

ianmstew:
This PR removes one of those stumbling blocks I keep hitting no matter how much I remind myself...  Keeping trigger arguments consistent up the childEvents proxy chain is a massive win!  Ensuring all framework-triggered events also pass a reference to the triggering object handles the "who triggered this" question, and the developer can then decide for him/herself whether to pass the view or not for custom events.

PR #2862, opened 2015-12-18
[docs-fix] showChildView

tsm91:
As mentioned in http://marionettejs.com/docs/v2.4.4/marionette.layoutview.html#basic-usage reference a region with the `showChildView` method.

### DIFF
--- a/docs/marionette.view.md
+++ b/docs/marionette.view.md
@@ -90,8 +90,8 @@ var LayoutView = Marionette.LayoutView.extend({
      Section: 'section'
    },
    onShow: function() {
-      this.Header.show(new Header());
-      this.Section.show(new Section());
+      this.showChildView('Header', new Header());
+      this.showChildView('Section', new Section());
    }
 });
 ```


paulfalgout:
:+1:

paulfalgout:
Thanks!

tsm91:
yw, one thing to note. On http://marionettejs.com/docs/v2.4.4/marionette.view.html the docs is still the same. I see there is a `compile docs` grunt task in the website's repo that i assume would update http://marionettejs.com/docs/ from this repo's docs folder.

Some kind of git post-merge hook should be introduced here i think. For Example if the title contains [docs-fix] you could run a shell script that goes to website repo, run grunt compile docs and push to remote.

paulfalgout:
I was pretty sure the site rebuilds anytime something is pushed to master..  But I assume the docs are specific to the release tag.
@samccone maybe we need to add an additional @latest or something to the docs that is whatever is on master.  hrmmmm

PR #2868, opened 2015-12-31
Remove Backbone Boilerplate reference

duffn:
The Backbone Boilerplate link was invalid and
the new project no longer follows this pattern.


paulfalgout:
I believe it just moved here: https://github.com/tbranyen/backbone-boilerplate  Though I don't know if it is worth it to list here or not anyhow.

Issue #2680, opened 2015-07-26
Deprecate region.reset and remove in v3


paulfalgout:
Because https://github.com/marionettejs/backbone.marionette/issues/2668 and https://github.com/marionettejs/backbone.marionette/issues/1852

paulfalgout:
https://github.com/jquery/jquery/commit/0ea8c32863af31fb5cfc184e8d513bbae35583e8

@marionettejs/marionette-core I hate adding one more thing, but if we want to support jQuery v3 in a month we need to remove or rebuild reset

paulfalgout:
Is fixing reset as easy as just caching the el from the constructor on `._el` and then running getEl again on it?

ahumphreys87:
fixed with #2869

Issue #2668, opened 2015-07-10
Use of deprecated $.selector does not work with Zepto


PhilippMi:
[Region.reset()](https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L272) uses [$.selector](https://api.jquery.com/selector/) which was deprecated in jQuery 1.7. 

Furthermore, Zepto returns an Array instead of a String when this.$el.selector is invoked. Hence, Region.reset() does not work with Zepto.

I made a little fiddle illustrating the issue:
Zepto-Version: http://jsfiddle.net/PhilippMi/vzoo95so/
jQuery-Version: http://jsfiddle.net/PhilippMi/vzoo95so/5/

IMHO this issue is closely related to #1379, though it provides a different view on the problem.

paulfalgout:
Also see https://github.com/marionettejs/backbone.marionette/issues/1852

JSteunou:
One left https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L197

JSteunou:
Or maybe another, I cannot see what's `definition` https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/regions.js#L92

paulfalgout:
Yeah.. hmm the one in the error.. I suppose the error could also attempt to display the initEl if it's a string.. 
The second one is ok.  it is just bad naming and not related to $.selector.  I believe the use of selector here will be deprecated for removal in v4 as you should simply use "el"  For most people just using a region on a view without customization, they won't see this anyways.

JSteunou:
Should be ok now, #2906 was the only one I found

PR #2869, opened 2016-01-02
Refactor Region#reset and call it in Region#destroy

paulfalgout:
I think this might be as simple as caching the initial value?

I know we had talked about removing it https://github.com/marionettejs/backbone.marionette/issues/1852 https://github.com/marionettejs/backbone.marionette/issues/2680, but it seems like we want that functionality, at least for now for layout.

reset currently breaks non-jQuery and upcoming jQuery v3  https://github.com/marionettejs/backbone.marionette/issues/2668 and https://github.com/marionettejs/backbone.marionette/pull/2850

@ianmstew @marionettejs/marionette-core think this is a solution?

With this I went ahead and resolved https://github.com/marionettejs/backbone.marionette/issues/2379

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -17,7 +17,7 @@ var Region = MarionetteObject.extend({
 
   constructor: function(options) {
     this._setOptions(options);
-    this.el = this.getOption('el');
+    this._initEl = this.el = this.getOption('el');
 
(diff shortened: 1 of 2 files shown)


ahumphreys87:
this gets my :+1:, simple solution and all tests pass

PR #2870, opened 2016-01-02
Radio unbind renames

paulfalgout:
These were incorrectly attached.

We're testing bind/unbind but not after they're attached to something like Object.

This is also true for the entity versions of these functions..

hmmm.


ahumphreys87:
:+1:

rafde:
Not sure why the merge button hasn't been pressed yet.

ahumphreys87:
@rafde we usually like to get approval from 2 team members before merging PRs. For small ones like this though we can make an exception

Issue #2860, opened 2015-12-14
Reorg/Combine mixin extends


paulfalgout:
In the very least we should combine the `_.extend` calls.
But I also think the View mixin should probably have it's mixin dependencies extended into it, instead of adding them to the separate libraries.  Additionally we should attempt to make sure the mixins generally don't have dependencies on each other, unless they are composing mixins together.
This may not always be possible, but it is likely a worth-while goal as far as organization and understanding of the code.

PR #2871, opened 2016-01-02
Various v3 mixin cleanup

paulfalgout:
I can break this up if need be.  This is mainly some cleanup regarding separation of concerns and mixins.  I think ideally a mixin shouldn't be aware of other mixins unless it is composing mixins together

- moved `_getImmediateChildren` to the `View` from the `Region` mixin.  :notes: It was not like the others. :notes: 
- Created a `delegate-entity-events` mixin so that behavior and views share the same functionality
- Created a `triggers` mixin so that behaviors could share the view's functionality.  Now the ensure view error is the only thing keeping a behavior from working with a `Backbone.View`  (we might be able to extract it, but I don't want to step on https://github.com/marionettejs/backbone.marionette/pull/2850 any more than I already am)
- `destroyBehaviors` was not passing all of the destroy arguments correctly
- Made the `ViewMixin` mixin its dependencies instead of the views themselves. Resolves https://github.com/marionettejs/backbone.marionette/issues/2860

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -8,6 +8,8 @@
 
 import _                  from 'underscore';
 import MarionetteObject   from './object';
+import DelegateEntityEventsMixin      from './mixins/delegate-entity-events';
+import TriggersMixin      from './mixins/triggers';
 import UIMixin            from './mixins/ui';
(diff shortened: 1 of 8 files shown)


ahumphreys87:
this lgtm :+1:

PR #2880, opened 2016-01-15
Update devDependencies

ivarni:
I've updated the devDependencies based on the output from
running the `npm outdated` command.

The only dependency I've left alone is on sinon. Going to
any version of sinon >1.10.3 breaks `behaviors.spec.js`
and I was not able to figure out exactly why as my experience
with sinon is very limited.

Wether you choose to merge or ignore this PR, at least now
you know ahead of time that there's an issue with it.

I've left jsdom at 3.1.2 since starting from 4.0.0 it requires 
node to be at least io.js and Travis seems to run an older version.


test/unit/behaviors.spec.js
```diff
@@ -315,7 +315,7 @@ describe('Behaviors', function() {
       });
 
       it('should call onShow when already shown and reset', function() {
-        this.collection.reset([{id:1}, {id: 2}]);
+        this.collection.reset([{id: 1}, {id: 2}]);
```
ivarni: The newest version of jscs complained about the missing space here, and it seems most other singleline objects have the space so I added it.

test/unit/setup/node.js
```diff
@@ -12,15 +12,14 @@ global.sinon = sinon;
...
-  global.document = jsdom('<html><head><script></script></head><body></body></html>', null, {
+  global.document = jsdom('<html><head><script></script></head><body></body></html>', {
     FetchExternalResources   : ['script'],
     ProcessExternalResources : ['script'],
     MutationEvents           : '2.0',
     QuerySelector            : false
   });
 
-  global.window = document.parentWindow;
+  global.window = document.defaultView;
```
ivarni: This was changed in jsdom v4 (https://github.com/tmpvar/jsdom/blob/master/Changelog.md#400)

paulfalgout:
@ivarni Thanks for looking into this!  `v3` on the `next` branch is looking to create a release candidate very soon.  As such I think that represents the future of the tooling and dev dependencies.  Any chance you'd be up for doing this same investigation over there?  I suspect it will be better, but it is quite possible there are some outdated things.  I know for sure we'd solved the jsdom by upping node to 4+.  I don't know about sinon, but I'd like to figure out a way to solve that too if it is an issue over on `next`

ivarni:
@paulfalgout Yes, things seem to be more up to date on `next` but I spot an old version of babel and a few minor packages with newer major versions. I can try to bump them and open a PR. Sinon is updated on `next`.

I suppose it makes sense to close this PR since `patch` and `next` seems to have diverged quite a lot? There doesn't seem like there's any point in merging these changes.

ivarni:
@paulfalgout I've got the tests running with babel 6 but because of [this](https://github.com/jscs-dev/babel-jscs/issues/24) issue it means running with `babel-jscs@3.0.0-beta1`.

I have however not been able to get the build-task to run because [rollup and commonjs modules](https://github.com/rollup/rollup-plugin-babel#configuring-babel) does not play nice. and if I use the preset that excludes `modules-commonjs` the import statements in the gulpfile throws a SyntaxError.

I'm not entirely sure how to handle that.

paulfalgout:
@ivarni I figured out the rollup problem here: https://github.com/marionettejs/backbone.marionette/pull/2890

Maybe now with that we can get other dependencies up?  jsdom in particular just went to v8 and it is suppose to be much improved.

Not sure what else needs updating specifically.. as for jscs, can we get around babel-jscs for now by just sticking to the gulp-jscs one?  I know we won't get additional es6 stuff necessarily, but I think for the most part we're relying on eslint anyhow.

Thoughts?  Still interested in doing some of these updates?

paulfalgout:
@marionettejs/marionette-core with a `2.4.5` being a thing now, anyone opposed to just merging this?

samccone:
lets land it!

paulfalgout:
Great! :+1:

ivarni:
@paulfalgout I can give it a go but I don't seem to have a stellar record with these things so far :) I'll probably have some spare time this weekend so unless anyone else has looked at it the meantime I should be able to have a poke at it. I see no API changes in their changelog but it sounds like they changed quite a bit under the hood.

ivarni:
@paulfalgout I applied the changes from your open PR to my local fork and bumped everything that was reported by running `npm outdated` and both the tests and the build seems to run fine from where I'm sitting so you can probably just amend the already open PR with that.

PR #2881, opened 2016-01-16
Update .travis.yml

jmeas:
Tests against new Nodez


paulfalgout:
unfortunately the old jsdom tests won't work with new nodez right?  FWIW `next` is on stable with the newer jsdom

jmeas:
ah okay. cool. I'll close then

PR #2887, opened 2016-01-27
Fix #2886 : Remove Jquery from bower dependencies

### LINKED ISSUE #2886: Jquery as a devDependency ?
In the `package.json` I was looking for jquery in the dependencies but I surprisingly found it in the devDependencies. It may be a silly question but is it deliberate ?

ThomasCrvsr:


paulfalgout:
:+1: :clap:

PR #2889, opened 2016-02-01
lodash 4 support

megawac:
I think thats all that's needed for this repo (collection-method mixins may need another look). Dependent repos will also need to be looked at (Babysitter, backbone)

Also the travis-sh scripts will need to be added.

Tests still currently fail with this changeset using v4

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -62,9 +62,9 @@ var AppRouter = Backbone.Router.extend({
 
     var routeNames = _.keys(appRoutes).reverse(); // Backbone requires reverted order of routes
 
-    _.each(routeNames, function(route) {
+    _.each(routeNames, route => {
       this._addAppRoute(controller, route, appRoutes[route]);
-    }, this);
+    });
   },
 
   _getController: function() {
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -142,19 +142,19 @@ const CollectionView = Backbone.View.extend({
   // Calculate and apply difference by cid between `models` and `previousModels`.
   _applyModelDeltas(models, previousModels) {
     const currentIds = {};
-    _.each(models, function(model, index) {
+    _.each(models, (model, index) => {
       const addedChildNotExists = !this.children.findByModel(model);
       if (addedChildNotExists) {
         this._onCollectionAdd(model, this.collection, {at: index});
       }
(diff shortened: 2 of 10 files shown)


paulfalgout:
So for collection-method mixins.. do you think there'll be a way to keep support for pre & post v4 or are we just screwed?

This looks good though.

@ianmstew not sure how this might mess with what you're doing over on #2850

megawac:
For the collection mixins we'll be running into the same issues backbone will with using lodash 4 and using lodash in general -- api contracts will be different

paulfalgout:
but so it'll just be on the user to understand what functions are mixed in and what their API based on what `_` they load?  Or are there other issues?

samccone:
question... does this change ever make sense until backbone has 4.0 support?

ps you will want to mod this file in the PR also
https://github.com/marionettejs/backbone.marionette/blob/master/.travis.yml#L20

paulfalgout:
@samccone yeah so we really can't update travis until there's full support for v4 in backbone and the other dependencies.

megawac:
@samccone backbone will support lodash 4 in v1.2.4 (coming soon™)

samccone:
ok, lets xlink that PR here... and add it to the grid --- we will merge when it lands

paulfalgout:
https://github.com/jashkenas/backbone/pull/3921 Changelog for 1.2.4

jasonLaster:
:+1:

paulfalgout:
BB 1.3.2 is out.
This'll need a rebase and likely a check of collectionview and region since quite a few changes went in.  Might as well bump bb for the v3 release.

paulfalgout:
Ok this is GTG I think.  However we can't test lodash 4 yet in the matrix until we update that bash script as well as update the Mn tests and dependencies.

samccone:
xref https://github.com/marionettejs/travis-ci-setup/pull/8

paulfalgout:
This can probably be merged now and we can deal with the lodash compatibility of the dependencies and tests in another issue.  https://travis-ci.org/paulfalgout/backbone.marionette/jobs/115601113  We should just make sure these tests don't reflect anything missed in the library itself.

paulfalgout:
Merging this and opening https://github.com/marionettejs/backbone.marionette/issues/2940

PR #2890, opened 2016-02-02
Update to Babel6

paulfalgout:
Previously I couldn't get the rollup preset to work in the build for v6, but this solved it.


test/setup/setup.js
```diff
@@ -5,7 +5,7 @@ module.exports = function() {
   Backbone.$ = jQuery;
   Backbone.ChildViewContainer = require('backbone.babysitter');
   Backbone.Radio = require('backbone.radio');
-  var Marionette = require('../../src/backbone.marionette');
+  var Marionette = require('../../src/backbone.marionette').default;
```
JSteunou: You can get rid of this by using https://github.com/59naga/babel-plugin-add-module-exports
I was super annoying by this when moving to babel v6

jasonLaster:
:+1:

paulfalgout:
wait on a merge here.. I need to re-try the browser test..  I think it was failing, but I didn't notice because the previous test build was loaded in the browser

paulfalgout:
Ok.  I think this guy is ready to go.

test/setup/setup.js
paulfalgout: I agree it is annoying.. not sure about needing the plugin at the moment.  If it comes up again for some reason, certainly.

ahumphreys87:
:+1:

PR #2901, opened 2016-02-24
Fixed CollectionView::filter example

jvitela:
The example with CollectionView's filter was not working because in the function Marionette.getOption the 'options' property has precedence over the target's own properties.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -985,7 +985,7 @@ and a falsey value if it should not.
   cv.render();
 
   // change the filter
-  cv.filter = function (child, index, collection) {
+  cv.options.filter = function (child, index, collection) {
     return child.get('value') % 2 !== 0;
   };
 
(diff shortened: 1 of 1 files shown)


paulfalgout:
:clap:  Good find!
This might be indicative of a larger issue however.  We're implementing a `setFilter` and I don't know that even it accounts for this use-case.  This is likely a candidate for a `2.4.5` fix, and possibly a change to `v3` @ianmstew ?

docs/marionette.collectionview.md
```diff
@@ -993,9 +993,7 @@ and a falsey value if it should not.
   cv.render();
 
   // remove the filter
-  // note that using `delete cv.filter` will cause the prototype's filter to be used
-  // which may be undesirable
-  cv.filter = null;
+  cv.options.filter = null;
```
maniator: Maybe keep the note?

paulfalgout:
So now that `getOption` isn't used internally, this is no longer an issue for `v3`.

@scott-w should we merge this in the meantime?

scott-w:
I'm happy with that.

PR #2906, opened 2016-03-02
Remove use of deprecated $.selector

JSteunou:
Mostly to initiate a discussion on that the thing.

linked to #2668

What is the nature of `this.el` at this point? Still a String? An Object? An Object with a nice `toString`?

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -112,7 +112,7 @@ const Region = MarionetteObject.extend({
       if (this.getOption('allowMissingEl')) {
         return false;
       } else {
-        throw new MarionetteError('An "el" ' + this.$el.selector + ' must exist in DOM');
+        throw new MarionetteError(`An "el" must exist in DOM for this region ${this.cid}`);
       }
     }
     return true;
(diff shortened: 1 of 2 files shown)


paulfalgout:
unfortunately it could be a string or a dom element.  It may be ok that it doesn't attempt to say what selector was used for the el...  For that matter the region has it's own cid now, so that could used as an identifier.  We should probably be utilizing those cids in errors more consistently.

JSteunou:
Something like that?

paulfalgout:
that works for me.

JSteunou:
squashed

paulfalgout:
This just needs the error's test updated.

JSteunou:
meh tests are failing 

> Error: Cannot find module 'estraverse-fb'

paulfalgout:
This is eslint.  We might need to set that dependency, though this was supposedly going to be fixed today.

paulfalgout:
@JSteunou mind squashing this into a single commit, and I'll get this in.  Thanks!

JSteunou:
aaaand squashed

PR #2908, opened 2016-03-04
Allow to use Marionette with Backbone 1.3

just-boris:
Backbone 1.3 is out! According to [the changelog](https://github.com/jashkenas/backbone/blob/master/index.html#L4304), the new version has only a few fixes and adds two new methods to `Collection`.

It seems that latest Backbone can be safely used with current Marionette.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
     - env: UNDERSCORE=1.4.4 BACKBONE=1.1.0
     - env: UNDERSCORE=1.5 BACKBONE=1.1
     - env: UNDERSCORE=1.7 BACKBONE=1.2.3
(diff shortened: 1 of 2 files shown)


just-boris:
As I understand, to make it work the version should be updated in `backbone.wreqr` and `backbone.babysitter` as well. 

It can send pull requests there as well, if you want.

jridgewell:
Sit on this for a bit, I'm going to be pushing out `v1.3.2` which'll fix my screwed up `v1.3.0` release.

just-boris:
Good to hear that you are aware about that and ready to make these updates.

paulfalgout:
Bb `1.3.2` will be going out in Mn `2.4.5` shortly

Issue #2474, opened 2015-04-08
Official support for other templating engines


paulovieira:
One of the selling points for Backbone (and Marionette) is the independence relatively to the templating engine. Marionette comes with support for [underscore templates](http://underscorejs.org/#template) out of the box, but it's not difficult to use other engines. 

For instance, some time ago I switched to [Nunjucks](https://mozilla.github.io/nunjucks/) (for pythonians around here, it's a port of [Jinja2](http://jinja.pocoo.org/docs/dev/) for javascript). I've been using marionette with nunjucks without any problem. In fact, I cleaned up my Marionette code a lot because nunjucks can automatically handle the final visual details (like joining an array of strings).

All this to say: even though it's not difficult to adapt [`Marionette.Renderer`](http://marionettejs.com/docs/v2.4.1/marionette.renderer.html) to work with other engines, the users still needs to know some internals of both Mn and the engine to make it work. I think we would add great value if that was something readily available and officially supported. I'm not saying Mn should handle other engines out-of-the-box, but we should have official "adapters" (not sure if this is a good term) for other engines. Users should only have to include another module (besides the engine itself) to be able to use another engine immediately.

Here is an example of how I see it working from the perspective of the user:

A simple post model:

``` javascript
var postM = new Backbone.Model({
    title: "an awesome title!",
    body: "lorem ipsum",
    tags: ["politics", "health"]
});
```

The template (nunjucks syntax) could be something like:

``` html
<script type="text/template" id="blog-post">

    <h1>{{ title | upper }}</h1> 

    {{ body }}

    <span>tags: {{ tags | join(',')}}</span>

</script>
```

The view:

``` javascript
var PostIV = Mn.ItemView.extend({
    // here I am using the new long form for the template property
    template: {
        engine: "nunjucks"
        selector: "#blog-post"
    }
});

var postIV = new PostIV({
    model: postM
});

postIV.render();
var html = postIV.$el.html()
```

The `html` variable should be:

``` html
<h1>AN AWESOME TITLE</h1> 
lorem ipsum
<span>tags: politics, health</span>
```

Under the hood, `Marionette.Renderer` would use the nunjucks adapter instead of the (default) underscore adapter. If the user prefers to use underscore templates, as he/she have always been doing, the short form for the template property could be used instead:

``` javascript
var MyView = Mn.ItemView.extend({
    // the short form is the same as we currently use
    template: "#my-selector"
});
```

The template property in the above definition would be equivalent to this:

[...]

Issue #2858, opened 2015-12-11
Reorg util files


paulfalgout:
So we have "classes", we have utils, now mixins, and an unnamed group of things.
Those will be:
- bind-entity-events
- bind-radio-events
- error
- monitor-view-events
- renderer
- template-cache
- trigger-method

I also think possibly `utils/deprecate` belongs in this list.

These things to me are more than a simple util, but they're not classes.
Best name I could come up with is module.  It also seems like each of these things could potentially be pulled out as submodule dependencies and be ok.  I got that idea from this: https://github.com/marionettejs/backbone.marionette/issues/2327

Although I suspect eventually triggerMethod will just be part of the `Marionette.Events` mixin... so there are grey areas for sure.. the two binds could also be mixins..

@marionettejs/marionette-core thoughts?

jmeas:
Sounds like a fine idea to me. One note on the name: the word 'module' carries too much meaning in JavaScript, I think. For instance, each individual file in the Marionette repo is a module now, as we're using ES2015 modules. I'd choose any word other than module.

JSteunou:
tools ? extracts ? pieces ? components ?

jasonLaster:
I think many of these could be in utils bind-entity-events, bind-radio,events, monitor-view-events, triggerMethod. the others could stay in the main folder

paulfalgout:
I'm ok with that too.  My only concern is that the utils dir is getting quite large and, other than `deprecate`, is now comprise of only single function utilities.  This re-org could also possibly wait until we land somewhere on documentation.  How we end up describing things may be even more telling.

ianmstew:
I'm used to a `utils` folder containing a number of utility _types_ and each utilities *type* exports a number of utility methods.  ES6 modules/export syntax makes this really nice.  E.g.,

utils/event-util.js
```
export function MonitorViewEvents() ...
export function triggerMethod() ...
export function triggerMethodOn() ...
```

utils/options-util.js
```
export function getOption() ...
export function getValue() ...
export function mergeOptions() ...
```

paulfalgout:
I like that.  
As far as exporting..  You could be adding this to a Bb.View right?

ianmstew:
Cool, that could help with the size of the utils folder.

Hmm, well we are now monitoring view events for Bb.View from within Region#show and CV#addChild.  Does that remove the need?

paulfalgout:
[...]

Issue #1968, opened 2014-10-04
Put the Renderer method inside View directly.


jamesplease:
We have an Object in Marionette called [Renderer](https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.renderer.js). It is a single method, render, which is where our logic is for rendering templates. The only objects in Marionette that use this logic are the Views. So why not just put it in the View object directly?

This has a few benefits, some more tangible than others:
1. The existence of an extra file for this makes Marionette seem more complicated than it is. By removing that file, both from the src folder and the docs sections, I think we'll be more portraying a more accurate representation of what Marionette is.
2. With fewer things to learn, people will 'get the hang' of Marionette quicker. I remember how long it took for me to read the Renderer doc, and it just kind of hung over my head as some unknown feature.
3. It makes the documentation clearer. Rather than saying Views delegate off to some random object, you can simply say...views have their own render method.
4. For better or worse, it gives users more power. By overriding the render method on a pre-view basis, you could use multiple types of templates in a single app (for...whatever reason).

PR #2911, opened 2016-03-06
custom renderer - issue #2474

paulovieira:
This pull request makes a small change to `_renderTemplate` which allows the user to define a custom renderer. Please see issue #2474 for more details about the main objective.

A concrete example of this new feature (using the [Nunjucks](http://mozilla.github.io/nunjucks/) templating engine with pre-compiled templates) is below:
```js
var MyView = Mn.View.extend({

    template: "something/templates/hello-world.html",

    renderer: function(template, data) {

        if (!template) {
            throw new Marionette.Error({
                name: 'TemplateNotFoundError',
                message: 'Cannot render the template since its false, null or undefined.'
            });
        }

        var output = "";
        try {
            // nunjucks will look for the pre-compiled template at window.nunjucksPrecompiled;
            // more details here: https://mozilla.github.io/nunjucks/api.html#browser-usage
            output = nunjucks.render(template, data);
            return output;
        } catch (err) {
            throw new Marionette.Error({
                name: 'NunjucksError',
                message: err.message
            });
        }
    }
});
```

This example shows a custom renderer for a specific view. However the default renderer (Mn.Renderer.render) could also be easily overridden by defining the custom renderer at `Backbone.View.prototype.renderer`.

By the way, people are doing some interesting templating languages that compile to increment DOM. It would be exciting to use them with Marionette for views that require fast rendering:

https://github.com/google/incremental-dom/blob/master/ECOSYSTEM.md

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -63,8 +63,11 @@ var AppLayoutView = Marionette.LayoutView.extend({
 });
 var layoutView = new AppLayoutView();
 layoutView.render();
+// you can access your region by its name
+layoutView.menu.show(new MenuView());
+// also with the getRegion method
 layoutView.getRegion('menu').show(new MenuView());
-// or with a more concise syntax
+// or in this case reference and show it with the showChildView method
 layoutView.showChildView('content', new MainContentView())
 ```
 


paulfalgout:
note: babel-eslint broke the build here. https://github.com/eslint/eslint/issues/5476  Looks like that'll be fixed shortly and travis'll need to be re-run..  Or we can hard set the eslint dep

paulfalgout:
I really like this, particularly considering the IDOM work being done.. this'll be an easy way to implement it even partially in legacy projects.
Your thoughts on something like this though?
```js
var html = this.renderHtml(template, data);

...
// override this for a custom renderer
renderHtml(template, data) {
  return Renderer.render(template, data, this);
}
```

paulovieira:
Your suggestion makes the code more clean. On the other hand it's 1 more call in the stack when rendering. I like it both ways. Let's hear what other people have to say.

One other suggestion though: get rid of the `Renderer` object and module completely and move the code in `Renderer.render` to this new `renderHtml` method. It would make this new method a nice companion for the `attachElContent` and would make the overall strategy for custom rendering more consistent. 

```js
var View = Backbone.View.extend({

  ...

  // Render a template with data. The `template` parameter is
  // passed to the `TemplateCache` object to retrieve the
  // template function. Override this method to provide your own
  // custom rendering and template handling for all of Marionette.
  renderHtml: function(template, data) {
    if (!template) {
      throw new MarionetteError({
        name: 'TemplateNotFoundError',
        message: 'Cannot render the template since its false, null or undefined.'
      });
    }

    var templateFunc = _.isFunction(template) ? template : TemplateCache.get(template);

    return templateFunc(data);
  },

  attachElContent(html) {
    this.$el.html(html);

    return this;
  }

  ...
});
```

See also: #1968 and #2858

paulovieira:
It seems the eslint has been fixed. Meanwhile this pull request has become outdated but I'll updated it. However I'd also like to have feedback about my last comment: what about get rid of the "Renderer" object and move the html rendering logic to a new "renderHtml" method?

Let's make marionette simpler!

JSteunou:
Getting rid of Renderer.render now will be a breaking change. Maybe it's fine for v3. Just let the people know about it.

paulovieira:
Yes, this pull request is for v3. It is definitely a breaking change. "Renderer" is present in marionette probably since its first day :-)

ahumphreys87:
@paulovieira Im not sure I want to introduce such a big breaking change to v3 at this stage, we have now cut rc1.

A small break would be ok in my opinion but maybe removing the renderer should wait until the next major. We are aiming to do more regular major releases (which wont be as big as 3.0) from this point onwards.

Maybe a compromise would be to mark the Renderer as deprecated in 3.0 and implement @paulfalgout's suggestion? In the docs we could advise that this should be overridden rather than using a custom renderer class as the renderer will be removed in v4?

paulfalgout:
I think that is a pretty sound suggestion.

blikblum:
Until now i never needed to use a custom Renderer. But i'm wondering how would be implemented a global render strategy with the removal of Renderer?
Is there any example of custom Renderer in the wild?
I dont have a strong opinion favorable or against removal.

ahumphreys87:
you make a good point @blikblum. Removing the Renderer would mean users having to do this on every view.

I think im :-1: for removing now

rogovdm:
Currently I'm :-1: 

Not sure how to implement my global strategy without Render. I'm using it for my Hot Reloading module for a better development process.

And in my other project it is used because I have multiple template languages on a client :open_mouth: Useful during migration.

What is an alternative after removing Renderer, any thoughts?

paulfalgout:
I think we can have all of it today, even deprecating the current renderer without removing it right away.

In order to set the renderer everywhere without the current renderer class you'd need to change the prototype of the views.  This certainly isn't a very nice way to go about updating the renderer and it has been [argued against previously](https://github.com/marionettejs/backbone.marionette/issues/1968#issuecomment-57957594).

So let's create a nicer interface for people to do this using the `classProperties` of [`Backbone.View#extend](http://backbonejs.org/#View-extend).

Such that:
```js
Marionette.View = Backbone.View.extend({ 
  /* all our view stuff */ 
  renderHtml(template, data) {
    return Renderer.render(template, data, this);
  }
 }, {
  setRenderer(renderer) {
    _.extend(Marionette.View.prototype, {
      renderHtml: renderer
    });
  }
});
```
This would mean an additional call in the stack for now, however you could quite easily remove that by
```js
Marionette.View.setRenderer(Marionette.Renderer.render);
Marionette.CollectionView.setRenderer(Marionette.Renderer.render);
```

Now while this would require 2 sets to replace it everywhere, I think this is ideal.  It may be very common in the future to want a different renderer for a CollectionView than a View.

In the future after we remove Marionette.Renderer and the default render can just be the `renderHtml` function.  Then the user can override it per view, or they can override it at a library level for View or CollectionView.

So hopefully this will open the door for.
```js
import IDomRender from 'marionette-idom';
Marionette.View.setRenderer(IDomRender);
```

And I think a single function interface off of the View class is simpler to understand and explain than the renderer class.

paulovieira:
@paulfalgout Your approach is the best one so far! It makes a lot of sense to place this in the `classProperties`.

@ahumphreys87, I think you misunderstood the strategy here. People wouldn't have to define a render per view. A default one would always be there (the current one, based on `_.template`).

The main objective is to be able to define custom renderers for specific views, keeping some default renderer for all the other views.

scott-w:
@paulfalgout @paulovieira What's the current status on this PR? I like the ideas behind it - do we try to get it into v3 or shall we put it into a v3.1, assuming this doesn't introduce breaking changes?

paulovieira:
I haven't dedicated time to think about this, but I liked the last approach proposed by @paulfalgout (class properties). But I have some free time this weekend and I'll look at it again. 

Having custom renderers per view would add a lot of value to v3. One use very interesting use case is to use incremental dom templates to get better performance for views that are heavily used.

ahumphreys87:
@paulovieira be great to see this work get merged. The use case of incremental dom is something im particularly interested in.

We've probably missed the cut off for v3 but if we can keep it non-breaking we can push out in a 3.x

PR #2912, opened 2016-03-06
V3.0.0 rc doc

scott-w:
I've written up the `CollectionView` migration examples to migrate users from `CompositeView`. It currently just fits into the existing docs but I think that'll be sufficient for anyone using the v3 RC.

If you guys have anything you'd like to add, or any suggestions, please let me know and I'll get them in as soon as time allows.

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -35,9 +35,11 @@ will provide features such as `onShow` callbacks, etc. Please see
 * [Building `CollectionView`s](#building-collectionviews)
   * [Rendering Lists][#rendering-lists]
   * [Rendering Tables][#rendering-tables]
-    * [Tables Using Marionette 2 CompositeView][#tables-using-marionette-2-compositeview]
+    * [Tables Using Marionette 2][#tables-using-marionette-2]
     * [Tables Using Marionette 3][#tables-using-marionette3]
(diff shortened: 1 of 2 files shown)


docs/marionette.collectionview.md
```diff
@@ -457,6 +465,295 @@ myView.destroyChildren({checkEmpty: false}); // will not show emptyView
...
+  tagName: 'tr',
+  template: '#row-template'
+});
+
+var TableBody = Marionette.CollectionView.extend({
+  tagName: 'tbody',
+  childView: RowView
+});
+
+var TableView = Marionette.View.extend({
```
urosgruber: Isn't this LayoutView?

paulfalgout: Marionette v3 will only have View CollectionView and the deprecated CompositeView.  So you'll be able to add regions to any view.

scott-w: I'll have a look at the `View` documentation and clearly label it there.

I think we're going to need to keep the existing documentation with links to the updated versions of everything at least for the initial versions so ease the transition.

scott-w:
This is ready to merge once the test is passing

paulfalgout:
The failing test is an eslint dependency issue.  We can come back to this if there's a problem.

PR #2917, opened 2016-03-09
Updated view code to better match es6 paradigms

maniator:
Updated the views to have consistency to use `let` instead of `var` or `const`

Next should probably update all of the uses of `const` to have the variables be all CAPS


maniator:
I have no idea how to edit milestone or labels, but this is for 3.0

maniator:
oops. branch off the wrong branch

PR #2918, opened 2016-03-09
Updated view code to better match es6 paradigms

maniator:
Updated the views to have consistency to use let instead of var or const

Next should probably update all of the uses of const to have the variables be all CAPS

(same as #2917 but off the correct branch)

Also de-capitalized `MonitorViewEvents` as it is not a class but a function.

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -9,7 +9,7 @@ import getOption          from './utils/getOption';
 import normalizeMethods   from './utils/normalizeMethods';
 import deprecate          from './utils/deprecate';
 
-import MonitorViewEvents  from './monitor-view-events';
+import monitorViewEvents  from './monitor-view-events';
 import MarionetteObject   from './object';
 import Renderer           from './renderer';
 import TemplateCache      from './template-cache';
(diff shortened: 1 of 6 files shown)


maniator:
Again, I have no idea how to add a label or milestone...

paulfalgout:
const doesn't need to be all uppercase.  It actually does not indicate a constant value but a constant reference.  I personally only think `let` should be used if you're actually setting that value again.  Often times we are not and `const` is more appropriate.

paulfalgout:
Here's a reference for `const` https://mathiasbynens.be/notes/es6-const

paulfalgout:
Thanks a ton for this work!  That odd npm dependency is an eslint related bug they're trying to solve.. I was hoping that'd be taken care of already but, :disappointed: https://github.com/eslint/eslint/issues/5476

maniator:
@paulfalgout Updated as per your comments :-)

paulfalgout:
@maniator awesome!
Can you remove the `package.json` change?  That was fixed here by locking down eslint for now: https://github.com/marionettejs/backbone.marionette/pull/2923

Then if you don't mind squashing the other 2 commits.. Then this LGTM :clap: :+1: :boom: :dancers:

maniator:
@paulfalgout Thanks :-)

All squashed.

I guess merge when ness :-)

(I do not know the process for merging here, but I def would like to help more!)

ahumphreys87:
great work @maniator. Gets my :+1:

@paulfalgout would you like to do the honours?

PR #2922, opened 2016-03-10
fix gulp test-browser runner

elgubenis:
when I ran gulp test-browser it failed in chrome because the local mocha package was missing

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "isparta": "^4.0.0",
     "jquery": "^2.1.4",
     "jsdom": "^8.0.2",
+    "mocha": "^2.4.5",
     "rollup": "^0.25.2",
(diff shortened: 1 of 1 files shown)


paulfalgout:
hmm that's odd.. we're using `gulp-mocha` and it has a dependency on `mocha` so it _should_ have loaded fine without adding it to the `package.json`.  What OS / node version are you using?

paulfalgout:
Ah ha! I know what's up.  In Node 5, npm3 will install the dependencies flat, which is what the runner is expecting.  So for node < 5 this is needed.  :clap:

Since this is next and clearly a problem, I'm going to merge it now @marionettejs/marionette-core

elgubenis:
Cool

jasonLaster:
Nice! Thanks Linus!

PR #2923, opened 2016-03-10
Fix dependencies and html test runner

paulfalgout:
Locked eslint per https://github.com/eslint/eslint/issues/5476

Added two babel dependencies for node < 5 so the test runner can find the files in node_modules


paulfalgout:
this commit has an X for travis because I was quick on the merge I guess, but it passes now: https://travis-ci.org/marionettejs/backbone.marionette/builds/114965543

PR #2924, opened 2016-03-11
Remove extra variable passing from 'isEmpty'  call on collection-view

maniator:
Removed the passing of the collection into the `isEmpty` as it was not used from the parameters.

### DIFF
--- a/test/unit/collection-view.empty-view.spec.js
+++ b/test/unit/collection-view.empty-view.spec.js
@@ -325,10 +325,6 @@ describe('collectionview - emptyView', function() {
       it('should reference each of the rendered view items', function() {
         expect(this.collectionView.children).to.have.lengthOf(1);
       });
-
-      it('should pass the collection as an argument to isEmpty', function() {
-        expect(this.isEmptyStub).to.have.been.calledWith(this.collection);
-      });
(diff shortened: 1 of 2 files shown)


test/unit/collection-view.empty-view.spec.js
```diff
@@ -325,10 +325,6 @@ describe('collectionview - emptyView', function() {
       it('should reference each of the rendered view items', function() {
         expect(this.collectionView.children).to.have.lengthOf(1);
       });
-
-      it('should pass the collection as an argument to isEmpty', function() {
-        expect(this.isEmptyStub).to.have.been.calledWith(this.collection);
-      });
```
maniator: This test is not needed anymore because it would be testing nothing really.

jasonLaster:
:+1:

paulfalgout:
I looked into this a bit to make sure there wasn't some public API reason for this argument.

It is a breaking change, so best to do it now..

Oddly enough this was done before, but somehow didn't get transferred to `next`.  https://github.com/marionettejs/backbone.marionette/pull/1783

I couldn't find any issue or PR putting it back, so I think this is :+1:

PR #2925, opened 2016-03-11
Rewrite the template cache as an es6 module class

maniator:
Rewrite the template cache as an es6 module class

### DIFF
--- a/src/template-cache.js
+++ b/src/template-cache.js
@@ -5,22 +5,27 @@ import _               from 'underscore';
 import Backbone        from 'backbone';
 import MarionetteError from './error';
 
-// Manage templates stored in `<script>` blocks,
-// caching them for faster access.
-var TemplateCache = function(templateId) {
-  this.templateId = templateId;
-};
+class TemplateCache {
+  constructor(templateId) {
+    this.templateId = templateId;
+  }
 
-// TemplateCache object-level methods. Manage the template
-// caches from these method calls instead of creating
-// your own TemplateCache instances
-_.extend(TemplateCache, {
-  templateCaches: {},
+  static get templateCaches() {
+    if (!this._templateCaches) {
+      this._templateCaches = {};
+    }
+
+    return this._templateCaches;
+  }
+
+  static set templateCaches(cache) {
+    this._templateCaches = cache;
+  }
 
   // Get the specified template by id. Either
   // retrieves the cached version, or loads it
   // from the DOM.
-  get: function(templateId, options) {
+  static get(templateId, options) {
     var cachedTemplate = this.templateCaches[templateId];
 
     if (!cachedTemplate) {
@@ -29,7 +34,7 @@ _.extend(TemplateCache, {
     }
 
     return cachedTemplate.load(options);
-  },
+  }
 
   // Clear templates from the cache. If no arguments
   // are specified, clears all templates:
@@ -38,7 +43,7 @@ _.extend(TemplateCache, {
   // If arguments are specified, clears each of the
   // specified templates from the cache:
   // `clear("#t1", "#t2", "...")`
-  clear: function(...args) {
+  static clear(...args) {
     var i;
     var length = args.length;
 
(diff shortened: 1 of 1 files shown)


blikblum:
What is the gain here?
AFAIK the produced code will be bigger with no real benefit.

maniator:
@blikblum Not really much, but it uses the new es6 paradigms, so when we _eventually_ don't have to use babeljs etc, it is smaller and more readable code.

blikblum:
ES6 is a superset of ES5, so by all means,  current code is ES6.

The fact that new tools / semantics are available does not means that it should be used. This is a general concept, that gains more traction in the context of replacement of valid, clean code of a library.

Hope in the future marionette do not need works like  mentioned in https://github.com/mozilla/localForage/pull/538#issuecomment-193016758

paulfalgout:
Yeah, this makes me moderately nervous as well.  If the ES6 isn't making the code much more readable / maintainable I don't know we want to risk the overhead.  I have my hangups in general too with ES6 class.  That said some of the visual wins such as removing ': function` from the objects is nice.  That doesn't really change the transpiled code at all, but is aesthetically nice.

@marionettejs/marionette-core weigh in?

paulfalgout:
@blikblum also related to your concerns, we have been concerned with that too.  This is a nice resource for checking out various module compilation https://github.com/samccone/The-cost-of-transpiling-es2015-in-2016  `v3` is certainly smaller than `v2` was.. And I think we can improve upon that still.

blikblum:
@paulfalgout thanks for the feedback

src/template-cache.js
```diff
@@ -5,22 +5,27 @@ import _               from 'underscore';
...
+  static get templateCaches() {
+    if (!this._templateCaches) {
+      this._templateCaches = {};
+    }
+
+    return this._templateCaches;
+  }
+
+  static set templateCaches(cache) {
+    this._templateCaches = cache;
```
samccone: i feel like since we are using this is the body static might be the wrong type here

maniator: hmmmm yea. maybe I can make the _templateCaches as a local variable to this module.

jmeas:
> That said some of the visual wins such as removing ': function` from the objects is nice. That doesn't really change the transpiled code at all, but is aesthetically nice.

Fwiw – and you may already know this – but this particular change doesn't require classes to do.

```js
var x = {
  someFn() { console.log('hello') }
};

x.someFn();
```

---

If we introduce this (I don't feel strongly either way), then we should probably look at other things in the Mn codebase and see if it makes sense to turn those into Classes, too. The things that extend from Bb will likely continue to work that way (maybe not, though?), but we may have other constructors that we could transition.

maniator:
@jmeas I actually tried that.

It seems that es6 does not allow for extending another class without calling super first in the constructor, but in our code we need to do some setup beforehand so it would not work.

I will create another PR for some of this stuff to remove the class from here and maybe instead use the new function syntax throughout to be consistent,

jmeas:
Hey @maniator ! Sorry, I should have clarified! I think that any other work to make things more consistent could come in future PRs, on top of this one. As a standalone chunk of code, a quick look made this out to be a great little chunk of code to get us toward transitioning toward ES6 classes (I haven't thoroughly reviewed it, but nothing raised any red flags when I checked it out).

Rather than making you do more work right now, I think we could continue the discussion here to see what the best way forward is. Maybe it's merging this PR, or maybe it's making another one.

Does that make sense to you / everyone else?

maniator:
@jmeas Whatever you guys want, just let me know :-)

Ill re-open this PR for now then.

src/template-cache.js
samccone: or just not use "static" :)

paulfalgout:
Not sure if it's mostly the class or the statics or whatever but here's a gist of the babel output from both the old and the new 
https://gist.github.com/paulfalgout/2e0975947408344fa425

I believe the new is `4.3 KB` gzipped and the old is `3.5KB` gzipped.
I know `0.8KB` isn't much by itself, but this is relatively small file and it might add up if we determine we want this style across the src.

src/template-cache.js
maniator: @samccone Then they cannot be used as `TemplateCache.templateCaches`. you would have to "new" up a new one every time.

paulfalgout:
I added an issue to discuss using ES6 classes and I'm going to close this for now to clean up the PR list, but I'm not closing it as a rejection.

Issue #2888, opened 2016-02-01
Behavior event bind error when no handler is specified


happylinks:
Hi,

I spend some time debugging a bug I made when I removed a behavior method but left the event that referenced that handler.

The error is because here:
https://github.com/marionettejs/backbone.marionette/blob/master/src/behaviors.js#L61
```
var handler   = _.isFunction(behaviour) ? behaviour : b[behaviour];
_events[eventKey] = _.bind(handler, b);
```
it isn't checked if the handler is a function with _.isFunction.
If it is not we could return a error message so it's easier to debug.

Shall I make a pullrequest for this? I'm not sure if logging these kind of errors is encouraged.

JSteunou:
Backbone dont with `events` and marionette does not with `modelEvents` and `collectionEvents` but indeed could be a cool thing for debugging. Does [marionette.inspector](https://chrome.google.com/webstore/detail/marionette-inspector/fbgfjlockdhidoaempmjcddibjklhpka) help on this @paulfalgout?

paulfalgout:
So backbone bails with behaviors it looks like https://github.com/jashkenas/backbone/blob/1.2.3/backbone.js#L1278

We should really probably be doing whatever backbone is doing.  I'm not sure if inspector would help with this specifically, but it is an interesting idea.

paulfalgout:
Ah FWIW `v3` aligns with the Bb behavior https://github.com/marionettejs/backbone.marionette/blob/next/src/behavior.js#L102

I do think there might be a `2.4.5`.  @happylinks if you wanted to add the bail ala backbone for the non-existing event I would :+1:

happylinks:
Ok, thanks for checking it. I'll make a pullrequest for it!

paulovieira:
Uhmm, this was actually a nice feature of marionette, i'm sad to see it go. You could save some precious debugging time if marionette tells you right away that the method is missing (a common mistake when the name of method is long).

I don't see much sense in imitating backbone just for the sake of it, especially the questionable parts.

What about add a "DEBUG" flag to the `Marionette` object and warn the user about this kind of problems when it is true? This would align with Radio:

https://github.com/marionettejs/backbone.radio/blob/master/src/backbone.radio.js#l21

Issue #1910, opened 2014-09-18
Thoughts on implementing Marionette Warnings?


MeoMix:
Hiya,

I think it's fair to say that Marionette is incredibly flexible and people use it in a plethora of interesting scenarios. As such, strictly enforcing error conditions isn't always the best idea because it can become a hindrance when a developer has a scenario which is outside of normal use cases. This fact was seen quite recently in #1861.

That said, I think that there should still be the possibility of some hand-holding through a warning system. Perhaps warnings would be opt-in with a debug flag being set.

The first use case which comes to my mind is when refactoring a view. On more than one occasion I have had a UI object with cached references to nothing. For example:

    //  OLD:
    //  Template:
       <div class='buttons'></div>
    //  View:
       ui: { buttons: '.buttons' }

    //  NEW:
    //  Template:
       <div class='buttons-fancy'></div>
    //  View:
       ui: { buttons: '.buttons', fancyButtons: '.buttons-fancy }

It would be nice if Marionette could warn me and let me know that no elements were found for the ui.buttons selector. Maybe it's intended, but maybe it was just old code which was refactored out.

I'm sure there are lots of other scenarios where a little guidance could go a long ways in cleaning up people's code. What do you guys think?

ahumphreys87:
I kinda think we should put together a separate debug library for this - something that you'll only include during development. This way we avoid bloating the main library.

Either that or could this belong in inspector? @jasonLaster what you think?

jasonLaster:
I feel like another lib could go out of date. Maybe inspector maybe inline ala react Dev checks

PR #2928, opened 2016-03-14
Nonexistent event handlers now fail silently

paulfalgout:
Aligns with `Backbone` and `v3`.

Resolves https://github.com/marionettejs/backbone.marionette/issues/2888

### DIFF
--- a/src/behaviors.js
+++ b/src/behaviors.js
@@ -59,7 +59,7 @@ Marionette.Behaviors = (function(Marionette, _) {
 
           var eventKey  = eventName + selector;
           var handler   = _.isFunction(behaviour) ? behaviour : b[behaviour];
-
+          if (!handler) { return; }
           _events[eventKey] = _.bind(handler, b);
         }, this);
 
--- a/test/unit/behaviors.spec.js
+++ b/test/unit/behaviors.spec.js
@@ -156,6 +156,9 @@ describe('Behaviors', function() {
         baz: Marionette.Behavior.extend({
           events: {'click': 'handleClick'},
           handleClick: this.bazClickStub
+        }),
+        qux: Marionette.Behavior.extend({
+          events: {'click': 'noHandler'}
         })
       };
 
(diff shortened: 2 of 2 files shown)


paulovieira:
Uhmm, this was actually a nice feature of marionette, i'm sad to see it go. You could save some precious debugging time if marionette tells you right away that the method is missing (a common mistake when the name of method is long).

I don't see much sense in imitating backbone just for the sake of it, especially the questionable parts.

What about add a "DEBUG" flag to the `Marionette` object and warn the user about this kind of problems when it is true? This would align with Radio:

https://github.com/marionettejs/backbone.radio/blob/master/src/backbone.radio.js#l21

paulfalgout:
For almost any other hash, I think I'd agree.  I'm not sure the use-case for purposefully attempting to bind an event to a nonexistent function and having it `noop`.  Though I do wonder why it is not more consistent.

However I do think it would be fairly awkward in this circumstance considering it is essentially the exact same hash providing the same functionality for the view.

```js
var MyBehavior = Behavior.extend({
  events: function() {
    'click': 'missingFoo' // throw error?
  }
});

var MyBehavior2 = Behavior.extend({
  events: function() {
    return {
      'click': this.view.missingFoo // throw error?
    };
  }
});

var MyView = ItemView.extend({
  behaviors: {
    error: {
      behaviorClass: MyBehavior 
    },
    error2: {
      behaviorClass: MyBehavior2 
    }
  }
  events: {
    'click': 'missingFoo' // no error
  }
});

```

Here's backbone's rationale https://github.com/jashkenas/backbone/pull/2191
*TL;DR*
`$.on('test` 'missingFoo');` doesn't cause an error but `$.trigger('test');` will
The same is with `Backbone.Events.on` and `Backbone.View#events`
The error is thrown when triggered not when defined.

But I'm open to alternatives, and like the DEBUG idea.  I think in this case, this is fixing a bug in `2.4.4` but it'd be great to consider a new debugging feature for `v3`

samccone:
+1 to warn on this one via a console && console.warn ? console.warn() : console.log()

paulfalgout:
related: https://github.com/marionettejs/backbone.marionette/issues/1910

paulovieira:
> Here's backbone's rationale jashkenas/backbone#2191
> 
> TL;DR
> $.on('test' 'missingFoo'); doesn't cause an error but $.trigger('test'); will
> The same is with Backbone.Events.on and Backbone.View#events
> The error is thrown when triggered not when defined.

Ah, interesting. It makes some sense from that point of view.

On the other hand, a more pragmatic approach would have value as well, that is, save developer's time by warning the user right away that something is missing.

Another option to consider if v3 were using webpack instead of gulp would be something like this:

megawac:
Does this make sense in a patch release? I'd rather leave it. Also this needs unit tests, no?

paulfalgout:
I think this PR fixes a bug.  Anything else is potentially a feature for later.  I can certainly add a test for this.

megawac:
I'm = for this inclusion

Issue #2927, opened 2016-03-13
The next build is broken


paulfalgout:
Related https://github.com/rollup/rollup/issues/553

There are two issues with the builds.  

For one the global export from UMD was broken with a rollup update related to the `Mn` alias. I have a work-around, but would prefer to find support from rollup for this.

Second, the bundled build isn't correct. The dependencies need to be built in with rollup most likely.  The interopt functions aren't working since they don't exist yet..  we could potentially get around this by moving the dependencies outside of the UMD bundle.  This might work fine.. though we should consider again just axing the bundled build altogether.

PR #2932, opened 2016-03-14
Fix the v3 build

paulfalgout:
This arguably fixing a hack with a hack.

Issue opened with rollup: https://github.com/rollup/rollup/issues/553

Suggestions welcome @jmeas ?

### DIFF
--- a/gulp/build.js
+++ b/gulp/build.js
@@ -53,7 +53,7 @@ function _generate(bundle, type){
 
   return bundle.generate({
     format: 'umd',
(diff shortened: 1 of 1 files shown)


paulfalgout:
I added a commit to fix the bundled build.

So this resolves https://github.com/marionettejs/backbone.marionette/issues/2927

The bundled build is now an iife, but that isn't a problem right?  That is it's only use case?

https://gist.github.com/paulfalgout/4e3ceaa0318f34c0f3c1

ahumphreys87:
:+1:

PR #2936, opened 2016-03-15
Renderer.render() only accepts two params.

vernondegoede:
`Marionette.View` passes three parameters, where the third one seems unused.

### DIFF
--- a/docs/marionette.view.md
+++ b/docs/marionette.view.md
@@ -2,13 +2,7 @@
 
 # Marionette.View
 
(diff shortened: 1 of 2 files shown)


paulfalgout:
@vernondegoede check this commit https://github.com/marionettejs/backbone.marionette/commit/03159984c06ce5136e9090b958c3720cb8b67fad

paulfalgout:
Looks like that 3rd param has some utility in overriding.  Discussed this with @vernondegoede in gitter and for now I'll close this.

PR #2938, opened 2016-03-16
Allow hyphenated ui strings to bind to events

### LINKED ISSUE #2718: Marionette.View#ui does not work with hyphenated names
The following ui hash will not work:

```javascript
ui: {
    'some-name': '.some-class'
}
```

The hyphen makes it break, silently.

If this is intended behaviour, please put a note in the documentation that hyphenated strings do not work in the UI hash, or have the view constructor throw an Error when this happens, so it does at least not fail silently.

ahumphreys87:
Fixes #2718


paulfalgout:
:+1:

Issue #2933, opened 2016-03-14
Build tests with rollup


paulfalgout:
Possible with https://github.com/eventualbuddha/rollup-plugin-multi-entry

Biggest advantages are that both prod and the html runner would use the same bundler, and it greatly reduces the dev dependencies, so travis will be quite a bit faster.

PR #2946, opened 2016-03-19
Build browser tests with rollup

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2933

Travis might be ~30 seconds faster now per matrix test, but honestly that's pretty hard to guess at.  Running travis multiple times before and after produced varying results.. but on average this is _slightly_ faster I think.

Might not have really been worth the effort, but I still kinda like it.

@jmeas @megawac thoughts?


jmeas:
lgtm.

Speed aside, I prefer this just for the sake of being consistent about how we build the lib :P

Issue #2647, opened 2015-06-30
Create Marionette Radio docs 


jasonLaster:
Currently, people have to go to Marionette Radio to read the readme to understand how to use the lib.

Lets have some documention in Marionette for Radio. This is especially important w/ `next` as we introduce some helpful APIs on top of Radio.

The docs should have the methods that are available, but also useful Eventing patterns that are common best practices

## Todo List

* [x] Document channels
* [x] Document Request/Reply
* [x] Document Events
* [x] Clarify difference in when to use Events vs Request
* [x] Document Radio Integration with Marionette

scott-w:
Things to do for this:

* [ ] Document channels
* [ ] Document Request/Reply
* [ ] Document Events
* [ ] Clarify difference in when to use Events vs Request
* [ ] Document Radio Integration with Marionette

scott-w:
I'll close this for now. We can give them another run through later.

PR #2949, opened 2016-03-20
Backbone radio docs

scott-w:
I've started outlining the Radio documentation with a really quick example. I think we should outline the basic API here but I'm not sure whether we should include the full documentation, structure, diagrams, etc. for Radio here too or if we can leave it all in the docs in Radio itself.

As always, I'm happy to get some early feedback and I can incorporate it as I work through the docs. 

See #2647

### DIFF
--- a/docs/backbone.radio.md
+++ b/docs/backbone.radio.md
@@ -77,9 +77,11 @@ access to this API.
   * [Assigning Channels to Objects](#assigning-channels-to-objects)
 * [Event](#event)
   * [Listening to Events on Objects](#listening-to-events-on-objects)
+  * [When to use Events](#when-to-use-events)
 * [Request](#request)
   * [Returning Values from Reply](#returning-values-from-reply)
   * [Listening to Requests on Objects](#listening-to-requests-on-objects)
+  * [When to use Requests](#when-to-use-requests)
 
 ## Channel
 
@@ -214,6 +216,20 @@ var Star = Mn.Object.extend({
 This gives us a clear definition of how this object interacts with the `star`
 radio channel.
 
+### When to use Events
+
+The Event is a simple notification that _something happened_ and you may or may
+not want other objects in your application to react to that. A few key
+principles to bear in mind are:
+
+* If you don't know what could act on the event, or don't care, use an `Event`
+* If you find yourself calling it an action that occurred, use an `Event`
+* If it's fine for many objects to perform an action, use an `Event`
+* If you don't mind that no objects react, use an `Event`
+
+If your use case isn't covered here, consider whether you want to
+[use a request](#when-to-use-requests) instead.
+
 ## Request
 
 The Request API provides a uniform way for unrelated parts of the system to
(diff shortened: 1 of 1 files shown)


docs/backbone.radio.md
```diff
@@ -0,0 +1,131 @@
...
+message bus system designed to allow two otherwise unrelated objects to
+communicate and share information.
+
+Let's look at a simplified example in Marionette:
+
+```javascript
+var Mn = require('backbone.marionette');
+
+var NotificationView = Mn.View.extend({
+  channelName: 'notify',
```
paulfalgout: I think currently only things that extend from `Mn.Object` have the radio mixin... though seems like it'd be ok to add to views as well... I don't think that has been implemented.

paulfalgout: In the case that we did add the radio mixin to a view, or for an Object, the fun part of the Mn integration is the `radioEvents` and `radioRequests` such that instead of the `initialize` below you'd just have:
```js
radioEvents: {
  'show:success': 'showSuccessMessage',
  'show:error': 'showErrorMessage'
}
```

Although in this case I'd argue that, in this case, those should be `radioRequests` and not events as you are requesting something to happen.  Where-as events should strictly be things that occurred, ideally written in `verb:noun`

scott-w: Awesome, cheers @paulfalgout - I didn't see `radioEvents` or `radioRequests` in the source code. I'll get all that changed over :+1:

scott-w:
I'm removing [WIP] as I'm about happy for this to be reviewed and merged if everyone is happy the documentation is accurate.

docs/backbone.radio.md
```diff
@@ -0,0 +1,381 @@
...
+var User = require('./models/user');
+
+var notifyChannel = Radio.channel('notify');
+var userModel = new User();
+
+// The following will call Notification.showErrorMessage(message)
+notifyChannel.request('show:error', 'A generic error occurred!');
+
+// The following will call Notification.showProfileButton(user)
+notifyChannel.trigger('user:logged:in', userModel);
```
JSteunou: API speaking, if I have a subapp user and a subapp notify, notify would only by with `request`. subapp user would request a *show message* or *show button* but subapp notify would certainly **not** listen to subapp user. The relation is in the wrong way here.

Might be better to use a *sidebar* and / or *footer* object listening to *user logged in* event emit from subapp user.

ahumphreys87: @JSteunou I think notify is just a generic name for the channel in this case. @scott-w ???

JSteunou: It's not about the channel name, it's about making the example more relevant and conceptually more correct from an API pov.

scott-w: My aim for this was just to quickly demonstrate the syntax rather than a totally realistic example.

paulfalgout:
👍  We can come back around for any improvements.

Issue #2915, opened 2016-03-08
Update documentation for Application


scott-w:
The documentation for `Application` needs to be updated for Mn 3.

## Updates

* [x] The new root region
* [x] Document Methods
* [x] Link to the Marionette Object
* [x] Fix Region documentation referencing Application

Feel free to add topics to this list (or comment).

scott-w:
I'll close this for now. It seems the bulk of the work is merged in.

Issue #2490, opened 2015-04-13
Update copy to ensure people do not think that Regions are deprecated


jmeas:
Something somewhere is giving people the idea that Regions as a class are deprecated for v3. This simply is not the case!

Only Regions _**on the application**_ are deprecated. Someone should dig through stuff to make sure that there's nothing potentially misleading out there.

It might even good to update the Application documentation to explicitly say that Regions are not deprecated, but App regions are.

paulfalgout:
This looks to be moderately ok now.. and it will be very clear, very shortly.

PR #2953, opened 2016-03-23
Went through the region docs #2915

scott-w:
This PR replaces the App references with View in the `marionette.region.md` documentation. This is in response to #2490 and to really finalise #2915


docs/marionette.region.md
```diff
@@ -29,81 +29,67 @@ Using the `LayoutView` class you can create nested regions.
...
 
-## Defining An Application Region
+## Defining the Application Region
 
-You can add regions to your applications by calling the `addRegions` method on
-your application instance. This method expects a single hash parameter, with
-named regions and either jQuery selectors or `Region` objects. You may
-call this method as many times as you like, and it will continue adding regions
-to the app.
+The Application defines a single region using the `region` attribute, and access
```
paulfalgout: maybe 
```diff
- and access it
+ and it can be accessed
```

docs/marionette.region.md
```diff
@@ -29,81 +29,67 @@ Using the `LayoutView` class you can create nested regions.
...
-  navigationRegion: MyRegion
+var MyView = Marionette.View.extend({
+  regions: {
+    mainRegion: MyRegion
+  }
 });
 ```
 
-#### Object Literal
+### Object Literal
```
paulfalgout: This section needs some cleaning up at some point.  Not supported at all yet too until your issue gets in.. this is also another case where we're doing quite a bit of describing region functionality.

docs/marionette.region.md
```diff
@@ -255,21 +257,23 @@ parameter. Several events will also be triggered on the views; see
...
-MyApp.mainRegion.show(anotherView2, { preventDestroy: true });
+myView.mainRegion.show(anotherView2, { preventDestroy: true });
 ```
 
-NOTE: When using `preventDestroy: true` you must be careful to cleanup your old views
-manually to prevent memory leaks.
+NOTE: When using `preventDestroy: true` you must be careful to cleanup your old
+views manually to prevent memory leaks.
 
 #### forceShow
```
paulfalgout: I think `forceShow` is gone entirely now.

docs/marionette.region.md
```diff
@@ -398,21 +404,25 @@ This example will make a view slide down from the top of the screen instead of j
...
+    modalRegion: {
+      regionClass: ModalRegion,
+      el: '#modal-region'
+    }
   }
-})
+});
 ```
 
 ### Attach Existing View
```
paulfalgout: This is all gone now too.. you shouldn't set currentView on initialization either.  The behavioral change in v3 is that if the view you are showing is already rendered it won't re-render it.. so a showing a rendered view is just attaching it.

docs/marionette.region.md
```diff
@@ -492,43 +506,45 @@ view.supportsDestroyLifecycle = true;
...
 
-MyApp.mainRegion.on("show", function(view, region, options){
+mainRegion.on("show", function(view, region, options){
   // manipulate the `view` or do something extra
   // with the `region`
   // you also have access to the `options` that were passed to the Region.show call
 });
 
-MyApp.mainRegion.on("before:swapIn", function(view, region, options){
+mainRegion.on("before:swapIn", function(view, region, options){
```
paulfalgout: swapIn / swapOut is gone now too.   They were redundant somehow and aren't particularly useful for animation as initially intended.

docs/marionette.regionmanager.md
```diff
@@ -103,7 +103,7 @@ var rm = new Marionette.RegionManager();
 var regions = rm.addRegions({
   main: '#main-content',
   navigation: {
-    selector: '#navigation',
```
paulfalgout: Regionmanager no longer exists entirely :-)

scott-w: Excellent, I'll chop it out. I'll also put a deprecated notice at the top of `marionette.regionmanager.md` directing readers to `marionette.region.md`

docs/marionette.region.md
scott-w: It is so I'll take it out here :+1:

docs/marionette.regionmanager.md
paulfalgout: hmm we might want to deprecate it on `master`, but on `next` wouldn't it be removed entirely? If we find people were using `RegionManager` directly, which seems pretty unlikely, we can always spin up a separate module.  The interface `RegionManager` provided is now baked into `View`

scott-w: My thinking was to maintain links for a short time period and direct users to the new docs until we've got something better for regions.

PR #2958, opened 2016-03-28
Check options for replaceElement / allowMissingEl

### LINKED ISSUE #2957: Mn 3.0: replaceElement option isn't respected in most scenarios.
Hey,

I'm doing:

```
import { Region } from 'marionette';
import HeaderView from './headerView.js';

export default Region.extend({
  initialize() {
    this.listenTo(App.channels.body, 'rendered', this._onBodyRendered);
  },

  _onBodyRendered() {
    this.show(new HeaderView(), {
      replaceElement: true
    });
  }
});
```

and it doesn't replace the element.

This is because `this.getOption('replaceElement')` always returns false. Options passed to `show` aren't extended onto the object (doubtful that they should be?). The passed in options are just discarded.

MeoMix:
Previously, the Region object was only able to find `replaceElement` and `allowMissingEl` options if they were passed as constructor params. Now, the options object passed to `show` is also valid.

If both are provided, `show` takes precedence as it seems to make sense to override default values with method options.

I chose to continue using the same formatting seen in `_removeView` for handling check for options value.

Resolves #2957

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -37,7 +37,7 @@ const Region = MarionetteObject.extend({
   // method for you. Reads content directly from the `el` attribute. The `preventDestroy`
   // option can be used to prevent a view from the old view being destroyed on show.
   show(view, options) {
-    if (!this._ensureElement()) {
+    if (!this._ensureElement(options)) {
       return;
     }
     this._ensureView(view);
@@ -84,9 +84,9 @@ const Region = MarionetteObject.extend({
     }
   },
 
-  _attachView(view) {
+  _attachView(view, options = {}) {
     const shouldTriggerAttach = !view._isAttached && isNodeAttached(this.el);
-    const shouldReplaceEl = !!this.getOption('replaceElement');
+    const shouldReplaceEl = typeof options.replaceElement === 'undefined' ? !!this.getOption('replaceElement') : !!options.replaceElement;
 
     if (shouldTriggerAttach) {
       triggerMethodOn(view, 'before:attach', view);
@@ -102,14 +102,16 @@ const Region = MarionetteObject.extend({
     this.currentView = view;
   },
 
-  _ensureElement() {
+  _ensureElement(options = {}) {
     if (!_.isObject(this.el)) {
       this.$el = this.getEl(this.el);
       this.el = this.$el[0];
     }
 
     if (!this.$el || this.$el.length === 0) {
-      if (this.getOption('allowMissingEl')) {
+      const allowMissingEl = typeof options.allowMissingEl === 'undefined' ? !!this.getOption('allowMissingEl') : !!options.allowMissingEl;
+
+      if (allowMissingEl) {
         return false;
       } else {
         throw new MarionetteError(`An "el" must exist in DOM for this region ${this.cid}`);
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -239,7 +239,7 @@ describe('region', function() {
       beforeEach(function() {
         this.MyRegion = Backbone.Marionette.Region.extend({
           el: '#region',
-          onShow: function() {},
+          onShow: function() {}
         });
 
         this.MyView2 = Backbone.View.extend({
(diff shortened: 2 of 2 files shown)


src/region.js
```diff
@@ -84,9 +84,9 @@ const Region = MarionetteObject.extend({
     }
   },
 
-  _attachView(view) {
+  _attachView(view, {replaceElement} = {}) {
     const shouldTriggerAttach = !view._isAttached && isNodeAttached(this.el);
-    const shouldReplaceEl = !!this.getOption('replaceElement');
+    const shouldReplaceEl = !!replaceElement || !!this.getOption('replaceElement');
```
paulfalgout: actually.. doesn't this have to be something like `replaceElement !== false`  In this case if `replaceElement` is explicitly false, but `this.replaceElement` is true, the option won't override the class?

MeoMix: Ahh yeah, you're right. That's what I get for trying to copy the existing "!!" style without thinking.

MeoMix: Yeah, you're right. My mistake - I was caught up on trying to use the same destructuring syntax and missed the obvious logic error.

I think I prefer `replaceElement === true` rather than `replaceElement !== false` to avoid using double-negative. Should be equivalent result, though, yeah?

src/region.js
```diff
@@ -86,7 +86,7 @@ const Region = MarionetteObject.extend({
 
   _attachView(view, {replaceElement} = {}) {
     const shouldTriggerAttach = !view._isAttached && isNodeAttached(this.el);
-    const shouldReplaceEl = !!replaceElement || !!this.getOption('replaceElement');
+    const shouldReplaceEl = replaceElement === true || !!this.getOption('replaceElement');
```
paulfalgout: :-) my bad.. this doesn't work either..  really we probably want `options` and `_.default`.  if you've explicitly set `replaceElement` to false but it is `true` on the obj, it would still end up true.  Can't do `||`

MeoMix: Hahaha. This is what I get for trying to do a 'simple fix' from GitHub's IDE. Attempt #3 committed.

src/region.js
```diff
@@ -102,14 +104,18 @@ const Region = MarionetteObject.extend({
     this.currentView = view;
   },
 
-  _ensureElement() {
+  _ensureElement(options = {}) {
```
paulfalgout: options isn't being passed from show

src/region.js
```diff
@@ -84,9 +84,11 @@ const Region = MarionetteObject.extend({
     }
   },
 
-  _attachView(view) {
+  _attachView(view, options = {}) {
     const shouldTriggerAttach = !view._isAttached && isNodeAttached(this.el);
-    const shouldReplaceEl = !!this.getOption('replaceElement');
+    const shouldReplaceEl = _.defaults(options, {
```
paulfalgout: thoughts on `const shouldReplaceEl = _.isBoolean(options.shouldReplaceEl) ? options.shouldReplaceEl : !!this.getOption('replaceElement');`

MeoMix: Does kind of weird things if you pass in bad data.. I think I'd always expect `option.shouldReplaceEl`'s value to override if not undefined.

src/region.js
```diff
@@ -102,14 +102,16 @@ const Region = MarionetteObject.extend({
...
-  _ensureElement() {
+  _ensureElement(options = {}) {
     if (!_.isObject(this.el)) {
       this.$el = this.getEl(this.el);
       this.el = this.$el[0];
     }
 
     if (!this.$el || this.$el.length === 0) {
-      if (this.getOption('allowMissingEl')) {
+      const allowMissingEl = _.isUndefined(options.allowMissingEl) ? !!this.getOption('allowMissingEl') : !!options.allowMissingEl;
```
paulfalgout: ah yeah.. I agree witch checking undefined vs boolean.. however checkout https://github.com/marionettejs/backbone.marionette/pull/2693

paulfalgout:
This needs to be [squashed](https://egghead.io/lessons/javascript-how-to-squash-multiple-git-commits)

MeoMix:
Done

paulfalgout:
👍

PR #2961, opened 2016-03-31
Move backbone and underscore to peerDependencies

ahumphreys87:


paulfalgout:
Bb isn't 1.3.2 here.  Might as well be.

ahumphreys87:
@paulfalgout updated

PR #2969, opened 2016-04-03
Softwraps and fixes

rhubarbselleven:
Updated to use softwraps instead of hard and corrected a few syntax items and headings


ahumphreys87:
thanks @rhubarbselleven

Issue #2972, opened 2016-04-05
[v3.0.0-pre.3] `global` variable not defined in browser.


devel-pa:
Hi,
Loaded in browser you'll receive: Uncaught ReferenceError: global is not defined

https://github.com/marionettejs/backbone.marionette/blob/next/lib/backbone.marionette.js#L444

paulfalgout:
I've got this one.  This is the `Mn` alias that works with the UMD wrap, but the bundled fix for `pre.3` switched to IIFE so that it doesn't break the local dependency vars.

paulfalgout:
FWIW the quickfix if you're downloading the bundled version is to replace `global['Mn']` with simply `Mn`

devel-pa:
NPM installed, just before the scripts loading `global = window` and works.

Issue #2518, opened 2015-04-21
Test the bundled build


jasonLaster:
This is half question half issue. We should create a new specrunner that tests that the bundled build works, not sure if we need to run all the tests, but some would be nice :)

PR #2980, opened 2016-04-09
Fix the bundled build and test build files

paulfalgout:
This should resolve https://github.com/marionettejs/backbone.marionette/issues/2972 and https://github.com/marionettejs/backbone.marionette/issues/2518

The bundled build is now back to UMD which will help anyone using the lib/.min.js for requirejs.

Also there's a bit less magic going on. I was able to use the same rollup plugins used for the browser test to pull in babysitter and radio for bundle.  Way cleaner.

The tests themselves run after running `gulp build` they aren't a part of travis at the moment, but we also don't build the library.  I'm sure we can get travis to build and test somehow.


ahumphreys87:
@paulfalgout needs a rebase - lets get this in and cut pre.3

Issue #2966, opened 2016-04-03
Consistency of filtering 


rogovdm:
Use case:
- partial rendering of a collection via `filter` option
- then adding some children via `addChild` method

As far as I can see we don't filter those children which are rendering with a direct addChild method. So filter method does not call `filter` all the time. Is that okay? Just opening discussion.

I thought to propose to name filter => initialFilter, but it's not that good too.

paulfalgout:
Yeah, trying to maintain two totally different collections is difficult.  Most of the `filter` functionality did not take into account the rest of the `CollectionView` API.  My personal opinion is that it should be removed from Mn.  `addChild` is privatized in `v3`.  The best way to update a `CollectionView` in my opinion is to maintain the `collection` passed to it.  If it needs to be different that the source collection, just clone it before handing it to the `CollectionView`

rogovdm:
@paulfalgout  yeah, that seems to be the most valid approach.

I've spent two days experimenting with an idea of a partial rendering — when I have to a quite big collection but want to insert only a visible through UI part into DOM.

Most of my experiments was in changing how child views has been attaching. The result is a blast, but my neighbour is asking what's thats smelling :)

> addChild is privatized in v3.

Ooops :) So It means there will be no documented way to insert a specific child view into specific place manually?

PR #2981, opened 2016-04-09
buildChildView, removeChildView and introduce addChildView

paulfalgout:
`_addChild` `_buildChildView` and `_removeChildView` were made private for v3.

I've personally run into issues with needing `buildChildView` to show different views with different models than in the collection itself.  (essentially it uses the model in the collection to know which model type to get and show with a view)

I've also heard feedback that `addChild` would be missed.  However `_addChild` is doing a lot, and using it seems overly complex when you simply need to add a view as a child to the collectionView.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -100,7 +100,7 @@ const CollectionView = Backbone.View.extend({
   // get the child view by model it holds, and remove it
   _onCollectionRemove(model) {
     const view = this.children.findByModel(model);
-    this._removeChildView(view);
+    this.removeChildView(view);
     this._checkEmpty();
   },
 
@@ -191,7 +191,7 @@ const CollectionView = Backbone.View.extend({
       this._appendReorderedChildren(elsToReorder);
 
       // remove any views that have been filtered out
-      _.each(filteredOutViews, _.bind(this._removeChildView, this));
+      _.each(filteredOutViews, _.bind(this.removeChildView, this));
       this._checkEmpty();
 
       this.triggerMethod('reorder', this);
@@ -323,7 +323,7 @@ const CollectionView = Backbone.View.extend({
         emptyViewOptions = emptyViewOptions.call(this, model, this._emptyViewIndex);
       }
 
-      const view = this._buildChildView(model, EmptyView, emptyViewOptions);
+      const view = this.buildChildView(model, EmptyView, emptyViewOptions);
 
       this.triggerMethod('before:render:empty', this, view);
       this._addChildView(view, 0);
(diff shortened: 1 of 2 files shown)


paulfalgout:
@ianmstew @ahumphreys87 @scott-w thoughts?  If we truly find reasons to remove `buildChildView` and the others we can deprecate them in v3 and remove in v4 ?

scott-w:
If I understand correctly, are you thinking that you need to display models that may or may not be in `collection`? I run into similar issues though I mostly solve them by adding models to the collection to represent that data.

Obviously this makes the models and templates look a bit ugly `<%  if (modelProperty()) { %> do_a <% } else { %> do_b <% } %>`

What about something like `getChildView` or `getModel` that explicitly returns `View` classes and `Model` instances that can be passed into `_addChild` `_buildChildView` and `_removeChild`?

paulfalgout:
so `getChildView` is also gone 😉  But you can just use `ChildView` as a function.

The problem is context.. in `buildChildView` you are taking the `model` `ChildView` and `childViewOptions` and instantiating the `childView`..  You could potentially solve this problem at the points that the `ChildView` `Model` and possibly even `childViewOptions` are created, but you might have fairly complex logic in 3 places instead of 1. 

I'm also not sure how `getModel` would work. I _think_ that might be complicating things where `buildChildView` is easier to understand.  `getModel` would be getting the actual model for the `childView` based on the `model` for the `childView`

But yes I essentially have a collection of activity events that describe something that occurred like in my case, a comment or a task was completed or a survey was taken.  So from that activity feed, I'm not actually showing the activity, I'm showing the comment or the completed task. And I'm getting the model to display in the view from a request based on an id and a type on the activity model.

Also I too have simply attached a model to represent an added view, but after talking with one guy about how he used `addChild` / `removeChildView` to attach a view to his collectionView unrelated to the Collection, it had me intrigued.  It might actually be a cleaner solution.  Although I don't think `addChild` is the correct interface, that's overly complex for this job and a `addChildView` to match the `removeChildView` is better.

Regardless, if people are actively using many of these APIs we should talk about deprecating them for alternatives instead of removing them right away. I think to some degree we assumed a bit much that many of these functions were easily redundant or only used internally anyways, but I don't think it is the case.

scott-w:
I agree with the sentiment that we can't just remove a heavily-used API - we've seen a few things come through because of this.

Your description of `getModel` is how I'd imagine it working though I've had a good look through the source and it could be done without a new method.

rogovdm:
#2966

JSteunou:
I totally agree! I'm also sad `getChildView` is gone. Event if `childView` can be used instead, this introduce an API breaking change and remove a nice helper. This will lead to more boilerplate.

paulfalgout:
So for every instance of `getChildView` you can just swap in `childView` right?  How will this lead to more boilerplate?

JSteunou:
Will it work simple as that?

paulfalgout:
I _think_ so.  I mean assuming that your current `getChildView` does not internally use `this.childView`

JSteunou:
I just read the src to be sure, I did not see there is a check *is Backbone.View or is Function*. So API change but no loss, I like that, good job!

paulovieira:
About the `getChildView` vs `childView`-as-a-function: yep, it's as simple as that. Just search-and-replace. It make the API simpler and more consistent with other parts of Backbone/Marionette (for instance, the `events` hash can be given as an object or as a function that returns an object).

More details: https://github.com/marionettejs/backbone.marionette/pull/2420

scott-w:
Coming back, I've just remembered there's an existing `getChildView('region')` that returns the view at 'region' - of course I'd suggested something a little different here that the current implementation of `childView` more or less does what I'd like it to. Minor thing - I think the documentation would change `child` to `model` to make it clearer what its argument actually is.

paulfalgout:
FWIW even if we re-privatized `_removeChildView` I think `_addChildView` is a clearer representation of what is going on than the larger `_addChild`

I'm finding when we have a pattern like:
```
// some code
this.triggerMethod('before:something');
// some other code
this.triggerMethod('something');
// even more code
```
It indicates we need a `doSomething` function that begins with the `before:event` and ends with the `event`.  This helps clearly define what needs to be in the function and when it needs to occur.

scott-w:
I'm quite happy to 👍 this and get it in

ahumphreys87:
👍

PR #2985, opened 2016-04-15
[docs] fix event name

tjdavenport:
I noticed my listener for "collection:rendered" wasn't working. Looks like the example event name is wrong?  "collection:rendered" doesn't exist anywhere in the source


paulfalgout:
Thanks!

PR #2986, opened 2016-04-17
View docs

scott-w:
Pushed the docs on a little bit by adding some more information about models/collections and context.

The big thing here is the new ToC for View. I'd love some feedback on whether everyone thinks it'll serve as a good basis for how to structure this part of the docs. My thinking is we can take more advanced topics - renderers and third party templating engines - and put them into their own section.

*Todo* The contents don't match the ToC any more so that'll all need to be updated.


docs/marionette.layoutview.md
```diff
@@ -2,6 +2,10 @@
 
 # Marionette.LayoutView
 
+**_DEPRECATED The `LayoutView` has been removed from Marionette 3. See
+the [documentation for View](./marionette.view.md) for the latest documentation
+for managing layouts._**
+
```
ahumphreys87: I think we should just remove the file and include in the upgrade guide

scott-w: In which case I'll pull out the CompositeView docs as well. I can review the history to figure out what goes where.

ahumphreys87:
👍

PR #2994, opened 2016-04-27
Remove Marionette.Module doc for v3

andrewmnlv:


scott-w:
Hi @andrewmnlv Thanks for this. For the initial v3 release, I'd rather use a deprecation notice at the top to prepare readers for the removal of the feature.

andrewmnlv:
Hi @scott-w 
It was deprecated on v2.4.2  And may be better remove module doc totally as for controllers?)

scott-w:
The only issue I have is we've had no big *Deprecated* notice at the top. I'd be 100% behind removing all text in favour of just a deprecation/removed notice that would lead to the file being deleted in 3.1.

That way we can point people at proper module loaders from that documentation e.g. Webpack/Browserify

scott-w:
I've been discussing this and I've started to come round to your way of thinking, so I'll 👍 this

andrewmnlv:
@scott-w thanks :+1:

PR #2997, opened 2016-04-29
Update rollup to v.0.26.1

TerrenceLJones:
Versions prior to v1 don’t update the minor version with the `^`. So this update needs to be done manually. 

New version fixes bad sourcemap errors: https://github.com/rollup/rollup/issues/618.


paulfalgout:
Yep this was breaking `gulp test-browser`

PR #3000, opened 2016-05-03
Refactor/fully convert jscs to eslint

TerrenceLJones:


test/.eslintrc
```diff
@@ -2,5 +2,8 @@
   "extends": ["../.eslintrc", "./.globals.json"],
   "env": {
     "mocha": true
-  }
+  },
+  rules: {
+      "new-cap": 0,
```
paulfalgout: spacing is off here

.eslintrc
```diff
@@ -30,18 +42,35 @@
...
+    "one-var": [ 2, "never" ],
+    "operator-linebreak": [ 2, "after" ],
     "quotes": [ 2, "single" ],
     "semi": [ 0, "never" ],
+    "space-before-blocks": [ 2, "always" ],
+    "space-before-function-paren": [ 2, "never" ],
+    "space-in-parens": [ 2, "never" ],
+    "space-infix-ops": 2,
+    "space-unary-ops": [ 2,
+      {
```
paulfalgout: I _think_ the spacing is off here, though less so than the other place

paulfalgout:
Not super worried about the spacing in those rc files personally, but just noticed it.
LGTM 👍

TerrenceLJones:
Thanks @paulfalgout...those spacing issues should be fixed in my latest commit.

scott-w:
While we're on the topic - `.eslintrc` is deprecated - we should really be using `.eslintrc.json` or `.eslintrc.yaml` as per the docs.

TerrenceLJones:
Thanks @scott-w. I'll make that needed change.

scott-w:
Interesting, it seems a bunch of stuff depends on `.eslintrc` - feel free to revert them back, it's not exactly critical right now.

TerrenceLJones:
Ok this PR should be ready to go. 🎱

scott-w:
:+1:

Issue #2971, opened 2016-04-04
Upgrade guide from v2 to v3


jjanssen:
Forgive if this was already asked. I could not find anything relevant in the search.
As the final release of version 3 is getting closer. Will there also be an upgrade guide and/or tool available on its release?

scott-w:
Hi @jjanssen 
At the minute, we're putting the upgrade instructions throughout the documentation.

We'll also be doing some form of Release Notes with the top-level changes for 2-3 with instructions for how to do it.

gavrashenko:
Any new about instructions ?

jjanssen:
Thanks for the update @scott-w

scott-w:
We've merged an upgrade guide in from #3003 
@jjanssen it would be great to get some feedback on it at https://github.com/marionettejs/backbone.marionette/blob/next/docs/upgrade.md

If anyone has any other notes that I may have missed, ping them here and I'll get them added.

jjanssen:
Awesome! Thanks for the update. I'm going to take a look at it. Probably can give you an update earlier next week.

scott-w:
That's great @jjanssen, thanks for your help 👍

paulfalgout:
This should be moved to or at least linked from this file https://github.com/marionettejs/backbone.marionette/blob/next/upgradeGuide.md

scott-w:
Good shout 👍

scott-w:
Additional TODO: `childViewEvents` are only built and cached once - in v2 they were built on every event.

This means we can't create `childViewEvents` after the view has been created.

jjanssen:
I promised to get back with some feedback. But unfortunately I'm a little caught up with other stuff at work. I started replacing `addInitializer` and module loading, but I haven't made any progress regarding the details. Hope to get back to you as soon as possible.

scott-w:
No bother @jjanssen, thanks for letting me know :)

paulfalgout:
Noticed a breaking change I'm not sure is documented somewhere..

In v2 the View triggers hash will pass an object to the handler where the object is
{
  view: view,
  model :model,
  collection: collection
}

paulfalgout:
I think this covered in the upgrade guide.. between that.. the compat tool and the compat tool docs..  maybe there's room for improvement.. but I think this is done enough for release.. reopen if anyone disagrees.

PR #3003, opened 2016-05-07
Added an upgrade guide #2971

scott-w:
I’ve started working on a quick upgrade guide that should highlight all
the major areas where Mn 3 has changed. This will help readers find
removed features with instructions (or links) to how to upgrade their
code.

Let me know if I'm missing anything and I'll get it added. Alternatively, we could merge this and add anything we're missing after-the-fact. This might affect marionettejs/marionettejs.com#448

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -1,0 +1,66 @@
+# Upgrade Guide
+
+Marionette 3 introduces a number of breaking changes. This upgrade guide will go
+through the major changes and describe how to change your application to
+accommodate them.
+
+Where possible, we will document how to make the changes in Marionette 2.4.5 so
+that your applications can continue to work and to ease the burden of upgrading
+to Marionette 3.
+
+## Views
+
+The most noticeable change to Marionette 3 is the consolidation of `ItemView`
+and `LayoutView` into `View`. In addition, `CompositeView` has been removed and
+its functionality merged into `CollectionView` and `View`.
+
+### Removing `LayoutView` and `ItemView`
+
+Using separate `View` `LayoutView` and `ItemView` was complicating the API for
+Marionette needlessly. The new `View` replaces all of this and sets a clear
+recommendation for building layout trees.
+
+#### Upgrading for Marionette 2.4.5
+
+For updating in Marionette 2.4.5, it is recommended you change all instances of
+`ItemView` to `LayoutView`.
+
+#### Upgrading to Marionette 3
+
+Change all instances of `LayoutView` and `ItemView` to `View`. Any views that
+previously extended `View` with a custom `render` should work relatively
+unchanged.
+
+### Removing `CompositeView`
+
+The `CompositeView` was removed in favor of using `View` and `CollectionView`.
+
+See [`CollectionView`](./marionette.collectionview.md#rendering-collectionviews)
+for detail on upgrading to Marionette 3. This technique works in both Marionette
+2.4.5 and Marionette 3.
+
+## Events
+
+A number of lifecycle events were changed or removed from Marionette 3.
+Information on which ones were removed, changed, or added will be found here
+with recommendations on how to update your code.
+
+### `show` and `before:show`
+
+The `show` events were removed completely as they were redundant and were being
+used incorrectly to show child regions. The `show` event was fired after the
+view had been attached, meaning the DOM was being constantly updated, leading to
+deteriorated performance.
+
+#### Upgrading for Marionette 2.4.5
+
+Replace all instances of `onShow`, `on('show')`, `onBeforeShow` and
+`on('before:show')` to use the `render` and `before:render` events. This is the
+recommendation for Marionette 3 and ensures the DOM tree is built in-memory
+before modifying the DOM.
+
+#### Upgrading to Marionette 3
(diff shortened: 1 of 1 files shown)


docs/upgrade.md
```diff
@@ -0,0 +1,74 @@
...
+previously extended `View` with a custom `render` should work relatively
+unchanged.
+
+### Removing `CompositeView`
+
+The `CompositeView` was removed in favor of using `View` and `CollectionView`.
+
+See [`CollectionView`](./marionette.collectionview.md#rendering-collectionviews)
+for detail on upgrading to Marionette 3. This technique works in both Marionette
+2.4.5 and Marionette 3.
```
ahumphreys87: Mark this as deprecated and a "recommended" update but not immediately necessary

docs/upgrade.md
```diff
@@ -0,0 +1,74 @@
...
+
+The `show` events were removed completely as they were redundant and were being
+used incorrectly to show child regions. The `show` event was fired after the
+view had been attached, meaning the DOM was being constantly updated, leading to
+deteriorated performance.
+
+#### Upgrading for Marionette 2.4.5
+
+Replace all instances of `onShow`, `on('show')`, `onBeforeShow` and
+`on('before:show')` to use the `render` and `before:render` events. This is the
```
ahumphreys87: shouldn't this be `onAttach`??? @paulfalgout

ahumphreys87:
we need info on:
 *  `templateHelpers` being renamed `templateContext`
 *  `childEvents` are now `childViewEvents`.
 * regions have been removed from the View object: i.e. you cannot do `this.view.header.show` you need to do `this.view.getRegion('header').show` now

scott-w:
Added the information in the above commit

docs/upgrade.md
```diff
@@ -0,0 +1,66 @@
...
+### `show` and `before:show`
+
+The `show` events were removed completely as they were redundant and were being
+used incorrectly to show child regions. The `show` event was fired after the
+view had been attached, meaning the DOM was being constantly updated, leading to
+deteriorated performance.
+
+#### Upgrading for Marionette 2.4.5
+
+Replace all instances of `onShow`, `on('show')`, `onBeforeShow` and
```
paulfalgout: I don't really know a use-case for `onBeforeRender`  `onRender` is essentially equivalent to what it was and what `onBeforeShow` was for.

docs/upgrade.md
```diff
@@ -0,0 +1,66 @@
...
+`on('before:show')` to use the `render` and `before:render` events. This is the
+recommendation for Marionette 3 and ensures the DOM tree is built in-memory
+before modifying the DOM.
+
+#### Upgrading to Marionette 3
+
+Replace all instances of `show` and `before:show` with `render` and
+`before:render`. If you want the view to be visible in the DOM, then listen to
```
paulfalgout: We're going to at some point understand why you would use `onAttach`  This is not a simple feature.  It is understandable directly on a view, but if the view is a child or grandchild, onAttach may behave fairly differently.  It should almost only be used by views that _must_ do something with the DOM in the window.. like some jQuery plugin.. even then, `onDomRefresh is probably the better solution there as well.

scott-w: Thanks for this. Does that mean `onDomRefresh` fires the first time it's attached as well?

paulfalgout: Yep.  https://github.com/marionettejs/backbone.marionette/blob/next/src/monitor-view-events.js#L54

It in fact triggers after the `view.onRender` and `view.onAttach`, but it happens if either of them occur.  I still think there is some work to prevent some extra triggers of that for grandchild views.  The remaining v3 work in my mind revolves around view attaching... both for grandchildren and attaching a view to existing DOM. there's some odd behavior in a few places that we either need to be aware of and document, or fix.  I'm not sure which yet.

docs/upgrade.md
scott-w: I have to admit, I don't either. I find `onRender` is _okay_ for doing this stuff but it's not great.

docs/upgrade.md
rafde: > We're going to at some point understand why you would use onAttach

Animating the view the one time it's added to the DOM is when I use `onAttach`

docs/upgrade.md
```diff
@@ -0,0 +1,139 @@
...
+the removal of the ability to access regions as attributes
+
+### Removing `view.region`
+
+The `view.<region_name>` syntax has been removed in favor of `view.getRegion()`,
+`view.getChildView()` and `view.showChildView()`.
+
+#### Upgrading for Marionette 2.4.5 and 3
+
+Change all references to `view.region` to `view.getRegion('region')`. For
```
rafde: `view.<region_name>` to `view.getRegion('<region_name>')` makes it consistent with https://github.com/marionettejs/backbone.marionette/pull/3003/files#diff-0f792d05638b9c26bc7cd4d4d1496bf9R91

scott-w:
This should be ready to merge if anyone has any more feedback for me to get in here.

ahumphreys87:
:+1:

Issue #2959, opened 2016-03-28
Improve Region Documentation


scott-w:
The region documentation needs cleaning up in a few areas. Like Behaviors, Regions are best understood in the ecosystem supporting them i.e. View/Application. We need to come up with a structure that ensures readers of the View documentation are aware that Regions exist, with links to the region documentation in general.

We'll also need to clean out unsupported features.

# Todo

* [x] Remove unsupported region features from documentation
* [x] Region overview in View/Application
* [x] Link to region documentation form View/Application
* [x] Outline of features that Region supports
* [x] How to document `regions` and the `Region` class.

jcrben:
As I asked over at https://github.com/marionettejs/backbone.marionette/issues/2129 - is regionManager deprecated? Doesn't seem like it from http://marionettejs.com/docs/v3.0.0-pre.2/marionette.regionmanager.html but you've got the box `Remove unsupported region features from documentation` checked.

paulfalgout:
The documentation of the deprecation has been merged but not released

curzonj:
In the v3 region docs it lists events that should fire on the views. I don't know if it's a doc, library, or user error, but I couldn't get the `onShow` method to be called on my view. In 2.4.5 it worked fine.

curzonj:
I understand now that it's just a doc issue, that `onShow` has been removed from mn3:

```
Sean Anderson @MeoMix May 06 15:26
beforeShow doesn't exist in mn3
show is toast, too
it's all about that render/attach life
```

scott-w:
Thanks for pointing this out, this will be covered in the upgrade guide - would it be helpful to mention it here as well?

scott-w:
I'm considering chopping out a lot of information from the `marionette.region.md` file including:

1. onBeforeAttach & onAttach
2. `renderView`
3. `shouldDestroyView`
4. Set How View's `el` Is Attached
5. Region Events And Callbacks
6. Custom Region Classes

Essentially, that's everything around customising the `Region` class, as the general feeling is that it should be hidden from the public API.

Would there be any objections to this? If people feel the `Region` is really useful, I'd love to hear about use cases so I can look into how to that would be documented.

denar90:
Can we close this?

scott-w:
Yep

JSteunou:
Ho I missed that... I just open this one because I found weird the lack of documentation around `regionClass`. This feature is still useful for me. Maybe you should have a *advanced* section then. #3152

PR #3006, opened 2016-05-08
Region Docs #2959

scott-w:
I’ve started chopping up the region docs to be a bit more detailed on
regions and the different ways of using them.

I'll need some help with the following:

1. What events get fired and where
2. Whether `Region` is deprecated in public docs

If `Region` isn't deprecated, I feel we should move it to the bottom as an "advanced" topic.


jasonLaster:
👍 

I really like changing the titles from the method to the purpose

docs/marionette.region.md
```diff
@@ -74,183 +74,87 @@ var MyView = Marionette.View.extend({
...
 
-var MyView = Marionette.View.extend({
-  regions: {
-    contentRegion: {
-      regionClass: MyRegion,
-      myRegionOption: 'bar',
-      myOtherRegionOption: 'baz'
-    }    
+  onRender: function() {
+    this.showChildView('tooverwrite', new OverWriteView());
```
rafde: 😕 Shouldn't this be `this.showChildView('main', new OverWriteView());`
Or should the `main` key be renamed `tooverwrite` in `regions`?

scott-w: Your first comment is correct - it was an oversight on my part.

ahumphreys87:
Really great work :+1:

denar90:
Thanks @scott-w 👍

Issue #2920, opened 2016-03-09
New docs for View


scott-w:
We need to document the `View` class as the basic view completely replaces `ItemView` and `LayoutView`.

## Todo List

* [x] Write the `View` documentation containing:
  * [x] Templates and rendering
    * [x] Rendering a model
    * [x] Rendering a collection
    * [x] Link to advanced rendering doc
  * [x] Regions
  * [x] Organising views (e.g. ui)
  * [x] Events
    * [x] Model and Collection events
    * [x] Listening to child events
    * [x] Link through to detailed event doc
* [x] Remove references to `AbstractView`

If anyone has any extra suggestions, either add them to the list or comment and I'll add them.

## Replaces

* #2477

vernondegoede:
I definitely would like to document these subjects.

ahumphreys87:
great @vernondegoede, help is greatly appreciated

vernondegoede:
@scott-w Let's have a Gitter chat tomorrow on how to start :)

curzonj:
When reading the v3 View docs, I would have liked to see the events that regions would fire on the view in the view docs in addition to the region docs.

scott-w:
Does that include things like the `showChildView('region', view)` lifecycle?

paulfalgout:
We need to go through every file and consider how to remove references to `AbstractView` seeing as it doesn't exist..  now the view's are composed with mixins instead of extending from a shared view class.

scott-w:
I've checked the last items off the above list against the documentation and this can be closed now.

paulfalgout:
I'm not 100% sure of https://github.com/marionettejs/backbone.marionette/blob/next/docs/marionette.view.md#region-availability

It says:
> Any defined regions within a View will be available to the View or any calling code immediately after instantiating the View

which is kind of true.  If the view has DOM (ie: was already there and therefore pre-rendered) then there are regions that will work.. 
Even without DOM the regions are _available_ (like you could setup listeners) but.. they can't be used to show a child if the view isn't rendered.. so the next sentence is confusing:

> This allows a View to be attached to an existing DOM element in an HTML page, without the need to call a render method or anything else, to create the regions.

The next paragraph clarifies, but somehow flipping this around and starting with the common case.. the view has no DOM.. might be clearer..

PR #3008, opened 2016-05-09
Document View Events #2920

scott-w:
Improved the event handling documentation:

* Created a simple create/destroy lifecycle table
* Documented the change from `childEvents` to `childViewEvents`
* Describe the transformation of `event:name` to `onEventName`


docs/marionette.view.md
```diff
@@ -569,84 +628,99 @@ By prefixing with `@ui`, we can change the underlying template without having to
...
+| Order |       Event       |
+| :---: |-------------------|
+|   1   |  `before:destroy` |
+|   2   |  `before:detach`  |
+|   3   |  `detach`         |
+|   4   |  `destroy`        |
+
+#### View Creation Events
+
+These events are fired during the view's creation and rendering in a region.
```
rafde: I wouldn't have guessed this to be the order
I would have thought

1. before:detach
2. detach
3. before:destroy
4. destroy

paulfalgout: The region emptying simply destroys the view.  The view destroy internally calls detach which is why the detach events are wrapped in the destroy events.

If the region is setup to prevent destroying the view it would instead call detach, but there's no need to call it first as view.destroy will handle it.

ahumphreys87:
More good docs :+1:

docs/marionette.view.md
```diff
@@ -482,6 +484,63 @@ In this example, the doubly-nested view structure will be rendered in a single p
...
+```javascript
+var Mn = require('backbone.marionette');
+
+let MyView = Mn.View.extend({
+  regions: {
+    product: '.product-hook'
+  },
+
+  childViewEvents: {
+    'add:item': function(child, item) {
```
denar90: @scott-w should we use `var` instead of `let`?

paulfalgout: Probably.  If we es6 things we should be consistent.  Plus this is a case for `const` anyhow

denar90: @paulfalgout de we have to use es6 for all docs?

scott-w: Yeah, we should use `var` for now. Thanks for spotting this.

denar90:
@scott-w thank you for cool docs 👍

PR #2308, opened 2015-02-09
feat(behavior): override behavior ui with view ui

stephanebachelier:
Rebase on major as discussed in #2265 with explanations in #2266

Issue #2266, opened 2015-02-04
Override behavior ui with view ui


stephanebachelier:
Don't know if it's only me, by I don't like how the `ui` is initialized for a behavior. It you looks [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/behaviors.js#L52) you can see that the ui is built by extending the view.ui and then the behavior.ui

```
        // Construct an internal UI hash first using
        // the views UI hash and then the behaviors UI hash.
        // This allows the user to use UI hash elements
        // defined in the parent view as well as those
        // defined in the given behavior.
        var ui = _.extend({}, viewUI, behaviorUI);
```

I think it breaks the idea of the behavior being small reusable unit of code, that a view use and can override. IMO it should be written the opposite which is : 

```
var ui = _.extend({}, behaviorUI, viewUI);
```

This will solve an issue that I faced, where the behavior ui define a default ui element that the view may want to override which is not possible with the existing form.

An example:

```
// close icon behavior
var CloseIconBehavior = Marionette.Behavior.extend({
  events: {
    'click @ui.closeIcon': 'closeView'
  },

  ui: {
      closeIcon: '.icon-close'
    };
  },

  closeView: function () {
    this.view.destroy();
  }
});

// the view
Marionette.ItemView.extend({
  ui: {
     closeIcon: '.icon-cancel'
  },

  behaviors: {
    iconClose: {
      behaviorClass: CloseIconBehavior
    }
  }
});
```

In the example above the `CloseIconBehavior.closeView` will never be called because the `CloseIconBehavior` will bind the click event to `click .icon-close` and not `.icon-cancel`.

Using my proposed approach it will "behave" correctly.

As I use really a lot the behaviors for my applications, - they are really powerful! -, this is a major pain.

See PR #2308

jasonLaster:
Hmm. Interesting point. I think I dig it

stephanebachelier:
@jasonLaster I can add more explanations if you need. I've seen some issues about some core developers not really liking behaviors, but after digging into it, I find they are a really powerful tool.
I use them a lot in two mobile applications and they are some kind of smart tools.

jasonLaster:
In

PR #2265, opened 2015-02-04
Override behavior ui with view ui

stephanebachelier:
See description in #2266

PR #3009, opened 2016-05-12
Behaviors ui override master

stephanebachelier:
The idea of this small change is to think the behaviors as small unit that can be reused for different views in a same project or shared between project. I've used the major branch on some project without any issues. 

This change was previously merged in major with the PR #2308  

As stated in the comment it was discussed in #2265 with explanations in #2266

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -252,6 +252,58 @@ Marionette.Behavior.extend({
 });
 ```
 
+### ui
+
+Same thing as documented in Marionette.View, `ui` hash let you maps UI elements to their jQuery selectors.
+Thus a behavior can have its own selectors and as in a view, you will be able to attach events to these ui elements.
+
+```js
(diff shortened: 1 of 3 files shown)


paulfalgout:
So this is a breaking change which is why is was merged into major.  This work is already in `next` I think.  It varies slightly from this PR, but I'm pretty sure all of this functionality survives in `v3`

https://github.com/marionettejs/backbone.marionette/blob/next/src/behavior.js#L35

stephanebachelier:
@paulfalgout you're right. I'm the author of that line :)
So now I'm stuck to use the branch in this PR while the 3.0.0 is not out ?

paulfalgout:
I think in `v2.x` it would be moderately easy to achieve the functionality you are looking for by overriding
`getBehaviorsUI` on the behavior?  That seems like a pretty clear hook to backport this functionality in the meantime.

Issue #2937, opened 2016-03-15
triggerMethod should support a space-delimited list of events


paulfalgout:
If we want to replace `trigger` with `triggerMethod` https://github.com/marionettejs/backbone.marionette/issues/2508 ...

it will really need to support the same API.

`this.trigger('foo bar')` will trigger both `foo` and `bar`

`this.triggerMethod('foo bar')` will not call both `onFoo` and `onBar`

PR #3010, opened 2016-05-19
Add support of space-delimited list of events

rise2semi:
Add support of space-delimited list of events for triggerMethod.
The problem is described in #2937.


docs/marionette.functions.md
```diff
@@ -173,6 +173,9 @@ All arguments that are passed to the triggerMethod call are passed along to both
 
 `triggerMethod("foo", bar)` will call `onFoo: function(bar){...})`
 
+To trigger several events and corresponding methods split the events by the space. Example:
+* `triggerMethod("render ready")` fires the "onRender" ans the "onReady" functions
```
paulfalgout:
```diff
- ans
+ and
```

paulfalgout: actually can we use `triggerMethod('foo bar')`?  Triggering render manually is probably not a great idea.

rise2semi:
"render" event was chosen just as an example. I've changed the event names to more abstract.
Yes, we can use ```triggerMethod('foo bar')```. You can check it in this [jsfiddle](https://jsfiddle.net/b2Lrxqwr/1/) (I created a build with my changes).

ahumphreys87:
@rise2semi this is great work! Could you please write a test for events bubbling up the view tree? to prove that childEvents can listen for events triggered like this.

paulfalgout:
childViewEvents 😂

paulfalgout:
Is there any perf concerns here? This is a pretty hot function

JSteunou:
On the *map* / trigger multiple part I dont see any perf downside, at least no more than manually triggering multiple events at once from the view. Should not be *visible*.

My concern is more about the 1st test, if it cost a little, it's a little x all the time triggerMethod is called, and I think it is called a lot. How many does cost this line `if (eventSplitter.test(events)) {` ?

rise2semi:
@paulfalgout I've added tests for childViewEvents.

I'm not sure that there will be performance issue with `eventSplitter.test(events)`, Backbone has the same logic for [events](https://github.com/jashkenas/backbone/blob/master/backbone.js#L148) and I haven't hear about performance issues in this place.

scott-w:
Should we look to resolve these conflicts and bring this feature in for 3.1 or (potentially) 3.2?

src/trigger-method.js
```diff
@@ -38,6 +35,31 @@ export function triggerMethod(event, ...args) {
...
+// `this.triggerMethod("foo:bar")` will trigger the "foo:bar" event and
+// call the "onFooBar" method.
+//
+// `this.triggerMethod("foo:bar baz:qux")` will trigger at first the "foo:bar"
+// event and call the "onFooBar" method and then trigger the "baz:qux" event
+// and call the "onBazQux" method
+export function triggerMethod(events, ...args) {
+  var result;
+
+  if (eventSplitter.test(events)) {
```
chchrist: do we need this if statement? ```"foo".split(/\s+/)``` returns an array

src/trigger-method.js
```diff
@@ -38,6 +35,31 @@ export function triggerMethod(event, ...args) {
...
+// call the "onFoo" method.
+//
+// `this.triggerMethod("foo:bar")` will trigger the "foo:bar" event and
+// call the "onFooBar" method.
+//
+// `this.triggerMethod("foo:bar baz:qux")` will trigger at first the "foo:bar"
+// event and call the "onFooBar" method and then trigger the "baz:qux" event
+// and call the "onBazQux" method
+export function triggerMethod(events, ...args) {
+  var result;
```
chchrist: why not a const?

src/trigger-method.js
```diff
@@ -7,20 +7,17 @@ import getOption from './utils/getOption';
 // split the event name on the ":"
 const splitter = /(^|:)(\w)/gi;
 
+// split the list of events on the space
+var eventSplitter = /\s+/;
```
chchrist: const?

rafde:
@scott-w I am not a big fan of this implementation. triggerMethod is my 2nd perf woe and I am sure this implementation is going to create a hit in the perf. If I address this, it's pretty much wipes this code out.
I am leaning on closing this and starting over. @marionettejs/marionette-core any objections?

scott-w:
Fair enough @rafde, I definitely see a lot of `triggerMethod` stuff happening when I run collection views inside collection views hammering the performance.

rafde:
@rise2semi this branch is out of sync at the moment. If you are still interested in helping, please rebase and resolve conflicts.

paulfalgout:
closing as it is stale

PR #3014, opened 2016-05-22
Monitor for child already attached

paulfalgout:
#Helps to resolve https://jsfiddle.net/599dtsb7/6/ where child views attached during `onAttach` get their `attach` events triggered twice.

Solution: https://jsfiddle.net/599dtsb7/8/

This still needs:
- [x] Tests
- [ ] ~~Documentation? (I had previously recommended against using onAttach)~~
- [x] Research failing tests

~~Unfortunately this broke other tests which is concerning.  I'm not sure why this change would break any expected behaviors. Hopefully this will just show light on tests that weren't exactly what we expected vs simply moving this bug down the chain~~

Turns out I just legit had a bug.

### DIFF
--- a/src/monitor-view-events.js
+++ b/src/monitor-view-events.js
@@ -5,22 +5,32 @@ import { triggerMethodOn } from './trigger-method';
 import _ from 'underscore';
 
 // Trigger method on children unless a pure Backbone.View
-function triggerMethodChildren(view, event, beforeEachTrigger) {
+function triggerMethodChildren(view, event, shouldTrigger) {
   if (!view._getImmediateChildren) { return; }
   _.each(view._getImmediateChildren(), child => {
-    if (beforeEachTrigger) {
-      beforeEachTrigger(child);
-    }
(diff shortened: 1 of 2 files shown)


paulfalgout:
@marionettejs/marionette-core  I am pretty sure we want to do this for `v3` 

@ianmstew since this is all in the monitor stuff you wrote, mind checking it out specifically?

ahumphreys87:
👍 from me - would be good to get a test showcasing the bug though

paulfalgout:
@ahumphreys87 for certain.  Hoping to work on tests tonight

paulfalgout:
Alright I think this is good to go.

@scott-w anywhere you can think of where I recommended we don't use `onAttach` we might be ok?  I think it _should_ work now in many cases.

ahumphreys87:
@paulfalgout I think a valid use case is for animating in a view when it first gets shown. I think that can be covered in a docs PR though

ahumphreys87:
👍

PR #3017, opened 2016-05-26
Update CONTRIBUTING.md

denar90:
Update contributing guide due to removed grunt.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,8 @@ the process of getting changes merged in and released.
 
 * Run `npm install` to make sure you have all Marionette dependencies locally.
 
+* Run `npm run build` to build source files.
+
 ## Reporting a bug
 
 In order to best help out with bugs, we need to know the following information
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -110,14 +112,11 @@ PR. That would at least get us started on creating the failing test in the code.
 
 ## Running Tests
 
-There are 3 ways you can run the tests.
-
 * via command-line by running `npm test`
 
-* in the browser by opening `SpecRunner.html`
+* in the browser by opening `test/runner.html`
```
paulfalgout: you'll first have to built the test assets with `gulp test-browser`  I don't know that we have an npm alias for that

paulfalgout: It would be sorta slick to add `npm run test-browser` which would run the gulp to build the assets and then afterwards run something like 
```js
const opn = require('opn');

opn('./test/runner.html');
```

denar90:
Due to @paulfalgout comments was added `npm run test-browser`. I've added `opn` which will open `test/runner.html` after running `gulp test-browser`.

ahumphreys87:
:+1:

PR #3022, opened 2016-05-28
Fix childView show

Corey-Maler:
I don't know, BUG is it or a feature, but this is not logical and uncomfortably.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -43,7 +43,7 @@ const Region = MarionetteObject.extend({
     this._ensureView(view);
     if (view === this.currentView) { return this; }
 
-    this.triggerMethod('before:show', this, view, options);
+    this.triggerMethod('before:show', view, options);
 
     MonitorViewEvents(view);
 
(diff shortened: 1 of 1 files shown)


Corey-Maler:
May be we should call `onShow` both on showing view and current region? Or use `onShowView` or `onShowChild` in current region?

paulfalgout:
the first argument of every `triggerMethod` in `v3` is itself.  There were only a few cases in `v2` where this wasn't the case, but now they are all consistent.
In v3 `show` was removed from the view.  The view's concerns should mostly be `render` or in edge-cases `attach`
Unless you're writing a fairly specific region I don't know if `show` is super useful.

Do you have a particular use-case where this is causing trouble?

Corey-Maler:
So, I see now. Thank you. 
Sorry, a lot of undocumented breaking changes here, this one for example confusing me after recommendation to use `onShow` in v2.

paulfalgout:
Yep.  Working on the documentation as well as some migration tools.  Thanks for working with us during the transition.

PR #3023, opened 2016-05-29
Remove API docs

paulfalgout:
While great efforts were made here, they're not inline with v3, and they don't build with gulp.  If we need this, it is in github, but I think for now we should remove them so they don't confuse new v3 users.  Additionally we have enough on our plate right now getting v3 out to worry about maintaining this as well.


ahumphreys87:
yup good idea! 👍

denar90:
👍

PR #3024, opened 2016-05-29
Remove getValue and internal getOption

paulfalgout:
`getValue` was very under-utilized.  In most cases `_.result` was fine.  In the others, the code is now more explicit.

`getOption` can have odd side-effects https://github.com/marionettejs/backbone.marionette/pull/2901#issuecomment-188317691


ahumphreys87:
👍

blikblum:
How about removing options property altogether?

paulfalgout:
I think we should deprecate `getOption` and _possibly_ the attaching of `options` in `v3.x` and remove it for `v4`

However `triggerMethod` handlers can currently be passed as an option automagically, and that needs to be discussed prior to removal.

paulfalgout:
This is missing importing some underscores into various modules

paulfalgout:
Fixed.  We could use some linter improvements  or something to prevent using non-imported vars.  Right now everything seems to work ok, but the build file will contain multiple versions of `_`

ahumphreys87:
@paulfalgout we should probably move to eslint given that jscs is now
unmaintained. I'll start to work on it

paulfalgout:
`next` is already on eslint :-)
Just needs a new rule I think.

JSteunou:
removing `this.options` would be a huge change. Following Marionette doc and first practice I have it a lot in my code.

scott-w:
I agree with @JSteunou on this - I've been under the impression that `this.getOption` is a preferred way to handle custom attributes.

paulfalgout:
Just to be clear for this instance, getOption causes odd bugs and patterns internally.  At the moment I am not push for the removal entirely.

JSteunou:
@scott-w I was just talking about `this.options` as @paulfalgout wrote about a *possible* deprecation of `options` attaching.

About `getOption` itself I'm totally ok for a removal of all use internally and a deprecation of the public API for v3

ahumphreys87:
I think this is good now?

PR #3026, opened 2016-05-29
Make internal triggerMethods consistent

paulfalgout:
Now every `triggerMethod`'s first argument is `this` internally.

This was almost true already.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -49,8 +49,8 @@ var Application = MarionetteObject.extend({
 
   // kick off all of the application's processes.
   start: function(options) {
-    this.triggerMethod('before:start', options);
-    this.triggerMethod('start', options);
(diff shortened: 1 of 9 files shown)


denar90:
@paulfalgout Do we have to pass `this` here https://github.com/marionettejs/backbone.marionette/blob/next/src/collection-view.js#L633-L637
?

paulfalgout:
Nope.  That is the exception.  This is a trigger proxy, so the `this` has already been attached by the child that originally triggered the event.

denar90:
So I can 👍 then)

ahumphreys87:
nice work @paulfalgout, needs a rebase then 👍

paulfalgout:
rebased :-)

PR #3028, opened 2016-05-30
CollectionView#reorder will remove an already displayed emptyView

paulfalgout:
This regression was introduced in `2.4.4`

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -165,6 +165,9 @@ Marionette.CollectionView = Marionette.View.extend({
   reorder: function() {
     var children = this.children;
     var models = this._filteredSortedModels();
+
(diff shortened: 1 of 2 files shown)


JSteunou:
For historic and comparison, which commit introduced the regression ?

paulfalgout:
https://github.com/marionettejs/backbone.marionette/commit/0d913a524a71a9b4427256b0ba1b3236abeee9d6#diff-091acc5f607e2622bd6cabf705443fffR186

This line check's the view's children which can be the `emptyView` and later removes it.  But it doesn't unset the `_showingEmptyView` so `checkEmpty` does nothing.

JSteunou:
nice catch, tricky one!

ahumphreys87:
👍

PR #3030, opened 2016-05-30
return this consistently

paulfalgout:
Most functions that don't return something return `this`

This is mainly for chaining.  I'm not a personal fan of chaining, but I am a fan of consistency.


paulfalgout:
The coverage issue here is adding a `return true;` to an already not covered function.
We should handle that here: https://github.com/marionettejs/backbone.marionette/issues/2581

ahumphreys87:
yup agreed - 👍

PR #3031, opened 2016-05-30
When a view is attached to an existing element, `isRendered()` should reflect `true`

### LINKED ISSUE #2859: When a view is attached to an existing element, `isRendered()` should reflect `true`
The argument here is that `isRendered()` intuitively means a view does not need to be rendered, which is the case when it is attached to an existing element.  Currently, `isRendered()` is only set during a normal `render()`.  It would be helpful if, as a view is bound to an existing element, `_isRendered` were also set to `true`.

https://github.com/marionettejs/backbone.marionette/pull/2850#discussion_r47439238

/cc @paulfalgout

paulfalgout:
Resolves #2859

The view.child-views.spec change was necessary as now that the rendered flag is set correctly, on a re-render the regions get re-init'd  But init'd regions don't have an $el until the first show. There are potentially other improvements to be made for this instance, but this PR improves the use-case from this test already.  No need to "render" a template: false view.

### DIFF
--- a/src/mixins/view.js
+++ b/src/mixins/view.js
@@ -3,6 +3,7 @@
 
 import Backbone from 'backbone';
 import _ from 'underscore';
+import isNodeAttached from '../utils/isNodeAttached';
 import MarionetteError from '../error';
 import BehaviorsMixin from './behaviors';
 import CommonMixin from './common';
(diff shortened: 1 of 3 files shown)


denar90:
👍

ahumphreys87:
nice work! 👍

Issue #3027, opened 2016-05-29
Calling view's parent constructor with only the passed options instead of arguments


popovicsandras:
Hi,

### Description
I'm creating a pluggable Service Locator library which I'd like to use with Backbone and Marionette. The problem I found (at least a problem for me) is that the base View and LayoutView are absorbing any additional parameter I pass to the constructor, because they are calling their parent's constructor with only the first argument (options).

Is there any particular reason for calling the parent constructor differently from the different types of Marionette.View's descendant? Sometimes the constructor is called with the "arguments", sometimes with only the options (first parameter).

Parent constructor calls in v2.4.6:
- **View**: Backbone.View.call(this, this.options);
- ItemView: Marionette.View.apply(this, arguments);
- CollectionView: Marionette.View.apply(this, arguments);
- CompositeView: Marionette.CollectionView.apply(this, arguments);
- **LayoutView**: Marionette.ItemView.call(this, options);

### Expected behavior

If every type of view were passing arguments instead of only the first argument (options), the Views would be more easy to reuse without monkey patching.

### Actual behavior

I think passing only the options can be problematic, since this way the View and LayoutView absorb all the other parameters, which would be good to reach from the initialize method. Currently the base View and LayoutView only pass the options.

Backbone does this in the View's contructor:
this.initialize.apply(this, arguments);

### Environment
1. Marionette version: 2.4.6
2. Backbone version: 1.3.3

paulfalgout:
The reason for this inconsistency is a bad one I think.

I have no idea why this was design this way, but you can set options on the view's definition as well as passing them in at instantiation

Such that:
```js
const View = Marionette.ItemView.extend({
  options: {
    className: 'some-class'
  }
});

const view = new View({
  tagName: 'span'
});

view.render();  // view.el == <span class="some-class"></span>
```

So the options passed to the view's constructor are combined prior to sending.  I think this is super weird myself and would be all for using arguments everywhere.

@marionettejs/marionette-core  ?

ahumphreys87:
Yeah, I haven't come across this myself but it definitely sounds like a bug. I'm happy to go with arguments.

@popovicsandras how would you feel about submitting a PR with a PR against the `next` branch? We can get this into 3.0

paulfalgout:
I suppose if we wanted to keep the options on the definition functionality the way to go about that would be to:
```
arguments[0] = this.options;
```
immediately before the
[...]

PR #3032, opened 2016-05-30
Calling Backbone.View constructor with arguments (#3027)

popovicsandras:
### Proposed changes
 - If every type of view were passing arguments to Backbone.View constructor function instead of only the first argument (options), the Views would be more easy to reuse without monkey patching.

Link to the issue:
#3027


paulfalgout:
👍 thanks!

PR #3033, opened 2016-05-31
Region empty edge-case fix

paulfalgout:
An edge-case in which something related to a view's destroy might also empty the view's region will force region empty to happen twice. This will prevent that edge case.


denar90:
👍

ahumphreys87:
👍

PR #3035, opened 2016-05-31
Region needs to detach previous html if there is no view

paulfalgout:
If a region had html in it on it's first show, the new view would get appended to the preexisting html.

This fix also makes a convienent place to override the jQuery detach

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -196,11 +196,16 @@ const Region = MarionetteObject.extend({
 
   // Destroy the current view, if there is one. If there is no current view, it does
   // nothing and returns immediately.
-  empty(options) {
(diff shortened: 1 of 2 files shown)


ahumphreys87:
good catch. Just needs a test to prove the bug then 👍

paulfalgout:
Tests that fail without `allowMissingEl: true`
- Behaviors behavior UI within a layout "before each" hook for "should call onBeforeAttach"
- region when removing a region "before each" hook for "should be removed from the view"
- layoutView manipulating regions should trigger correct region remove events

ahumphreys87:
👍

denar90:
👍

PR #3040, opened 2016-06-02
Add auto deploy for www

denar90:
### Proposed changes
 -  changed travis build, after creating tag latest build for www will be refreshed.

Link to the issue:
https://github.com/marionettejs/marionettejs.com/issues/456

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,12 @@ install:
   - curl -L https://rawgit.com/marionettejs/travis-ci-setup/master/ci.sh | bash
 after_success:
   - if [[ $MAINRUN == true ]]; then npm run-script coverage; fi
+before_deploy:
+  - npm install travis-ci
+deploy:
+  provider: releases
+  api_key: $GH_TOKEN
+  on:
+    tags: true
+after_deploy:
+  - node trigger-deploy-mn-com.js
--- a/trigger-deploy-mn-com.js
+++ b/trigger-deploy-mn-com.js
@@ -1,0 +1,32 @@
+var Travis = require('travis-ci');
+var repo = 'marionettejs/marionettejs.com';
+var travis = new Travis({
+  version: '2.0.0',
+  headers: {
+    'User-Agent': 'Travis/1.0'
+  }
+});
+
+travis.authenticate({
+  github_token: process.env.GH_TOKEN
+}, function (err, res) {
+  if (err) {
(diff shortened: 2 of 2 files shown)


scott-w:
I think this is excellent, I'm a huge fan of auto-deploy on build 👍

paulfalgout:
What are the steps for setting up the token?

scott-w:
Does Travis have support for things like environment variables? I know for CircleCI I store the keys/passwords in the config on the server (outside the repo) and just load them in the deploy step.

denar90:
Yeap, just need to add `GH_TOKEN` to Env variables.

paulfalgout:
Alright. 👍   I added the `GH_TOKEN`  

Someone confident enough to merge this in and fix the website if it breaks can merge it :-D

scott-w:
@jdaudier @denar90 We should book some time we can collectively dedicate to rolling this change out, checking it and a plan for fixing it before too many people notice if things go wrong ;)

scott-w:
Just thought I'd chase this up. I'm off work all next week so I can make some time to push this out. My plan to do so is:

1. Merge this change
2. Go through every page on the website
3. Count the number of non-working pages
4. If we're below the threshold, fix those pages and roll another change
5. If we go above the threshold, run a revert commit

Reverting this may require us to manually roll-back the website, however, so I'd need access to do that. I don't know enough about where/how it's hosted.

denar90:
The main problem is about testing this feature. To run this script we have to create some test tag in scope of `backbone.marionette` repository. But it can be rough to delete this tag because of `jsdelivr`. We have to find the way hot to delete versions tagged version from `jsdelivr` and after that test it. 
@ahumphreys87 please correct me if I'm wrong.

These changes don't do any commits to any repos. This functionality just triggering `Travis CI` build for our www.
About rolling back our www it's not a problem at all.  We have to just click reload build process in `Travis CI` if something goes wrong.
I'm really eager to merge this stuff but we have to be confident that we can test it immediately.

scott-w:
I'm happy to do the testing myself. Like I say, I've off work all week, so I could make some time tomorrow to go through the docs and make sure everything works fine.

denar90:
Cool. Just ping me if you get some question.

scott-w:
Will do 👍

denar90:
🎉

PR #3042, opened 2016-06-03
Remove var from code

denar90:
### Proposed changes
 - make Mn consistent with using `var`, `const`, `let`. Somewhere we have only const/let, somewhere var/const, somewhere only var. I propose to refuse `var` at all and use only const/let.

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -45,8 +45,8 @@ import {
   triggerMethodOn
 } from './trigger-method';
 
-var previousMarionette = Backbone.Marionette;
-var Marionette = Backbone.Marionette = {};
+const previousMarionette = Backbone.Marionette;
+let Marionette = Backbone.Marionette = {};
 
 // This allows you to run multiple instances of Marionette on the same
 // webapp. After loading the new version, call `noConflict()` to
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -46,7 +46,7 @@ import {
 } from './trigger-method';
 
 const previousMarionette = Backbone.Marionette;
-let Marionette = Backbone.Marionette = {};
+const Marionette = Backbone.Marionette = {};
 
 // This allows you to run multiple instances of Marionette on the same
 // webapp. After loading the new version, call `noConflict()` to
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -102,7 +102,7 @@ const Behavior = MarionetteObject.extend({
   getEvents: function() {
     // Normalize behavior events hash to allow
     // a user to use the @ui. syntax.
-    let behaviorEvents = this.normalizeUIKeys(_.result(this, 'events'));
+    const behaviorEvents = this.normalizeUIKeys(_.result(this, 'events'));
 
     // binds the handler to the behavior and builds a unique eventName
     return _.reduce(behaviorEvents, function(events, behaviorHandler, key) {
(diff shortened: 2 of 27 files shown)


src/backbone.marionette.js
```diff
@@ -45,8 +45,8 @@ import {
   triggerMethodOn
 } from './trigger-method';
 
-var previousMarionette = Backbone.Marionette;
-var Marionette = Backbone.Marionette = {};
+const previousMarionette = Backbone.Marionette;
+let Marionette = Backbone.Marionette = {};
```
paulfalgout: This can also be `const` I think?

src/behavior.js
```diff
@@ -102,7 +102,7 @@ var Behavior = MarionetteObject.extend({
   getEvents: function() {
     // Normalize behavior events hash to allow
     // a user to use the @ui. syntax.
-    var behaviorEvents = this.normalizeUIKeys(_.result(this, 'events'));
+    let behaviorEvents = this.normalizeUIKeys(_.result(this, 'events'));
```
paulfalgout: I think `const` works here.. assuming `_.reduce` is immutable.

src/error.js
```diff
@@ -5,9 +5,9 @@ import _      from 'underscore';
 import extend from './utils/extend';
 import {version} from '../package.json';
 
-var errorProps = ['description', 'fileName', 'lineNumber', 'name', 'message', 'number'];
+const errorProps = ['description', 'fileName', 'lineNumber', 'name', 'message', 'number'];
 
-var MarionetteError = extend.call(Error, {
+let MarionetteError = extend.call(Error, {
```
paulfalgout: I think `const` works here

src/mixins/regions.js
```diff
@@ -89,9 +89,9 @@ export default {
   },
 
   _buildRegionFromObject: function(definition) {
-    var RegionClass = definition.regionClass || this.regionClass;
+    const RegionClass = definition.regionClass || this.regionClass;
 
-    var options = _.omit(definition, 'regionClass');
+    let options = _.omit(definition, 'regionClass');
```
paulfalgout: I think `const` works here.  Since options is an object the "constant" is just the reference to the object.

src/mixins/regions.js
```diff
@@ -151,7 +151,7 @@ export default {
   // Empty all regions in the region manager, but
   // leave them attached
   emptyRegions: function() {
-    var regions = this.getRegions();
+    let regions = this.getRegions();
```
paulfalgout: I think `const` works here

src/object.js
```diff
@@ -16,7 +16,7 @@ const ClassOptions = [
 
 // A Base Class that other Classes should descend from.
 // Object borrows many conventions and utilities from Backbone.
-var MarionetteObject = function(options) {
+let MarionetteObject = function(options) {
```
paulfalgout: I think `const` works here

src/template-cache.js
```diff
@@ -7,7 +7,7 @@ import MarionetteError from './error';
 
 // Manage templates stored in `<script>` blocks,
 // caching them for faster access.
-var TemplateCache = function(templateId) {
+let TemplateCache = function(templateId) {
```
paulfalgout: I think `const` works here

paulfalgout:
Nice!  Just had a few questions about some `let`s

denar90:
Good points. Thanks.

scott-w:
Excellent stuff, I'm a fan!

PR #3046, opened 2016-06-05
emptyView can accept a function that returns a view

paulfalgout:
In reviewing the documentation for v3's `CollectionView` this seemed to be fairly inconsistent.  It is also a moderately small change.  I'm fiddling with the `CollectionView` docs now on another branch, and if this is accepted I can make the doc updates there or it can be done in a subsequent PR.

This removes the need for `getEmptyView`

This makes they `emptyView` API match the `childView`

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -350,7 +350,7 @@ const CollectionView = Backbone.View.extend({
   // Internal method to show an empty view in place of a collection of child views,
   // when the collection is empty
   _showEmptyView() {
-    const EmptyView = this.getEmptyView();
+    const EmptyView = this._getEmptyView();
 
     if (EmptyView && !this._showingEmptyView) {
       this._showingEmptyView = true;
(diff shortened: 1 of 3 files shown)


scott-w:
I'm wondering whether this could be called in a number of instances:

1. The collection initializes to empty
2. The collection is updated from populated to empty
3. When a `sync` completes - the populated was empty before and is still empty

My thinking behind 3 is to allow the `emptyView` method to return a loading spinner or "there is nothing here" message depending on whether the fetch request completes.

paulfalgout:
@scott-w yeah probably could. To some degree this would be possible now with `getEmptyView` in v2.  But in v3 we removed `getChildView` and replaced it with `childView` accepting a function, but didn't follow that pattern for `emptyView`  This PR is more for consistency than it is a new feature, but if there's a use-case to achieve here, we can certainly do it.. But mainly I'm focused on making v3 clean before launch more than accounting for anything new.

scott-w:
Fair enough, I'm just musing aloud at this point. Looks like a solid change overall.

JSteunou:
Good call! :+1:

denar90:
👍

PR #3048, opened 2016-06-06
collectionView should only destroyChildren once rendered

paulfalgout:
While going through the `collectionView` docs I noticed `destroy:children` was getting triggered on the first render.  This is an unnecessary action as there will never be anything to destroy if nothing is rendered.

The composite view test that was modified helps prove the previous functionality was overkill.


denar90:
👍

JSteunou:
:+1:

Issue #3045, opened 2016-06-05
v3 childViewEvents destroy events can't trigger destroy


paulfalgout:
This API worked in v2 but not in v3-pre.4

https://jsfiddle.net/599dtsb7/9/

denar90:
Found the problem. So when [destroy](https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/view.js#L172) is calling it's calling region `empty` method which is deleting [view._parent](https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L223).
So when `_triggerEventOnParentLayout` was called childView doesn't have parent so all [this](_triggerEventOnParentLayout) logic which is responsible for `childViewEvents` wasn't run.

paulfalgout:
Yep you are correct.  A possible solution is to move the view._parent deletion until after the empty event is triggered, but I haven't tried that yet

denar90:
What do you think about [this](https://github.com/denar90/backbone.marionette/blob/ffa8dbe2e38f775f951fe1e8f3b88959cd723199/src/region.js#L286-L288) solution?

denar90:
.. should be `delete this.currentView._parent`

paulfalgout:
Oh I don't think the regions destroy is part of this.  In fact a region currently is almost never destroyed.  Something that I think should change in the future.

denar90:
fixed by #3049

PR #3049, opened 2016-06-06
delete view._parent after destroying view

denar90:
### Proposed changes
 - `delete view._parent` should be after `destroy` because when [destroy](https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/view.js#L172) is calling, it's calling region `empty` method which is deleting [view._parent](https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L223).
So when `_triggerEventOnParentLayout` was called childView doesn't have parent so all [this](_triggerEventOnParentLayout) logic which is responsible for `childViewEvents` wasn't run.

Link to the issue:
#3045

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -220,8 +220,6 @@ const Region = MarionetteObject.extend({
       this._removeView(view, options);
     }
 
-    delete view._parent;
-
     this.triggerMethod('empty', this, view);
     return this;
   },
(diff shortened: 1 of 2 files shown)


paulfalgout:
I think the most important part of this PR will be making a test to prove the solution.  I'm not certain that this change will fix the issue.  It seems like the delete is essentially happening at the same time.

paulfalgout:
👍

JSteunou:
@denar90 could you squash your commits please?

paulfalgout:
Github will let you squash on merge now fwiw

JSteunou:
Hoooo that's a very nice feature!

JSteunou:
Loving it! I already forgot how to do it in command line :D

Issue #3043, opened 2016-06-03
CollectionView elements added at incorrect index.


jaridmargolin:
### Description

See: https://jsfiddle.net/fLv382y7/3/
1. CollectionView does not mirror sorted collection order when adding new items. The side effects are hidden due to Marionette calling `render` or `reorder` after all items are added to the list.
2. Update CollectionView `_onCollectionAdd` handler to correctly updateIndices and set index of newly add items.
3. Currently when adding an item, its index is determined by its index within the collection. Unfortunately the view indices have not been updated to mirror the collection so there are inconsistencies in where the new view is actually added.

I have what feels like a pretty bulky solution implemented here. Current tests are passing however I have not added any new tests for the updates 😳 (I will happily write the tests assuming core team agrees withe the overall direction): https://github.com/jaridmargolin/backbone.marionette/commit/fd11f2798cc4cef1702a2b86f402724609dc1d82
### Expected behavior

I would expect the items to be added at the correct index, and not just "fixed" with a call to `render` or `reorder`...

In reference to the output provided by the jsfiddle above:

```
------------ 
add:child
collection: 34
view: 
------------ 
add:child
collection: 34
view: 
------------ 
render
collection: 34
view: 34
------------ 
add:child
collection: 1234
view: 234
------------ 
add:child
collection: 1234
view: 1234
```

where the `reorder` actually does not need to take place.
### Actual behavior

In reference to the output provided by the jsfiddle above:

```
------------ 
add:child
collection: 34
view: 
------------ 
add:child
collection: 34
view: 
------------ 
render
collection: 34
view: 34
------------ 
add:child
collection: 1234
view: 324
------------ 
add:child
collection: 1234
view: 1324
------------ 
reorder
collection: 1234
view: 1234
```
### Environment
1. Marionette version: Latest
2. Backbone version: Latest
3. Additional build tools, etc: N/A

PR #3050, opened 2016-06-06
Fix incorrect ordering on addition of item in sorted collectionview.

jaridmargolin:
Proposed solution for https://github.com/marionettejs/backbone.marionette/issues/3043

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -101,11 +101,20 @@ Marionette.CollectionView = Marionette.View.extend({
   _onCollectionAdd: function(child, collection, opts) {
     // `index` is present when adding with `at` since BB 1.2; indexOf fallback for < 1.2
     var index = opts.at !== undefined && (opts.index || collection.indexOf(child));
+    var models = collection.models;
 
     // When filtered or when there is no initial index, calculate index.
-    if (this.getOption('filter') || index === false) {
-      index = _.indexOf(this._filteredSortedModels(index), child);
+    if (this.getOption('filter') || this.getViewComparator() || index === false) {
+      models = this._filteredSortedModels(index);
     }
+
+    // find adjusted index
+    _.reduce(models, function(counter, model) {
+      var view = this.children.findByModel(model);
+      if (view) { return view._index; }
+
+      return model === child ? index = counter + 1 : counter;
+    }, -1, this);
 
     if (this._shouldAddChild(child, index)) {
       this.destroyEmptyView();
--- a/test/unit/sorted-views.spec.js
+++ b/test/unit/sorted-views.spec.js
@@ -55,6 +55,28 @@ describe('collection/composite view sorting', function() {
       });
     });
 
+    describe('when adding multiple models', function() {
+      beforeEach(function() {
+        this.sinon.spy(this.collectionView, 'render');
+        this.sinon.spy(this.compositeView, 'render');
+
+        this.collection.add([
+          new Backbone.Model({foo: 5, bar: 0}),
+          new Backbone.Model({foo: 4, bar: 1})
+        ]);
+      });
+
+      it('should have the order in the dom', function() {
+        expect(this.collectionView.$el).to.have.$text('12345');
+        expect(this.compositeView.$el).to.have.$text('12345');
+      });
+
+      it('should not require additional render', function() {
+        expect(this.collectionView.render.callCount).to.equal(0);
+        expect(this.compositeView.render.callCount).to.equal(0);
+      });
+    });
+
     describe('when adding a model with the "at" option', function() {
       beforeEach(function() {
         this.model = new Backbone.Model({foo: 0, bar: 1});


jaridmargolin:
Note: PR was opened up against patch, but is also relevant in v3.

src/collection-view.js
```diff
@@ -101,12 +101,21 @@ Marionette.CollectionView = Marionette.View.extend({
...
+      models = this._filteredSortedModels(index);
     }
 
+    // find adjusted index
+    _.reduce(models, function(counter, model) {
+      var view = this.children.findByModel(model);
+      if (view) { return view._index; }
+
+      return model === child ? index = counter + 1 : counter;
+    }, -1, this);
```
scott-w: The logic here feels a little tricky to follow. Can you tell me if I've understood it correctly?

1. You go through until you find the view to insert after
2. On subsequent passes, you just return that index

The reduce call doesn't get assigned to anything so I'm assuming that the index variable set on line 103 is already correct in this case?

The other case here is the view is new, so `view` is always `undefined`, in which case:

1. Compare model to child
2. When you find it, set index to the next item

I'm assuming that this second path is how the view gets added in the right place?

jaridmargolin: Hey @scott-w,

The reduce is used just to hold the counter as the models are iterated through. As for what is happening internally:

On every iteration we attempt to find the view corresponding to the iterated model. The `_index` of that view is then set as our current counter value so that when we finally hit our newly added model, we have an accurate representation of where it should lie relative to other rendered views.

JSteunou: Is the bug more likely to be inside `_filteredSortedModels`? I have the feeling this PR bring a patch on a single point but not on the root cause.

jaridmargolin: @JSteunou I believe that this is the correct location of the patch. The models are sorted correctly. The issue is specific to the index at which the view is added to the DOM.

paulfalgout:
All of the sort / filtering stuff still feels pretty brittle.  This may very well be a move in a good direction to fix an issue, but revisiting babysitters role in relation to sorting and filtering seems necessary sooner than later.. we seem to be keeping multiple indexes and collections and so when something changes too many other things have to update intelligently and things break on the edges.

src/collection-view.js
JSteunou: I'm sorry I need more lights; if the `models` returned by `_filteredSortedModels` why using `indexOf` on it do not suffice?

jaridmargolin:
@paulfalgout - I agree with you that the code is brittle when making changes. What I initially thought was going to be a simple fix took quite some time to get working as expected. On the bright side, the test suite is pretty robust, and caught a few issues as I was attempting to find the best place to introduce this patch.

With that said, I think merging this PR is still beneficial in the time being. "Revisiting" babysitters sounds like a large undertaking...

src/collection-view.js
jaridmargolin: @JSteunou the model index is not identical to the view `_index`. Imagine this use case quickly: I have a collection with two models:

```
{ val: 3 } # model index: 0 # view _index: 0
{ val: 4 } # model index: 1 # view _index: 1
```

now I add two more models:

```
{ val: 2 } # model index: 1
{ val: 1 } # model index: 0
```

At the point where the patch was introduced the state of the system looks like:

```
# we are inside of onCollectionAdd
# child = model where { val: 2 }

{ val: 1 } # model index: 0
{ val: 2 } # model index: 1
{ val: 3 } # model index: 2 # view _index: 0
{ val: 4 } # model index: 3 # view _index: 1
```

Now if we utilized only the model index as our "index" it would insert `{ val: 2 }` in the DOM in between `{ val: 3 }` and `{ val: 4 }`. We must utilize the view index to ensure the node is added at the correct location in the DOM....

jaridmargolin: As the code stands right now (prior to this PR), the node is added at the incorrect location in the DOM. It is then fixed down the road, by a render call triggered by `sortViews`. You can see this by the test case outlined below (remove the patch and you will see the tests below fail).

jaridmargolin: More than just saving a `render` call, this patch allows for client logic to be accurately applied in the CollectionView `onAddChild` method.

JSteunou: Got it, thank you for the explanation!

JSteunou:
Got my :+1:

scott-w:
Thanks for the explanation @jaridmargolin. Would it be worth putting a summary in commented form in there so it's easier to follow what's happening when if/when we revisit this?

I agree with @paulfalgout that we need to get our heads together for a clearer long-term plan for sorting/filtering. In the meantime I'm happy to 👍 this.

paulfalgout:
Yeah I'm not saying this shouldn't get in, but this isn't the first PR to solve a similar issue and there will be more unless we address the root.

However, I'm unsure about putting effort into another patch.  If someone wants to take responsibility for another `2.x` release at the moment, fine, but this'd be more important to get into `3.x`

jaridmargolin:
I will go ahead and add some context to why this happening via some nicely written comments.

As for the release... I would love to see a `2.x` as we are not currently utilizing `3.x`, but if the time required to prepare a release is too large, I can just continue to work off of my fork. I assume it can still be merged into both 2.x & 3.x?

rafde:
well... the thing about this in 3.x is I will end up resolving this issue since I have to deal with `_onCollectionUpdate` and `options.changes.added` which will have to update each index correctly. So this work in 3.x might be for naught.

scott-w:
@rafde What's the status of this PR?

rafde:
@scott-w this PR isn't valid anymore after 3.0
the update added PR is pending till I (or anyone with ideas) can think of a cleaner implementation. Haven't given that time, but my subconscious is still going.

@jaridmargolin I think this can be closed.

jaridmargolin:
I am still utilizing the 2.X build. Upgrading to 3.X is not going to be in the near future. If all development on 2.X has stopped then I am fine closing this. But if bug patches will continue to roll out on 2.X I would appreciate if this could be merged.

paulfalgout:
It's not inconceivable that 2.x could be updated, but so far we haven't seen enough interest to really push for it.  To actually fix all of the problems requires breaking changes that can never be applied to 2.x. Personally I think the best action here would be to write up a gist on how to patch 2.x and link to it in this issue.  If we had a lot more resources, maybe things will be different, but it clearly doesn't seem realistic that this will get the support it need to move forward.  @marionettejs/marionette-core ?

rafde:
I agree with the gist approach. We gotta move forward in improving Marionette from version 3 onward.

Issue #3047, opened 2016-06-05
Remove apply:filter event from v3 CollectionView#setFilter


paulfalgout:
### Description

This event only occurs when using `setFilter` and the filter is attempted.  The new filter may not actually change the collection. And if you prevent the render or the view has never been rendered, this event won't fire.  I think the use-case for this event is nearly non-existent and it has edge cases that are odd to explain.  This event is new to v3 and I think we should remove it now.  We can always add it later if there is need, but I do not think we should add to the API / documentation if we don't have to.

@marionettejs/marionette-core thoughts?

### Environment

1. Marionette version: v3

JSteunou:
Gotcha, :+1: to remove it, better now than having a cross to bear for all v3

scott-w:
Happy with that, I can make a note on the `CollectionView` ticket that I'll (eventually) raise to not document that event.

paulfalgout:
https://github.com/marionettejs/backbone.marionette/issues/3051 @scott-w 😀

scott-w:
:D

PR #3051, opened 2016-06-06
V3/collectionview docs

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3047

I can re-add the `apply:filter` event if necessary, but it was less work to remove it.  It is an untested event that is only in `v3` and has a limited use-case.

Some of the docs were far outdated.  The CompositeView docs described API that existed prior to my using Marionette. I attempted to reduce the docs as much as possible.  Particularly all of the `before:` and `onBeforeEvent` `onEvent` documentation felt repetitive. Now that events should be consistent across Marionette there should be less reason to document general event behavior for each event.

I also regrouped functionality first by API / option `childView` and related, `children` and related, `emptyView` and related, `render`ing and `destroy` then `filter` and last `sort`.

Notably there is still work to be done, particularly around shared view functionality. This functionality is currently described as `Marionette.AbstractView` but `AbstractView` does not exist. Once this shared documentation is resolved, some of the general events, render and destroy can most likely just reference the documentation there. Also greater detail should be given so that users fully understand that the shared functionality is available to the `CollectionView`

### DIFF
--- a/docs/marionette.collectionview.md
+++ b/docs/marionette.collectionview.md
@@ -9,13 +9,9 @@ then append the results of the child view's `el` to the collection view's
 `el`. By default the `CollectionView` will maintain a sorted collection's order
 in the DOM. This behavior can be disabled by specifying `{sort: false}` on initialize.
 
-CollectionView extends directly from Marionette.AbstractView. Please see
+`CollectionView` has the base functionality provided by the View Mixin. Please see
 [the Marionette.AbstractView documentation](marionette.abstractview.md)
 for more information on available features and functionality.
-
-Additionally, interactions with Marionette.Region
-will provide features such as `onShow` callbacks, etc. Please see
-[the Region documentation](./marionette.region.md) for more information.
 
 ## Documentation Index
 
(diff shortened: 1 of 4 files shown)


scott-w:
If this is still open at the weekend I'll make some time to go through it all.

docs/marionette.collectionview.md
```diff
@@ -300,44 +298,62 @@ var ParentView = Marionette.CollectionView.extend({
...
-  addChild: function(child, ChildView, index){
-    if (child.shouldBeShown()) {
-      Marionette.CollectionView.prototype.addChild.apply(this, arguments);
-    }
+  childView: ChildView,
+
+  childViewTriggers: {
+    'show:message': 'child:show:message',
+    'submit:form': 'child:submit:form'
+  },
```
scott-w: Is this specific to `CollectionView`?

docs/marionette.collectionview.md
```diff
@@ -451,649 +512,587 @@ var CollectionView = Marionette.CollectionView({
...
-var ListView = Marionette.View.extend({
-  tagName: 'ul',
-  className: 'list-unstyled',
+When a model is removed from a collection (or destroyed / deleted), the collection
+view will destroy and remove that model's child view.
 
-  childView: ListItemView
-});
+When the collection for the view is sorted, the view will automatically re-sort its child views.
+if the [`reorderOnSort`](#collectionviews-reorderonsort) option is set it will attempt to reorder the DOM and do this without a full re-render, otherwise it will re-render if the order has changed. Please Note that if you apply a filter to the collection view and the filtered views change during a sort then it will always re-render.
```
scott-w: Typo: `if` -> `If`

docs/marionette.collectionview.md
```diff
@@ -451,649 +512,587 @@ var CollectionView = Marionette.CollectionView({
...
-Tree structures are extremely useful layouts for nesting the same type of data
-over and over. A common example of this would be the Windows Explorer file
-picker.
+### CollectionView's `setFilter`
 
-#### Trees in Marionette 2
+The `setFilter` method modifies the `CollectionView`'s filter attribute, and
+renders the new `ChildViews` in a efficient way, instead of
+rendering the whole DOM structure again.
+A `preventRender` option will prevent the view to be rendered.
```
scott-w: This would read better as: Passing `preventRender: true` in the `options` argument will prevent the view being rendered.

scott-w:
👍 I'm happy to merge this in now and just roll the changes into the stuff I'm currently working on.

docs/marionette.collectionview.md
paulfalgout: agreed

docs/marionette.collectionview.md
paulfalgout: works for both

PR #3053, opened 2016-06-06
Reset removeChildview to be consistent with AddChildView

wesmangum:
### Proposed changes
 - The CollectionView's `removeChildView` doesn't return the view like it's sibling function, `addChildView`
 - Reseting `removeChildView` to return the childView keeps the functions consistent with each other
 - `collection-view.spec.js` is also reverted to test for the childView being returned

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -525,7 +525,7 @@ const CollectionView = Backbone.View.extend({
   // in the collection in order to keep the children in sync with the collection.
   removeChildView(view) {
     if (!view || view._isDestroyed) {
-      return;
+      return view;
     }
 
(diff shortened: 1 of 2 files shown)


paulfalgout:
As mentioned this was how v2 behaved.  It got removed while `_removeChildView` was briefly privatized.  No reason it shouldn't behave this way

👍

paulfalgout:
BTW @wesmangum is helping out over on https://github.com/paulfalgout/marionette-v3-compat so we might run into some other inconsistencies along the way.

JSteunou:
Nice, thank you for your help @wesmangum !

denar90:
thanks @wesmangum 👍

PR #3054, opened 2016-06-07
make function declaration consistent

denar90:
### Proposed changes
 - in some places we have `foo: function() {...}` in some `bar() {...}`, so I proposing to make it consistent in format `bar() {...}`

Link to the issue:
I think It's a bit related to https://github.com/marionettejs/backbone.marionette/issues/2858
So for Classes we can have `bar() {...}` declaration and for utils 
```
function foo() {
  ...
}

function bar() {
  ...
}
export {foo, bar}
```
as it is now.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -54,7 +54,7 @@
     "no-undef": 2,
     "no-unused-vars": [ 1, { "args": "none" } ],
     "no-with": 2,
-    "one-var": [ 2, "never" ],
+    "no-var": 2,
     "operator-linebreak": [ 2, "after" ],
     "quotes": [ 2, "single" ],
     "semi": [ 0, "never" ],
@@ -71,6 +71,7 @@
     "strict": 0,
     "valid-jsdoc": 2,
     "valid-typeof": 2,
-    "wrap-iife": [ 2, "inside" ]
+    "wrap-iife": [ 2, "inside" ],
+    "object-shorthand": [ 2, "methods" ]
   }
 }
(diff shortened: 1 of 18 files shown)


paulfalgout:
Works for me.  We could enforce this with the method option of the eslint rule: http://eslint.org/docs/rules/object-shorthand#options

JSteunou:
+1 on the eslint rule

.eslintrc
```diff
@@ -71,6 +71,7 @@
     "strict": 0,
     "valid-jsdoc": 2,
     "valid-typeof": 2,
-    "wrap-iife": [ 2, "inside" ]
+    "wrap-iife": [ 2, "inside" ],
+    "object-shorthand": [ 2, "methods" ]
```
paulfalgout: sorry to be picky, but can you add this alphabetically?

.eslintrc
```diff
@@ -54,7 +54,7 @@
     "no-undef": 2,
     "no-unused-vars": [ 1, { "args": "none" } ],
     "no-with": 2,
-    "one-var": [ 2, "never" ],
+    "no-var": 2,
```
paulfalgout: same here, can you move it up one?

.eslintrc
denar90: Sure. Fair enough. It was me who didn't notice that they are sorted alphabetically.

paulfalgout:
👍

JSteunou:
:+1:

Issue #2989, opened 2016-04-22
Feature request — bindEntityReplies


rogovdm:
`bindEntityReplies` is an analogue to `bindEntityEvents`, but will bind / unbind within a channel using `.reply` and `.stopReplying` functions accordingly.

paulfalgout:
https://github.com/marionettejs/backbone.marionette/blob/next/src/bind-radio-requests.js#L34

rogovdm:
@paulfalgout oh, cool.

Are we sure that it's exactly bind-radio-*requests*? What we are binding in that code are replies, not requests

rogovdm:
Also, why to "limit" it's usage to only radio objects by naming it like that? It will work with any object, that has reply method with specific signature.

So my proposal to name this util functions like this: 
- bindEntityReplies
- unbindEntityReplies

paulfalgout:
I _think_ the naming comes from the key and not the value such that `{ "request:name": "replyHandler" }` we're binding the "request:name" to the reply handler.  The same could be said about "events" where the event is the key of the hash and the value is the handler.

It is true it'll work with any object with the reply API. I don't disagree with your suggestion either. However I think the naming is related to promoting Radio.  I'd have to dig back through the issue related to radio a bit, but for instance `bindEntityEvents` will be aliased as `bindRadioEvents` as well.  https://github.com/marionettejs/backbone.marionette/blob/next/src/backbone.marionette.js#L63

@marionettejs/marionette-core I'd actually support dropping the radio specific naming of the new API in favor of just using `bindEntityRequest` and not introducing an alias.  Thoughts?

Issue #3055, opened 2016-06-08
Update Documentation for Functions


denar90:
We should look through removed methods and docs for new one.

Were removed:
 - `proxyGetOption` 
 - `proxyBindEntityEvents`
 - `proxyUnbindEntityEvents`
 - `normalizeUIKeys`
 - `normalizeUIValues`
 - `actAsCollection`

Were added:
 - `bindRequests`
 -  `unbindRequests`
 - `deprecate`
 - `isEnabled`
 - `setEnabled`
 - `monitorViewEvents`

scott-w:
Are these related to any topic in particular? It'll be good to have a think about how this fits into #3066. It should be fine to put this into the `marionette.functions.md` file for now, however.

denar90:
I think we should leave this file as it is between `Templates and Rendering` and `Errors` topics. What do you think about name `API` instead of `Functions`?

scott-w:
API could work, or as part of "Advanced Topics"?

paulfalgout:
soo..
I don't think we necessarily need to document `deprecate`

`monitorViewEvents` can be / will be documented as part of the View lifecycle
`isEnabled` / `setEnabled` should go along with installation / configuration I would think.. this is for feature flags...  _although_ we might want feature flags to have its own section now, as depending on how often they get used, that could be a significant bit of documentation.
bind/unbind seems perhaps like advanced topics.  I'm indifferent to where they go, but they're definitely for someone who is digging deeper..  Though a lot of it will be anything but "advanced" as far as difficulty or complexity..  for instance `isNodeAttached` is a very simple utility.  but as far as "advanced" means anything beyond the basics of what you need to know to get started.. 👍

denar90:
resolver by #3068

PR #3056, opened 2016-06-08
rename bindEntityEvents->bindEvents

denar90:
### Proposed changes
 - bindEntityEvents became bindEvents, 
 - bindRadioRequests became bindRequests

Link to the issue:
#2989 

Also maybe we should think about removing `bindRadioEvents`, `unbindRadioEvents` because they are identical to `bindEntityEvents/unbindEntityEvents`?  What do you think guys?

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -60,8 +60,6 @@ Marionette.noConflict = function() {
 // Utilities
 Marionette.bindEvents = proxy(bindEvents);
 Marionette.unbindEvents = proxy(unbindEvents);
-Marionette.bindRadioEvents = proxy(bindEvents);
-Marionette.unbindRadioEvents = proxy(unbindEvents);
 Marionette.bindRequests = proxy(bindRequests);
 Marionette.unbindRequests = proxy(unbindRequests);
 Marionette.mergeOptions = proxy(mergeOptions);
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -31,14 +31,14 @@ import {
 } from './config/features';
 
 import {
-  bindEntityEvents,
-  unbindEntityEvents
-} from './bind-entity-events';
+  bindEvents,
+  unbindEvents
+} from './bind-events';
 
 import {
-  bindRadioRequests,
-  unbindRadioRequests
-} from './bind-radio-requests';
+  bindRequests,
+  unbindRequests
+} from './bind-requests';
 
 import {
   triggerMethod,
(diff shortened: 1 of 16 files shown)


paulfalgout:
Oh yeah.. I think if we're doing this bindRadioEvents doesn't need to exist.  can just use `bindEvents`

src/bind-events.js
```diff
@@ -39,7 +39,7 @@ function iterateEvents(target, entity, bindings, actionName) {
   if (!_.isObject(bindings)) {
     throw new MarionetteError({
       message: 'Bindings must be an object.',
-      url: 'marionette.functions.html#marionettebindentityevents'
+      url: 'marionette.functions.html#marionettebindEvents'
```
paulfalgout: I don't think the url will have an uppercase E

test/unit/bind-radio-request.spec.js
```diff
@@ -76,14 +76,14 @@ describe('Marionette.bindRadioRequests', function() {
...
-        Marionette.bindRadioRequests(this.target, this.channel, 'replyFooStub');
+        Marionette.bindRequests(this.target, this.channel, 'replyFooStub');
       }.bind(this);
     });
 
     it('should error', function() {
       expect(this.run).to.throw(Marionette.Error, new Marionette.Error({
         message: 'Bindings must be an object.',
-        url: 'marionette.functions.html#marionettebindradiorequests'
+        url: 'marionette.functions.html#marionettebindRequests'
```
paulfalgout: again these'll match links on the website that are generated from the `docs/*.md` and so I think those `#hash` are all lower case.

That does point out something though.. the docs need to be updated as well to reflect these changes.

denar90: What do you think if I change doc for `bindEvents` but for others in scope of task #3055 ?

src/backbone.marionette.js
```diff
@@ -58,12 +58,12 @@ Marionette.noConflict = function() {
...
 // Utilities
-Marionette.bindEntityEvents = proxy(bindEntityEvents);
-Marionette.unbindEntityEvents = proxy(unbindEntityEvents);
-Marionette.bindRadioEvents = proxy(bindEntityEvents);
-Marionette.unbindRadioEvents = proxy(unbindEntityEvents);
-Marionette.bindRadioRequests = proxy(bindRadioRequests);
-Marionette.unbindRadioRequests = proxy(unbindRadioRequests);
+Marionette.bindEvents = proxy(bindEvents);
+Marionette.unbindEvents = proxy(unbindEvents);
+Marionette.bindRadioEvents = proxy(bindEvents);
```
paulfalgout: I think it is perfectly ok to remove `bindRadioEvents` here as it matches the renaming of `bindRadioRequests`

paulfalgout:
👍  I personally like how this turned out a lot.  And it reduces the v3 API which is a big plus.

JSteunou:
Totally in, that's a great simplification! :+1:

Issue #3037, opened 2016-06-01
Marionette v3 should use any patch version as the peer dependency


paulfalgout:
The peerDependencies for `Bb` and `_` should be `1.3.x` and `1.8.x`

JSteunou:
peerDependencies or direct dependencies?

That aside I agree, bb and _ team seems to be more careful about semver than 2 years ago :)

paulfalgout:
`v3` is peerDependencies.  Bb and _ still isn't following semver, but if there's breaking changes they're saying it won't be on the "patch" number.

JSteunou:
With some `import from`, do bb and _ should be direct dependencies?

paulfalgout:
because Bb and _ do not follow semver, in order to follow semver with Mn, direct dependencies make it a nightmare anytime either Bb or _ update.  People will install newer versions of Bb or _ regardless of what versions Mn defines as a direct dependency, and then not understand why their app breaks.  With peerDependencies, people will set Bb and _ as dependencies in their own projects, which for the most part people are doing already anyways, and if they attempt to use a version outside of the officially supported versions, it will let them and it might work just fine.. but it will warn them.

JSteunou:
Right! It seems we already discussed about that because it rings a bell :D

ahumphreys87:
I think they are following semver a bit better these days cant we go for `^1.3.0` etc?

JSteunou:
Might be to lose. Wont go this far.

denar90:
@ahumphreys87 wouldn't `^1.3.0` cost us problems if in future Bb will release v2 with breaking changes?

paulfalgout:
for Bb v2 we'd probably want to do a major bump anyhow.. but `^1.3.0` === `1.x.x` and is set to v1.  But I don't think Bb is really following semver so much.. they're attempting to not introduce breaking changes.. particularly into the patch version, but there's absolutely no saying that at the last moment of building a `v2` that ashkenas wouldn't make the call that it wasn't worth a major version and would call it `1.5` and break everyones apps.

However.. with peerDependencies people can run _any_ version they want.. It'll just warn them if they're outside the "supported" versions.  I like this personally as things will probably work, but if there are breaking changes in Bb we don't have to officially support the change right away.

PR #3058, opened 2016-06-10
use any patch version as the peer dependency

denar90:
### Proposed changes
 - use any patch version as the peer dependency

Link to the issue:
#3037


JSteunou:
:+1:

PR #3059, opened 2016-06-10
add more es6

denar90:
### Proposed changes
 - add more consistency of es6 instead of es5.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -339,7 +339,7 @@ const CollectionView = Backbone.View.extend({
 
   _sortModelsBy(models, comparator) {
     if (typeof comparator === 'string') {
-      return _.sortBy(models, function(model) {
+      return _.sortBy(models, (model) => {
         return model.get(comparator);
       });
     } else if (comparator.length === 1) {
@@ -482,7 +482,7 @@ const CollectionView = Backbone.View.extend({
     }
 
     // update the indexes of views after this one
-    this.children.each(function(laterView) {
+    this.children.each((laterView) => {
       if (laterView._index >= view._index) {
         laterView._index += increment ? 1 : -1;
       }
@@ -587,7 +587,7 @@ const CollectionView = Backbone.View.extend({
   // Create a fragment buffer from the currently buffered children
   _createBuffer() {
     const elBuffer = document.createDocumentFragment();
-    _.each(this._bufferedChildren, function(b) {
+    _.each(this._bufferedChildren, (b) => {
       elBuffer.appendChild(b.el);
     });
     return elBuffer;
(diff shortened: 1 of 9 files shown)


src/utils/deprecate.js
```diff
@@ -25,7 +25,7 @@ const deprecate = function(message, test) {
 
 deprecate._console = typeof console !== 'undefined' ? console : {};
 deprecate._warn = function() {
-  const warn = deprecate._console.warn || deprecate._console.log || function() {};
+  const warn = deprecate._console.warn || deprecate._console.log || Function.prototype;
```
JSteunou: Is `Function.prototype` es6 specific? Cause es6 new "methods" wont be translated by babel.

paulfalgout: actually in v3 since we don't support really old underscore this could change to `_.noop`

denar90: @JSteunou no) I just like `Function.prototype` more then `function() {}`. I can return how it was.

paulfalgout: Like this deprecate was pulled over from `backbone-metal` and it originally used `_.noop` until it was tested against older underscore:  https://github.com/marionettejs/backbone-metal/commit/d16fe23cb2dfe0bd9308a5e2c440c2247a9e7788#diff-2b7c460ee7b16a2db70ff8cc661ee9cdL482

src/collection-view.js
```diff
@@ -339,7 +339,7 @@ const CollectionView = Backbone.View.extend({
 
   _sortModelsBy(models, comparator) {
     if (typeof comparator === 'string') {
-      return _.sortBy(models, function(model) {
+      return _.sortBy(models, (model) => {
```
yethee: Here, you can use short form of arrow function:

```js
return _.sortBy(models, model => model.get(comparator));
```

blikblum:
As i already pointed in other PR, ES6 is a superset of ES5 meaning that the old syntax is perfect valid ES6 and did not became deprecated or obsolete.

Personally i would do such changes only if there's a increase in code clarity.

IMO changing from `function(model) {` to `(model) =>` does not add in this matter. 

Anyway, unlike the PR i replied, this does not the change / increase the compiled code, so in the end is a matter of taste.

src/collection-view.js
denar90: @yethee of course, but i tried to write in one style like it was done before in other places
e.g.
```js
models = _.filter(models, (model, index) => {
        return this._shouldAddChild(model, index);
      });
```

denar90:
It's only about consistent code.

scott-w:
I'm in favour of this so I'll add my 👍

Issue #2895, opened 2016-02-16
Region and CollectionView housekeeping


ianmstew:
A bit of housekeeping following #2850.

- [x] Add test to check that detach for an entire tree occurs before descendants are destroyed
- [x] ~~Shore up test coverage on CollectionView and Region.~~
- [x] Move `monitor-view-events` to `utils` or [some alternative] (https://github.com/marionettejs/backbone.marionette/pull/2850#discussion_r47015703).
- [x] Decapitalize MonitorViewEvents.
- [x] Create docs update ticket for API changes
- [x] Finalize [`_isDestroyed` call](https://github.com/marionettejs/backbone.marionette/pull/2850#discussion_r52967579) and potentially remove comments [like these](https://github.com/marionettejs/backbone.marionette/blob/next/src/object.js#L44).

paulfalgout:
Test coverage can go https://github.com/marionettejs/backbone.marionette/issues/2581

JSteunou:
Might be room for region optimization. I see in the constructor `this.el` [can be String or Object](https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L23) then `this.$el` [is created](https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L32), but can be `length === 0`. Why [calling `getEl` again](https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L107) in the `_ensureElement` called by `show` and why condition it to `this.el` being a String? `this.el` could be a String but as I see in the constructor `this.$el` could already be fine. This make region use jQuery selector a recreate `this.$el` one time too many. I would understand a condition with `this.$el.length === 0`.

paulfalgout:
I've been looking into this a bit..  as it is, it does attempt to not recall the `getEl` inside `_ensureElement`, but it is almost never effective at doing so.
most of the time the `$el` isn't available when the region is created, and the regions stick around on re-renders of the layout so it has to get recalculated possibly at the moment you need it.

I think it may actually be better to remove `_ensureElement` and have the region require the $el in DOM at instantiation..  and this would mean a smarter layout that would build the regions on the view's render, or even better perhaps at the point of `getRegion` The 2nd option is nice, but it might mess with people who are using regions beyond a simple container.  Then instead of keeping the region around a resetting it, they're just destroyed and rebuilt on subsequent renders.. (or again even better. detached / reattached and destroyed if they can't find the el they need)

[...]

PR #3060, opened 2016-06-11
check if detach was called before destroy

denar90:
### Proposed changes
 - Add test to check that detach for an entire tree occurs before descendants are destroyed

Link to the issue:
#2895
think after merging this PR we can close issue.

### DIFF
--- a/test/unit/on-attach.spec.js
+++ b/test/unit/on-attach.spec.js
@@ -17,6 +17,7 @@ describe('onAttach', function() {
       sinon.spy(this, 'onBeforeAttach');
       sinon.spy(this, 'onDetach');
       sinon.spy(this, 'onBeforeDetach');
+      sinon.spy(this, 'onDestroy');
     },
     onAttach() {
       return !!this._isAttached;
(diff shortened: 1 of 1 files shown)


paulfalgout:
Ah sorry a little bit more info here.  I think the test needs to cover descendants.. so like:

```js
const MyDetachView = Mn.View.extend({
  template: false,
  onDetach() {
    console.log('gets detached before destroyed');
  },
  onDestroy() {
    console.log('after detach');
  }
});
var MyChildView =  = Mn.View.extend({
  template: _.template('<div id="region"></div>'),
  regions: {
    'region': '#region'
  },
  onAttach() {
    this.showChildView('region', new MyDetachView());
  },
  onDetach() {
    console.log('gets detached before destroyed');
  },
  onDestroy() {
    console.log('after detach');
  }
});
var MyCollectionView = Mn.CollectionView.extend({
  childView: MyChildView
});

// show collectionview in a region and then empty the region.
```

paulfalgout:
:±1:

paulfalgout:
merging this test to move 3.0 along.  We can adjust it as necessary if need be in a new PR.

PR #3061, opened 2016-06-11
Advanced event doc

scott-w:
### Proposed changes
 - Adds event docs in their own where they can be linked from `View` and `CollectionView`
 - Adds the missing `events` and `triggers` attributes to `View`
 - Added notice to the Upgrade Guide

Link to the issue: #2920 #2971

### DIFF
--- a/docs/marionette.events.md
+++ b/docs/marionette.events.md
@@ -408,3 +408,9 @@ var GrantParentView = Marionette.View.extend({
   }
 });
 ```
+
+## Lifecycle Events
+
+Marionette Views fire events during their creation and destruction lifecycle.
+For more information see [`View` Lifecycle](./marionette.view.md#lifecycle-events)
+and [`CollectionView` Lifecycle](./marionette.collectionview.md#collectionview-events-and-callbacks) events.
--- a/docs/marionette.events.md
+++ b/docs/marionette.events.md
@@ -24,6 +24,7 @@ responding to events.
 * [Child View Events](#child-view-events)
   * [Event Bubbling](#event-bubbling)
   * [Explicit Event Listeners](#explicit-event-listeners)
+  * [Triggering Events on Child Events](#triggering-events-on-child-events)
 * [Lifecycle Events](#lifecycle-events)
 
 ## Triggering and Listening to Events
@@ -173,7 +174,7 @@ var Item = Mn.View.extend({
 var Collection = Mn.CollectionView.extend({
   tagName: 'ul',
 
-  childEvents: {
+  childViewEvents: {
     'select:item': 'itemSelected'
   },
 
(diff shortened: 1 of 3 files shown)


scott-w:
This should be in a decent spot to merge once we're happy with the changes

docs/marionette.events.md
```diff
@@ -0,0 +1,416 @@
...
+The traditional [event handling system in Backbone](http://backbonejs.org/#Events)
+is also supported in Marionette. Marionette, however, provides an alternative
+event system using the `triggerMethod` method on `Marionette.Object` - the key
+difference between the two is that `triggerMethod` triggers magically named
+event handlers on views. This section covers how `triggerMethod` works and how
+listeners are set up to handle it.
+
+### View `triggerMethod`
+
+The `triggerMethod` method fires the named event on the view - any listeners
```
rafde: this is one of those formatting questions on whether it should be 
> the view - any

or

> the view - any listeners will then ...

docs/marionette.events.md
```diff
@@ -0,0 +1,416 @@
...
+  triggers: {
+    'click a': 'select:item'
+  }
+});
+
+var Collection = Mn.CollectionView.extend({
+  tagName: 'ul',
+
+  childViewEvents: {
+    'select:item': 'itemSelected'
```
rafde: This is probably bikesheddy but is there a reason `select:item': 'onSelectItem'` was not used instead. People tend to look for a standard naming convention for this stuff to avoid having to over think naming stuff.

docs/marionette.view.md
```diff
@@ -882,6 +886,149 @@ view.addRegion("foo", "#bar");
...
+[Magic Method Binding](./marionette.events.md#magic-method-binding) technique
+discussed in the [events documentation](./marionette.events.md).
+
+##### Defining Listeners
+
+Listeners are defined as:
+
+```javascript
+eventname jqueryselector
+```
```
rafde: from [here](https://github.com/marionettejs/backbone.marionette/pull/3061/files#diff-754a14ba135eb5c5b22e5325d9d4092fR1001)
to [this line](https://github.com/marionettejs/backbone.marionette/pull/3061/files#diff-754a14ba135eb5c5b22e5325d9d4092fR1029) looks the same if no similar to 
https://github.com/marionettejs/backbone.marionette/pull/3061/files#diff-754a14ba135eb5c5b22e5325d9d4092fR926

docs/upgrade.md
```diff
@@ -55,6 +55,10 @@ Bubbled child events no longer pass the `childView` implicitly and only pass the
 arguments passed as part of `triggerMethod`. This means that the arguments
 passed to `onEvent` and `onChildviewEvent` are now identical.
 
+In Marionette 2, `childEvents` where bound on every event. In Marionette 3,
+`childViewEvents` are bound once and cached. This means that you cannot add new
```
rafde: should
> This means that you cannot add new events after the view has been created.

be
> This means that the parent views cannot listen to new childViewEvents after the view has been created.

Since thats the case, is there a way for someone to bind and cache them themselves?

docs/marionette.view.md
```diff
@@ -882,6 +886,149 @@ view.addRegion("foo", "#bar");
...
+var Mn = require('backbone.marionette');
+
+var MyView = Mn.View.extend({
+  triggers: {
+    'click a': 'link:clicked'
+  },
+
+  onLinkClicked: function() {
+    console.log('Show the modal');
+  }
```
rafde: `Backbone.View` accepts `events` to get passes to it on instantiation, does Marionette allow for the same behavior?
example
```
new MyView({
  events:{
    'click a': 'noProblem'
  },
  triggers: {
    'click a': 'allowable' //?
  }
});
```

docs/marionette.events.md
scott-w: I usually find it best to name the callbacks the thing that will happen when the event fires, so there's no real standard naming convention here.

It also reinforces that there's no relation to the `triggerMethod` result that could be `select:item`.

docs/marionette.view.md
scott-w: I'm not sure what you mean here. Are you asking if you can bind both an event and a trigger to the same selector?

scott-w: Ahh I see. I'll check the code over the weekend.

rafde: spoiler alert
looks like there is https://github.com/marionettejs/backbone.marionette/blob/v3.0.0-pre.4/src/view.js#L23

But isn't anything indication that you can do that in the docs

scott-w: This might be best as part of the `basics.md` file being worked on in #3076.

docs/upgrade.md
scott-w: That's a good spot. My understanding is that you can do it manually so perhaps this should change to "Marionette will not automatically listen to any new new events after the view has been created". What are your thoughts?

rafde: That's sounds a lot more explicit. I just noticed that > where bound

Should be

> were bound

I suggest mentioning that it is still possible to bind events outside of using the hash. I can provide an example tomorrow. Typing this on my phone.

rafde: Actually, nm. It's out of context for this file.

rafde:
Welp, I got nothing. It'd 👍 if it was worth anything.

denar90:
👍

PR #3064, opened 2016-06-11
Reorg source files

paulfalgout:
This one might be a hot topic, or it could be easy.  Had the idea to use "common" after discussing documentation strategies with @scott-w 

Resolves: https://github.com/marionettejs/backbone.marionette/issues/2858

**Utils:** anything strictly internal
**Common:** shared functions that aren't necessarily a mixin and may be exposed on `Marionette` itself
**Config:** Anything separated for the purpose of being overridden.
**Mixin:** grouped methods for mixing into classes
Root `src/` dir is documented classes.  `Error` and `TemplateCache` being the outliers, though certainly stand-alone classes that _could_ be used directly by others.


paulfalgout:
To make this PR easier to understand this moves the files into the following structure:

- 📂  src
  - 📁 common
  - 📁 config
  - 📁 mixin
  - 📁 utils
  - app-router
  - application
  - **backbone.marionette**
  - behavior
  - collection-view
  - composite-view
  - error
  - object
  - region
  - template-cache
  - view

- 📂 common
  - bind-events
  - bind-requests
  - getOption
  - isNodeAttached
  - mergeOptions
  - monitor-view-events
  - normalizeMethods
  - trigger-method

- 📂 config
  - behavior-lookup
  - features
  - renderer

- 📂 mixins
  - behaviors
  - common
  - delegate-entity-events
  - radio
  - regions
  - triggers
  - ui
  - view

- 📂 utils
  - deprecate
  - destroyBackboneView
  - extend
  - getUniqueEventName
  - invoke
  - proxy
  - setOptions

I will say one inconsistency we could fix here.. is some files are dash-separated and some are camelCased.  Thoughts there?  might be worth renaming as part of this PR.

@marionettejs/marionette-core

denar90:
I really like it. Great job. I'm for 👍 dash-separated way.

paulfalgout:
Good deal I added a commit to make them all dash-separated.  If it is contested or even holding up this PR I can pull it and we can deal with it later

denar90:
Definitely 👍

ahumphreys87:
:+1: kebab-case ftw - case sensitivity gets me all the time when I switch platform!

PR #3067, opened 2016-06-12
Upgrade guide

scott-w:
Added some comments about child views and triggers to the upgrade guide.

Link to the issue: #2971

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -45,10 +45,11 @@ for detail on upgrading to Marionette 3. This technique works in both Marionette
 
 The `childEvents` attribute was renamed to `childViewEvents`.
 
-Child view events now only bubble a single level to their direct parent. To
-pass events further up the chain, you must explicitly using `childViewTriggers`
-to convert the event from the child into an event on the parent. These can be
-chained all the way up to the level you require them to be.
+Child event bubbling above one level is now removed in most instances and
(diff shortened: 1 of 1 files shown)


docs/upgrade.md
```diff
@@ -45,6 +45,20 @@ for detail on upgrading to Marionette 3. This technique works in both Marionette
 
 The `childEvents` attribute was renamed to `childViewEvents`.
 
+Child view events now only bubble a single level to their direct parent. To
```
paulfalgout: Actually this unfortunately isn't true.  multi-level bubbling still occurs in some instances, although this was never meant to happen.

scott-w: In that case, would it be more accurate to state that "Child event bubbling above one level is now removed in most instances and deprecated pending removal in others. This can no longer be relied upon." perhaps?

paulfalgout: So Marionette has never intended on including event bubbling beyond one level.  It was added by accident.. only for certain situations when combining layouts with collectionviews. https://github.com/marionettejs/backbone.marionette/issues/2667#issuecomment-130537382  This accidental behavior has yet to be removed.  Personally I'm hoping in the future to entirely remove the auto-proxying and have people rely on `childViewTriggers` for explicit.. way less noisy.. events.

denar90:
👍

Issue #3066, opened 2016-06-12
Restructure Documentation


scott-w:
Now that we have a good chunk of documentation in place, it's becoming clearer that we need to break up the documentation into different topics that can be more easily referenced from each other. The idea is to make the individual documentation pages shorter (Both `View` and `CollectionView` are over 1000 lines) and minimise duplication within the documentation.

### Topics

1. Install/Getting Started
2. Basic and Common concepts
3. Views
4. Collection Views
5. Events
6. Regions
7. Application
8. Behaviors
9. Radio
10. View Lifecycle
11. Routing
12. Object
13. Templates and Rendering
14. Errors
15. Upgrade from 2 to 3

### Discoverability

I think it's going to be really important to ensure that readers can find the documentation they need. This is going to involve some duplication - both the `View` and `CollectionView` docs will contain simple examples from the Events, Regions, Behaviors, and Radio documentation. We'll need to be conscious of this when we're updating features and documentation for those features.

### Discussion Points

I'd appreciate feedback on the topics - is there anything missing or unnecessary here? Does the order look alright?

paulfalgout:
Might be worth putting application, object, and radio together..  The rest of the docs are view related.   We also need a doc for routing.  Possibly suggesting options, but AppRouter is still in v3.

scott-w:
That's a good point that I've missed the `AppRouter`. We could roll `Object` into an Advanced topics for now with #3055. We can roll `Radio` into both `Application` and `Object` with a link to the official `Radio` documentation.

Another option could be to get a brief description of `Application` into the Getting Started overview with all the extra detail going into the Application/Object/Radio documentation merged as you suggest.

rafde:
Mission Accomplished, @scott-w ?

scott-w:
Yeah, I think it's safe to close this and we can open specific tickets for the extra bits needed.

Issue #3044, opened 2016-06-05
Deprecate Behaviors (not a behavior) API in v3


paulfalgout:
We should deprecate pretty much everything in this doc.
https://github.com/marionettejs/backbone.marionette/blob/next/docs/marionette.behaviors.md

I suggest in v3 we talk about the best ways to use behaviors in the `behavior.md` file and steer people away from the `behaviors.md`

We should look into eventually removing the behaviors lookup altogether.

Module loaders have really made the behavior lookup mostly unnecessary and we can quite easily provide a simple recipe for those who really want it.  

Similarly a behavior array pattern was introduced to simplify things which is also been made unnecessary in ES6.

In the simplest case, this is all a behavior needs.
```js
import someBehavior from 'some-behavior';

const MyView = Mn.View.extend({
  behaviors: {
    someBehavior
  }
});
```

scott-w:
Is that meant to be:

```javascript
import someBehavior from 'some-behavior';

const MyView = Mn.View.extend({
  behaviors: {
    keyName: someBehavior
  }
});
```

Or:

```javascript
import someBehavior from 'some-behavior';

const MyView = Mn.View.extend({
  behaviors: [
    someBehavior
  ]
});
```

paulfalgout:
Well the array notation was added to prevent having to write out a key, but in es6 with shorthand object notarion the key is implied and takes the value's variable name.  Makes the array notation unnecessary

JSteunou:
Unnecessary in es6, but have mercy for those still in es5 ;)

paulfalgout:
Well certainly... this'd just be deprecation for the next major.. which is probably early 2017?  Then those in es5 would just need to have to add a `keyName:` to each entry.  Probably worth removing the extra API / code in the behaviors..  We can check it out later, but I think the source gets way simpler.

scott-w:
In terms of docs, we're doing everything in ES5 syntax anyway.

scott-w:
Just making some minor amends to the current documentation and, if array notation is supported, we might want to make it the default in all cases. I've not found the hash key that important in applications I've made.

paulfalgout:
The key is used to determine which behavior to load from the Behavior lookup.  And it is of course necessary if you want to pass any additional options to the behavior.

scott-w:
[...]

PR #3072, opened 2016-06-18
Cleaned Behaviors

scott-w:
### Proposed changes
 - Remove redundant documentation with links to View/Event docs
 - A top-level view of how to use `Behavior`
 - Complete example of how to change a `View` into `Behavior`s

Link to the issue: #3044 #3066

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -4,12 +4,15 @@ not be accurate or up-to-date_**
 # Marionette.Behavior
 
 
-A `Behavior` is an  isolated set of DOM / user interactions that can be mixed into any `View` or another `Behavior`. Behaviors allow you to blackbox `View`-specific interactions into portable logical chunks, keeping your Views simple and your code DRY.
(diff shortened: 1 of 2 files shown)


docs/marionette.behavior.md
```diff
@@ -3,77 +3,166 @@ not be accurate or up-to-date_**
...
 
-These interactions tend to be chunks of logic that you want to use in multiple views.
-
-## Usage
-
-Here is an example of a simple `View`. Let's take a stab at simplifying it, and abstracting Behaviors from it.
-
-```js
+```javascript
 var MyView = Marionette.View.extend({
```
denar90: Would you like to use `var Mn = require('backbone.marionette');` instead if `Marionette` from global scope? In examples below `var Mn = require('backbone.marionette');` was used. So I think it'll be cool to make it same for all file)

scott-w: Thanks for that, I must have missed this one 👍

denar90:
👍

Issue #3073, opened 2016-06-19
Remove References to AbstractView


scott-w:
### Description

 1. The `AbstractView` was removed from Mn 3 but references are still in the documentation. It also forces people to look at the documentation there, away from the actual API they're using.
 2. We need to remove all references to `AbstractView` and move the detail into the respective documentation.
 3. This will update the documentation in line with the codebase and help readers find the information they need.
 4. This is related to #3066.

### Environment

1. Marionette version: 3.0.0-pre4
2. Backbone version: All
3. Additional build tools, etc: N/A

denar90:
@scott-w should this issue wait until `View Lifecycle`, `View`, `Events` etc docs is done, to add right references?

scott-w:
If there's anything to be pulled into other docs, we could start by getting that all put into roughly the right places then just ditch this file.

After that, we can start figuring out how to cleanly split up the docs to make them easier to navigate. I don't see #3066 blocking Mn 3 as long as the content is there.

denar90:
merged #3074

PR #3074, opened 2016-06-25
Remove References to AbstractView

denar90:
### Proposed changes
 - remove `AbstractView` doc
 - move `TemplateContext` info from `AbstractView` into `Renderer`

Link to the issue:
#3073 

Also `AbstractView` had doc related to view lifecycle events which was duplicate of https://github.com/marionettejs/backbone.marionette/blob/next/docs/marionette.view.md#lifecycle-events
In future we want to have `View Lifecycle` doc file. 
@scott-w can you move info about lifecycle events from `View` into `View Lifecycle` doc? It's your child, so I think it will be better that you do this)


scott-w:
Looks good. Do you see the `Renderer` docs containing all the info for how to render templates? My thinking for templates is:

1. Basics (part of View or their own Templates doc)
2. Sharing template rendering logic using `Renderer`
3. Specific examples for integrating different rendering engines e.g. Handlebars

denar90:
Agree, but I just moved into `Renderer` info which can be reused from `AbstractView` (didn't want lose it), another stuff are duplicates. Do you want me to do this here?

scott-w:
I'm happy with this as-is for now 👍

Issue #3063, opened 2016-06-11
Backbone + Marionette Basics


scott-w:
I'm finding that a lot of documentation is repeating the same basic concepts e.g. attributes can take values or functions returning values, references can be methods on the view or function definitions.

I think it would be helpful to put together a "Backbone and Marionette Basics" doc that covers these cases that can be referenced everywhere. This gives two major benefits:

1. Readers can understand that some things are universal in the Marionette framework
2. We can assume base levels of knowledge when writing docs and shorten them a little

We can use this to introduce a basic Application/View so users can see the ideal case.

### Required Documentation

* [x] Setting functions for attributes
* [x] Setting options when creating a instance

Related to: #3066

paulfalgout:
👍

paulfalgout:
I think previously lack of consistency has kept this from being possible. :-)

scott-w:
I'm pretty happy this is in place and we can slowly move the rest of the documentation over to referring to this in time.

PR #3076, opened 2016-07-03
Common concepts

scott-w:
### Proposed changes
 - Initial index
 - Document that we use `extend` everywhere
 - Initial `getOption` work

Link to the issue: #3063 

### Feedback

Let me know what I've missed - I'll try to get as much as possible done before this gets merged.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -120,3 +120,41 @@ Marionette can set options when you instantiate a class. This lets you override
 many class-based attributes when you need to. You can also pass new information
 specific to the object in question that it can access through special helper
 methods.
+
+```javascript
+var Mn = require('backbone.marionette');
+
+var MyView = Mn.View.extend();
+
+var view = new MyView({
+  foo: 'some text'
+});
+
+view.checkOption();
+```
+
+### The `getOption` Method
+
+To access an option, we use the `getOption` method.
+
+```javascript
+var Mn = require('backbone.marionette');
+
+var MyView = Mn.View.extend({
+  className: function() {
+    var defaultClass = 'table';
+    var extraClasses = this.getOption('extraClasses') || '';
+    return defaultClass + ' ' + extraClasses;
+  }
+});
+
+var myView = new MyView({
+  model: new MyModel(),
+  extraClasses: 'table-striped'
+});
+```
+
+This only works for custom options - arguments that belong to the standard
+Backbone/Marionette attributes, such as `model` and `collection`, are not
+accessible via `getOption` and should be accessed as just `view.model` or
+`view.collection`.
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -1,0 +1,16 @@
+**_These docs are for Marionette 3 which is still in pre-release. Some parts may
+not be accurate or up-to-date_**
+
+# Common Marionette Concepts
+
+This document covers the basic usage patterns and concepts across Marionette.
+This includes things like calling conventions, setting attributes, common option
+patterns etc.
+
+## Documentation Index
+
+* [Setting Attributes](#setting-attributes)
+  * [Fixed Values](#fixed-values)
+  * [Functions Returning Values](#functions-returning-values)
+* [Setting Options](#setting-options)
+  * [The `getOption` Method](#the-getoption-method)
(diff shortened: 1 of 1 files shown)


docs/basics.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+When we instantiate `MyView`, each instance will be given a `.bg-success` class
+with a `myRegion` region created on the `.my-region` element.
+
+### Functions Returning Values
+
+In almost every instance where we can set a value, we can also assign a function
+to figure out the value at runtime. In this case, Marionette will run the
+function on instantiating and use the returned value:
```
paulfalgout: on instantiation

also might mention something like "with the context of the instance"

docs/basics.md
```diff
@@ -0,0 +1,160 @@
...
+As we can see, almost all of the attributes here can be worked out dynamically.
+In most cases, Marionette will call the function once at instantiation, or first
+render, and preserve the value throughout the lifetime of the View. There are
+some exceptions to this rule - these will be referred to with their respective
+documentation.
+
+There are some exceptions to this general rule that will be documented in the
+relevant places.
+
+## Setting Options
```
paulfalgout: I definitely think it is worth mentioning and perhaps prioritizing `mergeOptions` along with `getOption` as well.

scott-w:
I've made those changes @paulfalgout - let me know what you think

docs/basics.md
```diff
@@ -0,0 +1,203 @@
...
+
+You can now create instances of `MyView` with JavaScript's `new` keyword:
+
+```javascript
+var view = new MyView();
+```
+
+### Value Attributes
+
+When we extend classes, we can set a form of class attributes by setting new
```
rafde: I'm having a difficult time reading this description.
Would this description help?
> When we extend classes, we can provide class attributes with specific values by defining them in the object we pass as the first parameter.

I would also note that the values are shallow copied over to the classes's prototype object. All object type values are passed by reference, so changing any attribute values will result in affect all class instances or extending classes.
I am not sure how to phrase all that in a way that isn't overwhelming, but I'd consider thinking of a way to do so.

docs/basics.md
```diff
@@ -0,0 +1,203 @@
...
+render, and preserve the value throughout the lifetime of the View. There are
+some exceptions to this rule - these will be referred to with their respective
+documentation.
+
+There are some exceptions to this general rule that will be documented in the
+relevant places.
+
+### Function Context
+
+Each function call will receive the class instance as its `this` context - this
```
rafde: The formatting looks off.
> `this` context - this

or

> `this` context
- this gives you a handy way to ...

or

> `this` context - this gives you a handy way to ...

?

docs/basics.md
```diff
@@ -0,0 +1,203 @@
...
+```javascript
+var Mn = require('backbone.marionette');
+
+var MyView = Mn.View.extend();
+
+var view = new MyView({
+  foo: 'some text'
+});
+
+view.checkOption();
```
rafde: I don't know where `view.checkOption();` is coming from.

docs/basics.md
```diff
@@ -0,0 +1,203 @@
...
+This only works for custom options - arguments that belong to the standard
+Backbone/Marionette attributes, such as `model` and `collection`, are not
+accessible via `getOption` and should be accessed as just `view.model` or
+`view.collection`.
+
+### The `mergeOptions` Method
+
+The `mergeOptions` method takes two arguments: an `options` object and `keys` to
+pull from the options object. Any matching `keys` will be merged onto the
+object instance. For example:
```
rafde: `class instance` for consistency?

rafde:
Sorry for injecting myself in this PR, but I started reading and was hearing the voices of my co-workers who are good at asking questions when they don't understand something. I figure I might as well as commit to annoy you with my suggestions and questions.

docs/basics.md
scott-w: I like it, good call.

docs/basics.md
scott-w: Looks like a typo on my part - I'll look through it again and get it sorted.

scott-w:
Hi @rafde thanks for the comments, it's really helpful to get this feedback before we push it all live. These are exactly the sort of questions we're trying to help answer with this.

denar90:
👍

PR #3079, opened 2016-07-13
reset "destroyed" lifecycle flag via initialization

nmschulte:
> allowing views to be re-initialized (and thus re-used)

This is useful for certain use-cases, and doesn't break any behavior.

I searched for occurrences of `isDestroyed` in the source tree and found no logic in the implementations that cared about the state, so this shouldn't break any assumptions or change any behavior whatsoever.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -26,6 +26,10 @@ Marionette.View = Backbone.View.extend({
     Marionette.MonitorDOMRefresh(this);
   },
 
+  initialize: function() {
+    this.isDestroyed = false;
+  },
+
   // Get the template for this view
   // instance. You can set a `template` attribute in the view
   // definition or pass a `template: "whatever"` parameter in
--- a/test/unit/view.spec.js
+++ b/test/unit/view.spec.js
@@ -5,6 +5,7 @@ describe('base view', function() {
     beforeEach(function() {
       this.initializeStub = this.sinon.stub();
       this.viewConstructorSpy = this.sinon.spy(Backbone, 'View');
+      this.viewInitializeSpy = this.sinon.spy(Marionette.View.prototype, 'initialize');
 
       this.View = Marionette.View.extend({
         initialize: this.initializeStub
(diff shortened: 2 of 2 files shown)


nmschulte:
There should be a test or two to come along w/ this change.  As it's written, it doesn't quite work: calling `initialize` on an extended instance doesn't all `Marionette.View`s `initialize` implementation.  I'm not sure what would be a good approach to this given the project's goals.  My first idea was to do this as a `Behavior`, because (though I don't use them myself) that seems to resolve the issue w/ how Backbone/Marionette manage polymorphism.

paulfalgout:
I don't know the use case for this.. 99% of the time you just want to create a new view.. The other 1% you might preventDestroy.. but once a view is destroyed, it shouldn't be reinitialized.

nmschulte:
> but once a view is destroyed, it shouldn't be reinitialized.

Why is that?  Backbone doesn't impose this limitation, and AFAICT, Marionette doesn't seem to care (beyond only triggering `onBeforeDestroy` when the destroy is actually going to happen, and alerting the developer to a potential mistake w/ the throwing of `ViewDestroyedError`).  That's not to say that the other implementations are not making this assumption internally (and thus, gating things themselves, but possibly not accounting for the "reset" condition), but because they don't check `isDestroyed`, it's not clear.

The particular case that prompted this involves retaining "view state" within a single page application; there are lots of different ways one can press Backbone and Marionette into play, so I'm not sure how important it is that my case isn't the most common one.  Given that this doesn't really break any behavior, I'd love to see it considered.

paulfalgout:
https://github.com/marionettejs/backbone.marionette/pull/724

`View#close` was renamed to `View#destroy` to prevent people from reusing views.  You can handle your use case in many ways.. store the state outside of the view.. if you absolutely must prevent the destroy, tell the region not to destroy the view..  But simply re`initialize`ing the view will only do a small portion of what happens when a view is constructed.  So you will only have a partially initialized view if you destroy it and attempt to reuse it.

nmschulte:
I see that now (e.g. `View` and `LayoutView`'s `constructor` have setup); so... s/`initialize`/`constructor`/?

JSteunou:
I'm gonna close this, still don't see any relevant use cases. As @paulfalgout said, you can still prevent a region to destroy its view to keep your view intact. Even if Marionette is not that much opinionated, a best practice is to create and destroy view better than trying to cheat the tool.

nmschulte:
More discussion of view re-use, from the docu:
(http://marionettejs.com/docs/marionette.view.html#viewmodelevents-and-viewcollectionevents)

> The `modelEvents` and `collectionEvents` will be bound and
> unbound with the Backbone.View `delegateEvents` and `undelegateEvents`
> method calls. This allows the view to be re-used and have
> the model and collection events re-bound.

paulfalgout:
yep.  `modelEvents` and `collectionEvents` are removed from `delegateEvents` in v3 as `delegateEvents` is related to the `$el`.. so now `delegateEntityEvents` only occurs in the constructor which aligns with views being constructed and destroyed.

PR #3082, opened 2016-07-16
100% Region Test Coverage 

rafde:
### Region Test Coverage need to be covered 💯 %

Helps #2581

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -164,11 +164,17 @@ const Region = MarionetteObject.extend({
 
   // Restore the region's element in the DOM.
   _restoreEl() {
-    if (!this.currentView) {
+    // There is nothing to replace
+    if (!this._isReplaced) {
       return;
     }
 
     const view = this.currentView;
+
+    if (!view) {
+      return;
+    }
+
     const parent = view.el.parentNode;
 
     if (!parent) {
@@ -179,6 +185,7 @@ const Region = MarionetteObject.extend({
     this._isReplaced = false;
   },
 
+  // Check to see if the region's el was replaced.
   isReplaced() {
     return !!this._isReplaced;
   },
@@ -210,9 +217,7 @@ const Region = MarionetteObject.extend({
     view.off('destroy', this.empty, this);
     this.triggerMethod('before:empty', this, view);
 
-    if (this._isReplaced) {
-      this._restoreEl();
-    }
+    this._restoreEl();
 
     delete this.currentView;
 
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -23,6 +23,10 @@ describe('region', function() {
       this.optionRegionJquery = new Backbone.Marionette.Region({el: $(this.el)});
     });
 
+    it('should not have been replaced', function() {
+      expect(this.customRegion.isReplaced()).to.be.false;
+    });
+
     it('should work when el is passed in as an option', function() {
       expect(this.optionRegionJquery.$el[0]).to.equal(this.el);
       expect(this.optionRegionJquery.el).to.equal(this.el);
@@ -300,6 +304,7 @@ describe('region', function() {
 
         describe('when setting the "replaceElement" class option', function() {
           beforeEach(function() {
+            this.sinon.spy(this.region, '_restoreEl');
             // empty region to clean existing view
             this.$parentEl = this.region.$el.parent();
             this.regionHtml = this.$parentEl.html();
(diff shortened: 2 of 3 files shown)


test/unit/region.spec.js
```diff
@@ -328,6 +329,10 @@ describe('region', function() {
             it('should restore the region\'s "el" to the DOM', function() {
               expect(this.$parentEl).to.contain.$html('<div id="region"></div>');
             });
+
+            it('should not call _restoreEl', function() {
```
rafde: well this is wrong... will fix.

src/region.js
```diff
@@ -156,31 +156,19 @@ const Region = MarionetteObject.extend({
...
+      // Replace using library to ensure proper cleanup.
+      view.$el.replaceWith(this.el);
+      this._isReplaced = false;
     }
-
-    parent.replaceChild(this.el, view.el);
-    this._isReplaced = false;
-  },
-
-  isReplaced() {
```
paulfalgout: I don't think `isReplaced` should be removed?

src/region.js
```diff
@@ -156,31 +156,19 @@ const Region = MarionetteObject.extend({
...
-      return;
-    }
-
     const view = this.currentView;
-    const parent = view.el.parentNode;
 
-    if (!parent) {
-      return;
+    if (view && view.el && view.el.parentNode) {
+      // Replace using library to ensure proper cleanup.
```
paulfalgout: is this the behavior we really want? any events added to the regions el before the show would be lost in this case.  The view itself is likely to get cleaned up anyways because it is the el of the view and would get cleaned up on destroy...  

In the very least it'd probably be good to abstract `this.$el.replaceWith` to a private function such that one function could be overridden.. we're hopefully not introducing additional jquery dependencies that can't be easily overridden to v3.

paulfalgout:
So I _think_ I'm all for this direction.. but with some suggestions..  However.. this PR is more of a behavior change than test coverage.

src/region.js
rafde: I didn't see it get used anywhere in the code base. I could have it get used where I saw it should have been used. I'll do that

paulfalgout: ah it doesn't need to be used internally.. it's only a public check.  like `isDestroyed()` or `isRendered()`

rafde: Ok. I'll add it back and add some test for it.

src/region.js
rafde: > any events added to the regions el before the show would be lost in this case.

Trying to think of a test for this scenario.

In the mean time I can revert the conditions.

> In the very least it'd probably be good to abstract this.$el.replaceWith to a private function such that one function could be overridden

I agree. The code base should definitely being going that route. I'll look to see if there's an issue to create some sort of mixin that abstract DOM interaction. Otherwise, I'll outline a something later.

rafde:
@paulfalgout no problem. I'll skip https://github.com/marionettejs/backbone.marionette/pull/3082/commits/5425d73e937dca3335f31e3d29f953e7851371c8 and https://github.com/marionettejs/backbone.marionette/pull/3082/commits/96411f70a4855f890d778650cb7d6e7ed63929e8 then add a test for `isReplaced`

src/region.js
```diff
@@ -164,11 +164,17 @@ const Region = MarionetteObject.extend({
 
   // Restore the region's element in the DOM.
   _restoreEl() {
-    if (!this.currentView) {
+    // There is nothing to replace
+    if (!this._isReplaced) {
```
rafde: OK. Last attempt at "fixing" an issue that I saw.
There are places where this logic happened but I didn't see it making sense to only apply it there. I figure it should always apply in here to prevent unnecessary replacement. Naturally, this assumes `_isReplaced` was not modified externally.

rafde:
@paulfalgout please give my this update a look. Worse case, I can edit the region.js to specify why `isReplaced` is there, with a good description of it's intent.

rafde:
I noticed the coverage badge is at 99% and it picked at me to ping @paulfalgout @denar90 @scott-w @ahumphreys87 @ianmstew @MeoMix even @jmeas for ole times sake. 

If there is any contention towards https://github.com/marionettejs/backbone.marionette/pull/3082#discussion_r71273464 I can remove it if need be. So... close... to... :100: % :exclamation:

ahumphreys87:
Hey @rafde I think this looks good, let's merge

scott-w:
:+1:

denar90:
👍

denar90:
@rafde thank you!

rafde:
🎉what what⁉️ 💯%❗️ that's what‼️

Issue #3071, opened 2016-06-16
ui hyphen fix needs to be applied to v3


paulfalgout:
https://github.com/marionettejs/backbone.marionette/commit/729adf875a60230fced70d944744e2b87eca24da needs to be cherry-picked onto `v3`

@ahumphreys87 ping

PR #3087, opened 2016-07-30
Pulled fix + test onto v3 #3071

scott-w:
### Proposed changes
 - Adds test for hyphen in UI hash
 - Copies regex over from v2 fix
 - Lint fix

Link to the issue: #3071


denar90:
@scott-w did you cherry pick commit https://github.com/marionettejs/backbone.marionette/commit/729adf875a60230fced70d944744e2b87eca24da ?

scott-w:
@denar90 No, I had to do it manually. The tests are in different files on Mn 3 so it worked out easier to just copy + paste the fix on top and change the tests manually.

It's not ideal - I just extended an existing test to cover the functionality rather than making a specific test. I could change it and add a test just for that if you'd prefer?

denar90:
Thank you for an explanation. It's 👍 for me.

PR #3088, opened 2016-07-31
AppRouter Docs

scott-w:
### Proposed changes
 - Add an introduction to the AppRouter
 - Tell readers we can use Marionette.Object
 - Add information about the History API (Todo)

Link to the issue: #3066


denar90:
One more cool job for docs!!! 👍

scott-w:
Cheers, I've still got a bit work to do on it so I'll get that all in before I merge. I want to add a little info for the History and navigate first. Though I could see that moving into a separate section or just remaining in the Guide as it can get somewhat involved.

ahumphreys87:
@scott-w can you pull the `-` fix into a different PR?

This looks good to me other than that

Issue #3096, opened 2016-08-04
Update mocha to v3


denar90:
Do we have to update mocha to v3 ?

Issue #3070, opened 2016-06-15
TypeError: ChildViewContainer is not a constructor


Regnoy:
Simple example from documentation Marionette 3.0.0 - pre.4 + Require.js

 var MyChildView = Marionette.View.extend({});

 var Col =  Marionette.CollectionView.extend({
    childView: MyChildView
  });
  new Col();

I have console error.
> TypeError: ChildViewContainer is not a constructor	
> this.children = new ChildViewContainer();

denar90:
Looks fine https://jsfiddle.net/599dtsb7/13/

Regnoy:
jsfiddle use AMD?

Regnoy:
My solution temporary is:

`ChildViewContainer =  factory(Backbone, _);
				/*define(['backbone', 'underscore'], function(Backbone$$, _$$) {
					return factory(Backbone$$, _$$);
				});*/`
and comment this line
//var ChildViewContainer = (backbone_babysitter && typeof backbone_bab .....

After that all work fine.

JSteunou:
I dont understand fully your issue. Is there a problem using Marionette with requirejs?

Can you make fiddle demonstrating the issue?

paulfalgout:
It sounds like perhaps you want to be using the bundled version 
https://github.com/marionettejs/backbone.marionette/blob/next/lib/backbone.marionette.js

Instead of the core version:
https://github.com/marionettejs/backbone.marionette/blob/next/lib/core/backbone.marionette.js

The bundled version includes babysitter.  If you're using the core then you need to include babysitter separately.

paulfalgout:
Ahhhh. never mind.. after reviewing your last comment it looks like you are using the bundled version.  I don't see any particular reason the `define` wouldn't work.  The bundler is modifying the variable names so that it is clear that the included library is not reaching outside of its closure to get Backbone and _... I am not certain why there would be any issue.. a fiddle of this loading with require might be necessary to debug any issue

Regnoy:
Thanks for response, and sorry for bad english :)

I use Marionette 3.0.0 - pre.4 (Bundled Version)
with 
/*!
* Includes BabySitter
* https://github.com/marionettejs/backbone.babysitter/
*
* Includes Radio
* https://github.com/marionettejs/backbone.radio/
*/
+ Require.js

I attach zip file.
[git.zip](https://github.com/marionettejs/backbone.marionette/files/317822/git.zip)

I hope help you. Wait for response.

[...]

PR #3097, opened 2016-08-04
Remove bundled build for v3 and integrate babysitter.

paulfalgout:
This is kind of a last minute sizable change, but hopefully I can argue well enough for it.

Immediately this would resolve https://github.com/marionettejs/backbone.marionette/issues/3070 as there would be no bundled build.  Our bundled build is complicated.  It might be easier solved by switching everything to webpack, but I'm not 100% certain it'd be that much better.. we'd have to change everything quite a bit and our output would look much less readable for the annotated source. Furthermore rollup is practically purpose built for what we're doing.  It certainly has it's issues, but if we simplify the build we probably won't see them.

**But why remove the bundled build?** Is it just because it's too hard to build?  No. There have been arguments to remove it in the past: https://github.com/marionettejs/backbone.marionette/issues/2273

And I even attempted to remove it previously: https://github.com/marionettejs/backbone.marionette/pull/2485 (there's a lengthy argument about the bundled build's value there)

However! I can now argue additionally that we should "bundle" by default.  **This PR also pulls babysitter into the library.**  Babysitter is neglected. We build babysitter like logic into the `CollectionView` because it is abstracted and that causes issue after issue. (think filtering / sorting)  And it is one additional headache to deal with every time we make a release.. all babysitter is good for lately is updating its dependencies.

But wait doesn't that leave out Radio?  Yes, but the argument that we need to keep the bundled build was essentially that we wanted backbone users to be able to easily add one script and getting running.  You will be able to do this!  Radio is only a dependency of the Object's channel feature and isn't used internally.  So I added an error that is throw if that feature is attempted without the dependency.  Otherwise.. essentially core and bundled are the same.

Also resolves https://github.com/marionettejs/backbone.marionette/issues/3096


JSteunou:
I'm all in! Did not except this to be dropped in this soon, but this is for the best I think.

My only concern is about how explain this change to the user, and the impact it could have. But for the sake of Marionette, this is all good!

Great job @paulfalgout :+1:

ahumphreys87:
this looks good to me @paulfalgout

denar90:
@paulfalgout what to do with `backbone.babysitter` repo, will we mark it as deprecated?

paulfalgout:
I dunno what to do with babysitter..  I mean _someone_ could technically be using it directly...

denar90:
Created some discussion issue https://github.com/marionettejs/backbone.babysitter/issues/78 about this.

PR #3105, opened 2016-08-06
CollectionView and CompositeView removing views performance update.

rafde:
# CollectionView and CompositeView removing views performance update.
## _initialEvents
* Listen to `this.collection~event:update` instead.

## _onCollectionUpdate
* pass `options.changed.removed` to `this._removeChildViews`
* updated test to reflect additions

##  _removeChildViews
* use array of models to determine what views to remove if not already destroyed.
* Support `checkEmpty` for preventing empty collection checking

## _destroyChildren
* checks to see if there are any children to destroy before calling `_removeChildViews`

## removeChildView
* using _destroyChildView
* added test for trying to remove falsy and valid values

## ~~_onCollectionRemove~~
* test and references removed

## _findGreatestIndexedView
* finds the view with the greatest `_index` and returns it.

## _updateIndices
* accepts array of views for _findGreatestIndexedView

## [test] when removing the sorted view, the `view._index` should update.

im

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -100,7 +100,7 @@ const CollectionView = Backbone.View.extend({
   _initialEvents() {
     if (this.collection) {
       this.listenTo(this.collection, 'add', this._onCollectionAdd);
-      this.listenTo(this.collection, 'remove', this._onCollectionRemove);
+      this.listenTo(this.collection, 'update', this._onCollectionUpdate);
       this.listenTo(this.collection, 'reset', this.render);
 
       if (this.sort) {
@@ -126,11 +126,89 @@ const CollectionView = Backbone.View.extend({
     }
   },
 
-  // get the child view by model it holds, and remove it
-  _onCollectionRemove(model) {
-    const view = this.children.findByModel(model);
-    this.removeChildView(view);
-    this._checkEmpty();
+  // Handle collection update model removals
+  _onCollectionUpdate(collection, options) {
+    const changes = options.changes;
+    this._removeChildViews(changes.removed);
+  },
+
+  // Remove the child views and destroy them.
+  // This function also updates the indices of later views
+  // in the collection in order to keep the children in sync with the collection.
+  // "modelsOrViews" can be an array of mixed models or views that
+  // will be removed and destroyed from the CollectionView
+  _removeChildViews(modelsOrViews, {checkEmpty} = {}) {
+    const shouldCheckEmpty = checkEmpty !== false;
+
+    // Used to determine where to update the remaining
+    // sibling view indices after these views are removed.
+    const removedViews = this._getRemovedViews(modelsOrViews);
+
+    if (!removedViews.length) {
+      return;
+    }
+
+    this.children._updateLength();
+
+    // decrement the index of views after this one
+    this._updateIndices(removedViews, false);
+
+    if (shouldCheckEmpty) {
+      this._checkEmpty();
+    }
+  },
+
+  // "modelsOrViews" can be an array of mixed models or views that
+  // will be removed and destroyed from the CollectionView
+  // Returns the views that will be used for re-indexing views
+  // through CollectionView#_updateIndices.
+  _getRemovedViews(modelsOrViews) {
+
+    // Returning a view means something was removed.
+    return _.reduce(modelsOrViews, (removingViews, modelOrView) => {
+      let view = modelOrView;
+
+      if (modelOrView instanceof Backbone.Model) {
+        view = this.children.findByModel(modelOrView);
+      }
+
+      if (!view || view._isDestroyed) {
+        return removingViews;
+      }
+
+      this._removeChildView(view);
+
+      removingViews.push(view);
+
+      return removingViews;
+    }, []);
+  },
+
+  _findGreatestIndexedView(views) {
+
+    return _.reduce(views, (greatestIndexedView, view) => {
+      // Even if the index is `undefined`, a view will get returned.
+      if (!greatestIndexedView || greatestIndexedView._index < view._index) {
+        return view;
+      }
+
+      return greatestIndexedView;
+    }, undefined);
+  },
+
+  _removeChildView(view) {
+    this.triggerMethod('before:remove:child', this, view);
+
+    this.children._remove(view);
+    if (view.destroy) {
+      view.destroy();
+    } else {
+      destroyBackboneView(view);
+    }
+
+    delete view._parent;
+    this.stopListening(view);
+    this.triggerMethod('remove:child', this, view);
   },
 
   // Overriding Backbone.View's `setElement` to handle
@@ -193,12 +271,12 @@ const CollectionView = Backbone.View.extend({
       }
       currentIds[model.cid] = true;
     });
-    _.each(previousModels, (prevModel) => {
-      const removedChildExists = !currentIds[prevModel.cid] && this.children.findByModel(prevModel);
-      if (removedChildExists) {
-        this._onCollectionRemove(prevModel);
-      }
+
+    const removeModels = _.filter(previousModels, (prevModel) => {
+      return !currentIds[prevModel.cid] && this.children.findByModel(prevModel);
     });
+
+    this._removeChildViews(removeModels);
   },
 
   // Reorder DOM after sorting. When your element's rendering do not use their index,
@@ -238,8 +316,7 @@ const CollectionView = Backbone.View.extend({
       this._appendReorderedChildren(elsToReorder);
 
       // remove any views that have been filtered out
-      _.each(filteredOutViews, _.bind(this.removeChildView, this));
-      this._checkEmpty();
+      this._removeChildViews(filteredOutViews);
 
       this.triggerMethod('reorder', this);
     }
@@ -485,10 +562,12 @@ const CollectionView = Backbone.View.extend({
 
   // Internal method. This decrements or increments the indices of views after the added/removed
   // view to keep in sync with the collection.
-  _updateIndices(view, increment, index) {
+  _updateIndices(views, increment, index) {
     if (!this.sort) {
       return;
     }
+
+    const view = _.isArray(views) ? this._findGreatestIndexedView(views) : views;
 
     if (increment) {
       // assign the index to the view
@@ -565,22 +644,10 @@ const CollectionView = Backbone.View.extend({
       return view;
     }
 
-    this.triggerMethod('before:remove:child', this, view);
-
-    if (view.destroy) {
-      view.destroy();
-    } else {
-      destroyBackboneView(view);
-    }
-
-    delete view._parent;
-    this.stopListening(view);
-    this.children.remove(view);
-    this.triggerMethod('remove:child', this, view);
-
+    this._removeChildView(view);
+    this.children._updateLength();
     // decrement the index of views after this one
     this._updateIndices(view, false);
-
     return view;
   },
 
@@ -670,18 +737,15 @@ const CollectionView = Backbone.View.extend({
   },
 
   // Destroy the child views that this collection view is holding on to, if any
-  _destroyChildren({checkEmpty} = {}) {
-    this.triggerMethod('before:destroy:children', this);
-    const shouldCheckEmpty = checkEmpty !== false;
+  _destroyChildren(options) {
     const childViews = this.children.map(_.identity);
 
-    this.children.each(_.bind(this.removeChildView, this));
-
-    if (shouldCheckEmpty) {
-      this._checkEmpty();
+    if (childViews.length) {
+      this.triggerMethod('before:destroy:children', this);
+      this._removeChildViews(childViews, options);
+      this.triggerMethod('destroy:children', this);
     }
 
-    this.triggerMethod('destroy:children', this);
     return childViews;
   },
 
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -129,20 +129,20 @@ const CollectionView = Backbone.View.extend({
   // Handle collection update model removals
   _onCollectionUpdate(collection, options) {
     const changes = options.changes;
-    this._removeChildViews(changes.removed);
+    this._removeChildModels(changes.removed);
   },
 
   // Remove the child views and destroy them.
   // This function also updates the indices of later views
   // in the collection in order to keep the children in sync with the collection.
-  // "modelsOrViews" can be an array of mixed models or views that
+  // "models" is an array of models and the corresponding views
   // will be removed and destroyed from the CollectionView
-  _removeChildViews(modelsOrViews, {checkEmpty} = {}) {
+  _removeChildModels(models, {checkEmpty} = {}) {
     const shouldCheckEmpty = checkEmpty !== false;
 
     // Used to determine where to update the remaining
     // sibling view indices after these views are removed.
-    const removedViews = this._getRemovedViews(modelsOrViews);
+    const removedViews = this._getRemovedViews(models);
 
     if (!removedViews.length) {
       return;
@@ -158,19 +158,13 @@ const CollectionView = Backbone.View.extend({
     }
   },
 
-  // "modelsOrViews" can be an array of mixed models or views that
-  // will be removed and destroyed from the CollectionView
-  // Returns the views that will be used for re-indexing views
+  // Returns the views that will be used for re-indexing
   // through CollectionView#_updateIndices.
-  _getRemovedViews(modelsOrViews) {
+  _getRemovedViews(models) {
 
     // Returning a view means something was removed.
-    return _.reduce(modelsOrViews, (removingViews, modelOrView) => {
-      let view = modelOrView;
-
-      if (modelOrView instanceof Backbone.Model) {
-        view = this.children.findByModel(modelOrView);
-      }
+    return _.reduce(models, (removingViews, model) => {
+      const view = this.children.findByModel(model);
 
       if (!view || view._isDestroyed) {
         return removingViews;
@@ -276,7 +270,7 @@ const CollectionView = Backbone.View.extend({
       return !currentIds[prevModel.cid] && this.children.findByModel(prevModel);
     });
 
-    this._removeChildViews(removeModels);
+    this._removeChildModels(removeModels);
   },
 
   // Reorder DOM after sorting. When your element's rendering do not use their index,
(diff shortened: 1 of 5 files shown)


rafde:
hmmm... CI failed due to Backbone version being less than 1.3.3

src/collection-view.js
```diff
@@ -99,7 +99,7 @@ const CollectionView = Backbone.View.extend({
   _initialEvents() {
     if (this.collection) {
       this.listenTo(this.collection, 'add', this._onCollectionAdd);
```
paulfalgout: is there a reason to keep add?  Wouldn't update replace add and remove?

src/collection-view.js
```diff
@@ -471,7 +534,7 @@ const CollectionView = Backbone.View.extend({
   // view to keep in sync with the collection.
   _updateIndices(view, increment, index) {
     if (!this.sort) {
-      return;
+      return this;
```
paulfalgout: Since `_updateIndices` is private we don't need to return `this`.

That consistency is just for the public API.

src/collection-view.js
```diff
@@ -125,11 +125,75 @@ const CollectionView = Backbone.View.extend({
     }
   },
 
-  // get the child view by model it holds, and remove it
-  _onCollectionRemove(model) {
-    const view = this.children.findByModel(model);
-    this.removeChildView(view);
-    this._checkEmpty();
+  // Handle collection update model removals
+  _onCollectionUpdate(collection, {changes: {removed = []} = {}} = {}) {
```
paulfalgout: I dunno if the defaults are good in this instance because of the transpilation.

Could be
```js
_onCollectionUpdate(collection, options) {
  if(_.isArray(options.removed)) { // you already check the length in _removeChildViews
```

instead it's 
```js
_onCollectionUpdate(collection) {
    var _ref = arguments.length <= 1 || arguments[1] === undefined ? {} : arguments[1];

    var _ref$changes = _ref.changes;
    _ref$changes = _ref$changes === undefined ? {} : _ref$changes;
    var _ref$changes$removed = _ref$changes.removed;
    var removed = _ref$changes$removed === undefined ? [] : _ref$changes$removed;
if (_.isArray(removed) {
```

paulfalgout: I don't think I quite have that right.. but you get the idea.

src/collection-view.js
```diff
@@ -644,18 +693,15 @@ const CollectionView = Backbone.View.extend({
...
   // Destroy the child views that this collection view is holding on to, if any
-  _destroyChildren({checkEmpty} = {}) {
-    this.triggerMethod('before:destroy:children', this);
-    const shouldCheckEmpty = checkEmpty !== false;
+  _destroyChildren(options) {
     const childViews = this.children.map(_.identity);
 
-    this.children.each(_.bind(this.removeChildView, this));
-
-    if (shouldCheckEmpty) {
```
paulfalgout: I think maybe the shouldCheckEmpty should move back here.. this is the only place it is used in `_removeChildViews` so it is making that function more complicated than need be.

src/collection-view.js
```diff
@@ -125,11 +125,75 @@ const CollectionView = Backbone.View.extend({
...
+
+      // everything after this could be async destructed
+      this.children._remove(view);
+      if (view.destroy) {
+        view.destroy();
+      } else {
+        destroyBackboneView(view);
+      }
+
+      view._parent = void 0;
```
paulfalgout: I think the delete is ok.. we don't iterate over view keys so I don't think there's any issue with delete, and we use it elsewhere.

src/collection-view.js
```diff
@@ -125,11 +125,75 @@ const CollectionView = Backbone.View.extend({
...
+  _removeChildViews(modelsOrViews, {checkEmpty} = {}) {
+    if (!modelsOrViews.length) {
+      return modelsOrViews;
+    }
+
+    const shouldCheckEmpty = checkEmpty !== false;
+
+    // Used to determine where to update the remaining
+    // sibling view indices after these views are removed.
+    const removedViews = _.reduce(modelsOrViews, (removingViews, modelOrView) => {
```
paulfalgout: This is a large iterator.  It'd be nice to move it to it's own function so that this line is just `const removeViews = _.reduce(modelsOrViews, (removingViews, modelOrView) => this._removeChildView, []);`

paulfalgout: actually this might be best split up further where line 167 - 179 are `_removeChildView` and the iterator is named something else

src/collection-view.js
```diff
@@ -177,12 +241,12 @@ const CollectionView = Backbone.View.extend({
       }
       currentIds[model.cid] = true;
     });
-    _.each(previousModels, (prevModel) => {
-      const removedChildExists = !currentIds[prevModel.cid] && this.children.findByModel(prevModel);
-      if (removedChildExists) {
-        this._onCollectionRemove(prevModel);
-      }
+
+    const removeModel = _.filter(previousModels, (prevModel) => {
```
paulfalgout: this is `removeModels` right? plural?

src/collection-view.js
```diff
@@ -125,11 +125,75 @@ const CollectionView = Backbone.View.extend({
...
+    }
+
+    const shouldCheckEmpty = checkEmpty !== false;
+
+    // Used to determine where to update the remaining
+    // sibling view indices after these views are removed.
+    const removedViews = _.reduce(modelsOrViews, (removingViews, modelOrView) => {
+      let view = modelOrView;
+
+      if (modelOrView instanceof Backbone.Model) {
```
paulfalgout: We're either going to get an entire array of models or an entire array of views correct?  I wonder if this would get cleaned up a bit if we added a `_removeChildViewsByModel` and just fetched all the views and then just called `_removeChildViews` or something like that.

paulfalgout:
This is a great PR!  Thanks!  I made a bunch of comments.. apologies if it seems like a lot, there's a lot going on here. This is non-breaking as far as I can tell, so this'd probably be a 3.x to not hold up the release, but I'd like to work with you if you're up for it to get this in!  🎈 🍰 💃

src/collection-view.js
rafde: There isn't but I still working on a better way to do add.

paulfalgout: sounds good.  That was my first comment, but this is already a sizable PR so it makes sense to wait probably anyhow. 👏

src/collection-view.js
rafde: I did this as a micro-opt. Since this view is getting removed, I can reassign the variable to 🌵 and it'll still be fine.

src/collection-view.js
rafde: yup. will update.

src/collection-view.js
rafde: I get you. I will probably revert to `options` and destruct what I actually need from the param instead and pass the `options` to `_removeChildViews`

src/collection-view.js
rafde: That sounds like 2 loops that can be avoided the way this works now.

rafde:
sure @paulfalgout 
I'll address your comments and figure out a good time to go over it some more.

src/collection-view.js
rafde: actually, I am changing it to
```
// Handle collection update model removals
  _onCollectionUpdate(collection, options) {
    const changes = options.changes;
    if (changes) {
      const removed = changes.removed;
      if (_.isArray(removed)) {
        this._removeChildViews(removed);
      }
    }
  },
```
in case options doesn't have `options.changes` and only do stuff when removed is an array

src/collection-view.js
rafde: I reverted all the changes to this function, except the use `_removeChildViews`,  and several test regarding `_checkEmpty` broke. I think I will have to leave the changes I made.

rafde:
@paulfalgout I updated the PR.

src/collection-view.js
```diff
@@ -125,11 +125,86 @@ const CollectionView = Backbone.View.extend({
...
+    }
+  },
+
+  // Remove the child views and destroy them.
+  // This function also updates the indices of later views
+  // in the collection in order to keep the children in sync with the collection.
+  // Return array of views where the first view has the index that
+  // tells CollectionView#_updateIndices where to start syncing remaining
+  // view indices.
+  _removeChildViews(modelsOrViews, {checkEmpty} = {}) {
```
ahumphreys87: maybe add a comment to explain the naming of the var `modelsOrViews` might be confusing

src/collection-view.js
```diff
@@ -125,11 +125,86 @@ const CollectionView = Backbone.View.extend({
...
+    this._updateIndices(removedViews[0], false);
+
+    if (shouldCheckEmpty) {
+      this._checkEmpty();
+    }
+
+    return removedViews;
+  },
+
+  _getRemovingViews(modelsOrViews) {
```
ahumphreys87: maybe rename to `_getRemovedViews`?

ahumphreys87:
only 2 extra nitpicks from me, this is a great PR @rafde

rafde:
@ahumphreys87 done.

rafde:
@marionettejs/marionette-core would anyone else like to take a look? Or can I hit the merge button?

paulfalgout:
need to get 2 thumbs on this one.  would like to look into it again too.. but at the moment I'm more focused on getting 3.0 out. Hoping we can put this off and turn around a quick 3.1.

src/collection-view.js
paulfalgout: this is moderately concerning.  I'd kind of like to know why checkEmpty only used to be here, but now it is needed everywhere?

src/collection-view.js
```diff
@@ -125,11 +125,90 @@ const CollectionView = Backbone.View.extend({
...
+    this.triggerMethod('before:remove:child', this, view);
+
+    this.children._remove(view);
+    if (view.destroy) {
+      view.destroy();
+    } else {
+      destroyBackboneView(view);
+    }
+
+    view._parent = void 0;
```
paulfalgout: while this might be a micro-optimization we should change it stylistically across the code rather than introduce the change in a single location for this PR, so that future contributors know which to do.  Granted they are slightly different on the prototype.. however I doubt we rely on that difference in this lib.

src/collection-view.js
```diff
@@ -125,11 +125,90 @@ const CollectionView = Backbone.View.extend({
...
+        removingViews.push(view);
+      }
+
+      this._destroyChildView(view);
+
+      return removingViews;
+    }, []);
+  },
+
+  _destroyChildView(view) {
```
paulfalgout: I _really_ like how this got broken up.  The pattern of `before:event` function `event` I'm hoping is everywhere as opposed to seeing code on either side of the events.

src/collection-view.js
```diff
@@ -125,11 +125,90 @@ const CollectionView = Backbone.View.extend({
...
+    this.triggerMethod('before:remove:child', this, view);
+
+    this.children._remove(view);
```
paulfalgout: _Definitely_ not for now.. but as we're looking at integrating babysitter into the CollectionView.. we do a lot of things where we update the children and then do something to the child.. or we update the children indices... then the collectionViews..  hopefully we can move this into a more serial single responsibility type of thing like `_removeChildView` -> `this.children._remove` -> `onChildRemove` -> `view.destroy` or _something_ to that effect. where things update the children.. and based on what the children do we react.

src/collection-view.js
rafde: 17 refs to `delete`. OK, changing it to delete for consistency.

src/collection-view.js
rafde: No problem. We all have our plans. I just want this code to be fast. I'll hold off poking you all till after 3.0 release and maybe look at what else is preventing this from working faster.

src/collection-view.js
paulfalgout: https://github.com/marionettejs/backbone.marionette/pull/3113 🎉

src/collection-view.js
rafde: These are the failing test when I move the check back.

1) collectionview - emptyView when the last item is removed from a collection should append the html for the emptyView:
     AssertionError: expected [object Object] to have html '<span>empty</span>' but found ''
      at Context.<anonymous> (collection-view.empty-view.spec.js:155:47)

  2) collectionview - emptyView when the last item is removed from a collection should call "onBeforeRenderEmpty":
     AssertionError: expected onBeforeRenderEmpty to have been called exactly once, but it was called 0 times
      at Context.<anonymous> (collection-view.empty-view.spec.js:159:7)

  3) collectionview - emptyView when the last item is removed from a collection should call "onRenderEmpty":
     AssertionError: expected onRenderEmpty to have been called exactly once, but it was called 0 times
      at Context.<anonymous> (collection-view.empty-view.spec.js:165:7)

  4) collectionview - emptyView when the last item is removed from a collection should reference each of the rendered view items:
     AssertionError: expected { Object (_views, _indexByModel, ...) } to have a length of 1 but got 0
      at Context.<anonymous> (collection-view.empty-view.spec.js:171:52)

  5) collection view - filter when rendering a collection where some models pass the filter when all models passing the filter are removed from the collection should show the EmptyView:
     AssertionError: expected [object Object] to contain text 'empty' but found ''
      at Context.<anonymous> (collection-view.filter.spec.js:134:52)

  6) collection view when providing a custom render that adds children, without a collection object to use, and removing a child should show the empty view:

      AssertionError: expected 0 to equal 1
      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (collection-view.spec.js:631:60)

  7) collection view when providing a custom render that adds children, without a collection object to use, and removing a child should call "onBeforeRenderEmpty":
     AssertionError: expected onBeforeRenderEmpty to have been called at least once, but it was never called
      at Context.<anonymous> (collection-view.spec.js:635:7)

  8) collection view when providing a custom render that adds children, without a collection object to use, and removing a child should call "onRenderEmpty":
     AssertionError: expected onRenderEmpty to have been called at least once, but it was never called
      at Context.<anonymous> (collection-view.spec.js:639:7)

I'll take a look some more some time in the week.

src/collection-view.js
rafde: thats for render though, not update add. I am still trying to figure out how to deal with that because `options.at` is screwing up how I think the views should update and render. I might do a broken PR push just to give you all a chance to see what I am talking about.

src/collection-view.js
rafde: @paulfalgout I know why the test fail
since `_removeChildViews` replaced [_onCollectionRemove](https://github.com/marionettejs/backbone.marionette/pull/3105/files#diff-091acc5f607e2622bd6cabf705443fffL129),  `_removeChildViews` needs to run `this._checkEmpty();` when its done. `_destroyChildren` also uses `_removeChildViews` so I placed the `shouldCheckEmpty` logic in there also to avoid double execution. So I can't move it out anymore.

src/collection-view.js
```diff
@@ -125,11 +125,89 @@ const CollectionView = Backbone.View.extend({
...
 
-  // get the child view by model it holds, and remove it
-  _onCollectionRemove(model) {
-    const view = this.children.findByModel(model);
-    this.removeChildView(view);
-    this._checkEmpty();
+  // Handle collection update model removals
+  _onCollectionUpdate(collection, options) {
+    const changes = options.changes;
+    if (changes) {
```
paulfalgout: turns out there's no need to check for changes here.. it's always available when `update` is triggered.

rafde:
@marionettejs/marionette-core I am finished making changes to this PR. Please feel free to look over. But this is as optimal, readable, and maintainable I can get it... for now.

src/collection-view.js
```diff
@@ -125,11 +125,95 @@ const CollectionView = Backbone.View.extend({
...
+    const removedViews = this._getRemovedViews(modelsOrViews);
+
+    if (!removedViews.length) {
+      return;
+    }
+
+    this.children._updateLength();
+
+    if (this.sort) {
+      const greatestIndexedView = this._findGreatestIndexedView(removedViews);
```
rafde: I'm thinking I should move this into `_updateIndices`
where that fuction will look like 
```
_updateIndices: function _updateIndices(views, increment, index) {
	    if (!this.sort) {
	      return;
	    }

      let view = views;
      if(_.isArray(view){
        view = this._findGreatestIndexedView(view);
      }

	    if (increment) {
	      // assign the index to the view
	      view._index = index;
	    }

	    this.children.each(function (laterView) {
	        laterView._index += increment ? 1 : -1;
	      }
	    });
	  }
```
removing the need to do the check `_removeChildViews` and having the need in one place.

paulfalgout: Oh yeah.  Much nicer.  :clapping:

src/collection-view.js
```diff
@@ -125,11 +125,91 @@ const CollectionView = Backbone.View.extend({
...
+    }
+  },
+
+  // "modelsOrViews" can be an array of mixed models or views that
+  // will be removed and destroyed from the CollectionView
+  // Returns the view that will be used for re-indexing views
+  // through CollectionView#_updateIndices.
+  _getRemovedViews(modelsOrViews) {
+
+    // the "memo" has to be `undefined` to make sure
```
paulfalgout: this comment doesn't look right now.. it's an array.

Also should it be `removingViews` instead of `removingsView`?

src/collection-view.js
```diff
@@ -125,11 +125,91 @@ const CollectionView = Backbone.View.extend({
...
+    this._updateIndices(removedViews, false);
+
+    if (shouldCheckEmpty) {
+      this._checkEmpty();
+    }
+  },
+
+  // "modelsOrViews" can be an array of mixed models or views that
+  // will be removed and destroyed from the CollectionView
+  // Returns the view that will be used for re-indexing views
```
paulfalgout: Returns the views

src/collection-view.js
```diff
@@ -125,11 +125,91 @@ const CollectionView = Backbone.View.extend({
...
+      // Even if the index is `undefined`, a view will get returned.
+      if (!greatestIndexedView || greatestIndexedView._index < view._index) {
+        return view;
+      }
+
+      return greatestIndexedView;
+    }, undefined);
+  },
+
+  _destroyChildView(view) {
```
paulfalgout: any objection to calling this `_removeChildView`?  it would match the events.

rafde: no objections.

paulfalgout:
might be a waste of time, but I attempted to remove the modelsOrViews  https://github.com/rafde/backbone.marionette/pull/2

I _think_ it works.. may not be worth it.. I don't think the perf is worse.. possibly better on reorder.. but haven't fully vetted it.  This can be dropped if it's too much.. just attempting to simplify.

rafde:
Yeah... the modelsOrViews thing looks weird. I'll evaluate when I get home and see.

src/collection-view.js
```diff
@@ -128,20 +128,20 @@ const CollectionView = Backbone.View.extend({
...
     const changes = options.changes;
-    this._removeChildViews(changes.removed);
+    this._removeChildModels(changes.removed);
   },
 
   // Remove the child views and destroy them.
   // This function also updates the indices of later views
   // in the collection in order to keep the children in sync with the collection.
-  // "modelsOrViews" can be an array of mixed models or views that
+  // "models" is an array of models and the cooresponding views
```
scott-w: Typo on `corresponding`

paulfalgout:
👍 @marionettejs/marionette-core can we get this in?

JSteunou:
I'm in :+1: I guess the next step is merging `add` into `update`?

rafde:
🎉 There's only add. I want to work on update after add since add will affect updated.

PR #3107, opened 2016-08-06
Make specs compatible with jquery v3

### LINKED ISSUE #3095: Make specs compatible with jquery v3
We can't currently update the jquery dependency used in our tests because 6 tests are incompatible with jquery3.  Again this isn't Marionette's code, that works fine, but the tests themselves are using deprecated functionality.

```
1) composite view when rendering with a false template should bind the ui hash to jQuery selectors:
     AssertionError: expected undefined to equal '.test-element'
      at Context.<anonymous> (composite-view.spec.js:950:61)

  2) region when initializing a region and passing an "el" option should manage the specified el:
     AssertionError: expected undefined to equal '#foo'
      at Context.<anonymous> (region.spec.js:807:43)

  3) layoutView when defining region selectors using @ui. syntax should apply the relevant @ui. syntax selector to a simple string value:
     AssertionError: expected undefined to equal '.craft'
      at Context.<anonymous> (view.child-views.spec.js:558:64)

  4) layoutView when defining region selectors using @ui. syntax should apply the relevant @ui. syntax selector to selector in a region definition object:
     AssertionError: expected undefined to equal '.bros'
      at Context.<anonymous> (view.child-views.spec.js:562:66)

  5) layoutView when defining region selectors using @ui. syntax should apply the relevant @ui. syntax selector to el in a region definition object:
     AssertionError: expected undefined to equal '.toadstool'
      at Context.<anonymous> (view.child-views.spec.js:566:69)

  6) item view when rendering with a false template should bind the ui hash to jQuery selectors:
     AssertionError: expected undefined to equal '.test-element'
      at Context.<anonymous> (view.spec.js:89:52)
```

rafde:
### Proposed changes
Update specs to not do test on removed jQuery property `.selector`

Link to the issue:
fixes #3095

Marionette V3

### DIFF
--- a/test/unit/composite-view.spec.js
+++ b/test/unit/composite-view.spec.js
@@ -946,10 +946,6 @@ describe('composite view', function() {
       expect(this.bindUIElementsSpy).to.have.been.calledOnce;
     });
 
-    it('should bind the ui hash to jQuery selectors', function() {
-      expect(this.compositeView.ui.testElement.selector).to.equal('.test-element');
-    });
-
     it('should trigger a before:render event', function() {
       expect(this.triggerSpy).to.have.been.calledWith('before:render', this.compositeView);
     });
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -804,7 +804,7 @@ describe('region', function() {
     });
 
     it('should manage the specified el', function() {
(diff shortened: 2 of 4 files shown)


scott-w:
This looks good to me 👍

paulfalgout:
Well..  So now our dev dependencies should be jquery v3

Issue #3103, opened 2016-08-05
Separate Templates from View


scott-w:
The template rendering logic is mixed in with the View documentation. We should separate this into its own file so it's easier for the reader to find the specific information they need. It'll also give us a clear place to put all the extra little bits from `marionette.templatecache.md` and `marionette.renderer.md` into.

All three of these sections should link heavily to the initial template document so we can direct the user to find it more easily.

denar90:
resolved #3108

PR #3108, opened 2016-08-07
Separate template docs

scott-w:
### Proposed changes
 - Create a template docs
 - Move all template docs into template
 - Re-link docs into template instead of view

Link to the issue: #3103


denar90:
One more 👍 Thank you @scott-w!

Issue #3110, opened 2016-08-07
Remove link to Backbone.BabySitter in CollectionView doc


denar90:
Since we don't have dependancy from `Backbone.BabySitter` I propose remove link on it form our doc
https://github.com/marionettejs/backbone.marionette/blob/next/docs/marionette.collectionview.md#collectionviews-children
and inject doc of it into Mn doc. It can be separate file or just extending our CollectionViewDoc file.

@scott-w @paulfalgout what do you think?

scott-w:
Sounds like a plan. We can just do a copy and paste job initially and work it around later.

ahumphreys87:
fixed via #3111

Issue #3052, opened 2016-06-06
CollectionView Documentation


scott-w:
The documentation for `CollectionView` needs to be updated in line with Marionette 3 standards.

## Todo

* [x] Outline `CollectionView` features
  * [x] How to show child views
  * [x] Empty view behaviour
  * [x] Example rendering for different views
* [x] Document `CollectionView` specific events
  * [x] Move Collection View events into the `events.md` file
  * [x] Create a separate set of files for View Lifecycle events
* [x] Separate in-depth rendering topics
  * [x] Build/add/render child views
  * [x] Managing child views (the Babysitter docs)
  * [x] Sorting and filtering the collection

paulfalgout:
https://github.com/marionettejs/backbone.marionette/issues/3051

scott-w:
I'm just looking this over and the `marionette.collectionview.md` doc is pretty long. I'm thinking we need to decide on a base set of functionality to document and pull the more advanced topics into their own section.

ahumphreys87:
@scott-w I agree, cover common use cases and maybe just mention other functionality with links to the associated docs for it

ahumphreys87:
Maybe, simple rendering and auto re-rendering, rendering a table (with a layout), sorting/filtering and reordering. Thoughts?

paulfalgout:
@scott-w the events moving seems reasonable.  The rendering at the bottom could possibly be standalone as well.. I don't really know what would be advanced topics?  There are some things people will never touch.. _maybe_ sorting/filtering is more like secondary functionality.. it's pretty hard to break up the API reasonably in my opinion. :-/

denar90:
Agree with you guys. I also noticed that we have some example of rendering tables using v2 and v3, should this section also be moved (e.g. upgrade guide) ?

scott-w:
Good stuff, I'll update the description here.

denar90:
Updated description due to #3117

jonas-db:
Might be relevant to put into the documentation?https://github.com/marionettejs/backbone.marionette/issues/458#issuecomment-241187938

paulfalgout:
@jonas-db https://github.com/marionettejs/backbone.marionette/blob/v3.0.0-pre.5/docs/marionette.collectionview.md#tables-using-marionette-3

PR #3111, opened 2016-08-08
Add backbone.babysitter doc

denar90:
### Proposed changes
 - Added babysitter doc to collection view 

Link to the issue:
#3110


docs/marionette.collectionview.md
```diff
@@ -22,7 +22,14 @@ in the DOM. This behavior can be disabled by specifying `{sort: false}` on initi
 * [CollectionView's children](#collectionviews-children)
   * [CollectionView's `buildChildView`](#collectionviews-buildchildview)
   * [CollectionView's `addChildView`](#collectionviews-addchildview)
+  * [CollectionView: Retrieve Child Views](#collectionviews-retrievechildview)
+    * [CollectionView's: `findByCid`](#collectionviews-retrievechildview-findbycid)
```
paulfalgout: I think the menu items should specify that it's the children somehow I think.. Otherwise just skimming the contents you would assume you could `myCollectionView.findByCid(1)`

ahumphreys87:
@scott-w this kinda clashes with your work on the CollectionView docs, the babysitter docs would be a candidate for an advanced topic to sit outside of the main CollectionView doc.

@denar90 what are your thoughts on the suggestion over in #3052 ?

paulfalgout:
@ahumphreys87 it does clash, but I wonder if it'll be easier to just dump it all in now and the reorg it all?  To me this is 👍 and then #3052

denar90:
As you wish guys, it's not a problem for me to make changes into right place. Where would you suggest to move this part?

ahumphreys87:
Yeah I agree with @paulfalgout, get all the docs written in 1 place then we can move around.

👍

PR #3113, opened 2016-08-09
CollectionView and CompositeView rendering views performance increase

### LINKED ISSUE #2873: CollectionView#showCollection rendering performance issue
Marionette 2.4.4
BackboneJS 1.2.3 
underscore 1.8.3
jQuery 2.1.3

http://codepen.io/rafde/pen/GorLyg?editors=101

perf issue

When Marionette renders a collection view of 3000 models, the view takes (on my computer) between 2.5 and 3 seconds to render the entire collection.

Using Chrome Dev tools, I saw these lines affect rendering performance.

https://github.com/marionettejs/backbone.marionette/blob/v2.4.4/lib/backbone.marionette.js#L84
https://github.com/marionettejs/backbone.marionette/blob/v2.4.4/lib/backbone.marionette.js#L2634
https://github.com/marionettejs/backbone.marionette/blob/v2.4.4/lib/backbone.marionette.js#L2577

To go around this, I ended up extending the collection view and overwriting the Marionette implementation
https://github.com/marionettejs/backbone.marionette/blob/v2.4.4/lib/backbone.marionette.js#L2414
```js
showCollection: function() {
      var ChildView;
      var models = this._filteredSortedModels();
      var childViewOptions = this.getOption('childViewOptions');
      var firstView;
      var firstIndex;

      if (models.length) {
        ChildView = this.getChildView(models[0]);

        _.each(models, function(child, index) {
          var computedChildViewOptions = Marionette._getValue(childViewOptions, this, [child, index]);
          var view = this.buildChildView(child, ChildView, computedChildViewOptions);

          if (_.isUndefined(firstView)) {
            firstView = view;
            firstIndex = index;
          }

          this.triggerMethod('before:add:child', view);
          this._addChildView(view, index);
          this.triggerMethod('add:child', view);

          view._parent = this;
        }, this);
        //UPDATE index after everything is done adding.
        this._updateIndices(firstView, true, firstIndex);
        this.children._updateLength();
      }

      return this;
},
// Internal Method. Add the view to children and render it at
// the given index.
_addChildView: function(view, index) {
      // Only trigger attach if already shown, attached, and not buffering, otherwise endBuffer() or
      // Region#show() handles this.
      var canTriggerAttach = this._isShown && !this.isBuffering && Marionette.isNodeAttached(this.el);
      var nestedViews;

      // set up the child view event forwarding
      this.proxyChildEvents(view);

      view.once('render', function() {
        // trigger the 'before:show' event on `view` if the collection view has already been shown
        if (this._isShown && !this.isBuffering) {
          Marionette.triggerMethodOn(view, 'before:show', view);
        }

        // Trigger `before:attach` following `render` to avoid adding logic and event triggers
        // to public method `renderChildView()`.
        if (canTriggerAttach && this._triggerBeforeAttach) {
          nestedViews = this._getViewAndNested(view);
          this._triggerMethodMany(nestedViews, this, 'before:attach');
        }
      }, this);

      // OVERWRITE what Marionette is doing to prevent calling
      // Backbone.ChildViewContainer#_updateLength
      this.children.addBulk(view);

      this.renderChildView(view, index);

      // Trigger `attach`
        nestedViews = this._getViewAndNested(view);
        this._triggerMethodMany(nestedViews, this, 'attach');
      }
      // Trigger `show`
      if (this._isShown && !this.isBuffering) {
        Marionette.triggerMethodOn(view, 'show', view);
      }
},
```js
and had to create a new function like 
https://github.com/marionettejs/backbone.marionette/blob/v2.4.4/lib/backbone.marionette.js#L72
but one that doesn't do `_updateLength`
```js
addBulk: function(view, customIndex) {
  var viewCid = view.cid;
  // store the view
  this._views[viewCid] = view;
  // index it by model
  if (view.model) {
    this._indexByModel[view.model.cid] = viewCid;
  }
  // index by custom
  if (customIndex) {
    this._indexByCustom[customIndex] = viewCid;
  }
  //REMOVED index updater
  return this;
},
```
This fixed the major bottlenecks but there are other parts that weren't addressed such as simple `ItemView` event delegation also slows down rendering when a `CollectionView` renders it. This fix is also a hack for the case that I need it to render.

I am looking for some input on a potential improved approach to dealing with major collectionView rendering.

rafde:
# CollectionView and CompositeView rendering views performance increase 

## addChildView
* using `_setupChildView`
* triggers `event~before:add:child`
* triggers `event~add:child`
* when buffering, will add view to list of children
  but will not update children length for performance increase
* using `_renderView`

## _updateIndices
* removed need to pass index

## _addChild
* removed passing ChildView param
* using `_createView`

## _onCollectionAdd
* using `_addChild`

## _showCollection
* using `_addChild`
* calls `children._updateLength`

## _showEmptyView
* removed unnecessary `view._parent` setting

## addChildView
* moved bulk of functions into `_addChildView`

## _createView
* creates view from model

## _renderChildView
* render views and triggers render events

## _setupChildView
* sets child view properties:
  parent, _index (conditionally), monitorViewEvents, and  _proxyChildEvents.

faster

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -122,8 +122,7 @@ const CollectionView = Backbone.View.extend({
 
     if (this._shouldAddChild(child, index)) {
       this._destroyEmptyView();
-      const ChildView = this._getChildView(child);
-      this._addChild(child, ChildView, index);
+      this._addChild(child, index)
     }
   },
 
@@ -402,28 +401,10 @@ const CollectionView = Backbone.View.extend({
     }
   },
 
-  _renderChildView(view) {
-    if (!view.supportsRenderLifecycle) {
-      triggerMethodOn(view, 'before:render', view);
-    }
-
-    // Render view
-    view.render();
-
-    if (!view.supportsRenderLifecycle) {
-      view._isRendered = true;
-      triggerMethodOn(view, 'render', view);
-    }
-
-    return view;
-  },
-
   // Internal method to loop through collection and show each child view.
   _showCollection(models) {
-    _.each(models, (child, index) => {
-      const ChildView = this._getChildView(child);
-      this._addChild(child, ChildView, index);
-    });
+    _.each(models, _.bind(this._addChild, this));
+    this.children._updateLength();
   },
 
   // Allow the collection to be sorted by a custom view comparator
@@ -497,10 +478,8 @@ const CollectionView = Backbone.View.extend({
       const view = this.buildChildView(model, EmptyView, emptyViewOptions);
 
       this.triggerMethod('before:render:empty', this, view);
-      this._addChildView(view, 0);
+      this.addChildView(view, 0);
       this.triggerMethod('render:empty', this, view);
-
-      view._parent = this;
     }
   },
 
@@ -563,11 +542,8 @@ const CollectionView = Backbone.View.extend({
   },
 
   // Internal method for building and adding a child view
-  _addChild(child, ChildView, index) {
-    const childViewOptions = this._getChildViewOptions(child, index);
-
-    const view = this.buildChildView(child, ChildView, childViewOptions);
-
+  _addChild(child, index) {
+    const view = this._createView(child, index);
     this.addChildView(view, index);
 
     return view;
@@ -586,13 +562,21 @@ const CollectionView = Backbone.View.extend({
   // children in sync with the collection.
   addChildView(view, index) {
     this.triggerMethod('before:add:child', this, view);
+    this._setupChildView(view, index);
 
-    // increment indices of views after this one
-    this._updateIndices(view, true, index);
+    // Store the child view itself so we can properly remove and/or destroy it later
+    if (this._isBuffering) {
+      // Add to children, but don't update children's length.
+      this.children._add(view);
+    } else {
+      // increment indices of views after this one
+      this._updateIndices(view, true);
+      this.children.add(view);
+    }
 
-    view._parent = this;
+    this._renderView(view);
 
-    this._addChildView(view, index);
+    this._attachView(view, index);
 
     this.triggerMethod('add:child', this, view);
 
@@ -601,17 +585,12 @@ const CollectionView = Backbone.View.extend({
 
   // Internal method. This decrements or increments the indices of views after the added/removed
   // view to keep in sync with the collection.
-  _updateIndices(views, increment, index) {
+  _updateIndices(views, increment) {
     if (!this.sort) {
       return;
     }
 
     const view = _.isArray(views) ? this._findGreatestIndexedView(views) : views;
-
-    if (increment) {
-      // assign the index to the view
-      view._index = index;
-    }
 
     // update the indexes of views after this one
     this.children.each((laterView) => {
@@ -619,21 +598,6 @@ const CollectionView = Backbone.View.extend({
         laterView._index += increment ? 1 : -1;
       }
     });
-  },
-
-  // Internal Method. Add the view to children and render it at the given index.
-  _addChildView(view, index) {
-    monitorViewEvents(view);
-
-    // set up the child view event forwarding
-    this._proxyChildEvents(view);
-
-    // Store the child view itself so we can properly remove and/or destroy it later
-    this.children.add(view);
-
-    this._renderView(view);
-
-    this._attachView(view, index);
   },
 
   _renderView(view) {
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -382,6 +382,42 @@ const CollectionView = Backbone.View.extend({
     }
   },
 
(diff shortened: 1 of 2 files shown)


src/collection-view.js
```diff
@@ -279,19 +268,55 @@ const CollectionView = Backbone.View.extend({
...
+  _createView(model, index) {
+    const ChildView = this._getChildView(model);
+    const childViewOptions = this._getChildViewOptions(model, index);
+    const view = this.buildChildView(model, ChildView, childViewOptions);
+    view._parent = this;
+
+    monitorViewEvents(view);
+
+    // set up the child view event forwarding
+    this._proxyChildEvents(view);
```
paulfalgout: just for funzies, what is the perf test difference if you comment out this line?

rafde: noproxy
without proxy (it fails several test, btw)

with proxy
withproxy

The difference isn't drastic (about 100ms for 3000 views)

src/collection-view.js
```diff
@@ -505,17 +502,24 @@ const CollectionView = Backbone.View.extend({
...
+    this.triggerMethod('before:add:child', this, view);
 
     // Store the child view itself so we can properly remove and/or destroy it later
-    this.children.add(view);
+    if (this._isBuffering) {
+      // Add to children, but don't update children's length.
+      this.children._add(view);
+    } else {
+      // increment indices of views after this one
+      this._updateIndices(view, true, index);
```
paulfalgout: this function no longer takes an `index`

src/collection-view.js
```diff
@@ -505,17 +502,24 @@ const CollectionView = Backbone.View.extend({
 
   // Internal Method. Add the view to children and render it at the given index.
   _addChildView(view, index) {
-    monitorViewEvents(view);
-
-    // set up the child view event forwarding
-    this._proxyChildEvents(view);
+    this._setupChildView(view, index);
+    this.triggerMethod('before:add:child', this, view);
```
paulfalgout: I think `before:add:child` should be the first thing to happen in the function before anything else occurs.

Also I don't think there's a need for `_addChildView` we can just put it all in `addChildView`?

paulfalgout: alternatively if we need want to keep the public api separate from the private api we should only use the private internally.

src/collection-view.js
rafde: oh yeah... I botched the merge conflict. I'll git that fixed.

rafde:
I'll rebase this later.

src/collection-view.js
```diff
@@ -549,14 +561,26 @@ const CollectionView = Backbone.View.extend({
   // This will also update the indices of later views in the collection in order to keep the
   // children in sync with the collection.
   addChildView(view, index) {
+    if (!(view instanceof Backbone.View)) {
```
paulfalgout: I'm on the fence here.  this prevents any error from occurring if you accidentally pass in something that isn't a view..  do we do this kind of bail out elsewhere?

src/collection-view.js
```diff
@@ -1,16 +1,16 @@
...
-import Backbone           from 'backbone';
-import destroyBackboneView from './utils/destroy-backbone-view';
-import isNodeAttached     from './common/is-node-attached';
-import childViewEventHandler from './common/child-view-event-handler';
-import monitorViewEvents  from './common/monitor-view-events';
-import { triggerMethodOn } from './common/trigger-method';
-import ChildViewContainer from './child-view-container';
-import MarionetteError    from './error';
-import ViewMixin          from './mixins/view';
+import _                      from 'underscore';
```
paulfalgout: spacing dealy again.  Sorry..  We should lint for this.

src/collection-view.js
rafde: > do we do this kind of bail out elsewhere?

Nope, but if this doesn't happen. The rest of the collection won't render.

I instinctively do these types of checks in my code.

Can can see either case: let it fail loudly or let it slide. Would be nice if we got another vote on leaving or keeping.

rafde: actually, after reading https://github.com/marionettejs/backbone.marionette/pull/3155#issuecomment-245181213 I'll remove.
If it becomes an issue and the check is need, easy to update.

paulfalgout: I think my preference would lean towards letting it crash or throwing an error vs essentially a noop... but there might be a use-case i'm overlooking.

paulfalgout: ah I was responding to the previous comment.  cool 👍

paulfalgout:
:+1: @marionettejs/marionette-core  need 1 more to merge

JSteunou:
:+1:

rafde:
🎉🎉onto update add 🎉🎉

PR #3114, opened 2016-08-09
Unexpose ChildViewContainer

paulfalgout:
I had originally exposed it because.. why not?  Babysitter’s was available on Backbone.. but if the goal is to change this such that it gets more heavily integrated into collectionview than we don’t want people using it directly.. removing ensures we can integrate it without introducing breaking changes.


denar90:
👍

ahumphreys87:
👍

Issue #3115, opened 2016-08-09
Update travis matrix to latest bb / _


paulfalgout:
We want to have Bb 1.3.3 features in `v3.x` so we should remove testing the old libs.

PR #3118, opened 2016-08-12
Update travis matrix to latest bb

denar90:
Link to the issue: #3115
@paulfalgout do we need to update `_` version to 1.8.3 or we can leave it at 1.8


paulfalgout:
I think `1.8` will install `1.8.3` so that is ok.. Bb `1.3.1` however was buggy I think..  not sure we need to test against it.   But.. now something else is broken.. probably jsdom.. but not sure.. it _was_ working not too long ago.. so possibly a recent patch on a dependency broke something 😿

denar90:
@paulfalgout so you think that 's why build is broken?

rafde:
Please test for 1.3.3. All changes forward for collection will rely on 1.3.3 stuff.
There isn't even a changelog for 1.3.0 - 1.3.2, only 1.3.3 http://backbonejs.org/#changelog

denar90:
Locally all tests are passed...

rafde:
Yeah... but seems odd to test for a version that isn't canonical.

PR #3119, opened 2016-08-12
cleanup region docs

denar90:
### Proposed changes
 - Cleanup some docs for regions due to @scott-w comment https://github.com/marionettejs/backbone.marionette/issues/2959#issuecomment-220850168

Link to the issue: #2959


docs/marionette.region.md
```diff
@@ -207,8 +196,7 @@ the DOM.
 If you replace the current view with a new view by calling `show`,
 by default it will automatically destroy the previous view.
 You can prevent this behavior by passing `{preventDestroy: true}` in the options
-parameter. Several events will also be triggered on the views; see
-[Region Events And Callbacks](#region-events-and-callbacks) for details.
+parameter. Several events will also be triggered on the views;
```
rafde: > Several events will also be triggered on the views;

I think this ends with a period now.

scott-w:
It looks like something has changed in the Travis config.

docs/marionette.region.md
denar90: @rafde sorry, what did you mean?

rafde: > Several events will also be triggered on the views.

sorry for the ambiguity.

scott-w:
Happy to merge this in. If we decide to reinstate `Region` docs later, we can have a wider ranging discussion about how to do it more effectively. :+1:

docs/marionette.region.md
```diff
@@ -234,67 +222,6 @@ mainRegion.empty({preventDestroy: true});
 **NOTE** When using `preventDestroy: true` you must be careful to cleanup your
 old views manually to prevent memory leaks.
 
-#### onBeforeAttach & onAttach
```
paulfalgout: this is gone

docs/marionette.region.md
```diff
@@ -234,67 +222,6 @@ mainRegion.empty({preventDestroy: true});
...
-```
-
-Or you can leave the events on by default but disable them for a single show.
-
-```javascript
-// This region will trigger attach events by default but not for this particular show.
-myRegion.show(myView, {triggerBeforeAttach: false, triggerAttach: false});
-```
-
-#### `renderView`
```
paulfalgout: this is gone

docs/marionette.region.md
```diff
@@ -234,67 +222,6 @@ mainRegion.empty({preventDestroy: true});
...
-
-var CachingRegion = Mn.Region.extend({
-  shouldDestroyView: function(view, options) { return false; },
-  renderView: function(view, options) {
-    if (!view.isRendered) { view.render(); }
-  }
-});
-```
-
-#### `shouldDestroyView`
```
paulfalgout: this is gone

docs/marionette.region.md
```diff
@@ -381,98 +308,3 @@ var MyView = Mn.View.extend({
   }
 });
 ```
-
-## Region Events And Callbacks
-
-A region will raise a few events on itself and on the target view when showing and destroying views.
-
-### Events Raised on the Region During `show()`
```
paulfalgout: this still exist

docs/marionette.region.md
```diff
@@ -381,98 +308,3 @@ var MyView = Mn.View.extend({
...
-
-* `before:show` / `onBeforeShow` - Called after the view has been rendered, but before its been displayed.
-* `show` / `onShow` - Called when the view has been rendered and displayed.
-* `before:empty` / `onBeforeEmpty` - Called before the view has been emptied.
-* `empty` / `onEmpty` - Called when the view has been emptied.
-
-```javascript
```
paulfalgout: really.. this is for supporting lifecycle events on a plain-ole `Backbone.View`

docs/marionette.region.md
```diff
@@ -381,98 +308,3 @@ var MyView = Mn.View.extend({
...
-view.supportsRenderLifecycle = true;
-view.supportsDestroyLifecycle = true;
-```
-
-## Custom Region Classes
```
paulfalgout: this still exist

docs/marionette.region.md
```diff
@@ -381,98 +308,3 @@ var MyView = Mn.View.extend({
...
-  }
-});
-```
-
-Note that a region must have an element to attach itself to. If you
-do not specify a selector when attaching the region instance to your
-Application or LayoutView, the region must provide an `el` either in its
-definition or constructor options.
-
-### Instantiate Your Own Region
```
paulfalgout: I definitely do this all the time in fiddles and what-not

denar90:
@paulfalgout I'll look at it and roll back removed docs which are still needed....

docs/marionette.region.md
scott-w: Out of curiosity, how do you use regions in a way that differs from attaching a behaviour? That'll help to structure the documentation so we can point people in the right direction for what they need to be using.

paulfalgout: Behaviors are very contextual to the view so your view has to be aware of what is applied to it.  Regions are containers that can provide a view with context.  The difference while slight makes the code easier to organize I think.  Aside from transitions which would be hard with a behavior, in my app a few things change structurally between desktop / tablet / phone..  The views don't know what context they are in or care, but they make a request to be shown in the sidebar region.  On desktop that region is always visible and can be resized.  On tablet it is fixed horizontally and collapses vertically.  On mobile it is collapsed by default and a menu button slides it into view.  And I can put dozens of different views into the region on the three different platforms and the container handles what happens to them

scott-w: Alright, so in terms of what knows about what, am I correct in assuming the following:

```
  Region
   |
Knows About
   |
   v
   View
   |  ^
   |  |
 Knows About
   |  |
   v  |
 Behavior
```

And you can build your application accordingly.

paulfalgout: Oh however in the case of adding widgets to a page..  A region is a much simpler application that starts the view tree life cycle.   So you might wait until the dom is ready and attach 3 different regions to places in the DOM and use the initialize of the regions to show their views.  Very very simplistic use case.

paulfalgout: Yeah I think that is a good graphic.   I mean you could do anything a custom region would do with a behavior, but it seems like in some cases it would be more complex..  But then you can do everything with a plain view and mixin in functionality..  that you can do with a view and a mn.behavior too..  I think there are some valid use cases for keeping both though.

Issue #3123, opened 2016-08-15
Event hash as methods are called without correct this context


dwt:
### Event hash as methods are called without correct this context

 1. We're building reusable views that have some common event handlers
 2. To achieve this we use methods instead of hashes for the `events` key on Mn.View which call their super implementation and aggregate the returned hashes.
 3. This is however foiled by https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/view.js#L94 which calls the view method without the proper this context, thus not allowing us to easily get to the super method without hardcoding the super object in the method. (We use klass: https://github.com/ded/klass/ to get nice super calls).

### Expected behavior

[_getEvents](https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/view.js#L94) should use apply to provide the proper this context.

### Actual behavior

It does not.

### Environment

1. Marionette version: 3.0 pre 5
2. Backbone version: 1.3.3

paulfalgout:
Good find. 👏

denar90:
merged #3126

PR #3125, opened 2016-08-15
Lock down dependencies

paulfalgout:
Some minor version broke the build.  We should just manually update dependencies when we need to.
And update jQuery to v3

PR #3126, opened 2016-08-15
Call events on view context

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3123


paulfalgout:
*note that the tests are failing because of https://github.com/marionettejs/backbone.marionette/pull/3125

PR #3129, opened 2016-08-17
Add a note about the upgrade tool

paulfalgout:
Also changed it to 2.x because it will be true for most if not all versions of v2 including if there is ever a `2.4.8` or `2.5`


denar90:
👍

Issue #2154, opened 2014-12-21
Differentiate between pre-existing nodes, and not pre-existing nodes


jmeas:
There are two general categories of views in Backbone. One category are pre-existing nodes that are unlikely to be rendered ever. An example of this would be creating a `rootView` and passing the `body` as the element, in a situation where the `body` already has a handful of nodes. The other category are views which may or may not have templates and are rendered throughout the application's lifecycle. The latter of the two categories is by far the most common.

Marionette does not distinguish between these two types of views, which creates some bad side effects. For instance, both sorts of LayoutViews are treated identically, which leads to [some nasty bugs](https://github.com/marionettejs/backbone.marionette/issues/1971).

One way to resolve situations like that are to differentiate between these two use-cases. We can infer that a view exists based on two criteria: it has no template, and it has children nodes.

In other words:

```js
_preExistingNode: function() {
  return !this.template && this.$el.children().length;
}
```

During the instantiation phase of any view, this would be used to dictate the behavior of the view. Take a LayoutView, for instance. Pre-existing nodes would have their regions instantiated instantly, and their UI would also be bound instantly. LayoutViews that aren't pre-existing would wait until the first render before doing these actions.

paulfalgout:
Pretty sure this is resolved in v3.1

PR #3135, opened 2016-08-19
Allow CollectionView to be populated with pre-rendered DOM

paulfalgout:
https://jsfiddle.net/t29dkss1/1/

I don't really know if this would be considered breaking..  I don't really think so..

~~This will need tests..~~ thanks @rafde 

Also one point to note.. overriding `buildChildView` was easy.. however that same function is used to build the emptyView which is kind of unfortunate.. it means that when an emptyView is built we're making an empty model to pass to this function..  and when you override it, we'll have to first detect if the view being passed is the childView or the emptyView before doing something out of the ordinary with the el.

### DIFF
--- a/test/unit/collection-view.spec.js
+++ b/test/unit/collection-view.spec.js
@@ -30,16 +30,21 @@ describe('collection view', function() {
 
   describe('when a collection view is DOM', function() {
     beforeEach(function() {
-      this.$fixtureEl = $('<div id="fixture-collectionview"></div>');
+      this.$fixtureEl = $('<div id="fixture-collectionview"><span id="el1">1</span></div>');
     });
 
     describe('and it\'s not attached to the document', function() {
       beforeEach(function() {
-        this.collectionView = new this.CollectionView({el: '#fixture-collectionview'});
+        this.collectionView = new this.CollectionView({el: this.$fixtureEl[0]});
+        this.view1 = this.collectionView.addChildView(new Backbone.View({el: this.$fixtureEl.children()[0]}), 0);
       });
 
       it('should have _isAttached set to false', function() {
         expect(this.collectionView).to.have.property('_isAttached', false);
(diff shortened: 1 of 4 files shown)


rafde:
looks reasonable. need help with the test?

paulfalgout:
sure thing.  this branch is on the core, so you can commit right to it if ya want.

rafde:
I'll git to it sometime today.

test/unit/collection-view.spec.js
```diff
@@ -28,6 +28,34 @@ describe('collection view', function() {
   // Collection View Specs
   // ---------------------
 
+  describe('when a collection view is DOM', function() {
+    beforeEach(function() {
+      this.$fixtureEl = $('<div id="fixture-collectionview"></div>');
```
ahumphreys87: can we change the fixture to cater for when childViews are pre-rendered as well?

rafde: something like
```
this.$fixtureEl = $('<div id="fixture-collectionview"><span id="el1">1</span></div>');

this.view1 = new this.ChildView({el: this.$fixtureEl.find('#el1')[0]});
```
then
```
this.collectionView.addChildView(this.view1, 0);
```
although I haven't tried that addChildView nor do I know if it will work

paulfalgout: it can also just be `new this.ChildView({ el: '#el1' });`  and I do think `addChildView` would work work.

rafde:
@paulfalgout @ahumphreys87 updated.

ahumphreys87:
Slightly related issue https://github.com/marionettejs/backbone.marionette/issues/2154

rafde:
:+1:

denar90:
👍

scott-w:
:+1:

PR #3143, opened 2016-08-27
Separate lifecycle events

scott-w:
Link to the issue: #3052

I’ve made a new viewlifecycle file that collects the `CollectionView`
and `View` documentation lifecycles in a single point.

The idea is that this should be easier to find and cuts a block of text
from the collectionview docs.

If you guys have any feedback, things that you'd like to see or that I'm missing out, let me know and I'll get it in.

This will also need a modification to the website for the documentation index ordering.


docs/viewlifecycle.md
```diff
@@ -0,0 +1,439 @@
...
+    console.log('el exists but is not visible in the DOM');
+  }
+});
+```
+
+#### View `before:attach`
+
+Triggered after the View has been rendered but just before it is first bound
+into the page DOM. This will only be triggered once per `region.show()` - if
+you want something that will be triggered every time the DOM changes,
```
rafde: Maybe don't include
>  - if you want something that will be triggered every time the DOM changes,  you may want `render` or `before:render`.

This leads me to think that if I do
`this.ui.myEl.append('<span></span>')` then the render event will trigger, when that's not the case.

docs/viewlifecycle.md
```diff
@@ -0,0 +1,439 @@
...
+you may want `render` or `before:render`.
+
+#### View `attach`
+
+Triggered once the View has been bound into the DOM. This is only triggered
+once - the first time the View is attached to the DOM. If you want an event that
+triggers every time the DOM changes visibly, you may want `dom:refresh`
```
rafde: > If you want an event that triggers every time the DOM changes visibly, you may want `dom:refresh`

I don't think you need this. `dom:refresh` section should explain itself.

docs/viewlifecycle.md
```diff
@@ -0,0 +1,439 @@
...
+
+#### View `dom:refresh`
+
+Triggered after the first `attach` event fires _and_ every time the visible DOM
```
rafde: `dom:refresh` is sort of a combination of `attach` and `render`
I think I would go with saying

> Triggered after the first `attach` event fires _and_ the view's `render` is called while the view is in the DOM.

Although the explanation isn't that much more insightful. I think providing an example would help.

```
const myView = new Mn.View({
  template: _.template('<span><%= count %><span>'),
  templateContext: function(){
    this.count = (this.count || 0) + 1;
    return {
      count: this.count
    };
  }
});

myView.on('render', ()=> console.log('render'));
myView.on('attach', ()=> console.log('attach'));
myView.on('dom:refresh', ()=> console.log('dom:refresh'));

// made up variable
layout.showChildView('myRegion', myView);
// render, attach, and dom:refresh should fire

myView.render();
// render, and dom:refresh should fire

File: docs/viewlifecycle.md (hunk 1 above)
Comment (scott-w) [inline docs/viewlifecycle.md]: Good point. While it's *recommended* you only update the DOM through `render` invocations, we can't reply on that 👍

File: docs/viewlifecycle.md (hunk 2 above)
Comment (scott-w) [inline docs/viewlifecycle.md]: My thinking here is that the reader might read this, think "oh, this is what I need" then not read `dom:refresh`.

File: docs/viewlifecycle.md (hunk 3 above)
Comment (scott-w) [inline docs/viewlifecycle.md]: Thanks for that, I'll put it in.

File: docs/viewlifecycle.md [hunk 4]
```diff
@@ -0,0 +1,478 @@
...
+|   3   |    `before:add:child`    |
+|   4   |       `add:child`        |
+|  5*   |      `render:empty`      |
+|  5+   |     `render:children`    |
+|   6   |          `render`        |
+|   7   |      `before:attach`     |
+|   8   |         `attach`         |
+|   9   |       `dom:refresh`      |
+
+The events marked with "\*" only fire on empty collections and events marked
```
rafde: "\*" intentional?

rafde:
👍

denar90:
@scott-w good job) Is it possible to add live examples or you think it's better to do it separately?

scott-w:
Probably best to do separately. I'm not great messing about with JSFiddle and all that stuff.

rafde:
@scott-w if you need help with those, feel free to ask. I'm volunteering @denar90 . Or I could do it later if you like.

denar90:
👍  then

PR #3146, opened 2016-08-29
Update installation.md

mariiapaniutina:
### Proposed changes
Installing marionette with specific version, because just `npm install backbone.marionette` or `npm install backbone.marionette latest-version` installing version 2.4.7 instead 3.0.0

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,6 +5,12 @@ a Marionette application. In this section we'll cover the most common ways.
 
 
 ## Quick start using NPM and Webpack
+[NPM](https://www.npmjs.com/) is the package manager for JavaScript.
+
+Installing with NPM through command-line interface
+```
+npm install backbone.marionette@3.0.0
+```
 
 [Webpack][webpack] is a build tool that makes it easy to pull your dependencies
 together into a single bundle to be delivered to your browser's `<script>` tag.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -9,7 +9,7 @@ a Marionette application. In this section we'll cover the most common ways.
 
 Installing with NPM through command-line interface
 ```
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -5,6 +5,12 @@ a Marionette application. In this section we'll cover the most common ways.
 
 
 ## Quick start using NPM and Webpack
+[NPM](https://www.npmjs.com/) is the package manager for JavaScript.
+
+Installing with NPM through command-line interface
+```
+npm install backbone.marionette@3.0.0
```
rafde: 1. `npm install backbone.marionette` is enough.
2. `npm install backbone.marionette --save` could also be good

I tested this on my machine, and 3.0.0 installed without forcing version install.
I want to guess that
> npm install backbone.marionette latest-version installing version 2.4.7 instead 3.0.0

might have happened because your package.json had ^2.4.x, which won't install 3.x version.

Anyhow, 1 or 2 might be a better alternative. I'll let other chime in regarding this message.

Thanks for the PR.

paulfalgout:
I don't want to suggest to people that they install at a particular version.  This sounds like either a package.json issue like @rafde suggested or possibly a local npm cache.  Either way I don't want to be constantly updating the docs when new versions come out.

denar90:
@mariiapaniutina thank you, but I'm closing due to comments.

paulfalgout:
I'm actually :+1: on this without the "@3.0.0"

rafde:
@mariiapaniutina please update 
`npm install backbone.marionette@3.0.0` 
to
`npm install backbone.marionette`

mariiapaniutina:
Thanks all. Updated without adding a specific version.

denar90:
👍

rafde:
:+1:

PR #3153, opened 2016-08-31
Fix Backbone.Radio version in bower.json

### LINKED ISSUE #3148: Fix Bower dependencies
Marionette 3.0.0:

- The Bower dependency for Backbone.Radio is still stuck at 2.0.0-pre.1. 
- Backbone and Underscore are missing. Bower doesn't have peerDependencies (it [wouldn't make sense](https://github.com/bower/bower/issues/292), dependencies are not nested anyway). So it appears that Backbone and Underscore should be listed as ordinary dependencies, which is nearly the same as npm peerDependencies.

  (I'd recommend a range of acceptable releases, not a "newest or higher" `~1.3.3`.)

Marionette 2.4.7:

- Installs Backbone 1.2.3, rather than 1.3.3. The version range in bower.json is ok, though. The problem seems to originate in Backbone.Babysitter as well as Backbone.Wreqr, which both use an invalid range format (`"backbone": ">=0.9.9 <=1.3.x"`). I've opened [issues](https://github.com/marionettejs/backbone.babysitter/issues/79) [there](https://github.com/marionettejs/backbone.wreqr/issues/83).

hashchange:
Fixes #3148 partially. Pending PRs for Backbone.Babysitter and Backbone.Wreqr do the rest.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -36,6 +36,6 @@
     "upgradeGuide.md"
   ],
   "dependencies": {
-    "backbone.radio": "2.0.0-pre.1"
+    "backbone.radio": "2.0.0"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -36,6 +36,6 @@
     "upgradeGuide.md"
   ],
   "dependencies": {
-    "backbone.radio": "2.0.0"
+    "backbone.radio": "^2.0.0"
   }
 }


paulfalgout:
👍

rafde:
👍

Issue #2640, opened 2015-06-29
Remove `preventDestroy` option from region `show` and `empty`


ahumphreys87:
`preventDestroy` is a bad idea. It will create memory leaks everywhere!

Lets remove the option in 3.0 and encourage people to create and dispose of views correctly

trezy:
:+1:

samccone:
there are still many usecases for this feature, keep in mind people can have expensive views that take time to rerender. The initial goal for the flag was to let users know they were opting into a dangerous thing. Since we prevent this behavior by default, and plenty of  (power) users use the flag I am totally fine keeping this.

bgaillard:
Hi, we use the `preventDestroy` option is several Cordova / Marionette applications. This is very useful when we need to agressively cache views on client side.

trezy:
@samccone That's a good point. There are valid reasons for using `preventDestroy` but our implementation is super risky. We could at least do a few things to make it less dangerous such as maintaining a list of views to be GC'd and providing mechanisms for a lower level automatic GC plus or easier resurrection. If this doesn't make sense I can elaborate on it.

The point is that while preventDestroy may have valid use cases, I believe we could offer a less risky alternative that still meets those needs.

paulfalgout:
I'd certainly like to explore alternatives.  Anecdotally it seems `preventDestroy` is very often used for poor or misguided reasons and it seems to complicate a lot of Mn development.

ahumphreys87:
In my haste to remove, im now happy we havent - had a case where this was needed recently for modals.

I'm gonna vote we keep as an option with it defaulting to off

paulfalgout:
Can you describe the use-case more?  I still think there's a potential to solve this a better way

PR #3155, opened 2016-08-31
Make a public detachView

paulfalgout:
This function removes the need for a `preventDestroy` option. https://github.com/marionettejs/backbone.marionette/issues/2640

Essentially if you want to `preventDestroy` just detach the view first. When you want to preventDestroy you already have to have a copy of the view.. and depending on what is doing the view emptying `preventDestroy` can become a round about API.

What you really want is
```js
const myView = myRegion.currentView;
myRegion.detachView();
myOtherRegion.show(myView);
```

Upon approval:
- [x] Tests
- [x] Docs

### DIFF
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -180,6 +180,10 @@ export default {
     return region.show(view, ...args);
   },
 
+  detachChildView(name) {
+    return this.getRegion(name).detachView();
+  },
+
   getChildView(name) {
     return this.getRegion(name).currentView;
   }
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -18,6 +18,7 @@ managing regions throughout your application.
 * [Showing a View](#showing-a-view)
   * [Hiding a View](#hiding-a-view)
   * [Preserving Existing Views](#preserving-existing-views)
+  * [Detaching Existing Views](#detaching-existing-views)
 * [Checking whether a region is showing a view](#checking-whether-a-region-is-showing-a-view)
 * [`reset` A Region](#reset-a-region)
 
@@ -218,6 +219,23 @@ mainRegion.empty({preventDestroy: true});
 
 **NOTE** When using `preventDestroy: true` you must be careful to cleanup your
 old views manually to prevent memory leaks.
+
+### Detaching Existing Views
+If you want to detach an existing view from a region, use `detachView`.
+
+```javascript
+var myView = new MyView();
+
+var myOtherView = new MyView();
+
+var childView = new MyChildView();
+
+// render and display the view
+myView.showChildView('main', childView);
+
+// ... somewhere down the line
+myOtherView.showChildView('main', myView.getRegion('main').detachView());
+```
 
 ### Checking whether a region is showing a view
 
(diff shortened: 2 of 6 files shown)


rafde:
how about 
```
var view = parentView.detachChildView('region');
myOtherRegion.show(myView);
``

which proxies detachView. That way people don't directly interact with the region.

**Comment by paulfalgout** [comment]
I think a `detachChildView` makes since, but I'd think it'd be added along side.. similarly to `showChildView` where that's the suggested API, but you can interact with the Region if need be.  And I think https://github.com/marionettejs/backbone.marionette/pull/3156 is a reason you may want to..

**Comment by JSteunou** [comment]
Hooo I like those ideas. @rafde your idea of an API from layout point of view is cool!

**Comment by rafde** [comment]
Yeah, I'm for this. Someone in gitter wanted something like this. @paulfalgout mind if I help with the `View#detachChildView` part? Or just general help? Or you got this?

**Comment by rafde** [comment]
```js
detachChildView(region){
  return this.getRegion(region).detachView();
}
```

```js
detachView() {
  const view = this.currentView;
  if(!view){
    return;
  }
  delete this.currentView;
  this._detachView(view);
  delete view._parent;
  return view; // I think this should return the view getting detached
},
```

paulfalgout:
@rafde you can take it over if you want.  

good call about the return if `!view`
And you're totally right about returning the view..  I almost did that at first, but every other function was returning `this`.. however add/remove functions elsewhere return the view.

rafde:
Ah right, @paulfalgout. I get that you don't want to write up the test, 😄 .
Can this work be placed in the marionettejs repo branch? Or should I just PR against your fork branch?

paulfalgout:
@rafde we can move it, or you can PR on my branch, or just close this and reopen another.. I'm up for whatever..  I mean I'll also do it eventually, but if you wanted to step in 👯

rafde:
I'll PR against yours for now.
I'm thinking we should start creating repo branches for collab work.

rafde:
https://github.com/paulfalgout/backbone.marionette/pull/4 forgot the docs, though.

ahumphreys87:
Nice i like this. Should we add a deprectation notice to `preventDestroy`?

src/mixins/regions.js
```diff
@@ -180,6 +180,10 @@ export default {
     return region.show(view, ...args);
   },
 
+  detachChildView(name) {
+    return this.getRegion(name).detachView();
```
rafde: forgot to do a region exist check
```js
  const region = this.getRegion(name);
  if (!region) {
    return;
  }
  return region.detachView(); 
```

paulfalgout: FWIW that check isn't being done in other places

rafde:
welp, I am done here. 👍

JSteunou:
Nice! Just concerned about that check on `getRegion('')`. I understand adding the check is not *a thing* in Marionette src code and it would silent catch an error making difficult to debug, but could checking + logging in debug mode be a nice feature?

rafde:
@JSteunou I am not sure what you are looking for. Snippet, please.

paulfalgout:
I'm in general more interested in consistency over completeness.   Even if it is decided that we should be throwing a warning in debug or an error or whatever.. we should handle that in a later issue and across the lib.

rafde:
I am for that sentiment.

paulfalgout:
@ahumphreys87 https://github.com/marionettejs/backbone.marionette/issues/3168

JSteunou:
@paulfalgout completely agreed, just throwing the ball. Wouldnt make this PR fuzzy with a lot of changes not related.

Nice work btw :+1:

scott-w:
I'll add a :+1: too

PR #3156, opened 2016-08-31
Amend region show to allow for async showing and animation

paulfalgout:
In v2 we introduced swap events to region, but since they were sync they weren’t so useful.

This allows the user to:

``` js
const MyRegion = Mn.Region.extend({
  thenShow(view, options) {
    var p = new Promise((resolve) => {
        this.currentView.fadeOut().done(resolve);
    });
    p.then(() => {
      this.empty(options);
      this.finallyShow(view, options);
    });
  }
});

```

Additionally this reorganized a couple of things to match the patterns being established here: https://github.com/marionettejs/backbone.marionette/pull/3113 by introducing `_setupView`

I also think that detachView https://github.com/marionettejs/backbone.marionette/pull/3155  has one of it's largest use-cases here. You could do quite a bit by detaching a view.. putting it somewhere temporarily for animating out while you simultaneously show the new view.  Not that you couldn't do it with `preventDefault`, but it is more direct.

Upon approval:
- [ ] Tests
- [ ] Docs


JSteunou:
Could help a lot for animation! I still have a custom region with a looooot of code to override its sync logic.

src/region.js
```diff
@@ -47,36 +47,53 @@ const Region = MarionetteObject.extend({
   // method for you. Reads content directly from the `el` attribute. The `preventDestroy`
   // option can be used to prevent a view from the old view being destroyed on show.
   show(view, options) {
-    if (!this._ensureElement(options)) {
-      return;
-    }
+    if (!this._ensureElement(options)) { return this; }
```
ahumphreys87: breaking change??

ahumphreys87:
Not sure about the public API. Might just be the naming of the methods. Here's a thought...

what about making them options passed to `show`?

```
region.show(view, {
  emptyAction: someFunc,
  showAction: someOtherFunc
});
```

`thenShow` and `finallyShow` could become private defaults?

src/region.js
paulfalgout: Seems more stylistic really.  We return this nearly everywhere for chaining purposes, but then in this case it will have thrown an error so I don't know that the return is ever really used. The only possible benefit other than just consistency I see here is that because we return this in so many places, this may gzip better?  But that could be bs.

rafde:
I like that idea. Possible suggestion, maybe pass the options to ask for a promise that the user can resolve?
```js
// returns Promise
var showPromise = region.show(view, {
  asPromise: true
}).then(()=>{ console.log('do something') };
//...

showPromise.resolve(); 
```

paulfalgout:
@ahumphreys87 the issue is that async must be introduced... so it isn't empty action / show action.. it is a function that asynchronously can control when the region empties and when the region shows.  So the user would never override the `finallyShow` they'd only override `thenShow`.. which could be accepted as an argument of show.. either way it'll need to call `this.empty` at some point and `this.finallyShow` at some point.

The alternative is what @rafde is suggesting where we pass promises.  However I don't know a super clean way to do that where it is allowing the user to control _both_ the empty and the show at different async times.  I suppose when passing `asPromise` the user can just understand that unless they call `region.empty` at some point it isn't going to happen.  Similar to overriding, but more abstract.

Also this would require a new dependency for Mn since promises aren't native everywhere.. plus people enjoy introducing their own promise libraries.  

It seems like an override (or passed function) that calls both `this.empty` and `this.finallyShow` is the simplest implementation?

JSteunou:
On the *promise support* point: not a big fan to add dependencies directly, more like ES5 few years ago, I would just advice some polyfill but that's would be up to the user to include it.

paulfalgout:
Polyfill is fine but breaking

denar90:
What if use `$.Deferred`, we already depend on jquery, so we will not add new one. I would really like to have less deps from other libs...

paulfalgout:
My preference would be to not introduce async at all, but provide an easy way to add it with a recipe or external library.

paulfalgout:
This issue seems to have support, but not the implementation.  As it's growing stale, I opened an issue so this PR can be closed.

Issue #3150, opened 2016-08-30
Lack of `childViewContainer` in Marionette 3 makes valid table layouts difficult.


jgerigmeyer:
### Description

I have a table layout (`<table>`) where I would like each item in a `collection` to render as a `<tbody>`. For example, the goal is markup such as:

```html
<!-- for a collection with two models -->
<table>
  <thead>
    ...
  </thead>
  <!-- first model -->
  <tbody>
    <tr>
      ...
    </tr>
    <tr>
      ...
    </tr>
  </tbody>
  <!-- second model -->
  <tbody>
    <tr>
      ...
    </tr>
    <tr>
      ...
    </tr>
  </tbody>
</table>
```

In Marionette 2, I was able to use a `CompositeView`:

```js
const Item = Marionette.ItemView.extend({
  tagName: 'tbody',
  template: '#item'
});

const List = Marionette.CompositeView.extend({
  template: '#list',
  childViewContainer: 'table',
  childView: Item
});
```

With the following `#item` template:

```html
<tr>
  ...
</tr>
<tr>
  ...
</tr>
```

And the following `#list` template:

```html
<table>
  <thead>
    ...
  </thead>
  <!-- collection will be appended here -->
</table>
```

Because I used the `table` as the `childViewContainer`, the children were simply appended.

With `CompositeView` now being deprecated in Marionette 3, I'm not sure how to get the same end result without adding an extra layer. And I can't really add an extra layer if I want valid table HTML. From reading [the documentation](http://marionettejs.com/docs/master/marionette.collectionview.html#rendering-tables), it seems like the approved new way to do this is to use a `View` wrapper that shows a `CollectionView` in a `region` (probably using `replaceElement: true`). But unless I'm missing something, that still adds one layer (because the `CollectionView` needs an `el` of its own). The examples in the docs assume that the CollectionView can be a `tbody`, and the individual items can be `tr`s. But if I need the items to be `tbody`s, I need a way to render the collection without any wrapper of its own.

It seems like the concept of a `CollectionView` that both renders a template _and_ renders children in a `childViewContainer` has been removed in Marionette 3, without a good replacement. I suppose I could override the `CollectionView` `render` function to render a template as well as the children?

What am I missing here? Is there a recommended way to do table layout like this?

### Environment

1. Marionette version: 3.0.0
2. Backbone version: 1.3.3

[...]

Issue #2560, opened 2015-05-15
Region/CollectionView Plan


ianmstew:
CollectionView has already become Region-like in many circumstances, but the conditionals that enable Region-like behavior are fragile.  The aim of this plan is to essentially promote CollectionView to a "List Region", the direction it has been heading naturally.

**Refactor Events in Region#show()**
1. Move `before:show` to front of `show()`, the timeline other `before` events are triggered (#2552).
   - Tests to verify `Region#show()` event sequencing.
2. Move `onBeforeAttach`/`onAttach` triggering on children to view instead of calling all descendants at once.  This reduces logical overlap with CollectionView.
3. Solve quirks around nested rendering in `onBeforeAttach`/`onAttach` (#2546).
4. Simplify/improve/replace onDomRefresh (#2299).
5. Update official recommendations for showing child views (#2539).

**Region Maintenance**
1. Destroying a Region should empty it as well (#2379)

**Promote CollectionView to "List Region"** #2545
1. Kill `_isShown` checks in CollectionView: no more logical dependency on Region.
2. Match `Region#show()` events in CollectionView.
   - Tests to verify event sequencing identical to `Region#show()`.

**Discuss: Refactor EmptyView**
1. Remove EmptyView from CollectionView (#2495).
2. Add `empty`/`some` events to CollectionView to facilitate alternative "empty view" handling in parent.
3. Behavior or alternative to simplify EmptyView handling in parent.

**Discuss: View "List Region" API**
1. Add “list” API to View to automate a nested CollectionView.
   - **Where does “Replace Region el" ability fit in?
2. Ensure compatibility with EmptyView solution.

**Breaking Changes**
- `before:show` moved to _beginning_ of `Region#show()`
- EmptyView refactoring

ahumphreys87:
@ianmstew Im interested in hearing your thoughts on whether you think we need all the events in `region.show` - it does an awful lot at the moment, any contenders for :skull:

ianmstew:
@ahumphreys87 Good question.  I have to think more, but off the top of my head I'd like to make sure the `swap` events aren't duplicated by anything.  One thing is for sure, I would like to push "attach" events down the chain rather than calling them once for entire view tree in `show`.

paulfalgout:
In relation to Region#show 4)  https://github.com/marionettejs/backbone.marionette/issues/2299

paulfalgout:
[...]

PR #3157, opened 2016-08-31
WIP Reimagining Region

paulfalgout:
This is a breaking WIP.. And it is a long way out.. just wanted to start the conversation.

Essentially the main idea will be to reduce the region API a lot.  Much of that is possible by changes to the region mixin..

The next steps possibly will be to modify the region mixin so that regions are instantiated with an `el` at the moment `getRegion` is used instead of at view instantiation.  Then instead of resetting regions we can destroy and re-instantiate them on render.  

And further we'll possibly be able to detach the views within the regions.. we can attempt to recreate known regions after render and reattach views.

This would also allow for further abstraction of regions where we could allow for `this.showChildView(this.ui.someRegion, myView);` or even `this.showChildView('.some-selector', myView);`

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -11,8 +11,7 @@ import MarionetteObject from './object';
 import MarionetteError from './error';
 
 const ClassOptions = [
-  'allowMissingEl',
-  'parentEl',
+  'el',
   'replaceElement'
 ];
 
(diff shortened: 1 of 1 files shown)


rafde:
[Regionifying CollectionView](https://github.com/marionettejs/backbone.marionette/issues/2560) not part of this? I am seeing more implicit region related issues with using regions as a way to append to rather than remove.

It seems like people (maybe) want to do, aside from showChildView, is a setChildView('region', view')
where the contents are kept but the region is used as the view's `el`.
The issue with this is: How does empty work?

> This would also allow for further abstraction of regions where we could allow for this.showChildView(this.ui.someRegion, myView); or even this.showChildView('.some-selector', myView);
 
I think we can do this now if we allow `showChildView` to check `regions` has first, then the hash value if it's defined, otherwise assume it's a selector/DOM/jQuery

but the issue with that is: how do I get the a region's view?

paulfalgout:
> where the contents are kept but the region is used as the view's el.

I think that is just `showChildView('region', myView, { replaceElement: true });` ?

rafde:
replaceChild is a https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L184
So the region's el gets replaced by the view's el
set is more like

```js
new View({
  el: this.getRegion('region').el
});
```
Which is similar to how I (maybe?) solved for https://github.com/marionettejs/backbone.marionette/issues/3150#issuecomment-243952941

paulfalgout:
I guess I just don't know why you'd want to set the region's el as the view's el.  It seems backwards if it is the life-cycle container.  I think it's still just `replaceElement` where the region and view are init'd with the same `el`.  If you want to attach pre-existing DOM to a view I'd think you'd do it like this:

```
MyView.extend({
  ui: {
    foo: '.some-selector'
  },
  regions: {
    region: '@ui.foo'
  },
  onRender(){
    const myChild = new Mn.View({ el: this.ui.foo });
    this.showChildView('region', myChild, {replaceElement: true}); 
  }
});
```

rafde:
The way I defined it preserves the contents. example
```html
<table data-ui="table">
  <thead>
    <!-- don't remove me, bro -->
  </thead>
  <!-- collection will be appended here -->
</table>
```
```js
new CollectionView({
  collection,
  el: this.ui.table[0],
  childView: TBodyView
});
```

this allowed for the thead to be kept while being able render the models with `tbody` as the `tagName`.

paulfalgout:
At that point, what is the purpose of having the region in that at all?

rafde:
Like the Region class binding to the DOM? Nothing, now that you point that out. I always saw regions as place holders for rending *View classes. What I didn't like was all the wrapping that came with showChildView, and in some cases I use the example I gave.

paulfalgout:
I think a Region's main goal is handling the lifecycle events.  I think that once you want some DOM and a view inside a "region" you now have a CollectionView.

What I don't want to start supporting with a single region is:
```html
<table data-ui="table">
  <!-- a view get's appended here -->
  <thead>
    <!-- don't remove me, bro -->
  </thead>
  <!-- a different view get's appended here -->
  <!-- a completely different view get's appended here -->
</table>
```

or even just
```html
<table data-ui="table">
  <!-- a view get's appended here -->
  <thead>
    <!-- don't remove me, bro -->
  </thead>
</table>
```

rafde:
yeah, that looks like a massive nightmare to support. Also, fragile. All thought provoking ideas, is all.

paulfalgout:
Closing this for now.  This PR did it's job mostly.. will reinvestigate when other things start to shake out.

Issue #2916, opened 2016-03-09
About performance


JSteunou:
We would have a forum it will be more a discussion than a real issue, but I would like to put some light on https://github.com/xlfishbone/marionette.morphdom project.

Besides the use of https://github.com/patrick-steele-idem/morphdom  @xlfishbone go on several DOM involved function to rewrite them. Maybe without going morphdom or vdom or idom, Marionette can get a boost on performance by going more native or working on detached elements.

jeffijoe:
So we already have a `Renderer` that we can patch to our hearts content, and I am using that extensively to provide global helpers for Jade.

If we abstracted everything that manages the DOM into a pluggable interface, providing VDOM/ IDOM/whateverDOM implementations should be more or less straightforward?

JSteunou:
Looking at marionette.morphdom, a lot of method are already exposed as public method; all `attach` things. There are mostly 4/5 private methods left.

paulfalgout:
I'm going to close this in favor of #3145 and it's related PR.

PR #3165, opened 2016-09-05
Separate Mn DOM interaction into a mixin for DOM plugin ease.

### LINKED ISSUE #3145: Separate Mn DOM interaction into a mixin for DOM plugin ease.
# Separate Mn DOM interaction into a mixin for DOM plugin ease.

If you want to add a VDOM system or don't want to use jQuery, there isn't an easy way to override DOM interacting functions.
## Expected behavior

Provide an easy way to find functions that do DOM interaction in order to override them.

Place all DOM interacting functions into a mixin that can be used to reference for overriding for plugin needs. 
example

https://github.com/marionettejs/backbone.marionette/blob/8604cd11b7ca25bcc870184886590b076f80ac8d/src/region.js#L178-L184
https://github.com/marionettejs/backbone.marionette/blob/8604cd11b7ca25bcc870184886590b076f80ac8d/src/region.js#L200

``` javascript

...

const parent = view.el.parentNode;
if (!parent) {
  return;
}
parent.replaceChild(this.el, view.el);

...

this.el.appendChild(view.el);
```

would be replaced by a function from a mixin that looks like

``` javascript
// from mixin
replaceChild(newEl, oldEl){
  const parent = oldEl.parentNode;

  if (!parent) {
    return;
  }

  parent.replaceChild(newEl, oldEl);
},

appendChild(el, childEl) {
  el.appendChild(childEl);
}

...
// region file

// https://github.com/marionettejs/backbone.marionette/blob/8604cd11b7ca25bcc870184886590b076f80ac8d/src/region.js#L178-L184

this.replaceChild(this.el, view.el);

...
// https://github.com/marionettejs/backbone.marionette/blob/8604cd11b7ca25bcc870184886590b076f80ac8d/src/region.js#L200

this.appendChild(this.el, view.el);
```

Now you can override `appendChild` and `replaceChild` for VDOM usage.

this idea came from https://github.com/marionettejs/backbone.marionette/pull/3082#discussion_r71100068

> In the very least it'd probably be good to abstract this.$el.replaceWith to a private function such that one function could be overridden.
### Environment
1. Marionette version: 3.x
2. Backbone version: 1.3.3

rafde:
# Separate Mn DOM interaction into a mixin for DOM plugin ease.

closes #3145 

**mixins/dom.js**
- containsEl
- findEls
- createBuffer
- appendChildren
- replaceEl
- detachContents
- setInnerContent
- beforeEl
- removeEl

**composite-view.js**
- findEls

**region.js**
- findEls

**template-cache.js**
- findEls

**src/collection-view**
- appendChildren
- beforeEl
- createBuffer
- test update

**src/composite-view**
- appendChildren

**src/view.js**
- setInnerContent

**src/mixins/view**
- removeEl

**src/region**
- replaceEl
- detachContents
- appendChildren

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -357,7 +357,7 @@ const CollectionView = Backbone.View.extend({
   // Internal method. Separated so that CompositeView can append to the childViewContainer
   // if necessary
   _appendReorderedChildren(children) {
-    this.$el.append(children);
+    this.appendChildren(this.el, children);
   },
 
   // Internal method. Separated so that CompositeView can have more control over events
@@ -674,14 +674,14 @@ const CollectionView = Backbone.View.extend({
 
   // You might need to override this if you've overridden attachHtml
   attachBuffer(collectionView, buffer) {
-    collectionView.$el.append(buffer);
+    this.appendChildren(collectionView.el, buffer);
   },
 
   // Create a fragment buffer from the currently buffered children
   _createBuffer() {
-    const elBuffer = document.createDocumentFragment();
+    const elBuffer = this.createBuffer();
     _.each(this._bufferedChildren, (b) => {
-      elBuffer.appendChild(b.el);
+      this.appendChildren(elBuffer, b.el);
     });
     return elBuffer;
   },
@@ -716,7 +716,7 @@ const CollectionView = Backbone.View.extend({
     }
 
     if (currentView) {
-      currentView.$el.before(childView.el);
+      this.beforeEl(currentView.el, childView.el);
       return true;
     }
 
@@ -725,7 +725,7 @@ const CollectionView = Backbone.View.extend({
 
   // Internal method. Append a view to the end of the $el
   _insertAfter(childView) {
-    this.$el.append(childView.el);
+    this.appendChildren(this.el, childView.el);
   },
 
   // Internal method to set up the `children` object for storing all of the child views
--- a/src/common/is-node-attached.js
+++ b/src/common/is-node-attached.js
@@ -1,11 +1,9 @@
 // Marionette.isNodeAttached
 // -------------------------
 
-import Backbone from 'backbone';
-
 // Determine if `el` is a child of the document
 const isNodeAttached = function(el) {
-  return Backbone.$.contains(document.documentElement, el);
+  return document.documentElement.contains(el && el.parentNode);
 };
 
 export default isNodeAttached;
--- a/src/mixins/dom.js
+++ b/src/mixins/dom.js
@@ -1,0 +1,44 @@
+// DomMixin
+//  ---------
+
+import Backbone from 'backbone';
+
+export default {
+  createBuffer() {
+    return document.createDocumentFragment();
+  },
+
+  appendChildren(el, children) {
+    Backbone.$(el).append(children);
+  },
+
+  beforeEl(el, sibling) {
+    Backbone.$(el).before(sibling);
+  },
+
+  replaceEl(newEl, oldEl) {
+    const parent = oldEl.parentNode;
+
+    if (!parent) {
+      return;
+    }
+
+    parent.replaceChild(newEl, oldEl);
+  },
+
+  detachContents(el) {
+    Backbone.$(el).contents().detach();
+  },
+
+  setInnerContent(el, html) {
+    Backbone.$(el).html(html);
+  },
+
+  removeEl(el) {
+    Backbone.$(el).remove();
+  },
+
+  findEls(selector, context) {
+    return Backbone.$(selector, context);
+  }
+};
--- a/src/mixins/view.js
+++ b/src/mixins/view.js
@@ -11,6 +11,7 @@ import TriggersMixin from './triggers';
 import UIMixin from './ui';
 import View from '../view';
 import MarionetteError from '../error';
+import DomMixin from './dom';
 
 // MixinOptions
 // - behaviors
@@ -137,8 +138,7 @@ const ViewMixin = {
     this.unbindUIElements();
 
     // remove the view from the DOM
-    // https://github.com/jashkenas/backbone/blob/1.2.3/backbone.js#L1235
-    this._removeElement();
+    this.removeEl(this.el);
 
     if (shouldTriggerDetach) {
       this._isAttached = false;
@@ -248,6 +248,6 @@ const ViewMixin = {
   }
 };
 
-_.extend(ViewMixin, BehaviorsMixin, CommonMixin, DelegateEntityEventsMixin, TriggersMixin, UIMixin);
+_.extend(ViewMixin, DomMixin, BehaviorsMixin, CommonMixin, DelegateEntityEventsMixin, TriggersMixin, UIMixin);
 
 export default ViewMixin;
--- a/src/region.js
+++ b/src/region.js
@@ -10,6 +10,7 @@ import isNodeAttached from './common/is-node-attached';
 import { triggerMethodOn } from './common/trigger-method';
 import MarionetteObject from './object';
 import MarionetteError from './error';
+import DomMixin from './mixins/dom';
 
 const ClassOptions = [
   'allowMissingEl',
@@ -154,16 +155,15 @@ const Region = MarionetteObject.extend({
   // Override this method to change how the region finds the DOM element that it manages. Return
   // a jQuery selector object scoped to a provided parent el or the document if none exists.
   getEl(el) {
-    return Backbone.$(el, _.result(this, 'parentEl'));
+    return this.findEls(el, _.result(this, 'parentEl'));
   },
 
   _replaceEl(view) {
     // always restore the el to ensure the regions el is present before replacing
     this._restoreEl();
 
-    const parent = this.el.parentNode;
+    this.replaceEl(view.el, this.el);
 
-    parent.replaceChild(view.el, this.el);
     this._isReplaced = true;
   },
 
@@ -180,13 +180,8 @@ const Region = MarionetteObject.extend({
       return;
     }
 
-    const parent = view.el.parentNode;
+    this.replaceEl(this.el, view.el);
 
-    if (!parent) {
-      return;
-    }
-
-    parent.replaceChild(this.el, view.el);
     this._isReplaced = false;
   },
 
@@ -198,7 +193,7 @@ const Region = MarionetteObject.extend({
   // Override this method to change how the new view is appended to the `$el` that the
   // region is managing
   attachHtml(view) {
-    this.el.appendChild(view.el);
+    this.appendChildren(this.el, view.el);
   },
 
   // Destroy the current view, if there is one. If there is no current view, it does
@@ -281,7 +276,7 @@ const Region = MarionetteObject.extend({
 
   // Override this method to change how the region detaches current content
   detachHtml() {
-    this.$el.contents().detach();
+    this.detachContents(this.el);
   },
 
   // Checks whether a view is currently present within the region. Returns `true` if there is
@@ -309,5 +304,7 @@ const Region = MarionetteObject.extend({
     return MarionetteObject.prototype.destroy.apply(this, arguments);
   }
 });
+
+_.extend(Region.prototype, DomMixin);
 
 export default Region;
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -110,7 +110,7 @@ const CompositeView = CollectionView.extend({
   // You might need to override this if you've overridden attachHtml
   attachBuffer(compositeView, buffer) {
     const $container = this.getChildViewContainer(compositeView);
-    $container.append(buffer);
+    this.appendChildren($container, buffer);
   },
 
   // Internal method. Append a view to the end of the $el.
@@ -118,7 +118,7 @@ const CompositeView = CollectionView.extend({
   // childViewContainer
   _insertAfter(childView) {
     const $container = this.getChildViewContainer(this, childView);
-    $container.append(childView.el);
+    this.appendChildren($container, childView.el);
   },
 
   // Internal method. Append reordered childView'.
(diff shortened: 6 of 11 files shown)


src/utils/destroy-backbone-view.js
```diff
@@ -11,6 +12,7 @@ export default function destroyBackboneView(view) {
     triggerMethodOn(view, 'before:detach', view);
   }
 
+  dom.removeEl(view.el);
```
paulfalgout: Is this going to work?  Might be unnecessary?  Seems like if someone is using a vanilla Bb view they'll need to modify the `_removeElement` as directly after the plug `removeEl` the plain ole jquery call is going to happen still in the `view.remove`?

src/mixins/view.js
```diff
@@ -221,9 +222,15 @@ const ViewMixin = {
       }
       parent = parent._parent;
     }
+  },
+
+
+  // Overwrite https://github.com/jashkenas/backbone/blob/b8c054aec729cdc1d93e55fbe3e9f23016ed4da7/backbone.js#L1267
+  _removeElement() {
```
paulfalgout: actually this _could_ just replace line `143` now?  Also if we want to keep this function for some reason can we move it up with `destroy`?

src/region.js
```diff
@@ -9,14 +9,15 @@ import isNodeAttached from './common/is-node-attached';
...
+import domMixin from './mixins/dom';
 
 const ClassOptions = [
   'allowMissingEl',
   'parentEl',
   'replaceElement'
 ];
 
-const Region = MarionetteObject.extend({
+const Region = MarionetteObject.extend(_.extend({
```
paulfalgout: At the moment all of the other mixins are added in an extend afterwards.  Can we do the same here and if there's a reason to change it, do it for all of them?

src/collection-view.js
```diff
@@ -1,16 +1,16 @@
...
-import Backbone           from 'backbone';
-import destroyBackboneView from './utils/destroy-backbone-view';
-import isNodeAttached     from './common/is-node-attached';
-import childViewEventHandler from './common/child-view-event-handler';
-import monitorViewEvents  from './common/monitor-view-events';
-import { triggerMethodOn } from './common/trigger-method';
-import ChildViewContainer from './child-view-container';
-import MarionetteError    from './error';
-import ViewMixin          from './mixins/view';
+import _                      from 'underscore';
```
paulfalgout: soooo.. we haven't enforced this or fixed it or even documented it.. but somewhere along the way of v3 in gitter we chatted about not aligning the imports.  There is definitely some aligning still around, but I think the predominate argument against keeping up the alignment was that adding 1 import may mean a change to every import line and it just jacks with git blame.

src/region.js
rafde: I'll keep with the trend.

src/utils/destroy-backbone-view.js
rafde: I actually couldn't figure out how to handle this.
I could leave this line out and let plain Backbone.View do it's own thing if the user decides to use it as is.

I couldn't decide to either override the plain view's `_removeElement` but that should be left up to the user.

src/mixins/view.js
rafde: so don't override Backbone.View's _removeElement?

ahumphreys87:
this is gonna help me sooo much!

👍

rafde:
@ahumphreys87 I still wonder if I took it far enough.
1. Should this also handle https://github.com/jashkenas/backbone/blob/master/backbone.js#L1463
2. Am I missing other parts to DOM manipulation? 
3. Is this good for now?

paulfalgout:
I honestly think this PR should _only_ be concerned with how to change Marionette.  People may already have their native solutions for backbone stuff, but more importantly I think we can make plugins that are aware of both libraries and can update dependently.. such that if Backbone changes `_setAttributes` Mn would not have to make a major version bump to support the new Bb lib, only the plugins.

src/utils/destroy-backbone-view.js
paulfalgout: Just a note.  Doesn't particularly matter now that it's changed, but `$.remove` also unbinds events.. so for a vanilla view (or for when we override remove) `this.undelegateevents` will be necessary.

src/mixins/dom.js
```diff
@@ -0,0 +1,34 @@
...
+  detachContents(el) {
+    Backbone.$(el).contents().detach();
+  },
+
+  attachHTML(el, html) {
+    Backbone.$(el).html(html);
+  },
+
+  removeEl(el) {
+    Backbone.$(el).remove();
```
paulfalgout: while I know other implementations will query the `el` passed in, does the internal plugin need to?  it seems to be called with `this` context every time, so the Mn implementation `$el.remove` or `$el.append` etc would work.. is there perf costs to re-`Backbone.$()`ing?

src/mixins/dom.js
```diff
@@ -0,0 +1,34 @@
...
+  appendChildren(el, children) {
+    Backbone.$(el).append(children);
+  },
+
+  beforeEl(el, sibling) {
+    Backbone.$(el).before(sibling);
+  },
+
+  replaceEl(newEl, oldEl) {
+    Backbone.$(oldEl).replaceWith(newEl);
```
paulfalgout: This is breaking.  `$.replaceWith` removes event listeners..  so to handle the `preventDestroy` case 😿 we have to use the vanilla that leaves the events alone.

paulfalgout:
Re: whether this covers enough..

I don't really know.

Might want to look at `isNodeAttached`

Any plugin will have to override what `View.$` is for ui binding..  and maybe `Backbone.$`  also there's a number of places that look at `$el.length` or `$el[0]` or check if `el` is and instance of `Backbone.$` so there's other places that in some ways depend of some form of "$"

src/mixins/dom.js
rafde: the cost of [Backbone.$](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/core/init.js#L101) looks trivial.
 
I went with `Backbone.$(el).remove();` to give as much override opportunity as possible, in case VDOM diffing might treat removing differently.

paulfalgout: Ah it does look pretty trivial.

And I totally get the overriding thing.. but this internal mixin could be just ignoring the argument
```
  removeEl(el) {
    this.$el.remove();
  }
```

src/mixins/dom.js
rafde: NOOOOOOooooOOOOOOooOOOO!!!!!

ah right. will update.
Gonna have to come up for a test for that some day.

src/mixins/dom.js
rafde: I get that but I am looking at it in terms of removing other elements other than `view.$el`

I can make a `removeEl` for `this.$el.remove();` and another one for other DOM like `removeDOM(dom)` if you like.

src/mixins/dom.js
paulfalgout: I believe we can change this back in v4 after we remove `preventDestroy` as people can just `detachView` before the empty I think..

src/mixins/dom.js
paulfalgout: Like people would call functions directly off the mixin somehow?  Or they'd do like `myView.removeEl(someOtherEl)`? What's the use-case for `removeDOM`?

rafde:
@paulfalgout I was thinking this mixin would be for DOM manipulation. But if this mixin also has to cover places that jQuery touches, then this will probably merge with getUi and others.
@ahumphreys87 any opinion on whether this mixin should cover the DOM touching?
all @marionettejs/marionette-core welcomed to chime in.

src/mixins/dom.js
rafde: > Or they'd do like myView.removeEl(someOtherEl)?

The quoted case is what I am covering for.

paulfalgout:
ah good point about this being scoped to DOM manipulation.. so for some forms of DOM manipulation the ui and other jquery things will likely break, right?  Like if you use some form of virtual dom you might have to avoid using `ui`?  At some point if we do have some sort of plugin ecosystem we will probably want to provide some sort of way for the authors to throw appropriate errors or whatever..  but for now this works for me 👍

rafde:
> Like if you use some form of virtual dom you might have to avoid using ui?

In the cases of `ui` they would need to use[view._bindUIElements](https://github.com/marionettejs/backbone.marionette/blob/master/src/mixins/ui.js#L67)
or re-bind the ui they are affecting.

Anyhow, this has 2 thumbs already but I want to see what @ahumphreys87 has to say before this is merged.

ahumphreys87:
hey @rafde I think this covers most stuff. Let me get the Native mixin im working on. I have an app running so have overridden most areas.

https://github.com/BedeGaming/orchestra/blob/no-jquery/src/mvc/native/index.js

So I included https://github.com/akre54/Backbone.NativeView and https://github.com/akre54/Backbone.Fetch

ahumphreys87:
The one tricky thing is behavior events, they will currently be overridden if you use the same selector for an event in the view.

paulfalgout:
Related https://github.com/marionettejs/backbone.marionette/issues/2916

rafde:
Made another update that adds 2 more functions for `$.contains` and `Backbone.$().find`

src/common/is-node-attached.js
```diff
@@ -1,11 +1,11 @@
 // Marionette.isNodeAttached
 // -------------------------
 
-import Backbone from 'backbone';
+import domMixin from '../mixins/dom';
 
 // Determine if `el` is a child of the document
 const isNodeAttached = function(el) {
-  return Backbone.$.contains(document.documentElement, el);
+  return domMixin.containsEl(document.documentElement, el);
```
paulfalgout: I'm not sure how this works.  how do you override the `domMixin` later?

blikblum: Another point is that this $ call can be replaced by

 `return document.documentElement.contains(el)` 

So no need to abstract this function at all

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains), is pretty safe to use it

blikblum: Assuming Node.contains is implemented (even ie5 has it according to MDN) 

Passing document.documentElement and el to $.contains,  following code path will be executed: 

`adown.contains( bup )`

where adown is document.documentElement and bup is el.parentNode

So the only difference would be, that using $.contains,  el.parentNode will be passed to Node.contains instead of el. Probably because of [this IE bug](https://connect.microsoft.com/IE/feedback/details/780874/node-contains-is-incorrect)

In fact using bellow code would execute the same code path as using jquery

`return document.documentElement.contains(el && el.parentNode)`

blikblum: Another reason to use el.parentNode is that Node.contains returns true when comparing a element to itself

paulfalgout: Thanks for looking into it.  I can't stand the `adown`s and `bup`s.

I'm cool with:
`return document.documentElement.contains(el && el.parentNode)`

@rafde ?

rafde: If the test pass, no problem.

rafde: done

src/region.js
```diff
@@ -197,8 +188,13 @@ const Region = MarionetteObject.extend({
 
   // Override this method to change how the new view is appended to the `$el` that the
   // region is managing
-  attachHtml(view) {
-    this.el.appendChild(view.el);
+  attachHtml(view, shouldReplace) {
+    if (shouldReplace) {
```
rafde: `domixin.attachHTML` is a name I am not attached to. If you have a better name, I am open to suggestions.

paulfalgout: This is actually different now.  `shouldReplace` is not apart of `attachHtml`

src/mixins/dom.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    parent.replaceChild(newEl, oldEl);
+  },
+
+  detachContents(el) {
+    Backbone.$(el).contents().detach();
+  },
+
+  attachHTML(el, html) {
```
paulfalgout: re: @blikblum comment below this could be `attachContents` to align with `detachContents`.. it's not quite the same, but they are the opposite functions.

src/region.js
rafde: change to

`
attachHtml(view, shouldReplace) {
   this.appendChildren(this.el, view.el);
}
```
?

File: src/region.js [hunk 10]
```diff
@@ -103,11 +104,7 @@ const Region = MarionetteObject.extend({
       triggerMethodOn(view, 'before:attach', view);
     }
 
-    if (shouldReplaceEl) {
-      this._replaceEl(view);
-    } else {
-      this.attachHtml(view);
-    }
+    this.attachHtml(view, shouldReplaceEl);
```
paulfalgout: Ah.. this is putting it back.. the left side is correct

src/mixins/dom.js
rafde: could name it `setInnerContent` since any html pass will completely change the element's inner.

paulfalgout:
I believe this resolves https://github.com/marionettejs/backbone.marionette/issues/2846

rafde:
@paulfalgout all changes were reverted.

src/mixins/view.js
```diff
@@ -248,6 +248,6 @@ const ViewMixin = {
   }
 };
 
-_.extend(ViewMixin, BehaviorsMixin, CommonMixin, DelegateEntityEventsMixin, TriggersMixin, UIMixin);
+_.extend(ViewMixin, domMixin, BehaviorsMixin, CommonMixin, DelegateEntityEventsMixin, TriggersMixin, UIMixin);
```
paulfalgout: The last nit-pick I just noticed.  All of the other Mixins have the first letter uppercase.  Can we do that here too and if there needs to be a stylistic change, do it afterwards?

rafde:
all done @paulfalgout

PR #3173, opened 2016-09-12
Listen to Collection#event:update and use options.changes.added for CollectionView

### LINKED ISSUE #2979: Listen to Collection#event:update instead of Collection#event:add/remove in CollectionView
https://github.com/marionettejs/backbone.marionette/blob/0d913a524a71a9b4427256b0ba1b3236abeee9d6/src/collection-view.js#L90-L91

Collection#event:add and Collection#event:remove works on adding Views one at a time when they should be added using something like https://github.com/marionettejs/backbone.marionette/blob/0d913a524a71a9b4427256b0ba1b3236abeee9d6/src/collection-view.js#L252-L254 for bulk adding [options.added](https://github.com/jashkenas/backbone/blob/master/backbone.js#L930) and probably a bulk remove using [options.removed](https://github.com/jashkenas/backbone/blob/master/backbone.js#L931)

Listening to [Collection#event:update](https://github.com/jashkenas/backbone/blob/master/backbone.js#L934) will allow you to start doing bulk add/remove views instead of doing it 1 +/- at a time.

### LINKED ISSUE #3043: CollectionView elements added at incorrect index.

rafde:
# Listen to Collection#event:update and use options.changes.added for CollectionView

This is as close to non-breaking as I can get this enhancement to be.
Please look over, help, review since this PR needs some extreme scrutiny to get it in good shape.

## _onCollectionUpdate
* Add models to collection view

## _applyModelDeltas
* simplified filter
* using _onCollectionUpdate

## _filteredSortedModels
* if `at` and there are models being added, add model correctly

## _updateIndices
* wrapped statement with condition for easier testing for code coverage

## _attachView
* `_isUpdating` check to not trigger attach event

## addChildView
* `_isUpdating` check to add but not update ChildViewContainer

## attachHtml
* `_isUpdating` check to skip adding to buffer or dom

## getChildViewContainer
* allow testing and call for CompositeView

## [test] collection-view
* ~~when firing an `add` event on the collection with `at` but without `index`, BB < 1.2 style~~ not supporting < 1.3.3

## [test] composite view child view
*`_addChildModels` usage

## [test] sorted-views test
* make sure _updateIndices does not update `view._index`
* adding models in correct order. fixes #3043

## [test] composite-view
 ~~shound send childView to getChildViewContainer~~ not sending single childView anymore

closes #2979


src/collection-view.js
```diff
@@ -724,6 +779,14 @@ const CollectionView = Backbone.View.extend({
     return false;
   },
 
+  getChildViewContainer() {
+    return this.el;
+  },
+
+  _appendEl(el, children) {
+    $(el).append(children);
```
paulfalgout: should be Backbone.$

src/collection-view.js
```diff
@@ -110,26 +117,19 @@ const CollectionView = Backbone.View.extend({
...
-    if (this._shouldAddChild(child, index)) {
-      this._destroyEmptyView();
-      this._addChild(child, index)
-    }
-  },
-
   // Handle collection update model removals
   _onCollectionUpdate(collection, options) {
     const changes = options.changes;
+    let added = changes.added;
```
scott-w: Minor nitpick but should this be `const`?

src/collection-view.js
```diff
@@ -158,6 +158,62 @@ const CollectionView = Backbone.View.extend({
...
+        view = this._addChild(model, index);
+        adding.push(view);
+        triggerOnChildren.push(view);
+        buffer.appendChild(view.el);
+      } else if (buffer.children.length) {
+        if (this.sort) {
+          view._index = index;
+        }
+
+        _.each(triggerOnChildren, child => {
```
scott-w: Is it possible to end up in a situation where `this.children.findByModel(model)` returns a view but the `_.reduce` isn't finished yet? My reading of this is that `before:attach` and `attach` could get triggered for the same view multiple times.

rafde: only if async (which can be said about most function calls)
or if an event call updates the collection in the middle of this function call.
I, honestly, don't like how most of this looks. It's not very DRY because of the `view.$el.before(buffer);` making it difficult to DRY. I am still kicking around how to make it nicer.

src/collection-view.js
```diff
@@ -586,18 +638,16 @@ const CollectionView = Backbone.View.extend({
   // Internal method. This decrements or increments the indices of views after the added/removed
   // view to keep in sync with the collection.
   _updateIndices(views, increment) {
-    if (!this.sort) {
-      return;
-    }
-
-    const view = _.isArray(views) ? this._findGreatestIndexedView(views) : views;
+    if (this.sort) {
```
rafde: had to do this wrap to make it easier to test for code coverage.

src/collection-view.js
rafde: > Is it possible to end up in a situation where this.children.findByModel(model) returns a view but the _.reduce isn't finished yet?

I could safe guard using something like `this._processingView = {};` and set/check against that until this is done, then reset to an empty object when done.

test/unit/composite-view.spec.js
```diff
@@ -457,10 +457,8 @@ describe('composite view', function() {
     beforeEach(function() {
       this.templateFn = _.template('composite <%= foo %>');
 
-      this.childViewTagName = 'span';
-
       this.ChildView = Backbone.Marionette.View.extend({
-        tagName: this.childViewTagName,
+        tagName: 'span',
```
rafde: Note to self: I have to remember why I did this.

rafde:
so.... every time I try a new way to do this PR I am always 1 breaking test away from making it work.

This version broke [`expect(this.compositeView.getChildViewContainer.getCall(0).args[1].tagName).to.equal(this.childViewTagName);`](https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/composite-view.spec.js#L527)
because [`const $container = this.getChildViewContainer(this, childView);`]( https://github.com/marionettejs/backbone.marionette/blob/master/src/composite-view.js#L120) 
doesn't happen anymore. 
Not using [getChildViewContainer(containerView, childView)](https://github.com/marionettejs/backbone.marionette/blob/master/src/composite-view.js#L134) breaks the public API expectation that a view will get sent, but that will never happen in this version of adding new views.

A different version had the code go through the normal route minus using `this.children._updateLength()` until it was completely done processing the models to view it order to not affect performance. 

the test for [expect(this.order).to.deep.equal(['1', '2', '3', '4', '5', '6']);](https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/collection-view.spec.js#L488-L490) broke at that point because [`const findPosition = this.sort && (index < this.children.length - 1);`](
https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L613-L629) relies on the childview container's length to be up to date (the cause of performance problems).

TL; DR:
1. CompositeView getChildViewContainer is hindering moving forward
2. ChildViewContainer's length is used to determine view attach position when it can't be until all views are processed.

rafde:
closing in favor of https://github.com/marionettejs/backbone.marionette/pull/3244

PR #3175, opened 2016-09-13
Clarify Underscore templates are used by default

webxl:
The redundant "will look up" was replaced with specifics about what render actually does with Underscore.

### DIFF
--- a/docs/marionette.renderer.md
+++ b/docs/marionette.renderer.md
@@ -14,8 +14,8 @@ without data.
 ## Basic Usage
 
 The basic usage of the `Renderer` is to call the `render` method. This method
-returns a string containing the result of applying the template using the `data`
-object as the context.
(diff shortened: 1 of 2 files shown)


scott-w:
This looks pretty good. Is it worth adding a little bit at the end that links through to other rendering methods (the `marionette.renderers.md` file if I recall correctly)?

webxl:
Not sure. It's already mentioned a bunch in marionette.templatecache.md.

The only place I think it would fit in Renderers is in the Basic usage section. I just tacked that on.

In  my original commit, I say "template element" but that might get confused with `<template>`. Not sure what to call the overloaded script element though.

scott-w:
Looks good to me 👍

JSteunou:
Should this be base on the `next` branch ?

Issue #3162, opened 2016-09-03
No "before:render" or "render" events when showing view in region


DenMesh:
### Description

Consider the following situation. There is a region attached to an existing DOM tree so that the region's "el" has some child nodes. Then a view is attached to the region, with "el" property set to an immediate child of the region's element and "template" property set to false. According to the docs (View => Managing an Existing Page) it should fire "render" and "before:render" events if shown in the region, even though no actual rendering happens. This is not so. According to the code, if there is an "el" property, the "_isRendered" property is set to true (view mixin, line 58) and if the "_isRendered" property is true, then the view's "render" method is never called (region, line 81).

### Environment

1. Marionette version: 3.0.0
2. Backbone version: 1.3.3
3. Additional build tools, etc: Webpack

scott-w:
Hi @DenMesh, thanks for reporting.

I understand correctly:

```javascript
var MyView = Mn.View.extend({
  el: '#my-selector',
  template: false,

  onRender: function() {
    console.log('this will never fire');
  }
});

var app = new Mn.Application({...});
app.getRegion().show(new MyView());
```

The `render` event shouldn't fire in this case because no rendering happens. Is that correct?

DenMesh:
Exactly this, yes.

scott-w:
Thanks, I'll update this ticket as an issue with the docs.

paulfalgout:
I don't think this is a bug.  In fact I find with v3 there is extremely little use-case for `template:false` If you never intend to render this view because the DOM exists and the template is false, use `initialize`  v3 was modified so that the constant unwanted re-rendering was removed.

DenMesh:
Never said it was a bug. It is an inconsistency between the docs and the code. Namely, 

> Though the View has no template,
you can still listen to the `before:render` and `render` events that will fire
as usual _when render is called_ - or when you execute `region.show(view)`.

(emphasis mine)

scott-w:
Ahh that was my bad, I'll take the bug label off

PR #3181, opened 2016-09-14
ensure onBeforeRender and onRender are fired when showing views with …

broood:
### Proposed changes
 - Remove `view._isRendered` check in Region::_renderView (originally added in this commit: https://github.com/marionettejs/backbone.marionette/commit/e92273ec29ccbb8b4d6a07220b8ff2548b56b385)

From the issue: 

> Consider the following situation. There is a region attached to an existing DOM tree so that the region's "el" has some child nodes. Then a view is attached to the region, with "el" property set to an immediate child of the region's element and "template" property set to false. According to the docs (View => Managing an Existing Page) it should fire "render" and "before:render" events if shown in the region, even though no actual rendering happens. This is not so.

Link to the issue: https://github.com/marionettejs/backbone.marionette/issues/3162

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -78,10 +78,6 @@ const Region = MarionetteObject.extend({
   },
 
   _renderView(view) {
-    if (view._isRendered) {
-      return;
-    }
-
     if (!view.supportsRenderLifecycle) {
       triggerMethodOn(view, 'before:render', view);
     }
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -1159,4 +1159,33 @@ describe('region', function() {
       });
     });
   });
+
+  describe('when showing a view with `template: false`', function() {
+    beforeEach(function() {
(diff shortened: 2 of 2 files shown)


paulfalgout:
Thanks for this.  I missed the original issue, but I'm personally :-1: on this change.  I think the docs are incorrect now.  If the view is already rendered because it is given and `el` and it is template: false so it is never intended to be rendered, just do the `onRender` type work in `initialize`

broood:
Thanks @paulfalgout .. I think that makes sense.  Looking back at the issue again, perhaps @danmesh and @scott-w were just pointing out that docs state (incorrectly) that the `onRender` would fire in such cases -- rather than it should be fired as I initially interpreted.  Feel free to close.

rafde:
closing per request.

rafde:
@broood thank you for the intent.

Issue #3168, opened 2016-09-07
Deprecate preventDestroy option on Region


paulfalgout:
With `Region#detachView` (https://github.com/marionettejs/backbone.marionette/pull/3155) coming I don't think `preventDestroy` has added value.  Plus it's removal will reduce the code's complexity.

PR #3183, opened 2016-09-16
Deprecate preventDestroy in favor of Region#detachView

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3168


JSteunou:
:+1:

Issue #2289, opened 2015-02-05
Remove selector key for region definitions


jfairbank:
We support `el` and `selector`. We should just use `el`. Recommend PR after `Region` mixin is added.

``` js
var MyView = Marionette.ItemView.extend({
  regions: {
    foo: {
      selector: '#foo',
      regionClass: MyRegionClass
    }
  }
};

// should be

var MyView = Marionette.ItemView.extend({
  regions: {
    foo: {
      el: '#foo',
      regionClass: MyRegionClass
    }
  }
};
```

thejameskyle:
:+1:

jmeas:
:+1:

samccone:
:+1:

instanceofpro:
:+1:

jmeas:
:+1:

jmeas:
:+1:

jfairbank:
:+1:

samccone:
![](http://i.imgur.com/8DgP4.jpg)

thejameskyle:
![](http://media0.giphy.com/media/wTzPSxZBqSove/giphy.gif)

samccone:
![](http://cdn0.dailydot.com/uploaded/images/original/2013/3/7/uj7beoj.gif)

buob:
![article-1283511293727-0b0681d5000005dc-493565_466x380](https://cloud.githubusercontent.com/assets/986404/6066582/22eafa0a-ad3c-11e4-9b3b-beabef74133f.jpg)

scott-w:
Chasing turtles here - did this ever get resolved?

paulfalgout:
@scott-w selector is still in the code, but we can break it for v4

scott-w:
:+1:

paulfalgout:
https://github.com/marionettejs/backbone.marionette/pull/3465

PR #3184, opened 2016-09-16
Deprecate Region definition selector

paulfalgout:
Gets ready for https://github.com/marionettejs/backbone.marionette/issues/2289


JSteunou:
:+1:

Issue #3187, opened 2016-09-17
Region#detachView should work similarly to Region#empty


rafde:
### Expected behavior

When using Region#detachView, it should

```
detachedView.off('destroy', this.empty, this);
this.triggerMethod('before:empty', this, detachedView);
this._restoreEl();
delete this.currentView;
delete detachedView._parent;
this.triggerMethod('empty', this, detachedView);
```
Somewhere in Region#detachView

### Environment

1. Marionette version: 3.x

PR #3190, opened 2016-09-17
Empty region on view detach

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3187

It's funny this is turning into `preventDestroy` with an enforced view being returned.  So this practically does `preventDestroy` under the hood but still allows us to remove that option in the future with very little behavior duplication (ie: if `empty` changes we don't have to remember to apply the change to `detachView` as well, etc)

FWIW there is some awkward side-effects in v3 currently if your view had an `allowMissingEl` or `preventDestroy` defined on them as `empty`'s `options` argument would be the view in the case of `view.on('destroy', this.empty`.  This PR'll fix that issue as well.

Still need to add some tests and probably improve the docs a little.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -63,7 +63,7 @@ const Region = MarionetteObject.extend({
     // We need to listen for if a view is destroyed in a way other than through the region.
     // If this happens we need to remove the reference to the currentView since once a view
     // has been destroyed we can not reuse it.
-    view.on('destroy', this.empty, this);
+    view.on('destroy', this._empty, this);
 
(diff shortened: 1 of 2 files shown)


rafde:
probably need some test. i'll include them in a while.

rafde:
💥 PR collaboration. Test added.

JSteunou:
Good question, I will wait another approval before merging.

paulfalgout:
I will approve the tests..  That way we have 2 :+1:s on everything

Issue #2353, opened 2015-02-21
Add cross browser testing.


samccone:
@megawac I know you are doing this for Backbone and Underscore, mind giving it a go on this repo?

megawac:
Where's QUnit?!!

samccone:
Nooooooooo

samccone:
but really, ping @megawac any interest in dropping some knowledge on this front for us? I know you setup the basic stuff on backbone, would love to have some help on this if you can find the time :)

thanks

ahumphreys87:
Im guessing this will involve bringing karma back in? It provides a simple saucelabs plugin.

megawac:
Grunt-saucelabs may work as well. I spent about an hr trying to get it
working but ran into some issues with the build steps (@sam and I were
supposed to figure this out a couple weeks ago. May look at it again
sometime this week, my schedule has calmed a bit

PR #3198, opened 2016-09-24
Add cross browser testing

denar90:
### Proposed changes
 - Added cross browser testing using https://saucelabs.com
 
Link to the issue:
https://github.com/marionettejs/backbone.marionette/issues/2353

Sometimes test are not passed because of `ngrok` failing with error - to many connections.
Another questions about platforms. Which one do we need. Currently added:
```
    ["Windows 7", "internet explorer", "9.0"],
    ["Windows 8", "internet explorer", "10.0"],
    ["Windows 8.1", "internet explorer", "11.0"],
    ["Windows 10", "MicrosoftEdge", "latest"],
    ["Windows 10", "chrome", "latest"],
    ["Windows 10", "firefox", "latest"],
    ["OS X 10.11", "chrome", "latest"],
    ["OS X 10.11", "firefox", "latest"],
    ["OS X 10.11", "safari", "9.0"]
```


package.json
```diff
@@ -54,6 +55,7 @@
     "backbone": "1.2.1 - 1.3.x",
     "chai": "3.4.0",
     "chai-jq": "0.0.9",
+    "easy-sauce": "^0.3.5",
```
rafde: would it be ideal to lock down the version in case of something randomly breaking

test/setup/browser.js
```diff
@@ -4,6 +4,6 @@ global.mocha.setup('bdd');
 global.onload = function() {
   global.mocha.checkLeaks();
   global.mocha.globals(Object.keys(mochaGlobals));
-  global.mocha.run();
-  require('./setup')();
+  var runner = global.mocha.run();
+  require('./setup')(runner);
```
rafde: strange, the CI doesn't seem to get `runner` passed to it.

denar90: Fixed this one.

package.json
denar90: Sound reasonable.

denar90:
Found out why problems with too many connections appear. `easy-sauce` lib ping status of each test to know if it's ok or not. If we have to many platforms for testing then ping requests increase (in our case to ~70). `ngrok` allow only 20 per 1 minute. Try to solve this issue by changing `easy-sauce` API - https://github.com/philipwalton/easy-sauce/issues/7
Another thing is that `easy-sauce` supports node >v6 ... Probably we should say no for using it, because we run our CI on different node env, lower then v6...

paulfalgout:
@denar90 https://github.com/marionettejs/backbone.marionette/issues/3199 open an issue about Node with the issues around upgrading there.

rafde:
So no version lock down of "easy-sauce": "^0.3.5"?

denar90:
@rafde I think we should wait until new version with fixes for stuff I wrote above will be added. Then of course lock version.

denar90:
First, I removed livereload for test `runner.html` page (it was not used anyway). 
Second, badge from sauce labs was added, so readme will have a bit updated look https://monosnap.com/file/hDOvt131ydbmwDeaBGfhqul1qtDnKT
Third, Travis`Environment Variables` with saucelabs credentials were added for running cross browser testing.

Moreover, I wanna say thanks to @philipwalton for reviewing/improving/releasing [easy-sauce](https://github.com/philipwalton/easy-sauce).

gulp/test-browser.js
```diff
@@ -10,6 +10,23 @@ import commonjs from 'rollup-plugin-commonjs';
 import json from 'rollup-plugin-json';
 import opn from 'opn';
 
+const sauceConf = {
+  name: 'Marionette.js',
+  username: process.env.SAUCE_USERNAME,
+  key: process.env.SAUCE_ACCESS_KEY,
+  tests: '/test/runner.html',
```
denar90: I tested with all 3 services, everything was fine.

philipwalton: Yeah, looks like I missed a few instances in my find-and-replace. I've fixed it and released easy-sauce 0.4.1 with the changes.

paulfalgout:
> I removed livereload for test runner.html page (it was not used anyway). 

Pretty sure it is used.  It you run `gulp test-browser` it'll spawn a browser window.. then if you make any chances to the code or the tests, the browser will run the tests again.  Is there no way to keep this for development?

denar90:
@paulfalgout I got back livereload.

gulp/test-browser.js
```diff
@@ -43,47 +44,47 @@ function bundle() {
         exclude: 'node_modules/**'
       })
     ]
-  }).then(function (bundle) {
+  }).then(bundle => {
```
scott-w: Strictly speaking you don't need the excess `{` and `return` here:

```javascript
.then(bundle => bundle.write({
...
})).then(...);
```

denar90: Thanks 👍

Issue #3172, opened 2016-09-11
Document View element removal


scott-w:
Reported by @bazineta

~~In Mn 3, a DOM element is removed from the DOM tree before any events fire on its children. For example:~~

``` javascript
import {View} from 'backbone.marionette';

const SecondChild = View.extend({
  onBeforeDestroy() {
    console.log('This is already out the DOM');
  }
});

const FirstChild = View.extend({
  regions: {
    child: '.child-hook'
  },

  onRender() {
    this.showChildView('child', new SecondChild());
  }
});

const Root = View.extend({
  regions: {
    child: '.child-hook'
  },

  onRender() {
    this.showChildView('child', new FirstChild());
  }
});

app.getRegion().show(new Root()); // Renders everything

app.getRegion().empty();
```

In this example, by the time, the log hits, `SecondChild` is no longer visible in the DOM tree.

bazineta:
For our application, this isn't a documentation issue; rather, it's a showstopper.

Consider the case where SecondChild is using a plugin of some type, that provides a 'destroy' function. Examples; Bootstrap's Carousel widget, which must be paused before being removed from the DOM, Highcharts, which will leak without being destroyed, innumerable jQuery and Bootstrap plugins.

In 2.x, one could use 'onBeforeDestroy' for this purpose -- the plugin could then be torn down as the view was going away.

In 3.x, this isn't possible; the chances of a view not being in a hierarchy of views is practically zero, and thus by the time any of the view destruction lifecycle events are called on it, it's no longer in the DOM tree, and 'destroy' functions are going to likely be at best entertaining; for example, the Bootstrap Carousel widget actually re-instantiates itself, which is a bit ironic considering that we were trying to prevent leaks, not make more of them.

As it stands, the destruction lifecycle events seem to be of questionable value, if by the time the lifecycle event fires, the view is in a zombified state; by the time any lifecycle event fires, a leak has already occurred and nothing can be done about it. Other than completely reinventing destruction lifecycle via some type of pre-destruction notification all the way down an arbitrary nesting hierarchy, there seems no way in 3.x to prevent leaks.

bazineta:
As an example of the view destruction lifecycle events being odd in this situation, consider that for SecondChild, any onBeforeDetach event fires, from its perspective, after it's been detached. That will actually have occurred when Root was destroyed; both children however will receive a 'before:detach' event after the fact.

paulfalgout:
[...]

PR #3200, opened 2016-09-26
Docs fixes

paulfalgout:
So far I've done three things:
- Fixed doc menus
- Fixed some documentation errors mainly in general functions and added a couple of missing API.
- Added a doc linking common class functions to the general documentation

I'm going to keep going on this PR, but at any point in the commits, if this is ready to merge, I can pull out the later commits or whatever to get it in and continue on another PR.

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -310,15 +310,15 @@ var Mn = require('backbone.marionette');
 var myCollection = new Bb.Collection([{}]);
 
 var CollectionView = Mn.CollectionView.extend({
-childViewEventPrefix: 'some:prefix'
+  childViewEventPrefix: 'some:prefix'
 });
 
 var collectionView = new CollectionView({
-collection: myCollection
+  collection: myCollection
 });
 
 collectionView.on('some:prefix:render', function(){
-// child view was rendered
+  // child view was rendered
 });
 
 collectionView.render();
@@ -412,20 +412,6 @@ var MyCollectionView = Mn.CollectionView.extend({
     render: function() {
       console.log('A child view has been rendered.');
     }
-  }
-});
-
-// ...or as a function that returns a hash.
-var MyCollectionView = Mn.CollectionView.extend({
-
-  childViewEvents: function() {
-    return {
-      render: this.onChildRendered
-    }
-  },
-
-  onChildRendered: function () {
-    console.log('A child view has been rendered.');
   }
 });
 ```
--- a/docs/events.md
+++ b/docs/events.md
@@ -20,9 +20,15 @@ responding to events.
   * [View events and triggers](#view-events-and-triggers)
 * [Child View Events](#child-view-events)
   * [Event Bubbling](#event-bubbling)
+    * [Using CollectionView](#using-collectionview)
+  * [A Child View's Event Prefix](#a-child-views-event-prefix)
   * [Explicit Event Listeners](#explicit-event-listeners)
+    * [Attaching Functions](#attaching-functions)
+    * [Using `CollectionView`'s `childViewEvents`](#using-collectionviews-childviewevents)
   * [Triggering Events on Child Events](#triggering-events-on-child-events)
+    * [Using `CollectionView`'s `childViewTriggers`](#using-collectionviews-childviewtriggers)
 * [Lifecycle Events](#lifecycle-events)
+
 
 ## Triggering and Listening to Events
 
@@ -154,10 +160,10 @@ As before, all arguments passed into `triggerMethod` will make their way into
 the event handler. Using this method ensures there will be no unexpected
 memory leaks.
 
-### View events and triggers
+### View `events` and `triggers`
 
-Views can automatically bind DOM events to methods and View events with `events`
-and `triggers` respectively:
+Views can automatically bind DOM events to methods and View events with [`events`](./marionette.view.md#view-events)
+and [`triggers`](./marionette.view.md#view-triggers) respectively:
 
 ```javascript
 var Mn = require('backbone.marionette');
@@ -226,7 +232,7 @@ methods bound to the `childViewEvents` attribute. This works for built-in
 events, custom events fired with `triggerMethod` and bound events using
 `triggers`.
 
-When using implicit listeners, the `childview:*` event prefix is used which
+When using implicit listeners, the [`childview:*` event prefix](#childvieweventprefix) is used which
 needs to be included as part of the handler:
 
 ```javascript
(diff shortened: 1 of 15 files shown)


paulfalgout:
I should also note that I've been updating the live examples as I've gone.. soo... if there's something wrong with an update we'll have to revert..  but I'm not currently sure of a better way to do this.

docs/marionette.object.md
```diff
@@ -78,28 +74,14 @@ radioRequests: {
...
+In this example the object will listen for the `some:request` request on the `myChannel` channel, and run the 'requestHandler' method.  When using Radio Requests with Objects, the same rules and restrictions that normal Radio use implies also apply here: a single handler can be associated with a request, either through manual use of the reply functions, or through the Object API.
 
 ## Destroying A Object
 
 Objects have a `destroy` method that unbind the events that are directly attached to the
 instance.
 
-Invoking the `destroy` method will trigger a "before:destroy" event and corresponding
+Invoking the `destroy` method will trigger a `before:destroy` event and corresponding
 `onBeforeDestroy` method call. These calls will be passed any arguments `destroy`
```
scott-w: I'd be tempted to chop this bit out and just link through to the trigger binding docs.

paulfalgout: yeah.. I think the oddity here worth documenting is anything you pass as arguments to destroy get passed to the destroy events which is unique to `destroy`.

scott-w: Fair enough - we should still maintain the link and perhaps add a `**Note**` here?

paulfalgout: cool on the link.. I'll get that added what'd be in the `**Note**` ?

scott-w: I meant using the `**Note**` to tell readers that `destroy` will pass its arguments into its events and this is different to other event handling stuff.

paulfalgout:
Just a progress note.. I'm still looking to review Region, ChildViewContainer, and CollectionView docs, but I feel ok about the rest.  I'm not really taking into account too many open issues at the moment.. just going through the docs and fixing issues that stand out to me.

docs/marionette.functions.md
```diff
@@ -355,28 +413,31 @@ var View = Mn.View.extend({
...
   }
 
 });
+
+var myView = new MyView();
+myView.do('action:one');
+myView.do('action:two');
 ```
 
 [Live example](https://jsfiddle.net/marionettejs/zzjhm4p1/)
```
paulfalgout: I just hadn't saved the fiddle on accident.. I _think_ I've updated most of them already.

paulfalgout:
Ok mainly down to regions now..

paulfalgout:
I believe https://github.com/marionettejs/backbone.marionette/issues/3172 is resolved here

docs/events.md
```diff
@@ -360,12 +399,46 @@ var ParentView = Mn.View.extend({
...
+  }
+});
+
+// ...or as a function that returns a hash.
+var MyCollectionView = Mn.CollectionView.extend({
+
+  childViewEvents: function() {
+    return {
+      render: this.onChildRendered
+    }
```
scott-w: Is this necessary or can we just link through to the relevant documentation?

docs/events.md
```diff
@@ -428,6 +501,54 @@ var GrantParentView = Marionette.View.extend({
...
+
+```javascript
+var Mn = require('backbone.marionette');
+
+// The child view fires a custom event, `show:message`
+var ChildView = Mn.View.extend({
+
+  // Events hash defines local event handlers that in turn may call `triggerMethod`.
+  events: {
+    'click .button': 'onClickButton'
```
scott-w: It might be better to use an explicit name such as `showMessages` here?

docs/events.md
```diff
@@ -428,6 +501,54 @@ var GrantParentView = Marionette.View.extend({
...
+});
+
+// The parent uses childViewEvents to catch the child view's custom event
+var ParentView = Mn.CollectionView.extend({
+
+  childView: ChildView,
+
+  childViewTriggers: {
+    'show:message': 'child:show:message',
+    'submit:form': 'child:submit:form'
```
scott-w: This could be a bit confusing as it's close enough to the default `childview:show:message` events. I find a totally different convention is usually more helpful for new developers to separate the concepts when building applications.

docs/marionette.behavior.md
```diff
@@ -347,11 +378,11 @@ var ViewBehavior = Mn.Behavior.extend({
 
 [Live example](https://jsfiddle.net/marionettejs/8dmk30Lq/)
 
-#### Referencing the DOM
+### Referencing the DOM
 
 Behaviors, like views, have a `ui` attribute that can reference and cache DOM
 elements, just as in the `View`. For more detail, see the
-[`ui` documentation for views](./marionette.view.md#organising-your-view).
+[`ui` documentation for views](./marionette.view.md#organizing-your-view).
```
scott-w: Good spot, my Britishness must have slipped in :P

docs/marionette.functions.md
```diff
@@ -157,28 +176,44 @@ example has an undefined value for `f`.
...
+
+`triggerMethod('foo', bar)` will call `onFoo: function(bar){...})`
+
+
+```javascript
+var Mn = require('backbone.marionette');
+
+var MyObject = Mn.Object.extend({
+  initialize: function(){
+    this.triggerMethod('baz');
```
scott-w: Is this supposed to be `this.triggerMethod('foo', 'baz');`

docs/marionette.view.md
```diff
@@ -73,32 +73,43 @@ For more detail on how to render templates, see the
...
-### Setting a falsy `template` value
 
-When using an existing page, Marionette explicitly looks for `false` - any other
-falsy value will cause Marionette to raise an error when attempting to render
-the template.
+Marionette will [set the appropriate state of the view](./viewlifecycle.md#views-associated-with-previously-rendered-or-attached-dom).
+
+### Setting a `template` to `false`
+
+Setting the `template` to `false` allows for the view to create all of the bindings and trigger all view events without re-rendering the el of the view. Any other falsy value will throw an exception.
```
scott-w: I'd be tempted to wrap the last sentence in italics: `*Any other falsy value will throw an exception.*`

docs/events.md
paulfalgout: I just moved this from elsewhere.. although I typically follow naming my function names after the event in this way.  Then inside those functions it will procedurally call whatever needs to happen even if it is only `showMessages`.  Typically I follow this because when in the future the button needs to show message and then do something else do you rename the function?  This is also the functionality behind `triggerMethod` so I find it consistent.  Right now, I suppose, my naming structure can be dangerous because of auto-proxying, but hopefully that can be disabled soon.

docs/events.md
paulfalgout: Yeah I'm all for that. for event names I pretty much follow https://github.com/jmeas/backbone.class-event-spec

However perhaps we can make a second pass on this stuff?  I didn't really change anything style-wise I don't think.. most of this was changing incorrect or misleading things.

docs/marionette.behavior.md
paulfalgout: :-)  This was changed elsewhere by somebody, but this link got left out

docs/marionette.functions.md
paulfalgout: ayep.. :-(

docs/events.md
paulfalgout: good catch.. just copy-pasta'd  didn't notice the 2nd part

docs/events.md
scott-w: Yeah no worries, I might have time over the next week or two to run through it all.

paulfalgout:
Ok I _think_ this is done.  Moved some stuff on regions..  this is going to conflict a bit with the other region doc PR.. and it's definitely going to conflict with `next`

But I feel like after this is in we can do the doc work for v3.1 easier.

paulfalgout:
One more todo:  need to look at the view add:region remove:region events and API.

denar90:
@paulfalgout thanks)

Issue #3203, opened 2016-09-27
Backbone.Babysitter replacement missing .call and .apply


jgkinnear:
### Description
The replacement for Backbone.Babysitter (Container) doesn't have .call and .apply methods. The Container could have those methods added back in, or the upgrade guide detail the breaking change.

Means the following is no longer available
`view.children.call('doSometing')`

### Expected behavior
I would like the methods added back in purely for compatibility

### Environment

1. Marionette version: 3.0.0
2. Backbone version: 1.3.3

paulfalgout:
Definitely need to document the breaking change, even if we revert it.

However `call` is redundant.  `view.children.invoke('something');` is the same, but a better API handled by underscore.. now in lodash I believe `invoke` is now `invokeMap`..  so if we want to fully support that we should probably add it.. however `myBackboneCollection.invokeMap('something)` isn't going to work either, so I dunno..
`apply` is a little less convienent.  perhaps not with ES6 spread.. but otherwise it'd be something like `view.children.invoke.apply(view.children, ['something].concat(arguments));`

@marionettejs/marionette-core ?

Xref: https://github.com/marionettejs/backbone.babysitter/issues/57

rafde:
to be fair, that issue in the ref link hasn't been closed because it hasn't happened there yet. So the removal of `call` and `apply` did come out of nowhere without deprecation warning (despite their redundancy.

What would be fair is announcing their deprecation and remove for V4 in favor or `invoke` or whatever.

paulfalgout:
This is one among many changes in v3 that were not deprecated ahead of time. The babysitter library was and will likely continue to be woefully unmaintained. If we want to walk this one back that's fine, but it could be one among many. v3 was too large, too long, and with too little contribution. I'd lean towards documenting and not repeating the mistakes of v3 rather trying to resolve them fully.

jgkinnear:
I dont think its necessary to walk back, considering there are alternatives like 'invoke' that are available. The Container specifically would be made publicly accessible too...

`Marionette.Container = Container;`

paulfalgout:
The container was purposefully made not accessible so that `CollectionView` can be updated without concern for creating new dependencies on the container itself.

babysitter can be added separately to a project for handling use-cases that would need `Marionette.Container`

PR #3204, opened 2016-09-27
update emulate-collection to BB 1.3.3

paulfalgout:
I don’t know how valueable this is or if it really that much more performant or important.. but https://github.com/marionettejs/backbone.marionette/issues/3203 had me looking into if Bb supported lodash's `importMap` (they don’t) and then just recognizing how different the current implementation is..

I just noticed there's kind of a breaking change here at the moment.. `children.pluck` would have to be implemented differently.  Related https://github.com/jashkenas/backbone/pull/3865  However.. what the heck does `pluck` do for `children`?  Or maybe in this case, adding pluck to the list might actually work because we don't have the models to deal with.. I'm unsure why it didn't work for Backbone..
@jridgewell / @megawac ?

pros:
* up to date with Bb code and can just follow the leader there.
* apparently more performant

cons:
* way more difficult to grok
* includes code not in use (case 2 and 4)
* will have to add odd tests for coverage

todo:
- [ ] Add tests for coverage
- [ ] Update docs?
- [ ] Determine what to do with cases 2 and 4


rafde:
If there's any consolation regarding `.pluck`, it's still available via `.map`
`this.children.map('model')` theoretically works.

regarding this pr, I am indifferent main because I don't use this unless I am writing my own custom view filter that doesn't go through `CollectionView#filter` to toggle `display:none` on views that don't pass the filter. Which I seldom do anyhow.

If you are reading this and are not a "Marionette.js" member and want this, I suggest 👍 @paulfalgout initial comment.

paulfalgout:
FWIW we do use `each`, `filter`, `find`, and `map` on children internally.. but again, I don't know what perf comes with this change.. particularly to these 4 options.

paulfalgout:
The thing about `pluck` (and some of the others) is that it makes less sense to me..  In terms of making `children` behave like a Bb collection.. if you use pluck on a collection, you're plucking the attributes of the collections models.. 
```javascript
var c = new Backbone.Collection([{ id: 1 }, {id: 2}, {id: 3}]);
c.pluck('id'); // returns [1,2,3]
```
It _doesn't_ return models or other attributes on the collection..   I'm pretty sure the goal of most of this stuff was to make children behave like a collection.. but in this case if there isn't an equivalent behavior for children do we add the interface anyways?  I think I'd argue that the methods relating specifically to model data, should not be available to `children`..

Issue #1952, opened 2014-10-01
UI elements should be rebound on setElement


pimlottc:
When re-binding the view to a new element via `setElement`, the `ui` elements should be updated within the context of the target element.  You can do this manually by calling `bindUIElements` but IMHO this should happen automatically.

Related to #1940.

Test case: http://jsfiddle.net/pimlottc/xvcv8fcw/13/
Clicking anywhere within the PeopleView will toggle the colors on the header, but when the view is reset to a different element (when you hit the Dupe button), `ui.title` still refers to the original header.

jasonLaster:
I'd like to see this issue closed for the time being. I don't see us changing backbone.view behavior currently... If we start doing that, we can revisist these suggestions.

paulfalgout:
I'm not certain that this is changing `backbone.view` behavior anymore than we already are.  Currently `setElement` in backbone calls `delegateEvents`.. so when you `setElement` on a Marionette.View you do get your `events` and `triggers` setup, but since the ui hash is not rebound the ui is still the old jQuery look ups.

I don't know if this affects whether the triggers works or not if they include `@ui.`  I _think_ they'd be fine.. 

So in order to fix this we'd have to add to the currently overridden `_setElement` or perhaps `delegateEvents` something to the effect of:

```js
if(this._isRendered){
  this. bindUIElements();
}
```

Essentially anytime `delegateEvents` needs to be recalled, you also need `bindUIElements` or your `Mn.View` breaks.

jasonLaster:
Hmm persuaded me.

ianmstew:
How would this break apps?  I'd like to get this into `2.5` if possible.

JSteunou:
+1

eth0lo:
Also when calling ``setElement`` on a ``CollectionView`` does not rebound the event handlers on the ``childViews``

Example: If you have a already rendered CollectionView (for example it was rendered server side) and want to rebound using ``setElement`` the delegateEvents are not propagated to the childViews, nor the childViews are set within the collectionView. The events that are bound are those that are defined inside the collectionView only

paulfalgout:
https://github.com/marionettejs/backbone.marionette/blob/master/src/view.js#L100

PR #3207, opened 2016-09-28
When a preexisting node is given to the view it needs to bind the ui

paulfalgout:
One more change for `v3.1`.  While working on the docs I noticed one more inconsistency.

It is correctly setting `isRendered` and `isAttached` but not the `ui` hash.

This only affects `View`.  `CollectionView`'s `setElement` is different.


paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/1952

Issue #3210, opened 2016-09-30
Region el is not emptied when showing a second view


blikblum:
### Description

When a view is show in a region that is holding a previous view instance,  the region element is not emptied. Instead, the old view element is removed and and the new view element is appended.
This can lead to problems when a element is added to the region element after the view is show.
This behavior changed in V3  compared with V2 

In region document (http://marionettejs.com/docs/master/marionette.region.html), the described attachHtml implementation is different from the actual:

Mn.Region.prototype.attachHtml = function(view){
  this.$el.empty().append(view.el);
}

### Expected behavior

The region element should be emptied before showing another view

### Actual behavior

https://jsfiddle.net/98u073m0/31/

### Environment

1. Marionette version: 3.0
2. Backbone version: 1.3
3. Additional build tools, etc: Webpack

scott-w:
Just so I understand, the issue is that custom HTML attached to the region's `$el` (not the view's `$el`) isn't removed.

I've just had a play with the fiddle and the HTML *inside* the view is definitely removed. The question we need to decide on then is whether we want to support this behaviour or whether the documentation is now incorrect and should change.

blikblum:
> Just so I understand, the issue is that custom HTML attached to the region's $el (not the view's $el) isn't removed.

Yes

The issue is due to `empty` method not removing extra added elements:
See https://jsfiddle.net/98u073m0/32/

The actual usage is at https://bitbucket.org/blikblum/sharemonitor/src/e943c2c15c4e827f0b5eeb789c85a5a1123353e2/clients/ionic-mn/scripts/application.js?at=master&fileviewer=file-view-default#application.js-22

After the view is show i need to setup the scroll component that adds a element to hold the scrollbar, but each time a new view is show a new scrollbar element is added

I will play with workarounds and report later

paulfalgout:
Yeah I think if someone is adding DOM manually outside of the marionette view lifecycle, then it's best in those instances to find a way to clean it up manually.  Here's a very specific solution... though you could just add an `$el.empty` inside this particular region's onEmpty.

https://jsfiddle.net/marionettejs/98u073m0/33/

I think it's better to make sure marionette can optimize the main case and have well placed events and hooks for adding and cleaning up anything extraneous.

blikblum:
I fixed in my end by removing the added element

Anyway i think that a update in region docs is in order given that the attachHtml implementation described in it, suggest another behavior.

[...]

Issue #3209, opened 2016-09-29
Add more live examples


denar90:
We have a good start with making live examples, but we have to do more work to cover all docs by them. Here pages without live examples
- [ ] [Regions](http://marionettejs.com/docs/master/marionette.region.html)
- [ ] [CollectionViewAdvanced](http://marionettejs.com/docs/master/marionette.collectionviewadvanced.html)
- [ ] [ViewLifecycle](http://marionettejs.com/docs/master/viewlifecycle.html)

CharlyJazz:
I would like to collaborate on this issue, how can I create Live Examples?

paulfalgout:
It'd be finding the areas within the docs that need them and forking off of jsfiddle.net/marionettejs/adhv48ky/ to make the examples.  You can use your own url and we'll update them to the marionette account upon merge

CharlyJazz:
Should a live example be created for the 'Remove Region' and 'Adding Region' sections?

paulfalgout:
I think the only place that an example would not improve the docs is if one already exists :-D

CharlyJazz:
What you think about this live example for the section 'Adding Region'?

https://fiddle.jshell.net/charlyjazz/gze5ykov/16/

Do you think that is a good example? Or do we need an example more oriented to a real life scenario?

paulfalgout:
Hmm I think a more real life scenario would be better.
Perhaps something like:
```
getTemplate() {
  if (this.isAdmin) return AdminTemplate;

  return Template;
},
onRender() {
  if (this.isAdmin) {
   this.addRegion('admin', '#admin-panel');
   this.showChildView('admin', 'You are an admin!');
  }
}
```

CharlyJazz:
Hi @paulfalgout, what you think about this live example for the section 'Adding Region'?

http://jsfiddle.net/charlyjazz/gze5ykov/49/

Purlicue:
How can i view these live examples? In my here the site 'http://jsfiddle.net' can’t be reached.

paulfalgout:
Hmm maybe you'd need a VPN?  Not sure why jsfiddle wouldn't work

PR #3211, opened 2016-10-01
Updated the region docs

scott-w:
### Proposed changes
 - Document region el not emptied in docs
 - Document upgrade guide difference between 2 & 3
 - Link back to upgrade guide

Link to the issue: #3210

I'm not sure whether to put information in for emulating the behaviour or if we should be discouraging this. I'm leaning towards the latter, as most of the time, we don't want developers setting up their own lifecycle outside of views/regions.

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -174,6 +174,9 @@ Both forms take an `options` object that will be passed to the
 For more information on `showChildView` and `getChildView`, see the
 [Documentation for Views](./marionette.view.md#managing-sub-views)
 
+If you show a view in a region with an existing view, Marionette will
+[remove the existing View](#hiding-a-view) before showing the new one.
+
 ### Hiding a View
 
 You can remove a view from a region (effectively "unshowing" it) with
@@ -189,6 +192,7 @@ mainRegion.empty();
 
 This will destroy the view, cleaning up any event handlers and remove it from
 the DOM.
+[Howevever, any HTML that doesn't belong to the View will remain](./upgrade.md#changes-to-regionshow).
 
 ### Preserving Existing Views
 
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -261,18 +261,22 @@ This can be useful in unit testing your views.
 Override the region's `attachHtml` method to change how the view is attached
 to the DOM. This method receives one parameter - the view to show.
 
-The default implementation of `attachHtml` is:
+The default implementation of `attachHtml` works a bit like the following:
 
 ```javascript
 var Mn = require('backbone.marionette');
 
 Mn.Region.prototype.attachHtml = function(view){
-  this.$el.empty().append(view.el);
+  if (this.isShowingAView()) {
+    this.removeRenderedView();
+  }
+  this.$el.append(view.el);
 }
 ```
 
-This replaces the contents of the region with the view's
-`el` / content. You can override `attachHtml` for transition effects and more.
+If there's a view being displayed, this code replaces it and renders then
+attaches the new view's HTML. You can override `attachHtml` for transition
+effects and more.
 
 ```javascript
 var Mn = require('backbone.marionette');
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -267,16 +267,12 @@ The default implementation of `attachHtml` works a bit like the following:
 var Mn = require('backbone.marionette');
 
 Mn.Region.prototype.attachHtml = function(view){
-  if (this.isShowingAView()) {
-    this.removeRenderedView();
-  }
-  this.$el.append(view.el);
+  this.el.appendChild(view.el);
 }
 ```
 
-If there's a view being displayed, this code replaces it and renders then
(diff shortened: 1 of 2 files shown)


docs/upgrade.md
```diff
@@ -102,6 +102,14 @@ Replace all instances of `show` and `before:show` with `render` and
 `before:render`. If you want the view to be visible in the DOM, then listen to
 the `dom:refresh` event.
 
+### Changes to `region.show()`
+
+The `region.show()` method (and by extension `showChildView()`) made a change to
```
paulfalgout: This isn't true.  If a view is in a region and you empty the region, it _only_ removes the el of the view.. anything else that was added extraneous to the view, should be cleaned up similarly to how it was added..  But if you show a view inside of a region that has html in it, but no view, it _will_ clear out the existing html

https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L210

docs/marionette.region.md
```diff
@@ -189,6 +192,7 @@ mainRegion.empty();
 
 This will destroy the view, cleaning up any event handlers and remove it from
 the DOM.
+[Howevever, any HTML that doesn't belong to the View will remain](./upgrade.md#changes-to-regionshow).
```
blikblum: However

blikblum:
This statement in region.md also needs adjust since the behavior is different from the described:

The default implementation of attachHtml is:

var Mn = require('backbone.marionette');

Mn.Region.prototype.attachHtml = function(view){
  this.$el.empty().append(view.el);
}

docs/upgrade.md
scott-w: Cheers, I was intending to get that across but I've not communicated that clearly enough. I'll amend it now.

scott-w:
@blikblum Thanks for reminding me, I'll put that in 👍

docs/marionette.region.md
```diff
@@ -257,18 +261,22 @@ This can be useful in unit testing your views.
...
 
-The default implementation of `attachHtml` is:
+The default implementation of `attachHtml` works a bit like the following:
 
 ```javascript
 var Mn = require('backbone.marionette');
 
 Mn.Region.prototype.attachHtml = function(view){
-  this.$el.empty().append(view.el);
+  if (this.isShowingAView()) {
```
paulfalgout: I'm not sure this is true..

`attachHtml` is passed a flag as the 2nd argument that's used so that the new view's el can replace the current region's el.. otherwise it just appends..  it doesn't do anything regarding removing a view, and by this point there's no chance that a region has a view in it.

paulfalgout: I think the best example would simply be:
```javascript
Mn.Region.prototype.attachHtml = function(view){
  this.el.appendChild(view.el);
});
```

But we should probably document that if you change the default `attachHtml` then you cannot use `replaceElement`

docs/marionette.region.md
```diff
@@ -257,18 +261,22 @@ This can be useful in unit testing your views.
 Override the region's `attachHtml` method to change how the view is attached
 to the DOM. This method receives one parameter - the view to show.
 
-The default implementation of `attachHtml` is:
+The default implementation of `attachHtml` works a bit like the following:
 
 ```javascript
 var Mn = require('backbone.marionette');
```
rafde: Would it be troubling to place this example in a fiddle?

denar90: @rafde I created related task for examples https://github.com/marionettejs/backbone.marionette/issues/3209
Region doc has no live examples at all. Hence we can skip fiddle for this, if you are not against 😄

scott-w: I think this particular example is easy enough to keep in the docs.

I do agree that we should get some examples in place, however.

denar90:
@scott-w can you take care of merge conflicts?

docs/marionette.region.md
```diff
@@ -267,16 +267,12 @@ The default implementation of `attachHtml` works a bit like the following:
...
-  }
-  this.$el.append(view.el);
+  this.el.appendChild(view.el);
 }
 ```
 
-If there's a view being displayed, this code replaces it and renders then
-attaches the new view's HTML. You can override `attachHtml` for transition
-effects and more.
+The existing `attachHtml` takes an extra `shouldReplace` argument, however if
```
paulfalgout: Yeah.. that `shouldReplace` argument was once a `_shouldReplace` indicating it wasn't for public consumption..  I'm really thinking we want to pull the replaceEl stuff out of here..  I'm not convinced that overriding `attachHtml` should _break_ `replaceElement`.  I think we should move that flag and check outside of this function.  So I'm worried that documenting this argument would make changing it breaking.

scott-w: Cool, I'll remove the documentation before this is merged.

paulfalgout:
@scott-w there's still conflicts, but now that `next` is caught up to master it may be easiest on us to put this in `next` and prioritize the 3.1 release rather than merging into master and pulling it back over.

paulfalgout:
Actually I just manually merged this into `next`

PR #3214, opened 2016-10-03
Update .travis.yml

denar90:
### Proposed changes
 - fix broken CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,8 @@ before_install:
   - npm config set ca ""
 install:
   - curl -L https://rawgit.com/marionettejs/travis-ci-setup/master/ci.sh | bash
+  - npm install travis-ci
 after_success:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,12 +12,8 @@ before_install:
   - npm config set ca ""
 install:
   - curl -L https://rawgit.com/marionettejs/travis-ci-setup/master/ci.sh | bash
+  - npm install travis-ci
 after_success:
   - if [[ $MAINRUN == true ]]; then npm run-script coveralls; fi
-before_deploy:
-  - npm install travis-ci
-deploy:
```
scott-w: I'm not familiar with Travis configs. Why does the `deploy` step need to be removed?

denar90: It was my mistake adding deploy option, which cost us wrong release...
[Here](https://docs.travis-ci.com/user/customizing-the-build#Deploying-your-Code) some doc about deploy option, I think it will explain better then I..

Issue #2999, opened 2016-05-02
no destroy or before:destroy event on region in a view


JSteunou:
When using Region class it's hard to unbind or clean up things because there is no event to tell so from Region. Listening to parent view event do not work either because `stopListening` on region is called very soon. Add to override a bunch of method and listen to region manager to get it work.

rafde:
Are you asking in favor of triggering `before:destroy` on regions?
https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L255

JSteunou:
Yep, on regions. The loc you're linking is on view.

paulfalgout:
```js

JSteunou:
I cannot get it worked in a layoutView :( 
I saw some loc removing regions listeners, maybe it's too soon.

paulfalgout:
Ah yes, so what you're saying it that regions are not destroyed with the `LayoutView#destroy`

I think you can use `before:remove:region` on the layout to do what you're looking for.  Clearly not ideal.

I actually think in `v4` or possibly even `3.x` regions should be created on their first `getRegion` and should be recreated and destroyed with a lifecycle more like views instead of `Region#reset`  In this case there'd be much more use for `destroy`  Region never originally even had a destroy it was added incidentally when most things were made to extend `Object`

JSteunou:
Here is a better description of my pain & sadness around this missing feature ;)

https://jsfiddle.net/jhr7qqre/1/ 

none of the region `console.log` are called :(

paulfalgout:
I still think the `remove:region` events would cover this use-case.  I think in the future Regions should be created and destroyed instead of reset, but in the meantime, Regions aren't really meant to be destroyed.. they're "reset"

https://jsfiddle.net/jhr7qqre/2/

JSteunou:
I understand, but I was looking for an easier API, from dev PoV. When creating regions to add some animation behaviors, it should be usable as it is, like a plugin. Instead I have to hack it or add some code in each views using this region. Not the simplest for custom regions user.

paulfalgout:
I think perhaps this should be renamed.  The issue isn't the events it is that regions never get destroyed.  However I would suggest that we cannot do this right now.  When a region destroys, it empties itself, but that's probably not a performant action right now to tack on to view destruction.

JSteunou:
[...]

PR #3215, opened 2016-10-04
Destroy a region in region remove

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2999

The cost of the region destroy should be very minimal.

### DIFF
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -144,8 +144,7 @@ export default {
   _removeRegion(region, name) {
     this.triggerMethod('before:remove:region', this, name, region);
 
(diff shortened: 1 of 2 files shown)


paulfalgout:
There's a lot of 3.1 docs that need to happen once the docs and master settle.  Plus there's nowhere to add this to..  (I actually missed adding the add:region remove:region events) which is what this is specifically related to..  I'll make a note on the docs PR and try and get to that shortly

scott-w:
Cool, I'm happy to approve as-is as long as it's just additional behaviour and we can catch the docs up later :+1:

Issue #2056, opened 2014-11-05
Allow for custom sorting in a CollectionView


cpjolicoeur:
I'm working on a sample implementation to show as an example, but wanted to open a ticket to start a discussion and see if there was other interest in this.

It would be a nice feature to be able to pass a custom sort option to a CollectionView.  By default, a CollectionView of course keeps its children view rendered in the same sort order as the underlying collection.

In my use case, I need to render the same collection in a few places in my app, but sorted differently in each place.  My current solutions are to clone the original collection and then do manual work to keep the two collections in sync related to model changes, but each collection having its own sort comparator.  Or, In the collection views themselves, I can override the `attachHtml` method to insert the DOM where I want depending on how that view wants to sort things.

It would be nice to be able to pass a custom sort comparator to a collection view and have it use that for its children view sorting or default to the collection sort if no custom sort option is provided.

I'm not fully sure the implications or implementation details on this yet, as I'm still working through it but I'm curious if there is any input on something like this or if there is an existing agreed-upon preferred solution.

samccone:
merged, will go out with 2.4

PR #3216, opened 2016-10-04
[WIP] filter should attach/detach and leave children alone

paulfalgout:
This is barely a WIP.. just some playing..

But while trying to wrap my head around the complexities in https://github.com/marionettejs/backbone.marionette/pull/3173 I began pondering what is making things so complex.

I think Marionette could support filter functionality in a very similar way / API as it does now, but if handled very differently it could greatly reduce the code complexity.  But there are so many moving parts I want to float the idea before I go any further..

**Proposal:**

Marionette's filtering should not modify `children`.  Filter should merely decide which children are attached / detached.  Marionette should maintain children as a collection of views that match the backbone collection.  If a view inside the children is not filtered out, it would be rendered and attached.. if that view is ever filtered out, it would merely be detached.. if the filter changed and it could be displayed again it would simply be reattached and not re-rendered.  Views that were never attached would never be rendered.

I believe this reduces the complexity greatly.  As models are added and removed from the collection, adding them to children is trivial as we don't have to worry about filtering..

So the logic would be
collection events -> update children -> sort children -> update DOM from sort -> attach/detach views as per the filter.

If a view is _removed_ from children via the bb model, then we'd destroy the childview.. otherwise we're just detaching until the collectionview is destroyed.. at that point we can clean up everything else.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -119,10 +119,7 @@ const CollectionView = Backbone.View.extend({
       index = _.indexOf(this._filteredSortedModels(index), child);
     }
 
-    if (this._shouldAddChild(child, index)) {
-      this._destroyEmptyView();
-      this._addChild(child, index)
-    }
+    this._addChild(child, index)
   },
 
   // Handle collection update model removals
@@ -477,7 +474,8 @@ const CollectionView = Backbone.View.extend({
       const view = this.buildChildView(model, EmptyView, emptyViewOptions);
 
       this.triggerMethod('before:render:empty', this, view);
-      this.addChildView(view, 0);
+      this._setupChildView(view, 0);
+      this._addChildView(view, 0, true);
       this.triggerMethod('render:empty', this, view);
     }
   },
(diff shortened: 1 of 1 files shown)


scott-w:
Have you looked into www.scottwalton.codes/marionette.collectionhelper which I wrote to be a way of handling sorting, filtering and paging separately from Mn views?

paulfalgout:
I'm not against wrapping a collection for the purposes of sorting/filtering.  I'm not 100% sold on it needing to be apart of the API though.  I do think there's something to the views themselves handling this, but it is possible it is not reasonable... I don't like the current implementation.

But going back to the original issues for sorting and filtering, wrapping a collection was suggested:
https://github.com/marionettejs/backbone.marionette/issues/2056
https://github.com/marionettejs/backbone.marionette/issues/1596

I do wonder, if done correctly, if handling sorting / filtering at the view level can produce more performant collections.  But I don't know what the solution is.

However as @samccone mentions the filter functionality was meant to be "presentation only" https://github.com/marionettejs/backbone.marionette/issues/1596#issuecomment-48414718
I think attaching/detaching accomplishes this better and is far less of an issue than the current implementation..

Really though we need some good perf testing to understand the best direction.

But I'm not quite ready to dump everything and start over.  If a solution can be done with clean code and good performance using the same API, that'd be my preference.

scott-w:
How would you see this interacting with, say `Backbone.Paginator` and other helper utilities that hide the fact that there's a full collection of which the `CollectionView` only renders some?

paulfalgout:
@scott-w I really don't..  I think filter is better in a collectionview if it is presentation only as it was initially suggested..   I don't think collectionView should be messing with or needing to know much about the data aspect of it.. I think that is over-reaching and poorly located.. if that is the need.. you should filter the collection _prior_ to sending it to the collectionView, much like your library.  I think filter on the collection should be a very easy, very performant way of hiding and showing stuff that's in the collection.. and it should do very little else.

JSteunou:
I'm in favor of the initial suggestion, I did it that way before filtering was in Marionette, and indeed it is easier to maintain

paulfalgout:
@JSteunou sorry clarify please.. what is the initial suggestion?

JSteunou:
maintaining all children in the view and only play with detach / attach

paulfalgout:
closing in favor of https://github.com/marionettejs/backbone.marionette/pull/3244

PR #3218, opened 2016-10-07
Fix example for rendering lists

lvv83:
### Proposed changes
Replace `View` with `CollectionView`


scott-w:
Thanks for spotting this! :+1:

Issue #3219, opened 2016-10-08
Application start this.triggerMethod('start', this, options);


rmc33:
### Description

 In Marionette.Application start this.triggerMethod('start', this, options); is being called causing method onStart to be called with arguments (appObject, options) but according to documentation http://marionettejs.com/docs/v3.0.0/marionette.application.html#application-lifecycle onStart should only receive the options argument. 

I am able to get around this by defining to have two arguments onStart: function(appObject, options)  in the example below. I am confused whether the documentation should be updated to reflect this or if the start function should not pass "this, options" but just "options" since the additional reference to this is not needed in onBeforeStart or onStart but please correct me if I am wrong.

var App = Marionette.Application.extend({

  initialize: function(options) {
    console.log('Initialize');
  },

  //docs say just onStart(options) but in my version two params are passed one is the same object
  //using onStart: function(appObject,options) fixes this
  onStart: function(options) {
    console.log("options = ", options)
  }

});

MyApp = new App();

var cats = [{
  name: 'Wet Cat'
}, {
  name: 'Bitey Cat'
}, {
  name: 'Surprised Cat'
}];

MyApp.start({
  cats: cats
});

### Expected behavior
console output:

Initialize
options =  Object {cats: Array[3]}

### Actual behavior
console output:

Initialize
options =  n {options: Object, _region: undefined, cid: "mna1"}

### Environment

1. Marionette version: 3.0.0
2. Backbone version: 1.3.3
3. Additional build tools, etc:

https://jsfiddle.net/gmbpp0xh/2/

rmc33:
The following unit test supports that the marionette.application doc's should be updated to reflect onBeforeStart(app, options) and onStart(app, options)

https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/application.spec.js

PR #3220, opened 2016-10-08
Fix documentation for Marionette.Application onStart and onBeforeStart

rmc33:
### Proposed changes
Fix documentation references to Marionette.Application onStart and onBefore to include additional app argument. Updated jsfiddle in marionette.application.md.

 - Updating marionette.applcation.md 
 - Updating marionette.region.md

Link to the issue:
https://github.com/marionettejs/backbone.marionette/issues/3219

### DIFF
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -55,7 +55,7 @@ var RootView = require('./views/root');
 var App = Mn.Application.extend({
   region: '#root-element',
 
-  onStart: function() {
+  onStart: function(app, options) {
     this.showView(new RootView());
   }
 });
@@ -130,11 +130,11 @@ var App = Mn.Application.extend({
     console.log('Initialize');
   },
 
-  onBeforeStart: function(options) {
-    this.model = new MyModel(options.data);
+  onBeforeStart: function(app, options) {
+    this.model = new MyModel(this.options.data);
   },
 
-  onStart: function(options) {
+  onStart: function(app, options) {
     this.showView(new MyView({model: this.model}));
     Bb.history.start();
   }
@@ -144,7 +144,7 @@ var app = new App(options);
 app.start();
 ```
 
-[Live example](https://jsfiddle.net/marionettejs/yovad75L/)
+[Live example](https://jsfiddle.net/yovad75L/16/)
 
 As we'll see below, the `options` object is passed into the Application as an
 argument to `start`.
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -55,7 +55,7 @@ var RootView = require('./views/root');
 var App = Mn.Application.extend({
   region: '#root-element',
 
-  onStart: function(myApp, options) {
+  onStart: function() {
     this.showView(new RootView());
   }
 });
@@ -130,11 +130,11 @@ var App = Mn.Application.extend({
     console.log('Initialize');
   },
 
-  onBeforeStart: function(app, options) {
-    this.model = new MyModel(this.options.data);
+  onBeforeStart: function() {
+    this.model = new MyModel(this.options.data); 
   },
 
-  onStart: function(app, options) {
+  onStart: function() {
     this.showView(new MyView({model: this.model}));
     Bb.history.start();
   }
@@ -144,9 +144,9 @@ var app = new App(options);
 app.start();
 ```
 
-[Live example](https://jsfiddle.net/yovad75L/16/)
+[Live example](https://jsfiddle.net/yovad75L/17/)
 
-As we'll see below, the `options` object is passed into the Application as an
+As we'll see below, the `options` object can be passed into the Application as an
 argument to `start`.
 
 ## Starting An Application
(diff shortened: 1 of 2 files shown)


docs/marionette.application.md
```diff
@@ -55,7 +55,7 @@ var RootView = require('./views/root');
 var App = Mn.Application.extend({
   region: '#root-element',
 
-  onStart: function() {
```
scott-w: This part isn't necessary as `app` and `options` aren't used here. I think it's fine to just leave the arguments set below and remove them here.

docs/marionette.application.md
```diff
@@ -130,11 +130,11 @@ var App = Mn.Application.extend({
     console.log('Initialize');
   },
 
-  onBeforeStart: function(options) {
-    this.model = new MyModel(options.data);
+  onBeforeStart: function(app, options) {
+    this.model = new MyModel(this.options.data);
```
scott-w: I think the `this.options` should just be `options` here?

paulfalgout: Yep.. the real problem here is line 143.  `options` should be being passed to start and not at instantiation as mentioned in the note below it.

rmc33: Hi @scott-w  this.options is correct because the options are supplied in the constructor in the example http://marionettejs.com/docs/v3.0.0/marionette.application.html#application-lifecycle. Both app.options and this.options would be correct for this example. I could do the following to illustrate the difference between app.options (this.options) and options:

onBeforeStart: function (app, options) {
   if (options === undefined) {
       this.model = new MyModel(app.options.data); //same as this.options.data
   }
   else {
      this.model = new MyModel(options.data);
  }

rmc33:
@scott-w , @denar90 , @paulfalgout  Thank you for your comments, please review the updated pull request!

denar90:
@rmc33 can you resolve conflicts? Thanks.

docs/marionette.application.md
```diff
@@ -160,6 +160,21 @@ allows you to provide extra configuration for various parts of your app througho
 initialization sequence.
 
 ```javascript
+var App = Mn.Application.extend({
```
denar90: I think live example here will not spoil docs 😄 I'll add one.

docs/marionette.application.md
```diff
@@ -160,6 +160,21 @@ allows you to provide extra configuration for various parts of your app througho
...
+  initialize: function(options) {
+    console.log('Initialize');
+  },
+
+  onBeforeStart: function(app, options) {
+    this.model = new MyModel(options.data);
```
denar90: One more. If we are removing arguments `onBeforeStart: function(options) {` in example above, it will be nice to be consistent and use `this.options.data` here to.

docs/marionette.application.md
denar90: Made some really similar example https://jsfiddle.net/marionettejs/k05dctyt/ to one we have above.

docs/marionette.application.md
rmc33: Hi @denar90 , the second example should use options.data instead to show how to receive options passed in from app.start. Notice this example passes options in app.start instead of var app = new App(options); as in first example.

app.start({
  data: {
    id: 1,
    text: 'value'
  }
});

rmc33: @denar90  for the second example I believe this jsfiddle would be more accurate https://jsfiddle.net/k05dctyt/4/ 

and notice if you tried to change options.data to this.options.data it would not work

rmc33: I added the second live example now, please review

denar90: Yes, true. 👍

denar90: Here fixed example https://jsfiddle.net/marionettejs/k05dctyt/. Please, add it instead of yours.

rmc33: @denar90  ok that is done, thanks

denar90: Thanks.

scott-w: @denar90 Apologies, I was a bit unclear - I think using `options` here is perfectly correct as it demonstrates the use of arguments as @rmc33 points out. Removing it earlier was just to follow the style guide from `eslint`.

docs/marionette.application.md
rmc33: @scott-w  Please note currently on lines 58, 133 and 137 having options as the first argument is misleading since app is passed first so this is why I would prefer these lines either include both arguments or not include either since they are not being used.

paulfalgout:
I think this looks fine.. We should just make sure to squash when we push the merge

denar90:
@rmc33 can you resolve conflicts, one more time, please?
@scott-w can we merge it after?

rmc33:
@denar90  @scott-w, conflicts resolved

paulfalgout:
I'm 👍 on this.  Since it's docs, and adjustments can be made in a subsequent PR

Issue #3201, opened 2016-09-26
Support childViewEventPrefix: false


paulfalgout:
I believe that the magic auto-proxying events from children to parent views is bad.  It is inconsistent between `View` and `CollectionView` (CVs handle `trigger`, but V only handled `triggerMethod`) https://github.com/marionettejs/backbone.marionette/issues/2464

And we have issues with too much bubbling. https://github.com/marionettejs/backbone.marionette/issues/2667

The auto-proxying is also noisy.  It is throwing many, many events that are most likely never listened to.  And now in v3 we have `childViewEvents` and `childViewTriggers` which very easily allow for explicit ways to bubble events.

Beyond that, if you are naming your events in the same manner as you would elsewhere in Marionette, `childViewEvents` and `childViewTriggers` can be _dangerous_ when mixed with auto-proxying.

```js
const MyView = Mn.CollectionView.extend({
  childViewEventPrefix: 'user',
  childView: MyUserView,
  childViewEvents: {
    'select': 'onUserSelect'
  },
  onUserSelect() {
    console.log('This will get called twice!');
  }
});
```
https://jsfiddle.net/marionettejs/f5ndtmbe/

**Recommended change**
If we support passing `false` to `childViewEventPrefix` this could prevent the auto-proxying. 
In `v3.x` we could also add a feature flag that makes `false` the default instead of `childView`.
Then in v4 we could discuss making the flag false by default.

JSteunou:
:+1:

scott-w:
I agree, when I re-read my code-bases full of `onChildviewSomething` all I see is a total mess of methods proxying other methods and I can't pull it together. I think we should be nudging users to using `childViewEvents` and `childViewTriggers` and potentially start deprecating the magical stuff.

blikblum:
> In v3.x we could also add a feature flag that makes false the default instead of childView.
Then in v4 we could discuss making the flag false by default.

Agree. Just would consider removing auto proxying feature completely in v4

PR #3225, opened 2016-10-10
Add childViewEventPrefix false to disable magic auto-proxy

paulfalgout:
https://github.com/marionettejs/backbone.marionette/issues/3201

@marionettejs/marionette-core I think this is a small change with a big win, and I'd like to squeeze it into v3.1 if possible.  This still needs docs, but the doc changes need to be pulled back from master before they can be added. (The relevant section to modify doesn't exist atm)  But that'd make this one among almost all 3.1 features that'll need documentation.  I'm also willing to take that on.

Auto proxying is actually biting me right now because I have a very complicated view structure in one case and I am exceeding the call stack due to https://github.com/marionettejs/backbone.marionette/issues/2667

### DIFF
--- a/src/mixins/view.js
+++ b/src/mixins/view.js
@@ -223,10 +223,6 @@ const ViewMixin = {
   },
 
   _childViewEventHandler(eventName, ...args) {
-    const prefix = _.result(this, 'childViewEventPrefix');
-
-    const childEventName = prefix + ':' + eventName;
-
     const childViewEvents = this.normalizeMethods(this._childViewEvents);
 
     // call collectionView childViewEvent if defined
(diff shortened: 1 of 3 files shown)


src/mixins/view.js
```diff
@@ -242,7 +238,13 @@ const ViewMixin = {
       this.triggerMethod(childViewTriggers[eventName], ...args);
     }
 
-    this.triggerMethod(childEventName, ...args);
+    const prefix = _.result(this, 'childViewEventPrefix');
+
+    if (prefix !== false) {
```
rafde: can't this be `if (_.isString(prefix)) {`.
Seeing this makes me think that anything other than `false` is acceptable.

PR #3226, opened 2016-10-10
Updating the insert before to insert it in a non destroyed view.

CarlosCraviotto:
### Proposed changes

 -
 -
 -

Link to the issue:

In the current implementation we can insert the new view in a destroyed view. With this we will be sure the next view is not a destroyed one.


rafde:
I want to take a look and help this a little more since this will help me with one of the PR versions of https://github.com/marionettejs/backbone.marionette/pull/3173

Thanks for the PR. I will respond some time later since I don't have time right this minute.

paulfalgout:
A fiddle of the issue would be nice.  I'm not sure I grasp the use-case

paulfalgout:
I think I understand this now.. I think it's buggy behavior.. and I think fixing it might be breaking..

But I think essentially a fix for now is something like
```js
CollectionView.extend({
  onChildViewDestroy(childView) {
    this.removeChildView(childView);
  }
});
```

Honestly this situation is awkward because if a childview gets destroyed but remains in the collectionview that means that the views don't represent the data.. at that point doing anything based on the index will be off.

CarlosCraviotto:
I agree 👍

paulfalgout:
Alright, in that case I'm going to close this for now.  Thanks for the issue.  It inspired: https://github.com/marionettejs/backbone.marionette/pull/3244/files#diff-c4fcc4edadcafa174aa6de6124ecbeacR249

PR #3235, opened 2016-10-15
Remove unnecessary variable

38elements:
### Proposed changes
 - This removes unnecessary variable.

The `this` in the  following code is not used in [Renderer.(template, data)](https://github.com/marionettejs/backbone.marionette/blob/5f5e74b7936a35f3f0aebad20079434241360d06/src/config/renderer.js#L16-L27).

```
    const html = Renderer.render(template, data, this);
```

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -144,7 +144,7 @@ const View = Backbone.View.extend({
     const data = this.mixinTemplateContext(this.serializeData());
 
     // Render and add to el
-    const html = Renderer.render(template, data, this);
+    const html = Renderer.render(template, data);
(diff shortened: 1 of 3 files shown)


denar90:
I think current view argument might be useful for people who redefining Renderer. But from other side this possibility is not documented http://marionettejs.com/docs/master/marionette.renderer.html#custom-template-selection-and-rendering. Hence I dunno.

38elements:
Thank you for your reply.
I think your pointing out is correct.
I think that it is better to describe about current view in the document.

PR #3236, opened 2016-10-15
Fix document for Marionette.getOption

38elements:
This fixes document for `Marionette.getOption()`.

### DIFF
--- a/docs/marionette.functions.md
+++ b/docs/marionette.functions.md
@@ -115,8 +115,8 @@ myView.getOption('weight');  // light
 Retrieve an object's attribute either directly from the object, or from
 the object's `this.options`, with `this.options` taking precedence.
 All Marionette Classes, such as the Views, come with this method.
-`this.getOption` is used for Marionette class instance.
-`Marionette.getOption` is used for Backbone class instance.
+`this.getOption` is used for instance of Marionette class.
+`Marionette.getOption` is used for instance of Backbone class.
 
 ```javascript
 var Mn = require('backbone.marionette');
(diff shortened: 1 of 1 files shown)


38elements:
@denar90 
Would you tell me whether or not the following example is correct?

```javascript
var Obj = Mn.Object.extend({
  foo: 1,
  bar: 2,
  initialize: function() {
    console.log(this.getOption('bar'));
  }
});

var obj = new Obj({
  bar: 3,
  baz: 4
}); // => 3

console.log(Mn.getOption(obj, 'foo')); // => 1
console.log(Mn.getOption(obj, 'bar')); // => 3
console.log(Mn.getOption(obj, 'baz')); // => 4
```

38elements:
@denar90 
Thank you for your reply.
I made an example.

denar90:
@38elements thanks for adding changes 👍 
I also added one live example.

docs/marionette.functions.md
```diff
@@ -124,16 +126,35 @@ var Obj = Mn.Object.extend({
 
   initialize: function(attributes, options){
```
paulfalgout: actually this was incorrect to begin with.  `attributes` isn't a thing.  That'd be true for a `Backbone.Model` but not for an object...

docs/marionette.functions.md
```diff
@@ -124,16 +126,35 @@ var Obj = Mn.Object.extend({
...
   }
 });
 
-var obj1 = new Obj(); // => "bar"
+var obj1 = new Obj(); // => 'bar'
+var obj2 = new Obj({}, { foo: 'quux' }); // => 'quux'
+
+```
+
+[Live example](https://jsfiddle.net/marionettejs/271nn9ht/)
```
paulfalgout: this example is forcing the code to work by adding `this.options = options` but that's mostly breaking Marionette.

It is however necessary in a Backbone.Model

docs/marionette.functions.md
38elements: I am sorry.
I sent pull request #3238 .

Issue #2723, opened 2015-09-01
Create a smarter `render` method for `CollectionView`


ahumphreys87:
Currently CollectionView contains various features that require DOM manipulation (sorting, filtering). We have a performant "render" for sorting by enabling `reorderOnSort` but for `filter` the default is to re-render.

For 3.0.0 we should change the default behaviour of `render` to perform a smart render for filter and sorting.

https://github.com/marionettejs/backbone.marionette/issues/2703
https://github.com/marionettejs/backbone.marionette/issues/2596

thousand:
As an add-on to this, I would suggest that filtered views be "hidden" instead of removed from the list of children. This they do not have to be spun up and again on subsequent shows, allowing both better performance and ability to maintain state.

PR #3244, opened 2016-10-28
New CollectionView

paulfalgout:
This is an **experimental** rebuild of `CollectionView`. I believe it could be released along-side the current CollectionView, and then replace the current `CollectionView` for `v4` or `v5`.  This was inspired by all of @rafde's improvements around the collection `update` event and looking into [filtering views leaving children alone](https://github.com/marionettejs/backbone.marionette/pull/3216).  I was also interested in attempting to provide a clean/efficient way of keeping the current API.  While I'm not against filter and sorting at the collection, I do find that having both options available is useful.  They function quite differently and performance for some instances can be improved in the view.

This new CollectionView attempts to keep or improve upon the existing API without removing much, however there are a few breaking changes and modified behaviors.

I certainly need to do documentation and tests.

I've bastardized a perf test from @rafde using the new lib https://jsfiddle.net/marionettejs/x020k0a5/ 

In my initial testing of various CollectionView features, this new version is _much_ faster.  I think it is faster in every case, but additional tests are needed.  The biggest win comes from adding/removing to an already rendered collection.

nextcollectionview

---

**CollectionView Breaking Changes**

_There are probably more, but I tried to keep notes as I went along_
- filtering no longer changes `children`, it merely detaches filtered-out views. https://github.com/marionettejs/backbone.marionette/pull/3216
- `childViewOptions` no longer gets `index` as an argument but this was unstable anyways. Also `children.findIndexByView` was added.
- there is no `attachBuffer` only `attachHtml` (which is essentially what `attachBuffer` was) `attachHtml`s arguments changed fairly significantly
- when a childView is destroyed it is removed from children (I think a better solution than https://github.com/marionettejs/backbone.marionette/pull/3226 but equally as weird)
- ~~`add:child` and `remove:child` were removed - now that almost everything is in bulk these events don't really work.  Plus their original purpose is nearly redundant to `onChildViewRender` or `onChildViewDestroy` I do think a version of these could be added back-in, but it wouldn't really work with filtering. Perhaps what would be better is to look into https://github.com/marionettejs/backbone.marionette/issues/2570~~  **Edit** after looking into this further both events make sense and #2570 is solved with simply `render:children`
- `render:children` events are now triggered whenever the `children` change including sorting or filtering
- `children` are now really only rendered once in the entire life of a `CollectionView`.  If you were previously relying on a `CollectionView` over-rendering a `childview`, this would be breaking.
- `reorderOnSort` was removed as an option and is now the default behavior if `sortWithCollection` is true.
- `viewComparator` is now passed the view instead of the model as an option.  If you want to sort by the model use `view.model` However this also support views added that have no model.
- `getViewComparator` was renamed to `getComparator` to match the set/remove APIs
- `children.findByIndex` is now stable
- `children.add` and `children.remove` were made private (using them will totally break a collectionView)
- `children.findByCustom` was removed since it is unusable (`children` are never added with a custom index so this hasn't been useful to anyone)
- the `sort` option was renamed `sortWithCollection` as it is specifically for sorting when the collection sorts and nothing else.
- `reorder` and `resortViews` were removed and now there's just a `sort` method.  Also the `reorder` events are gone and replaced with `sort` events
- The `filter` option was renamed `viewFilter` and, like sort, `filter` is now a method that can be used to re-apply the current filter.
- Like `viewComparator`, `viewFilter` now also receives the view instead of the model.  The user can add views without models, and `viewFilter` needs to be aware of this.
- `emptyView` is no longer added to `children` so it is built different than a `childView` and doesn't affect `children.length`
- There are no longer events specifically related to the emptyView rendering, however the `emptyRegion` is exposed and will have the typical `show`/`empty` events.

@marionettejs/marionette-core we should probably get many 👍 s and internal testing before considering this.

I'm not 100% certain these resolves are true, but I'm betting on it.

Resolves https://github.com/marionettejs/backbone.marionette/issues/2723 https://github.com/marionettejs/backbone.marionette/issues/2829 https://github.com/marionettejs/backbone.marionette/issues/2863 https://github.com/marionettejs/backbone.marionette/issues/2896 https://github.com/marionettejs/backbone.marionette/issues/2156 https://github.com/marionettejs/backbone.marionette/issues/2948 https://github.com/marionettejs/backbone.marionette/issues/2966 https://github.com/marionettejs/backbone.marionette/issues/2979 https://github.com/marionettejs/backbone.marionette/issues/3043 https://github.com/marionettejs/backbone.marionette/issues/2570 https://github.com/marionettejs/backbone.marionette/issues/3274 https://github.com/marionettejs/backbone.marionette/issues/3298

Mostly resolves https://github.com/marionettejs/backbone.marionette/issues/2560 (I'm less interested now in "regionifying" CollectionView though there are still some similarities that could be shared)


scott-w:
Quick point about `add:child` and `remove:child` - is it worth replacing them with `children:added` and `children:removed` that only fire at most once per operation? Alternatively, is there a form of `update` trigger that lets a listener know that something happened to the view?

paulfalgout:
Yep I think doing something like you mention relating to https://github.com/marionettejs/backbone.marionette/issues/2570 is probably a good idea.  It will probably need to be limited to views added _prior_ to filtering.  So filtering may need to be a different event.

If we attempt to determine `add:children` and `remove:children` after filtering it will be a lot more complicated I think.  (ie: 3 items get added as children, but all 3 get filtered out or the opposite.. 3 items that were filtered out then get removed from the collection)

And we probably could throw an update:children delivering both added/removed backbone style. I'm not 100% sure of that yet.. but probably.

src/next-collection-view.js
```diff
@@ -0,0 +1,615 @@
...
+    }
+
+    this._destroyEmptyView();
+
+    this.triggerMethod('before:sort', this);
+
+    this._sortChildren();
+
+    const filteredViews = this._filterChildren(removedViews);
+
```
paulfalgout: @scott-w This might be a good place to throw a `before:update:children' that get's passed `filteredViews` so you know what views are about to get attached/detached.  However at this point `this.children` has already been updated with new models.. they're just not in the DOM.

src/next-collection-view.js
```diff
@@ -0,0 +1,615 @@
...
+    this._detachChildren(filteredViews.detaching);
+
+    // Check if all views have been filtered out
+    if (this.isEmpty(!filteredViews.attaching.length)) {
+      this._showEmptyView()
+    } else {
+      this._renderChildren(filteredViews.attaching);
+    }
+
```
paulfalgout: @scott-w This might be a good place to throw a `update:children' that get's passed `filteredViews` so you know what views were attached/detached.  Again though.. it's possible the no views were added or removed if this was just a sort.. or a filter was applied that made no changes.

Unlike `remove:child` the views will not have been destroyed yet, and some of them may just be detached because of a filter.

blikblum:
Regarding `sortWithCollection` name, i think that as long the behavior change is documented, is fine to keep `sort` name, that already has a mind share (no need to remember the new option).

The changes are big, so the dev migrating will have to review your CollectionView classes anyway

To make transition smoother, in addition to tests, i suggest to compile into pens/fiddles the different CollectionView use cases, common and edge ones, to make sure things are working

Furthermore, good work

paulfalgout:
@blikblum, good thoughts.

One reason for renaming the `sort` option was to allow for `sort` method as opposed to `resortView` / `reorder` with `reorder` events..  None of those quite describe what is really going on which is just plainly `sort` which also aligns with the Bb collection event / method.

I suspect that the `sort` option was not often used (or rather not modified), so I'm not personally too concerned about the mind share, but I'm certainly up for hearing differently.

I agree with the use-case fiddles for sure.  I'd like to do so with comparison examples.

rafde:
@paulfalgout this work is a great start.

I wanted to point something out about this quote
> (I'm less interested now in "regionifying" CollectionView though there are still some similarities that could be shared)

My reasoning for finding a solution to have`CollectionViews`act like a region that generates `View`s is to replace `CompositeView`s. I don't see this solving for that yet. It's still creates a wrapped container, and it doesn't solve edge cases like https://github.com/marionettejs/backbone.marionette/issues/3150

If this can solve for creating views while this new `CollectionView` serves as a container or a placeholder that gets replaced when children are created and restores the placeholder when the last child is removed from the document, then `CompositeView` can be replaced.

I would like to get this work to that state.

paulfalgout:
@rafde 

I think allowing for the removal of the CollectionView's `el` will be very difficult unless we no longer extend from a `Backbone.View`.  It seems like an entirely different approach that may introduce a lot of complexity mostly for an edge case.

Fortunately I think this new collectionView does solve for that issue rather nicely.. in fact you can quite easily duplicate how a CompositeView works (I suppose we could even add it back).  This is entirely because all content gets attached through a single place.
https://jsfiddle.net/marionettejs/v3xbvh6n/2/

A CompositeView is essentially just:
```js
  attachHtml(collectionView, els) {
    this.$el.contents().detach();
    this._renderTemplate();
    this.$(this.childViewContainer).append(els);
}
```

There _might_ be some caveats.. but I'm struggling to thing of any.

src/next-child-view-container.js
```diff
@@ -0,0 +1,108 @@
+import _ from 'underscore';
+import emulateCollection from './utils/emulate-collection';
+
+// Provide a container to store, retrieve and
+// shut down child views.
+const Container = function() {
+  this._views = [];
```
rafde: cool. doing the https://github.com/joefitter/backbone.enumerable/blob/master/src/backbone.enumerable.js#L9 approach.

scott-w:
@paulfalgout With that in mind, does it make sense to have `childViewContainer` to default to the value of `el` with an option to override it? Something like:

```js
const MyCompositeView = Mn.CollectionView.extend({
  tagName: 'table',
  childViewContainer: 'tbody',
  childView: MyChildView
});
```

So, when you `show` a `CollectionView`, it defaults to assigning `childViewContainer` to the value of `el`, and `attachHtml` always uses `childViewContainer` to determine where it renders its child views.

paulfalgout:
@scott-w my example may have been misleading.  By default this new collectionView doesn't have a template to render.. so the default is _only_ the `el`.  But any user could fairly easily override `attachHtml` and then append any html prior to appending the `els` of the children to that html.

This would be very hard if not impossible with the current collectionView because in one place we make a buffer.. and in one place we insert before or after one at a time.. that said it wouldn't be impossible.

Anyhow I think an external lib or a recipe for making it will be enough.

src/next-child-view-container.js
paulfalgout: Yeah.. I wish I had started there.. unfortunately I went the long way around starting with having the "buffer" in the collectionView available all of the time.. and then indexing it..and a bunch of other bad ideas before it was clear it was better to just do it in the container..  which enumerable knew all along.  @ahumphreys87 does anyone use this lib?  Maybe we should deprecate babysitter and point people to that lib.

ahumphreys87:
Hey @paulfalgout not that I'm aware of, was in my mind as an option to
replace babysitter though so that gets my +1

src/next-child-view-container.js
rafde: There are some things about it that I notice that will not work so well for the child rendering, such as https://github.com/joefitter/backbone.enumerable/blob/master/src/backbone.enumerable.js#L100

which will prevent more than one child constructor type from being added to the list.

I think adopting some of the logic from `Backbone.Enumerable` is best.

src/next-collection-view.js
```diff
@@ -0,0 +1,615 @@
...
+      });
+    }
+
+    comparator = _.bind(comparator, this);
+
+    if (comparator.length === 1) {
+      return _.sortBy(views, comparator);
+    }
+
+    return views.sort(comparator);
```
rafde: All other conditions return a copy of a sorted array except for this one, where it sorts the new childview container's `_view` array and returns it.
Should this also return a copy of the sorted array, or should all sorters update and mutate `this.children._views`?

src/next-collection-view.js
```diff
@@ -0,0 +1,615 @@
...
+
+  setFilter(filter, {preventRender} = {}) {
+    const canBeRendered = this._isRendered && !this._isDestroyed;
+    const filterChanged = this.filter !== filter;
+    const shouldRender = canBeRendered && filterChanged && !preventRender;
+
+    this.filter = filter;
+
+    if (shouldRender) {
+      this._showChildren();
```
rafde: `this._showChildren` will sort `this.children._views` again when it really needs to do line
https://github.com/marionettejs/backbone.marionette/pull/3244/files#diff-c4fcc4edadcafa174aa6de6124ecbeacR294 through https://github.com/marionettejs/backbone.marionette/pull/3244/files#diff-c4fcc4edadcafa174aa6de6124ecbeacR303

src/next-collection-view.js
paulfalgout: Oh interesting.. I didn't realize it mutated in all cases.. that's why I added the `_set` to begin with.  I think you're right.. this could just go straight into the children.. so you'd just pass it a comparator I guess.. and it'll mutate.  

The way backbone checks if it sorted is to loop through pretty much:
```
orderChanged = this.length !== set.length || _.some(this.models, function(m, index) {
   return m !== set[index];
});
```

I'm not sure if we need to do that though.. maybe?  Whether or not it sorted, I don't know that we'll be doing anything overly intensive if it did not sort.

src/next-collection-view.js
paulfalgout: yeah... so.. once we're to the point of setting the filter..I agree it could skip the sort.. but what about destroying the empty view?  Would we have to handle that in two places.. I suppose it could go in this function..  I haven't really loved `showChildren`..  I'll try breaking it up some more.. previously I was thinking most of the events would have to wait until after the collection was rendered into the collectionView `el` to throw an event, but I'm less certain that's a good concern now.

src/next-child-view-container.js
rafde: what new childviewcontainer has is good enough.

src/next-child-view-container.js
```diff
@@ -0,0 +1,107 @@
...
+    // add to end by default
+    index = index || this._views.length;
+    this._views.splice(index, 0, view);
+
+    this._updateLength();
+  },
+
+  _sort(comparator) {
+    if (typeof comparator === 'string') {
+      _.sortBy(this._views, view => {
```
paulfalgout: turns out `_.sortBy` doesn't mutate, so I'll reimplement `_set` momentarily

src/next-collection-view.js
```diff
@@ -390,18 +379,41 @@ const CollectionView = Backbone.View.extend({
...
+  },
+
+  _filterChildren() {
+    if (!_.isFunction(this.filter)) {
+      return this.children._views;
+    }
+
+    this.triggerMethod('before:filter', this);
+
+    const filteredViews = this.children.partition(_.bind(this.filter, this));
```
rafde: `partition`, fancy.

blikblum:
While playing with CollectionView (CV) sort and filter feature i realized that when CV is sorted the events 'add:child' and 'remove:child' are fired. The same when a child view is filtered.

Is this the expected behavior? IMO this is buggy behavior. 
Seems a good time to define if those events should be fired or not

The pen: http://codepen.io/blikblum/pen/XjLRbB?editors=1011

paulfalgout:
@blikblum in the current collection view children are added and removed in many cases which is super wierd. In the new one neither sorting or filtering will trigger these events

paulfalgout:
@marionettejs/marionette-core 
I added a breaking commit. It's specifically to address how to reapply a filter, and renaming the filter API to be more like `sort`..  Plus having `collectionView.sort` and `collectionView.filter` as methods is much better.
```js
collectionEvents: {
  'change:date': 'filter'
}
```

So with this we'd have `viewComparator` and `viewFilter`. Any thoughts on this?  I don't love adding another breaking change, but consistency and simplicity I think out-weigh the alternatives.

rafde:
Is the plan to have this implementation live side by side with the current CollectionView, if so then 🌮 👍

paulfalgout:
Yeah I think so.  Then for v4 or v5 we can just swap it.  I mean it's sizable, but I think it's worth getting it out there for public testing sooner than later.

scott-w:
I'd definitely like to give it a go. Do we want to give it a better name than `NewCollectionView` - or even potentially just call it `_CollectionView` so it can be road-tested with some scope for breaking changes until we settle on a final API?

paulfalgout:
Currently the actual name of the object is NextCollectionView.  I wasn't sure what to name it.  It should probably be obvious though?  Fwiw, we can also make a feature flag that overwrites the current cv with this one too..  I am just not sure when it will be production ready.

src/next-collection-view.js
```diff
@@ -0,0 +1,611 @@
...
+];
+
+// A view that iterates over a Backbone.Collection
+// and renders an individual child view for each model.
+const CollectionView = Backbone.View.extend({
+  // flag for maintaining the sorted order of the collection
+  sortWithCollection: true,
+
+  // constructor
+  // option to pass `{sort: false}` to prevent the `CollectionView` from
```
JSteunou: So is this `sort` or `sortWithCollection`?

paulfalgout: Ah good catch.  Probably need to scan. All of the code comments..

paulfalgout:
I need to pull in https://github.com/marionettejs/backbone.marionette/pull/3234

src/next-child-view-container.js
```diff
@@ -0,0 +1,129 @@
...
+// Container Methods
+// -----------------
+
+_.extend(Container.prototype, {
+
+  // Add a view to this container. Stores the view
+  // by `cid` and makes it searchable by the model
+  // cid (and model itself). Additionally it stores
+  // the view by index in the _views array
+  _add(view, index) {
```
rafde:
```js
_add(view, index = this._views.length) {
```
and delete line https://github.com/marionettejs/backbone.marionette/pull/3244/files#diff-24babd314506fb676ab7fdb23980804cR40

src/next-child-view-container.js
```diff
@@ -0,0 +1,129 @@
...
+    index = index || this._views.length;
+    this._views.splice(index, 0, view);
+
+    this._updateLength();
+  },
+
+  // Sort (mutate) and return the array of the child views.
+  _sort(comparator) {
+    if (typeof comparator === 'string') {
+      comparator = _.partial(stringComparator, comparator);
```
rafde: There isn't a `_.partial` in lodash.core in case that is support for it.
I can't think of anything better alternative, though.

src/next-collection-view.js
```diff
@@ -0,0 +1,684 @@
...
+    _.each(detachingViews, _.bind(this._detachChildView, this));
+  },
+
+  _detachChildView(view) {
+    const shouldTriggerDetach = !!view._isAttached;
+    if (shouldTriggerDetach) {
+      triggerMethodOn(view, 'before:detach', view);
+    }
+
+    this.detachHtml(view);
```
rafde: It doesn't seem like `this.detachHtml(view);` should this be outside of `if (shouldTriggerDetach) {` to me. What would be the point of detaching something that seems to be already detached. unless this is some safe guard.

src/next-child-view-container.js
paulfalgout: hmm yeah I dunno.  `_.partition` isn't core either.  I don't know how much I care about core since it didn't affect underscore.  Do you have a preference?  Would we need to list what lodash modules outside of core we're using?

src/next-collection-view.js
paulfalgout: This'll still detach this child view from this view without destroying the child.  the detach event is only concerned with if the view is in the DOM.. not if it is attached to some `el`.  I don't think we want to prevent child view management if the collectionview isn't in the DOM.

src/next-child-view-container.js
paulfalgout: partition would be easy to implement internally if need-be I suppose.

rafde: I wouldn't bother with supporting lodash.core since it wasn't meant to be supported by other lib usage.

paulfalgout: sounds good to me

paulfalgout:
Alright @marionettejs/marionette-core I think the code and tests are done here.

My plan to complete this work is to:
- [ ] Test out all of the issues this work might resolve
- [ ] Scan the current collectionview tests for any missing use-cases
- [ ] Document next collectionview functionality
- [ ] Blog about why the change

paulfalgout:
One things that's always bugged me a little bit about the `CollectionView` API is `childView` and `emptyView` are lowerCamel which to me makes me think of an instance and not a class.  @marionettejs/marionette-core anyone feel similarly, or prefer the current API to UpperCameling these 2 keys?

Part of me doesn't want to add another breaking change, but part of me would prefer to just get it over with now.

Thoughts?

RobertCZ:
UpperCameling childView: Yes, please.

paulfalgout:
Sadly, I might be talking myself out of it a little.  this'd make this collectionview inconsistent with `regionClass` and `behaviorClass`... This might just be worth a new issue where we can update all class keys at the same time for v4

blikblum:
There also Backbone Collection.model.

Those properties can be also a function

So i would not do so big break changes with no clear benefit

paulfalgout:
@marionettejs/marionette-core alright I need to test the docs and double check them, but if this could be reviewed that'd be great.  I'll work on the accompanying blog post, but otherwise this is super close to ready-to-roll

src/next-collection-view.js
```diff
@@ -0,0 +1,679 @@
...
+      viewComparator = viewComparator.bind(this);
+    }
+
+    this.children._sort(viewComparator);
+
+    this.triggerMethod('sort', this);
+  },
+
+  // Sets the view's `viewComparator` and applies the sort if the view is ready.
+  // To prevent the render pass `{ preventRender: true } as the 2nd argument.
```
rafde: `{ preventRender: true }`

src/next-collection-view.js
```diff
@@ -0,0 +1,679 @@
...
+  },
+
+  // Override this function to provide custom
+  // viewFilter logic
+  getFilter() {
+    return this.viewFilter;
+  },
+
+  // Sets the view's `viewFilter` and applies the filter if the view is ready.
+  // To prevent the render pass `{ preventRender: true } as the 2nd argument.
```
rafde: `{ preventRender: true }`

rafde:
@marionettejs/marionette-core : 2 👍 . Anyone else want to take a look before merging?

JSteunou:
Hell yeah! Good job @paulfalgout! This is a massive improvement as CollectionView is one of this biggest Marionette's feature, thank you so much!

Issue #3170, opened 2016-09-07
onAddChild callback params changed, never noted in Changelog.


KeyHui:
Before:

``` javascript
{
    onAddChild: function(childView) {}
}
```

Now:

``` javascript
{
   onAddChild: function(currentView, childView) {}
}
```

Have other function params changed ?

scott-w:
Thanks for this @KeyHui, I've updated your message with code blocks. There's likely a gap in our documentation in the arguments passed to event handlers here.

paulfalgout:
This actually true for all events.  All view events pass itself as the first argument now.  So also new for `add:region` `remove:region`

scott-w:
I'm thinking I'll document this in the tables for each event as a third column. I'll get a PR in for you guys to have a look over and see how it looks.

PR #3245, opened 2016-10-30
Document Event Args

scott-w:
### Proposed changes
- Add arguments to table
- Change Magic Methods to `onEvent` handlers
- Note the event changes in the upgrade guide

Link to the issue: #3170

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -74,7 +74,8 @@ parent. These can be chained all the way up to the level you require them to be.
 
 Bubbled child events no longer pass the `childView` implicitly and only pass the
 arguments passed as part of `triggerMethod`. This means that the arguments
-passed to `onEvent` and `onChildviewEvent` are now identical.
+passed to `onEvent` and `onChildviewEvent` are now identical. See the
+[documentation on event lifecycles](./viewlifecycle.md) for more information.
 
 In Marionette 2, `childEvents` where bound on every event. In Marionette 3,
 `childViewEvents` are bound once and cached. This means that you cannot add new
@@ -167,6 +168,13 @@ In Marionette 3, the HTML will be:
   <div>View contents</div>
 </div>
 ```
+
+### Arguments for Lifecycle Events
+
+The arguments to all lifecycle events were completely changed. All events will
+now receive the view that is emitting the event as the first argument. See the
+[documentation for view lifecycles](./viewlifecycle.md) for more complete
+information.
 
 ## Templates
 
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -171,10 +171,23 @@ In Marionette 3, the HTML will be:
 
 ### Arguments for Lifecycle Events
 
-The arguments to all lifecycle events were completely changed. All events will
-now receive the view that is emitting the event as the first argument. See the
-[documentation for view lifecycles](./viewlifecycle.md) for more complete
-information.
+The arguments for a number of lifecycle events were changed. For consistency,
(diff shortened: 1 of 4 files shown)


docs/viewlifecycle.md
```diff
@@ -303,19 +310,23 @@ The `CollectionView` creation lifecycle can go down two paths depending on
...
-|   7   |      `before:attach`     |
-|   8   |         `attach`         |
-|   9   |       `dom:refresh`      |
+| Order |           Event          |                Arguments                |
+| :---: |--------------------------|-----------------------------------------|
+|   1   |      `before:render`     | `this` - collection view being rendered |
+|  2*   |  `before:render:empty`   | `this` - collection view being rendered |
+|       |                          |    `view` - empty view being rendered   |
+|  2+   | `before:render:children` | `this` - collection view being rendered |
+|   3   |    `before:add:child`    | `this` - collection view being rendered |
```
blikblum: I just checked in one app and this and view points to the same object (collection) in onBeforeAddChild
http://codepen.io/blikblum/pen/XjLRbB

blikblum: I've got confused by wording of this. this in the table mean the first argument not this JS keyword.
Maybe replace this by collection? (I updated my pen arguments in onBeforeAddChild - and fixed a bug ;-))

docs/upgrade.md
```diff
@@ -168,6 +169,13 @@ In Marionette 3, the HTML will be:
 </div>
 ```
 
+### Arguments for Lifecycle Events
+
+The arguments to all lifecycle events were completely changed. All events will
```
paulfalgout: Kind of.  Most events already received it.  What changed was that all events were made consistent.

I believe all of the breaking events because of this change were (each having a `before:`):
- Object `destroy`
- Region `show`
- View's region add and remove events
- CollectionView's add and remove child events
- Application `start`

Every other event was either already passed `this` as the first arg, or had no args to begin with.

docs/viewlifecycle.md
scott-w: Good catch on that. I think `collectionView` as the first argument perhaps?

docs/upgrade.md
paulfalgout: I'm honestly not too worried about the comparison.. the args weren't super well documented before and for these lesser used events.. probably saw little use.  I just don't want to worry anyone looking to upgrade, when most of the heavily used events didn't change.

scott-w: Cool, I'll just list the ones that did change and direct the reader to the documentation for everything so people can check it out for themselves.

blikblum:
Seem that there are two typos in example of http://marionettejs.com/docs/master/viewlifecycle.html#reorder--beforereorder-events

```
myView.on({
  'remove:empty': function() {      -> 'reorder'
    console.log('the collection view has been reordered following its collection');
  },
  'before:remove:empty': function() {   -> 'before:reorder'
    console.log('the collection view is about to be reordered');
  }
});
```

scott-w:
I've updated that now 👍

docs/viewlifecycle.md
```diff
@@ -508,10 +508,10 @@ myView.render();
 myCol.comparator = function () { return this.get('foo'); };
 
 myView.on({
-  'remove:empty': function() {
+  'reoder': function() {
```
paulfalgout: This is smelly

paulfalgout: Well and `bfore:reoder` too

Issue #2508, opened 2015-04-17
We should extend Backbone.Events


jasonLaster:
Marionette has added new application level functionality to Backbone.Events in the form of triggerMethod. We've made it possible to invoke on methods. We've enabled event bubbling up the view hierarchy for layouts and collections... These features have been added to triggerMethod and have lead to inconsistencies with trigger. This has created some considerable confusion. These features have also lead to regressions such as on methods been called post `stopListening` calls.

As we prepare for a major release in Marionette 3, it'd be great to use the opportunity to remove the distinction. What I'm proposing, might seem radical at first, but I'd like to extend Backbone.Events to add triggerMethod functionality to trigger and then remove triggerMethod.

How would this work? If we go ahead with Backbone.Metal, this would mean defining Backbone.Events as a Mixin and Marionette.Events as a subclass of Backbone.Events. Then mxing in Marionette.Events into Mn.Object, Backbone.View, Backbone.Model...

If we do not go aheas with Metal, we could do something like this, but it'd be a little uglier...

What do you guys think?

paulfalgout:
I am :+1: for extending `Backbone.Events`  I'm not fully sure about how `triggerMethod` should "stop listening" yet, but without extending `Backbone.Events` it will be very complex to solve reasonably.

I also think this will be a useful extend for other things down the line

paulfalgout:
Note: The proposed metal implementation is already introducing a `Marionette.Events`, but with no particular additional utility

ianmstew:
:+1: The fact that we have to manage two separate implementations when we expect to have both function, for all intents and purposes, identically, is a huge argument for this.  For example #2464 and discussion on #2404 would both be simplified by consolidating our triggering impls.

taburetkin:
please remove obsolete bind and unbind )

PR #3247, opened 2016-11-02
View childViewEvents should support trigger

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2667 and https://github.com/marionettejs/backbone.marionette/issues/2464

This makes child events consistently work between both views for both `triggerMethod` and `trigger`  Currently some bubbling exists between view types.  This will add additional bubbling.  Previous to having `childViewEventPrefix: false` I would have been against the bubbling, but it can be managed easily now.

However we _may_ want to wait until `childViewEventPrefix: false` is the default?

@rafde any other perf concerns?  I like the consistency, but realize this might be doing more now.. however since it's regions on a view it is _probably_ not an issue.  If you think it is an issue perhaps we should consider breaking CollectionView such that it too only works with `triggerMethod` ?

I'm bringing this up now as we _could_ introduce this https://github.com/marionettejs/backbone.marionette/pull/3244

### DIFF
--- a/test/unit/view.child-views.spec.js
+++ b/test/unit/view.child-views.spec.js
@@ -287,11 +287,13 @@ describe('layoutView', function() {
   describe('when showing a childView as a View', function() {
     beforeEach(function() {
       this.layoutView = new this.View();
-      this.childEventsHandler = this.sinon.spy();
+      this.childEventsHandlerTrigger = this.sinon.spy();
+      this.childEventsHandlerTriggerMethod = this.sinon.spy();
 
       // add child events to listen for
       this.layoutView.childViewEvents = {
-        'content:rendered': this.childEventsHandler
+        'before:content:rendered': this.childEventsHandlerTrigger,
+        'content:rendered': this.childEventsHandlerTriggerMethod
       };
       this.layoutView.delegateEvents();
       this.layoutView.render();
@@ -299,6 +301,9 @@ describe('layoutView', function() {
       // create a child view which triggers an event on render
       var ChildView = Marionette.View.extend({
         template: false,
+        onBeforeRender: function() {
+          this.trigger('before:content:rendered');
+        },
         onRender: function() {
           this.triggerMethod('content:rendered');
         }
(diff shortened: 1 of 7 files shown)


rafde:
Perf concerns.... I will have to run test to find out. Which will goad me to get the perf checker into the repo. Btw, the perf checker will always test against src vs lib.

paulfalgout:
I wouldn't be so concerned about perf here since you probably don't have 1000 regions on a view like you might have on a collectionview which is already doing this.. except that.. for sufficiently deep views nesting on an existing app this is going to open the door to deep bubbling.. soo that could slow things down if not managed.

test/unit/view.child-views.spec.js
```diff
@@ -300,7 +300,7 @@ describe('layoutView', function() {
       var ChildView = Marionette.View.extend({
         template: false,
         onRender: function() {
-          this.triggerMethod('content:rendered');
```
scott-w: Is this signalling a move towards removing `triggerMethod` in favour of just using `trigger` everywhere instead?

paulfalgout: Not directly.  That'd be https://github.com/marionettejs/backbone.marionette/issues/2508

However if this were a childView of a `CollectionView` then `trigger` would work.  This PR is just about making sure that `trigger` works in both places.

rafde:
@paulfalgout regarding perf. no noticeable degradation in render time.

test/unit/view.child-views.spec.js
rafde: @scott-w would you prefer another childViewEvents https://github.com/marionettejs/backbone.marionette/pull/3247/files#diff-0f13482af07d71e12b54fc8c40be7af6L293 to cover the `triggerMethod` case?

paulfalgout:
I went ahead and added a test so this can get in.

Hopefully the new test and the one that changed doesn't last as they're all pretty awful. https://github.com/marionettejs/backbone.marionette/issues/3248

paulfalgout:
@marionettejs/marionette-core I just realized that this change warranted some additional code cleanup.  Can I get 👀 again?

rafde:
I think it all checks out @paulfalgout 
I am surprised by how much was removed to allow something very useful to work for both trigger and triggerMethod, it almost seems to good of a change.

Issue #1579, opened 2014-07-07
View/Behavior constructor/initialize order


thejameskyle:
Right now View/Behavior constructor/initialize methods run in this order:
- Behavior::constructor
- Behavior::initialize
- View::constructor
- View::initialize

I think this would be the ideal order:
- Behavior::constructor
- View::constructor
- Behavior::initialize
- View::initialize

https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.view.js#L19-L23
https://github.com/marionettejs/backbone.marionette/blob/master/src/marionette.behavior.js#L28

``` js
// View::constructor
if (_.isObject(this.behaviors)) {
  new Marionette.Behaviors(this); // < calls Behavior::initialize
}

Backbone.View.apply(this, arguments);
```

I think we should remove the explicit initialize call and add it to the the wrapped methods list:

``` diff
// marionette.behavior.js
- this.initialize.apply(this, arguments);

// marionette.behaviors.js
 Behaviors.wrap(view, behaviors, [
+  'initialize',
   'bindUIElements', 'unbindUIElements',
   'delegateEvents', 'undelegateEvents',
   'behaviorEvents', 'triggerMethod',
   'setElement', 'destroy'
 ]);
```

MeoMix:
Hey,

It doesn't look like this change ever landed? I'm trying out v3 right now and I'm not able to access "el" from a behavior's initialize.

paulfalgout:
@marionettejs/marionette-core we might want to consider this.

How about:

```js
const View = Backbone.View.extend({

  constructor(options) {
    this.render = _.bind(this.render, this);
    
    // no longer a need for this
    // this._setOptions(options);

    monitorViewEvents(this);

    const viewInitialize = this.initialize;
    this.initialize = function() {
      this._initBehaviors();
      this._initRegions();
      viewInitialize.apply(this, arguments);
    };

    Backbone.View.prototype.constructor.call(this, this.options);

    this.initialize = viewInitialize;

    this.delegateEntityEvents();
  },
```

paulfalgout:
😿 
So nope:

```js
var View = Backbone.View = function(options) {
    this.cid = _.uniqueId('view');
    _.extend(this, _.pick(options, viewOptions));

// For this to work we need to call _initBehaviors here.

    this._ensureElement();
    this.initialize.apply(this, arguments);
  };
```

Otherwise the `events` hash isn't correct for `_ensureElement`

[...]

PR #3250, opened 2016-11-05
Add events order + fixes to behavior doc 

blikblum:
### Proposed changes
 - Document order of events between behavior and view
 - Merge Using ui and referencing DOM since the subject is the same
 - Add note about triggering an event in the view

The example is created in an anonymous fiddle. It should be migrated to an marionettejs account?

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -17,7 +17,7 @@ collection interactions to be utilized across your application.
   * [Listening to View Events](#listening-to-view-events)
   * [Proxy Handlers](#proxy-handlers)
   * [Events Order](#events-order)
-  * [Using ui](#using-ui)  
+  * [Using `ui`](#using-ui)  
   * [View and el](#view-and-el)
 
 
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -16,7 +16,7 @@ collection interactions to be utilized across your application.
 * [View Proxy](#view-proxy)
   * [Listening to View Events](#listening-to-view-events)
   * [Proxy Handlers](#proxy-handlers)
-  * [Events Order](#events-order)
+  * [Events / Initialize Order](#events-initialize-order)
   * [Using `ui`](#using-ui)  
   * [View and el](#view-and-el)
 
@@ -321,17 +321,22 @@ including:
 * [`modelEvents`](./marionette.view.md#model-events)
 * [`collectionEvents`](./marionette.view.md#collection-events)
 
-### Events Order
+### Events / Initialize Order
 
 If both view and behavior are listening for the same event, this will be executed
-first in the view than in the behavior. 
+first in the view than in the behavior as below.
 
-Unlike events, the initialize method will 
-be called first in the behavior. 
+The View + Behavior initialize process is as follows:
+
+1. View is constructed
+2. Behavior is constructed
+3. Behavior is initialized with view property set
+4. View is initialized
+
+This means that the behavior can access the view during its own `initialize` method.
+The view `initialize` is called later with the information eventually injected by the behavior.
   
-> While initialize is called in behavior before than in view, the view property is already set, and can be accessed, inside behavior initialize 
-  
-[Live example](https://jsfiddle.net/9wtqgwgu/)  
+[Live example](https://jsfiddle.net/marionettejs/qb9go1y3/)  
 
 #### Using `ui`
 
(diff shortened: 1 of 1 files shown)


docs/marionette.behavior.md
```diff
@@ -16,8 +16,10 @@ collection interactions to be utilized across your application.
 * [View Proxy](#view-proxy)
   * [Listening to View Events](#listening-to-view-events)
   * [Proxy Handlers](#proxy-handlers)
+  * [Events Order](#events-order)
+  * [Using ui](#using-ui)
```
scott-w: Can you make this: 
```
  * [Using `ui`](#using-ui)
```

docs/marionette.behavior.md
```diff
@@ -316,10 +321,23 @@ including:
...
+
+If both view and behavior are listening for the same event, this will be executed
+first in the view than in the behavior. 
+
+Unlike events, the initialize method will 
+be called first in the behavior. 
+  
+> While initialize is called in behavior before than in view, the view property is already set, and can be accessed, inside behavior initialize 
+  
+[Live example](https://jsfiddle.net/9wtqgwgu/)
```
denar90: It's better to use `marionettejs` account for fiddle.

docs/marionette.behavior.md
blikblum: Done

docs/marionette.behavior.md
blikblum: How it can be done?

blikblum: While initialize is called in behavior before than in view, the view property is already set, and can be accessed, inside behavior initialize 

The above sentence, is not very good. I tried various versions. If someone has suggestions, please do.

scott-w: I usually do a heading of sorts - something like:

```markdown
#### How Behavior is Initialized

The View+Behavior initialize process is as follows:

1. View is constructed
2. Behavior is constructed
3. Behavior is initialized with attached view
4. View is initialized

This means that the Behavior can see an unitialized view during its own initialize setup and can inject information into the view to consider for its own initialize steps.
```

denar90: As addition to @scott-w comment some info can be grab from https://github.com/marionettejs/backbone.marionette/issues/1579

blikblum: @denar90 , thanks. #1579 was one of my source of information although the initial order is not valid anymore. I updated the fiddle to show constructor: https://jsfiddle.net/marionettejs/qb9go1y3/1/ .

This knowledge was one that i missed and had to test myself when i was learning how to use behavior. As the other changes i'm proposing, it is inspired by the lacks i observed reading the docs.

docs/marionette.behavior.md
```diff
@@ -16,8 +16,10 @@ collection interactions to be utilized across your application.
 * [View Proxy](#view-proxy)
   * [Listening to View Events](#listening-to-view-events)
   * [Proxy Handlers](#proxy-handlers)
+  * [Events / Initialize Order](#events-initialize-order)
```
paulfalgout: This is going to be `#events--initialize-order`

docs/marionette.behavior.md
```diff
@@ -306,6 +308,9 @@ triggered on a `View` are passed to all attached `behaviors`. This includes:
 These handlers work exactly as they do on `View` -
 [see the `View` documentation](./marionette.view.md#events)
 
+> Events triggered in behavior instance are not executed in view. To notify
```
paulfalgout: can we say "in the behavior instance" and "in the view" to match the next sentence?

docs/marionette.behavior.md
```diff
@@ -316,10 +321,28 @@ including:
 * [`modelEvents`](./marionette.view.md#model-events)
 * [`collectionEvents`](./marionette.view.md#collection-events)
 
+### Events / Initialize Order
+
+If both view and behavior are listening for the same event, this will be executed
+first in the view than in the behavior as below.
```
paulfalgout:
```diff
- than
+ then
```

docs/marionette.behavior.md
blikblum: Done

docs/marionette.behavior.md
blikblum: done

paulfalgout:
Just noticed this introduced doc linter errors: https://travis-ci.org/marionettejs/backbone.marionette/builds/173784807#L987

@denar90 we should probably make a doc linter error fail the build.

denar90:
@paulfalgout these are warning. But yeah it's better to fail build so we can avoid lint warnings in future.

blikblum:
I will open a new pr fixing them

Issue #2719, opened 2015-08-28
Rerendering collectionviews on `reset` events throw.


jeffijoe:
I've managed to reproduce this in a Fiddle:

http://jsfiddle.net/jeffijoe/wzxxpp06/2/

The issue arises when a LayoutView is hosting a collectionview, and the layoutview also
rerenders when the collection is reset.

This does not apply to removing all models from the collection, it's just resetting.

Code from Fiddle:

``` javascript
console.clear();
let Mn = Marionette, Bb = Backbone;

let Col = Bb.Collection.extend();

let data = new Col([{cat: 'Bob'}, {cat: 'Mittenz'}]);

let ListView = Mn.CollectionView.extend({
    childView: Mn.ItemView.extend({template: '#item'})
});

let MainView = Mn.LayoutView.extend({
    el: '#app',
    template: '#main',
    initialize() {
        this.listenTo(data, 'add reset remove', this.render);
    },
    templateHelpers() {
        return {
            hasAnyCats: data.length > 0
        };
    },
    regions: {
        list: '.list'
    },
    onRender() {
        if (data.length > 0) {
            this.showChildView('list', new ListView({collection: data}));
        }
    }
});

new MainView().render();

setTimeout(() => data.add({cat: 'Dog'}), 1000);
setTimeout(() => data.reset(), 2000);
```

jasonLaster:
@ianmstew any ideas?

@jeffijoe what marionette version?

jeffijoe:
@jasonLaster The latest on npm.

paulfalgout:
This is a dupe https://github.com/marionettejs/backbone.marionette/issues/1736

If a parent view and a child view is queued to listen to the same event, the parent view cannot do anything that will destroy the child, as the child view event will still be queued to run.

paulfalgout:
Actually this is a less of a dupe, and more of a very related issue.

The problem here is that there isn't a way to work around the "error" that marionette throws when you try and render a destroyed view.

https://github.com/marionettejs/backbone.marionette/blob/v2.4.2/src/collection-view.js#L92

```
this.listenTo(this.collection, 'reset', this.render);
```
should probably change to something like:
```
this.listenTo(this.collection, 'reset', function(){
  if(this.isDestroyed) return;
  this.render();
});
```

jeffijoe:
@paulfalgout that's the only way? Would not throwing internally "fix" it?

jeffijoe:
[...]

Issue #1744, opened 2014-08-19
ViewDestroyedError woes


paulfalgout:
This error is killing me.

I understand why it exists.  I don't at any point use a view after it's destroyed except when this happens:  https://github.com/marionettejs/backbone.marionette/issues/1736

When it does happen, the error is not very useful.  In these situations I may have many nested views working off the same model, and if something changing on the model causes a parent view to destroy I could have a nested view which renders on a model change as well, causing the error.  

The error stack however is simply the trigger that caused the error and there is no indication as to which view is causing the problem.  In the very least, I'd like to see the view's cid in the error thrown.

However I'd prefer the ability to show the error only under a debug flag and otherwise just prevent the render.  Otherwise literally anywhere a view calls render from event that has any chance of being listened to anywhere else will need `if(this.isDestroyed) return;`

It is far too easy to cause this error, particularly when the expectation would be that view events would not continue after a view is destroyed.

Here's a simplified example of this issue:  http://jsfiddle.net/paulfalgout/rg5u3g00/1/

Issue #1736, opened 2014-08-18
Unexpected ViewDestroyedError behavior


paulfalgout:
My initial expectations are that when a layout view is closed, any nested view listeners are cleaned up.  However this is not entirely the case and it's causing a ViewDestroyedError.  Though the listeners are cleaned up, it is possible that they are still handled after the view is destroyed.  I believe it's because the modelEvents are handled apart from the views, so if two views are handling the same modelEvent and one view closes another (as it should in the following case), the event still continues.

I can walk through the code and understand why this is happening, but I think there is some possible improvement needed somewhere.

As an example I have a layout view and a nested view both working off the same model. The nested view has

```
modelEvents: {
  ‘change:status’: ‘render’
}
```

The parent layout view is also listening to the status, and on a particular status destroys itself, which of course also destroys the nested view.. however even though the nested view gets destroyed, that ‘change:status’ model event happens, and it happens after the view is destroyed causing the view destroy error

Is ‘change:status’: ‘render’ just a bad pattern and now everywhere we should be doing this?

```
modelEvents: {
  ‘change:status’: function(){
    if(!this.isDestroyed) this.render();
  }
}
```

Or is it possible modelEvents not being cleaned correctly when a view inside a layout is closed?

Issue #2831, opened 2015-11-24
UI interpolation eagerly evaluated and cached


jiaweihli:
I've written a wrapper around a UI element that only injects an element into the DOM when it's clicked.
The selector to the injected element is getting evaluated / overwritten to empty array before I ever access it.

paulfalgout:
I think I follow, but some pseudo code could help

jasonLaster:
If you want you can recall bindUiElements

jiaweihli:
```typescript
class SidebarDropdownView<T> extends DropdownView<T> {
  ...

  onRender(): void {
    super.onRender();

    this.ui.select2Input
      .on('select2-open', () => {
        // :UPGRADE: (marionette) Selectors are auto eagerly-cached, so we can't use the UI hash.
        // :LINK: https://github.com/marionettejs/backbone.marionette/issues/2831
        // :KLUDGE: Add select2 dropdown styling specific for this view.
        $('#select2-drop').addClass('for-sidebar-dropdown');
      });
  }
}
```

When the dropdown (powered by select2) is opened, a new div `#select2-drop` is appended to `body` and a [`select2-open` event is fired](https://select2.github.io/select2/).

It feels inconsistent to sometimes call `bindUIElements` and sometimes not.

Maybe add an option to completely disable UI interpolation caching in a view?  I recognize per-selector disabling takes a little bit more thought and design :smile:

paulfalgout:
I'm not sure what the solution is here..   `bindUIElements` is called currently when `View#render` is called.  If DOM is added after the view is rendered it is unaware and can't bind the UI.

But the bigger problem here I think is that view ui is scoped within the view.  so if `#select2-drop` is appended to the `body` the `SidebarDropdownView` is never going to be able to append it to the `ui` hash.  You can only bind things that you can access via `this.$()`  So if you can replace `$('#select2-drop')` with `this.$('#select2-drop')` in your code and have it work.. then some sort of smarter binding _might_ be possible..  but in this case, I do not think that it is.

jiaweihli:
Making UI interpolation work for global elements is a separate concern - a possible solution would be to treat selectors beginning with `:root` as a global selector during interpolation.  (e.g. `:root #select2-drop`)

I don't use UI interpolation anywhere right now because I don't want to think about whether it applies to a dynamic element or not - a solution that works 90% of the time but subtly doesn't work the other 10% of the time isn't that useful.

[...]

PR #3252, opened 2016-11-07
Remove ViewDestroyError from Views

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2719

I’ve brought up this issue before https://github.com/marionettejs/backbone.marionette/issues/1736 https://github.com/marionettejs/backbone.marionette/issues/1744

In v3 showing a view in a region (or via collectionView) is very important for setting up the view’s lifecycle.  Even if you’re using a `Backbone.View`, you’re not getting much value out of Marionette if you are not using regions.  Even though we are working to abstract them further, they are the glue that make things work.  The region `show` [already checks for this error](https://github.com/marionettejs/backbone.marionette/blob/v3.1.0/src/region.js#L54).  And so it is very unlikely that the view’s error will get called outside of rendering the view directly.

And rendering it directly after it has been destroyed mean that most likely it’s not in the DOM (as destroying the view will remove it from any region)

Making the view `render` no-op after the view is destroyed is consistent with the view `destroy` function.

Making it a no-op is harmless.  The biggest issue this _might_ introduce is the error you would get from `getUI` will be less descriptive. However I do not think this is a strong enough reason to keep this around, and the error you’d get should be easily traceable.

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -82,7 +82,6 @@ const Behavior = MarionetteObject.extend({
   },
 
   getUI(name) {
-    this.view._ensureViewIsIntact();
     return this._getUI(name);
   },
 
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -221,7 +221,7 @@ const CollectionView = Backbone.View.extend({
   // Render children views. Override this method to provide your own implementation of a
   // render function for the collection view.
   render() {
-    this._ensureViewIsIntact();
+    if (this._isDestroyed) { return this; }
     this.triggerMethod('before:render', this);
     this._renderChildren();
     this._isRendered = true;
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -85,7 +85,7 @@ const CompositeView = CollectionView.extend({
 
   // Renders the model and the collection.
   render() {
-    this._ensureViewIsIntact();
+    if (this._isDestroyed) { return this; }
     this._isRendering = true;
     this.resetChildViewContainer();
 
--- a/src/mixins/view.js
+++ b/src/mixins/view.js
@@ -7,10 +7,10 @@ import { triggerMethod } from '../common/trigger-method';
 import BehaviorsMixin from './behaviors';
 import CommonMixin from './common';
 import DelegateEntityEventsMixin from './delegate-entity-events';
+import DomMixin from './dom';
 import TriggersMixin from './triggers';
 import UIMixin from './ui';
 import MarionetteError from '../error';
-import DomMixin from './dom';
 
 // MixinOptions
 // - behaviors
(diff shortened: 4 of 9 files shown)


blikblum:
I support removing the error thus giving the developer the freedom to decide if the behavior is harmless or not. On the other hand, the concerns raised in other issues are valid that this can lead to hard to track bugs.

I suggest that instead of throw an error, send a warning to the console. This warning could be protected by a condition to check if is a production build or no, like is done in vuejs:  https://github.com/vuejs/vue/blob/dev/src/core/instance/lifecycle.js#L45

The idea is that, in the build step, process.env.NODE_ENV is set to a constant and according to its value will become a dead code that is removed by js minifier.

Vue distributes the non minified with process.env.NODE_ENV set to 'development' and the minified to 'production' (no checks) but any combination is possible.

paulfalgout:
I'm interested to hear use-cases and/or code samples for the hard to track bugs.

While I'm not against adding some sort of warning.. I can't see where this error (or warning) is providing value.  With the exception of situations like #2719 where a no-op is what you want, I don't what a user would be trying to do to trigger this error

scott-w:
I'm in favour of this general direction, however I'd be interested in knowing what happens when `getUI` is called on a destroyed view without a guard.

paulfalgout:
@scott-w I think the unguarded is the best option.
https://jsfiddle.net/akaqmrbu/1/

It is most consistent with what happens before a view is rendered (which I think is a _much_ more likely scenario)

Not to mention I think more people probably just use `.ui` that `.getUI`

The vanilla errors in all cases are pretty easily traceable and explanatory I'd think.

If _anything_ `getUI` should only care about if the view `isRendered` which it would _not_ be after it is destroyed.

scott-w:
Looks good, I think we should move to documenting `getUI` as the default way of doing everything going forward personally, but we can worry about that later.

Should we look to bring this in for Mn 3.2 or 4.x?

paulfalgout:
I do not personally believe this is a breaking change.

I'd be open to moving to `getUI`.. particularly so we could query the ui element at its first use: https://github.com/marionettejs/backbone.marionette/issues/2831

But I think we'd need to do it over time.. deprecate `ui` and push for the new API.. maybe for a `v5`?  I dunno.. but as you said.. should be its own issue.

blikblum:
When a dev calls render method in an event handler, the expectation is that it will be triggered when the view is still in a state to be rendered as in #1736. 

By itself, doing a noop when rendering a destroyed view would not harm but would hide the fact that the event is being triggered after the view is destroyed. This situation can be not desirable depending of the context, e.g., is interacting with view dom in the handler.

Is up to dev evaluate if this is a problem or not. With the warning will be easier to spot such conditions without stopping those that evaluated as not an issue

paulfalgout:
@marionettejs/marionette-core it looks like this could move forward and other concerns handled later?

Issue #3248, opened 2016-11-02
Clean up tests


paulfalgout:
Looking over the tests many of them use older terms (some even Marionette v1).  There are a few tests that were modified to match the current behavior, but aren't in fact testing anything anymore.  And in general the tests could be updated to be cleaner and more consistent.  

Further I believe most of the tests should be organized by class and then by either a feature of that class or an integration with a mixin or other class.  We have too many one-off test files that it is difficult to tell where to add something.

I'll be playing with organization and style of tests while writing new ones for https://github.com/marionettejs/backbone.marionette/pull/3244 while examining what is currently there.  I just want to be thinking about what improvements can be made.

Additionally there's a Readme file in the test directory that is completely irrelevant to our current setup. 😞

paulfalgout:
As mentioned above https://github.com/marionettejs/backbone.marionette/pull/3244/commits/5603b0eb551cd43e9daaff69e0dc8a7682e4369b

rafde:
I agree with this, I think a clean up outline is in order.
@denar90 is tackling Application
There are so many test, I don't know where to start.

denar90:
@rafde moreover, I'm working with behaviors also.

paulfalgout:
I think the best way to approach this work might be to start testing the things with the least amount of dependencies.  So I'd think we could make a utils directory (just have the test folder structure mirror the src) and make a unit test for each file... most likely by pulling from and updating older tests..  I think we can possibly ignore CollectionView, CompositeView, and AppRouter stuff for now.
We definitely won't want to get rid of anything until we know there's a test that covers the same case.

denar90:
Sounds really good 👍 . It will be easy to find test file if it be a mirror of src folder. In addition, starting with behaviors was just experiment how it can goes, importing stuff from src etc.

paulfalgout:
I think for large classes with lots of integrations we can break up the tests into multiple files, but they can be put in a folder that matches the file name.  live `view/view.spec.js` `view/view-ui.spec.js` `view/view-behaviors.spec.js` etc.

rafde:
Some of the specs could probably be renamed also, like https://github.com/rafde/backbone.marionette/blob/e887d160ea8df451f36bd8474b17edd59648b14e/test/unit/view.triggers.spec.js#L1-L1 is actually testing from `view.$el.trigger`

denar90:
SGTM

paulfalgout:
[...]

PR #3255, opened 2016-11-10
Cleanup Application and Behavior tests

denar90:
### Proposed changes
 - Cleanup Application and Behavior tests

Link to the issue: #3248 

I have several concerns/questions about some stuff.
1. During cleanup I found out that it's hard to get rid of `this.sinon` or other `this. ...` variables. Idea was to import it like either `Application` or `View` but it's not so easy. We have [setup.js](https://github.com/marionettejs/backbone.marionette/blob/v3.1.0/test/setup/setup.js) file where stuff for sinon set  - https://github.com/marionettejs/backbone.marionette/blob/v3.1.0/test/setup/setup.js#L47-L52
So `sinon` has to became exported or smth.
2. Since `Behaviors` are deprecated we can 
either remove some of this functionality from tests - https://github.com/marionettejs/backbone.marionette/compare/next...denar90:tests-cleanup?expand=1#diff-31ea63470331c37566b6cd93d65fa730L20
or keep it with some small change 
https://github.com/marionettejs/backbone.marionette/compare/next...denar90:tests-cleanup?expand=1#diff-31ea63470331c37566b6cd93d65fa730L53
3. Some suits private functionality was removed, like https://github.com/marionettejs/backbone.marionette/compare/next...denar90:tests-cleanup?expand=1#diff-31ea63470331c37566b6cd93d65fa730L177, which cost coverage decreasing.
4. Test libraries are not included into build file for browser testing for example. Maybe we should make them part of the build.
Update:
5. Another this is mixing importing stuff and use global classes like Backbone or Marionette. Here good example https://github.com/marionettejs/backbone.marionette/pull/3255/commits/e5162ada55b691ceda6239896078507e1bcef3ae
There are `Behavior` importing at the top of the file, but test failed so `Marionette.Behavior` was used instead... Same for `Marionette.Behaviors.behaviorsLookup`...

### DIFF
--- a/test/unit/README.md
+++ b/test/unit/README.md
@@ -3,40 +3,105 @@
 
 ### Running unit tests
 
-1. via the command line
+1. Running just unit tests - `npm run test`
 
+2. Running coverage reporter - `npm run coverage`. 
+To check coverage, open `./coverage/lcov-report/index.html`.
+
+3. Running tests in browser - `test-browser`.
+
+
+### Common concepts for writing tests
+
+1. Test suites should cover only public API.
+
+2. Code style.
+
+- Each `describe` should have name of tested method. 
+
+> There are should be not more then 1 nested describe for on method.
+ 
+Wrong way.
+ 
+ ```javascript
+ 
+ describe('some events in myMethod', function() {
+   describe('some logic', function() {
+     it('do something', function() {
+      ...  
+     })
+     
+     describe('some other logic', function() {
+       etc...
+     })
+   })
+ })
 ```
-grunt test
+
+
+Wright way.
+
+
+```javascript
+ 
+  describe('#myMethod', function() {
+    describe('when some logic', function() {
+      it('should do something', function() {
+        ...  
+      })
+    })
+   
+    describe('when some other logic', function() {
+      it('should do something', function() {
+        ...  
+      })
+    })
+ })
 ```
 
-2. in the browser
-```
-open SpecRunner.html
-```
 
+- `before/beforeEach` should consist only some preparation logic 
+but not call methods you expect to test.
 
-### Adding a new test file
+Wrong way.
 
-New test files are added as a script tag to the SpecRunner.html
-
-```html
-<script src="test/unit/application.app-regions.spec.js"></script>
-<script src="test/unit/application.spec.js"></script>
-```
-
-
-### Adding a new source file
-
-When you want to add a new source file to be tested, the file needs to be added in two places
-
-1. SpecRunner.html
-
-```html
-<script src="src/item-view.js"></script>
+```javascript
+  describe('#myMethod', function() {
+    let myInstance;
+    
+    beforeEach(function() {
+      myInstance = new MyClass({
+        render: this.sinon.spy
+      });
+      myInstance.render();
+    })
+   
+    it('should do something', function() {
+      expect(myInstance.render).to.have.been.called;
+    })
+ })
 ```
 
-2. unit/setup/node.js file
+Wright way.
 
-```js
-requireHelper('composite-view');
+```javascript
+  describe('#myMethod', function() {
+    let myInstance;
+    let renderSpy;
+    
+    beforeEach(function() {
+      renderSpy = this.sinon.spy();
+      
+      myInstance = new MyClass({
+        render: renderSpy
+      });
+    })
+   
+    it('should do something', function() {
+      myInstance.render();
+      
+      expect(renderSpy).to.have.been.called;
+    })
+ })
 ```
+
--- a/test/unit/README.md
+++ b/test/unit/README.md
@@ -8,100 +8,124 @@
 2. Running coverage reporter - `npm run coverage`. 
 To check coverage, open `./coverage/lcov-report/index.html`.
 
-3. Running tests in browser - `test-browser`.
+3. Running tests in browser - `npm run test-browser`.
 
 
 ### Common concepts for writing tests
 
-1. Test suites should cover only public API.
+1. Test suites should cover public API.
+
+> In most cases it will be public API testing,
+but sometimes we should test something like: When models added to collection,
+hence, in this case we are adding needed suites.
 
 2. Code style.
 
-- Each `describe` should have name of tested method. 
+- Each `describe` should have name of tested method.
 
-> There are should be not more then 1 nested describe for on method.
+> If it's possible, there should be not more then one nested describe for one method.
  
-Wrong way.
+_Wrong way_
  
- ```javascript
- 
- describe('some events in myMethod', function() {
-   describe('some logic', function() {
-     it('do something', function() {
-      ...  
-     })
-     
-     describe('some other logic', function() {
-       etc...
-     })
-   })
- })
+```javascript
+  describe('#MyClass', function() {
+    describe('some events in myMethod', function() {
+      describe('some logic', function() {
+        it('do something', function() {
+          ...  
+        });
+         
+        describe('some other logic', function() {
+          // other nested describes
+        });
+      });
+    });
+  });
 ```
 
 
-Wright way.
-
+**Correct way**
 
 ```javascript
- 
-  describe('#myMethod', function() {
-    describe('when some logic', function() {
-      it('should do something', function() {
-        ...  
-      })
-    })
+  describe('#MyClass', function() {
+    describe('#myMethod', function() {
+      describe('when some logic', function() {
+        it('should do something', function() {
+          ...  
+        });
+      });
    
-    describe('when some other logic', function() {
-      it('should do something', function() {
-        ...  
-      })
-    })
- })
+      describe('when some other logic', function() {
+        it('should do something', function() {
+          ...  
+        });
+      });
+    });
+  });
 ```
 
+- In case of testing some behavior
+
+> behavior means not Marionette Behavior class
+
+**Correct way**
+
+```javascript
+  describe('#MyClass', function() {
+    describe('when some data was changed', function() {
+      it('should do something', function() {
+        ...
+      });
+    });
+  });
+```
 
 - `before/beforeEach` should consist only some preparation logic 
-but not call methods you expect to test.
+but inside it should not present calling methods you expect to test.
 
-Wrong way.
-
-```javascript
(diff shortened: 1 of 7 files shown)


paulfalgout:
1.  I think the global test libraries might be ok.. not sure of a better way to do it right now.
2. I definitely think we should keep tests for deprecated code
3. Removing testing of private stuff should not make it impossible to get coverage.  We should be able to achieve coverage by testing the public API.. otherwise that code is not in use.
4. I could go either way here.. seems similar to 1.)
5. I think this is just testing the constructor which could be done from `Behavior` correct?

I also think the behaviors spec should be two files.  One called behaviors in the mixin directory and one for testing a behavior class.  For the behavior mixin although all of the functions are denoted private with the `_` they're internally public.  private to marionette, but public for this module.  The only thing that looks private to the mixin is `this._behaviors`  So if we can avoid touching that directly, we'll be good I think.

paulfalgout:
FWIW I think it's also easier to write tests for public marionette using behavior descriptions `when you do something...`  where-as for our internal-only "public" APIs we can just reference the function name like: `describe('#_someMethod', ...`  Thoughts?

denar90:
I think I'm done here.
Need a deep look at it from you guys. 
From my side I can tell that file `behaviors.spec.js` was already present just added some small changes without changing core of the tests.
https://github.com/marionettejs/backbone.marionette/pull/3255/files#diff-31ea63470331c37566b6cd93d65fa730
But `application.spec.js` was rewritten, `behaviors.spec.js` created from scratch. 

Moreover `behaviors.spec.js` can be left for now as it is and later divided on integration and API tests. Right now its a mix. Thoughts?

scott-w:
It all looks decent from what I can see. I'm with @paulfalgout in that we should be able to test all `_doSomething` methods from just public methods. If we can't execute a code branch from a public method somehow then it can become a candidate for removal.

denar90:
@scott-w I tried to do that for `Behaviors` mixin and`Application`, but Implementing this approach for `Behavior` itself will be too much for reviewing this PR I think. Guys if you want me to do it for `Behaviors` I can do that :)

scott-w:
I'm happy to :+1: this as is. It's still a big step in the right direction.

test/unit/behaviors.spec.js
```diff
@@ -16,390 +24,480 @@ describe('Behaviors', function() {
...
 
     it('should call initialize when a behavior is created', function() {
-      expect(this.initializeStub).to.have.been.calledOnce.and.calledWith(this.behaviorOptions, this.view);
+      expect(initializeStub).to.have.been.calledOnce.and.calledWith(behaviorOptions, fooView);
     });
 
-    it('should set _behaviors', function() {
-      expect(this.view._behaviors.length).to.be.equal(1);
+    afterEach(function() {
+      Marionette.Behaviors.behaviorsLookup.restore();
```
paulfalgout: nice.  cleaning up.

test/unit/behaviors.spec.js
```diff
@@ -410,509 +508,600 @@ describe('Behaviors', function() {
...
+          expect(onBarClickStub).to.have.been.calledOnce.and.calledOn(fooBehavior);
         });
       });
     });
 
     describe('within a layout', function() {
       beforeEach(function() {
         this.setFixtures('<div id="layout"></div>');
-        this.ItemView = Marionette.View.extend({
+        let BarView = View.extend({
```
paulfalgout: This might just be personal preference, and I have seen arguments either way, but I do think the way Marionette is currently doing things is that if it can be a `const` it should be.  And this and `barView` below (I think) should be `const`.

There are other examples of using `let` when it could be `const`  I don't know how important, if at all, it is that it is changed.

test/unit/behaviors.spec.js
```diff
@@ -410,509 +508,600 @@ describe('Behaviors', function() {
...
     });
 
     it('should still be bound to "on" on destroy', function() {
-      expect(this.onFooStub).to.have.been.calledOnce;
+      expect(onFooStub).to.have.been.calledOnce;
     });
   });
 
   describe('behavior model events', function() {
+    let handleModelChangeStub;
```
paulfalgout: I don't think there's anything to change yet, but I'm trying to determine patterns..  and I'm starting to suspect that if there are large blocks of `let`s in a row that the tests should be split up.  But I'm not yet convinced this is true... just sharing my initial thoughts.

test/unit/mixins/behaviors.spec.js
```diff
@@ -0,0 +1,466 @@
...
+
+    it('should invoke unbindUIElements', function() {
+      behaviorsInstance._unbindBehaviorUIElements();
+
+      expect(FooBehavior.prototype.unbindUIElements).to.have.been.calledOnce;
+      expect(BarBehavior.prototype.unbindUIElements).to.have.been.calledOnce;
+    });
+  });
+
+  //@todo
```
paulfalgout: Do you want to get this merged with this one incomplete?  I'm cool with that, just wondering.

test/unit/application.spec.js
```diff
@@ -1,74 +1,219 @@
-describe('marionette application', function() {
-  'use strict';
+'use strict';
+
+import Application from '../../src/application';
+import View from '../../src/view';
+
+describe('Marionette Application', function() {
+
+  describe('#_initialize', () => {
```
denar90: Make sense, I just misunderstood `#_methodName` pattern of `describe` naming.

test/unit/application.spec.js
```diff
@@ -1,74 +1,219 @@
...
+      let fooArgs;
+
+      beforeEach(function() {
+        fooArgs = ['foo', 'bar'];
+      });
+
+      it('should call show method in region with additional arguments', function() {
+        app.showView(view, fooArgs);
+
+        expect(showViewInRegionSpy).to.have.been.called;
```
denar90: Totally agree, it isn't. I cheated here living just expecting that `region.show` will be just called. Sorry, I couldn't stub `region.show` so I had to leave some comment or smth. I'll try to make it works.

test/unit/behaviors.spec.js
```diff
@@ -16,390 +24,480 @@ describe('Behaviors', function() {
   });
 
   describe('behavior parsing with a functional behavior lookup', function() {
```
denar90: We can move this test from here into `config` folder, just for testing `behaviorsLookup` function. Thoughts?

test/unit/behaviors.spec.js
denar90: Thanks, but looking at it right now I think it's not need :) because of 
https://github.com/marionettejs/backbone.marionette/blob/master/test/setup/setup.js#L48
https://github.com/marionettejs/backbone.marionette/blob/master/test/setup/setup.js#L52

test/unit/behaviors.spec.js
denar90: This test is too big, I'll try to split it up. One more think, I'm still doubt about
```
let foo;
let bar;
let baz;
```
and
```
let foo, bar, baz;
```

test/unit/mixins/behaviors.spec.js
denar90: I just have no idea how to stub `triggerMethod` which was just imported, so I lived todo comment for that. Sinon doesn't have possibility for subbing functions. Maybe you know more?

denar90:
@paulfalgout thank you for so grounded code review and no problem about comments at all. All we want to do our best and comments during review help us to achieve that.

test/unit/README.md
```diff
@@ -3,40 +3,105 @@
...
+       etc...
+     })
+   })
+ })
+```
 
-### Adding a new source file
 
-When you want to add a new source file to be tested, the file needs to be added in two places
+Wright way.
```
scott-w: Right*

You could use "Correct way." instead

test/unit/README.md
```diff
@@ -3,40 +3,105 @@
...
 
+- Each `describe` should have name of tested method. 
+
+> There are should be not more then 1 nested describe for on method.
+ 
+Wrong way.
+ 
+ ```javascript
+ 
+ describe('some events in myMethod', function() {
```
paulfalgout: So there will be other describes that are necessary.  I'm not sure we can limit it to API functions.  Stuff like "when models are added to the collection"  However I agree it makes sense in most cases that we should try.

test/unit/README.md
```diff
@@ -3,40 +3,105 @@
...
 
-```html
-<script src="test/unit/application.app-regions.spec.js"></script>
-<script src="test/unit/application.spec.js"></script>
-```
+2. Code style.
 
+- Each `describe` should have name of tested method. 
+
+> There are should be not more then 1 nested describe for on method.
```
paulfalgout:
```diff
- There are should
+ There should
```
I'm ok with starting here, but I do wonder if at some point we'll need to allow for a third nest.  I'd like to think not, but there might be the case for one somewhere to keep from duplication overkill.  But I very much agree that shallow nesting is the goal.

test/unit/README.md
paulfalgout: Just a note that this is true in multiple places in this file.

test/unit/README.md
```diff
@@ -3,40 +3,105 @@
...
+      
+      myInstance = new MyClass({
+        render: renderSpy
+      });
+    })
+   
+    it('should do something', function() {
+      myInstance.render();
+      
+      expect(renderSpy).to.have.been.called;
```
paulfalgout: Can we change this to `calledOnce`?

test/unit/README.md
```diff
@@ -3,40 +3,105 @@
...
+    })
+ })
```
paulfalgout: sooo one other nit-pick style thing.. none of the closing `})`s have `;` and they probably should to match our specs.

scott-w: Does `eslint` cover the test directory?

paulfalgout: It does but this is markdown.

denar90:
@scott-w @paulfalgout I updated doc, there also might present some spelling errors. Don't hesitate to point on it 😃

denar90:
💃  thanks guys 👍

Issue #3254, opened 2016-11-10
Document DOM Mixin API


rafde:
### Document DOM Mixin functions available

 1. DOM mixin API awareness, why, where mixed in, and how to override.

https://github.com/marionettejs/backbone.marionette/blob/91c26af7d70527b9a6dcf8112f008e9c1bf275c5/src/mixins/dom.js

PR #3256, opened 2016-11-12
Dom API Docs

scott-w:
### Proposed changes
 - Add the documentation for DOM Mixin API

Link to the issue: #3254 

I'm not sure I've got the Howto on mixing in the new API so I'd like some feedback on that.


scott-w:
Opened this against `master` instead of `next`. I'll close and re-open

PR #3257, opened 2016-11-12
Dom API Docs

scott-w:
### Proposed changes
 - Add the documentation for DOM Mixin API

Link to the issue: #3254 

I'm not sure I've got the Howto on mixing in the new API so I'd like some feedback on that.

### DIFF
--- a/docs/dom.mixin.md
+++ b/docs/dom.mixin.md
@@ -60,7 +60,6 @@ To implement your own DOM API for `View`, override the provided functions to
 provide the same functionality provided and mix it in using `_.extend` as such:
 
 ```js
-var _ = require('underscore');
 var Mn = require('backbone.marionette');
 
 var MyDOMMixin = require('./mixins/mydom');
--- a/docs/dom.mixin.md
+++ b/docs/dom.mixin.md
@@ -12,7 +12,7 @@ defines the methods that actually attach and remove views and children.
 
 The default mixin depends on Backbone's jQuery `$` object however it does not
 rely on jQuery-specific behavior. This should make it easier to develop your own
-API.
+API. You will, however, [need to also override Backbone's jQuery integration](#backbone-jquery-integration).
 
 ### `createBuffer()`
 
@@ -49,10 +49,12 @@ Remove `el` from the DOM.
 
 ### `findEls(selector, context)`
 
-Lookup the `selector` string within the DOM node for `context`. The `context`
-argument should be a DOM node reference.
+Lookup the `selector` string within the DOM node for `context`. The optional
+`context` argument will come in as a DOM Node reference to run the `selector`
+search. If `context` hasn't been set, then `findEls` should search the entire
+`document` for the `selector`.
 
-## Providing Your Own DOM
+## Providing Your Own DOM API
 
 To implement your own DOM API for `View`, override the provided functions to
 provide the same functionality provided and mix it in using `_.extend` as such:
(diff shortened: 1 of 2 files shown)


docs/dom.mixin.md
```diff
@@ -0,0 +1,67 @@
...
+### `removeEl(el)`
+
+Remove `el` from the DOM.
+
+### `findEls(selector, context)`
+
+Lookup the `selector` string within the DOM node for `context`. The `context`
+argument should be a DOM node reference.
+
+## Providing Your Own DOM
```
rafde: I don't know if we should be suggesting anything yet.
There was discussion on how to provide an easy way to do the mixing in https://github.com/marionettejs/backbone.marionette/issues/3145#issuecomment-243033085

I am almost hesitant to do it that way, but in most cases, people want a quick plug for this stuff and in advance topics, we could provide an more involved solution.

docs/dom.mixin.md
```diff
@@ -0,0 +1,67 @@
...
+
+```js
+var _ = require('underscore');
+var Mn = require('backbone.marionette');
+
+var MyDOMMixin = require('./mixins/mydom');
+
+_.extend(Mn.View, MyDOMMixin);
+
+module.exports = Mn.View;
```
rafde: `module.exports = Mn.View.extend(MyDOMMixin);`

but if you don't prefer that, maybe.

`_.extend(Mn.View.prototype, MyDOMMixin);`

docs/dom.mixin.md
```diff
@@ -0,0 +1,67 @@
...
+Replace the contents of `el` with the HTML string of `html`. Unlike other DOM
+functions, this takes a literal string for its second argument.
+
+### `removeEl(el)`
+
+Remove `el` from the DOM.
+
+### `findEls(selector, context)`
+
+Lookup the `selector` string within the DOM node for `context`. The `context`
```
rafde: The `context` is an optional argument that should be a DOM node reference. Otherwise, it should default to `document.body`.

paulfalgout:
Any chance @marionettejs/marionette-core would be up for introducing this as an experimental feature?  I feel a little bit like it would be great to play with this in the real world before fully committing to a API.

scott-w:
Yeah, I think that's a good place to start with this so we can adapt it as we get use-cases in.

docs/dom.mixin.md
```diff
@@ -0,0 +1,67 @@
...
+jQuery and integrate with the DOM using a custom mixin.
+
+## Mixin Methods
+
+The DOM mixin manages the DOM on behalf of `View` and `CollectionView`. It
+defines the methods that actually attach and remove views and children.
+
+The default mixin depends on Backbone's jQuery `$` object however it does not
+rely on jQuery-specific behavior. This should make it easier to develop your own
+API.
```
scott-w: I think it would be, I'll add it in now.

docs/dom.mixin.md
```diff
@@ -0,0 +1,81 @@
...
+search. If `context` hasn't been set, then `findEls` should search the entire
+`document` for the `selector`.
+
+## Providing Your Own DOM API
+
+To implement your own DOM API for `View`, override the provided functions to
+provide the same functionality provided and mix it in using `_.extend` as such:
+
+```js
+var _ = require('underscore');
```
denar90: I think we should rather not to require `underscore` here or do smth like
```
module.exports = _.extend(View.prototype, MyDOMMixin);
```

blikblum:
Just found this doc: https://github.com/jashkenas/backbone/wiki/using-backbone-without-jquery 
I think it should be mentioned

Must be clarified also differences/need between overriding `view._removeElement` and `domMixin.removeEl`.

I understood, by the tests, that is not necessary to override view._removeElement if using only Mn.View but necessary if using vanilla Backbone.View.

paulfalgout:
👍

rafde:
:+1:

Issue #3208, opened 2016-09-28
CollectionView ui hash


paulfalgout:
So CollectionView… it’s a Backbone.View.. so it has `events` by default.. which is a little odd, but I guess you can add an `events` hash that applied to all children.. you can also then of course do this with `triggers`.. _however_ we don’t ever bind the `ui` hash of a `CollectionView` 

despite never doing it (and we never have) all of the code for that is in there.. you can manually call `bindUIElements`.. but even if we _did_ call it on `render` we definitely don’t do any work to maintain those bindings as we filter or add children or that kind of thing.  And I really doubt we want to.

And so CollectionView also has behaviors.. but again.. no ui bindings..  

I don’t really know if ui elements on a CollectionView are a great idea.. and previously it was decided against: https://github.com/marionettejs/backbone.marionette/issues/584   _however_ if we don’t do ui.. but keep triggers and behaviors we should document the caveat..  but then should we have `triggers` without `ui`?  Doesn’t make much sense except that we can't get rid of `events`...  

This came up again and was nearly implemented here: https://github.com/marionettejs/backbone.marionette/pull/1507

Now I pretty much agree that `ui` doesn't make any sense on a `CollectionView`.. but then I wonder if we should pull it out of the class.  And if `ui`... `triggers` too?  The same logic for the reasons not to support `ui` apply to `triggers`.  But then _behaviors_?  I can still see reasons to support behaviors, but they will only partially work.

I'm mainly opening this issue to ask: 
**How do we document this currently.**
**What can/should we remove or implement?**

If we are removing it, it isn't currently documented in v3 that it should work.. should we consider it breaking?  To some degree it doesn't really hurt much to have it in there. except that it is unsupported, partially broken functionality.

JSteunou:
Looking at my code base, I would only keep behaviors to CollectionView

scott-w:
From my perspective, it could be safe to consider it breaking. Part of the reason I removed chunks of documentation was to give us a chance to remove things after v3 was released. In this case, the feature itself was never documented at all. 

I agree with @JSteunou that Behaviors could be useful in CollectionView (and I'm sure I make use of them somewhere).

denar90:
Talking about behaviors. I haven't faced situation of using it with CollectionView (CV). 
@JSteunou in most cases CV has parent View, is there are any case of using behaviors in scope of it instead of using behaviors in CV?
[...]

PR #3258, opened 2016-11-13
Document bindUIElements

scott-w:
### Proposed changes
 - Add documentation for binding `ui`

Link to the issue: #3208

I've opened this against `next` though I can reopen against `master` if we'd rather put it there.

### DIFF
--- a/docs/marionette.collectionviewadvanced.md
+++ b/docs/marionette.collectionviewadvanced.md
@@ -424,11 +424,10 @@ var MyCollectionView = Mn.CollectionView.extend({
 
 ## Binding `ui`
 
-By default, `CollectionView` will not bind anything defined in `ui`. As it has
-no direct `template` of its own to manage, this isn't a particular issue.
-There may be instances where you want to bind the `ui` element to listen to
-children on the collection directly. For example, to bind `triggers` or
-act on `behaviors`.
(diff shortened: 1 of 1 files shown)


docs/marionette.collectionviewadvanced.md
```diff
@@ -419,3 +421,32 @@ var MyCollectionView = Mn.CollectionView.extend({
   }
 });
 ```
+
+## Binding `ui`
+
+By default, `CollectionView` will not bind anything defined in `ui`. As it has
+no direct `template` of its own to manage, this isn't a particular issue.
+There may be instances where you want to bind the `ui` element to listen to
+children on the collection directly. For example, to bind `triggers` or
```
paulfalgout: oddly enough `triggers` works without binding.  This is because it uses the text selectors to do the `@ui` stuff regardless if the ui has been bound.

https://jsfiddle.net/784y75ks/1/

This may only be for the `ui` hash and `getUI`

scott-w: Fair enough, though I'm not sure we should guarantee that behaviour in future development.

paulfalgout: Cool with leaving it out.. but in either case `bindUIElements` doesn't affect anything that uses `@ui.` anyhow.  The only thing it does is bind the ui hash.

Issue #2967, opened 2016-04-03
dom events leaking from child view to parent view


paulovieira:
When using nested views with events in both the parent and child, it might happen that a dom event triggered in the child view will call also call the event handler in the parent (besides calling the handler in the child, as it should).

This might happen if the selectors used in the `events` hash are the same in the parent and in child. Below is a simple example reproducing the situation (jsfiddle here: https://jsfiddle.net/paulovieira/kmt8kdmy/). 

If you click the save button in the child view, you see that the handler from the parent is called as well, contrary to my expectation (the child view shouldn't be aware that there is a parent)

```js
var Child = Mn.ItemView.extend({

    attributes: {
        style: "border: 1px solid red; margin: 5px;"
    },
    
    template: _.template(`
        <p>child view</p>
        <button class="btn-save">save</button>
    `),

    ui: {
        saveButtonChild: "button.btn-save"
    },

    events: {
        "click @ui.saveButtonChild": function(e){
            alert("this is the child view handler!");
        }
    },

});

var Parent = Mn.LayoutView.extend({

    attributes: {
        style: "border: 1px solid blue; padding: 5px;"
    },
    
    template: _.template(`
        <p>parent view</p>
        <button class="btn-save">save</button>
      
        <div class="some-region"></div>
    `),

    ui: {
        saveButtonParent: " button.btn-save",
        someRegion: "div.some-region"
    },

    events: {
        "click @ui.saveButtonParent": function(e){
            alert("this is the parent view handler!");
        }
    },

    onBeforeAttach: function(){
        this.getRegion("someRegion").show(new Child);
    },

    regions: {
        someRegion: '@ui.someRegion'
    }
});

var mainRegion = new Mn.Region({ el: "div#root-region"});
mainRegion.show(new Parent);
```
This is not a bug. It's simply a consequence of the way backbone uses event delegation in jquery, and the fact that the selectors are the same. There's an informative blog post in the marionette blog about this subject: http://blog.marionettejs.com/2015/02/12/understanding-the-event-hash/index.html

But I was wondering how have been people dealing with this issue. In a situation of deep nested views it might become problematic. Child views should not care about what is happening above them. If I return to some marionette code 1 year later to add a new child view, I shouldn't have to review the details in the parent to make sure the events are not clashing. Do you agree?

Having to add specific dummy classes or ids everywhere ("js-child-view-person", etc) just to avoid these clashes doesn't feel right. 

[...]

PR #3259, opened 2016-11-14
Add templateAsEl option to View

### LINKED ISSUE #2357: Use template as el

paulfalgout:
Resolves #2357 

Adds `templateAsEl` option to View

This allows the user to bypass the View’s `el`, `tagName`, `className`, `attributes` and use the template as the `el`.

Note that re-rendering will re-delegate events which is much more expensive than using the view’s `el`.

https://jsfiddle.net/9z76s1p1/2/

I'm not sure how much I really like this feature.  It seems like it will have unwelcomed pref issues for little value.  

What could be further addressed is delegating events on the document at which point https://github.com/marionettejs/backbone.marionette/issues/2967 will only get worse if not handled correctly.  I've seen mixed opinions on globally delegated events and performance.

In any case, here it is to play with.  If there is interested in seeing this to completion it'll need specs and docs.  I'd also love to see perf tests to fully understand the hit calling delegateEvents for each render will have.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -20,10 +20,13 @@ const ClassOptions = [
   'regionClass',
   'regions',
   'template',
+  'templateAsEl',
   'templateContext',
   'triggers',
   'ui'
 ];
+
+const TEMPLATE_AS_EL = 'TEMPLATE_AS_EL';
 
 // The standard view. Includes view events, automatic rendering
 // of Underscore templates, nested views, and more.
@@ -35,6 +38,10 @@ const View = Backbone.View.extend({
     this._setOptions(options);
 
     this.mergeOptions(options, ClassOptions);
+
+    if (this.templateAsEl) {
+      this.el = TEMPLATE_AS_EL;
+    }
 
     monitorViewEvents(this);
 
@@ -86,7 +93,9 @@ const View = Backbone.View.extend({
   // Overriding Backbone.View's `setElement` to handle
   // if an el was previously defined. If so, the view might be
   // rendered or attached on setElement.
-  setElement() {
+  setElement(el) {
+    if (el === TEMPLATE_AS_EL) { return this; }
+
     const hasEl = !!this.el;
 
     Backbone.View.prototype.setElement.apply(this, arguments);
@@ -145,6 +154,12 @@ const View = Backbone.View.extend({
 
     // Render and add to el
     const html = Renderer.render(template, data, this);
+
+    if (this.templateAsEl) {
+      this.setElement(html);
+      return;
+    }
+
     this.attachElContent(html);
   },
 


rafde:
How will this work with the template produces nodes that aren't contained, example:
```html
<span>hi</span>
<span>there</span>
```

paulfalgout:
Only the first node of the template is used.

rafde:
Most people are not going to expect. I am leaning on the no side to this because of the complications people will expect this to solve.
Too bad DOM API doesn't provide something like DOM array that you could bind events to without being part of the DOM.

paulfalgout:
Yep.. Besides overhauling backbone entirely I don't see this working with multiple nodes..  the `getEl` in Backbone uses `$(el)[0]`.  I think trying support multiple nodes is a bad idea anyhow.

I've never been a big fan of this feature... that said I think this might be it's best shot?  I dunno if there's anyone around who's still interested in seeing this.  In the very least this and the issue could be closed and reopened later if someone needs it.

src/view.js
```diff
@@ -20,11 +20,14 @@ const ClassOptions = [
   'regionClass',
   'regions',
   'template',
+  'templateAsEl',
   'templateContext',
   'triggers',
   'ui'
 ];
 
+const TEMPLATE_AS_EL = 'TEMPLATE_AS_EL';
```
paulfalgout: if for some reason this PR does go through, we should look into changing this into a `Symbol()`

blikblum:
The way i thought to overcome this limitation was with the aid of tooling. With a tool like [vue-loader](https://github.com/vuejs/vue-loader) is possible to design the markup with a root el and the view class would be configured with the according properties.

It has some benefits:
 - One file with markup and code
 - No changes in core Mn and Backbone
 - Allow to use CSS modules / scoped CSS

and drawbacks:
 - One file with markup and code
 - Only can export a view per file
 - Tied to webpack
 - Another dev tool

howardroark:
I like this @paulfalgout !  

Choosing the first element in the template makes sense... React also chooses the first element if more than one is present.  I think people get that a View is expecting a single element to work with. Maybe at first 50% assume a wrapper element is created, and the other 50% assume it grabs the first.

In terms of performance issue...  do you think it would compare at all to DOM repaints?

paulfalgout:
I don't think it compares to DOM repaints, but I don't know that this helps with DOM repaints at all either.. so at the moment it only increase perf issues... I think.  How much, I dunno.. would love someone to measure this somehow.

howardroark:
I'm thinking that dom repaints will most likely be the thing that effects the "perceived performance" when rendering a lot.  Not really directly related... but just kind of pointing out dealing with repaints will indirectly deal with this.  :P

scott-w:
Regarding the multiple root nodes: is it relatively straightforward to detect this and raise an error?

paulfalgout:
@scott-w FWIW Backbone and Marionette already has this restriction without errors.  The function causing the issue is in Backbone:
```js
_setElement: function(el) {
  this.$el = el instanceof Backbone.$ ? el : Backbone.$(el);
  this.el = this.$el[0];  // <- Selects only the first element
},
```
It would not be impossible to handle multiple `el`s I suppose, but it's definitely something you could run into now.

```html
<ul>
  <li>This'll be the view</li>
  <li>This won't</li>
  <li>Or this</li>
</ul>
<script>
const MyView = Marionette.View.extend({
  $el: $('li') 
});

const view = new MyView();

view.$el.text(); // This'll be the view
</script>
```

paulfalgout:
regarding multiple nodes.. this could always fallback to using the view `el` if the template's `length !== 1`

scott-w:
:+1: on that

JSteunou:
I did not see this one. I like it a lot, working on bel + morphdom I had to put some work to bypass view.el and use only DOM from template string.

paulfalgout:
Ok so I believe this can be done pretty easily using the `setRenderer` interface.  It'll be even easier in v4 when `attachElContent` doesn't get called if the renderer doesn't produce html..  Not sure we need to worry about this as an API option.

Issue #3128, opened 2016-08-16
Allow application with pre-generated html


dwt:
### Allow application with pre-generated html
1. We want to create an application where the initial view html is pre-generated on the server
2. This is relatively easy to achieve with Mn.View as it allows {template: false} to prevent it from trying to render anything
3. However, there seems to be no equivalent of this for regions. That is, it's not possible to call `app.showView(someView, {dontTouchTheFrickingHTML: true})`
### Expected behavior

I want a way to tell the application (really the region) that it shouldn't empty the element it's about to attach a view to.
### Actual behavior

It always empties it. Even with {preventDestroy: true} which seems to be the only relevant option that is available.
### Environment
1. Marionette version: 3.0a5
2. Backbone version: 1.3.3

scott-w:
Just so I understand, the definition of `someView` has `template: false` on it?

dwt:
of course

scott-w:
That sounds reasonable. I wonder if the best behaviour is to have the value of `template` tell the region manager whether to empty the `el` or not.

paulfalgout:
I think the way to accomplish this is to attach the view to the DOM element.  That will mark that view as "rendered" then when you "show" it in the region it won't attempt to re-render it.

paulfalgout:
oh actually I see the issue.. this is a recent bugfix..  it's not the view rendering.. it's the region empty on show.. ugh..
https://github.com/marionettejs/backbone.marionette/blob/next/src/region.js#L210
This is a fix for if there was dom inside a region but no view.. the new view was just appended to the existing dom..

so.. somehow the region needs to know that the view it is being handled is currently a child of its `$el` or something like that.. hmmm

paulfalgout:
Actually no it works as I described:
https://jsfiddle.net/gzyew8Lw/1/

While the region's el does get detached.. it gets reattached when the view is shown.. 

Is this what you're looking to accomplish?

dwt:
Indeed, setting the el does allow showView to be called without the existing dom being deleted.

Is this something that should be documented? Or should it be possible to tell a region that it shouldn't empty the existing dom?

I know I wasn't finding this solution from my trial and error and reading the docs + code, so maybe some documentation would be the solution?

What do you think?

paulfalgout:
Yeah I think documentation is in order..  not emptying the region could start all kinds of issues.

rafde:
This reads as if @dwt wants to treat the region as a collection view like. 
[...]

PR #3260, opened 2016-11-15
Don't detach pre-existing HTML when view's el is already in the region

### LINKED ISSUE #3227: Don't detach pre-existing HTML when view's el is already in the region
Related to https://github.com/marionettejs/backbone.marionette/issues/3128

When a view is attached to pre-existing DOM and then shown in a region, the view's el may already be inside of the region, but currently the region is unaware of this. So to allow for pre-generated HTML we detach the region HTML https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L212
Which is then immediately re-attached in the region's `show`.

Functionally this works, but this is likely causing extra paints and it is detaching/attaching a view without throwing events, so anything setup externally to this HTML will break with no use-able hooks to handle the situation.

We could solve this in a couple of ways.  The cheapest way is for a region to _not_ empty itself if it doesn't have a view but the view being passed in to show is already attached..  This way assumes that the view's el is already in the region.  **Can anyone think of any other reason an already attached view should be shown in a region?**  Attaching a view to 2 regions seems like a horribly bad idea.  But we could take it a step further to ensure the region contains the view el with something like `Backbone.$.contains(this.el, view.el);`   But then what are we protecting against?

Conversely `_attachView` would need to bail out if the view was already attached.  This pattern would I think also be applied to `CollectionView`  This makes attach and render much more similar so I think I like it.

paulfalgout:
Resolves #3227

This PR also moves some of `Region.show` into a `_setupChildView` function which mirrors what is happening in a CollectionView currently and helps reduce the length of the `show` function which grew to handle the should empty check.  Additionally moved the `currentView = view` to `show` to make the attach functions between CollectionView and Region more similar.

In the specs, there was a failing test regarding “when setting the "replaceElement" class option”.  I believe this test was failing because it was reusing `this.view` which wasn’t cleaned up in a prior test.  So effectively `this.view` was currently already shown in another region.  I switched this to `this.view1` which is instantiated as part of this `describe` block (or closer at least) and should be better for these tests.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -57,23 +57,32 @@ const Region = MarionetteObject.extend({
 
     this.triggerMethod('before:show', this, view, options);
 
-    monitorViewEvents(view);
+    // Assume an attached view is already in the region for pre-existing DOM
+    if (!view._isAttached) {
+      this.empty(options);
+    }
 
-    this.empty(options);
-
-    // We need to listen for if a view is destroyed in a way other than through the region.
-    // If this happens we need to remove the reference to the currentView since once a view
-    // has been destroyed we can not reuse it.
-    view.on('destroy', this._empty, this);
-
-    this._proxyChildViewEvents(view);
+    this._setupChildView(view);
 
     this._renderView(view);
 
     this._attachView(view, options);
 
+    this.currentView = view;
+
     this.triggerMethod('show', this, view, options);
     return this;
+  },
+
+  _setupChildView(view) {
+    monitorViewEvents(view);
+
+    this._proxyChildViewEvents(view);
+
+    // We need to listen for if a view is destroyed in a way other than through the region.
+    // If this happens we need to remove the reference to the currentView since once a view
+    // has been destroyed we can not reuse it.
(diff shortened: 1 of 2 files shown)


src/region.js
```diff
@@ -57,25 +57,34 @@ const Region = MarionetteObject.extend({
...
-    monitorViewEvents(view);
-
-    this.empty(options);
-
-    // We need to listen for if a view is destroyed in a way other than through the region.
-    // If this happens we need to remove the reference to the currentView since once a view
-    // has been destroyed we can not reuse it.
-    view.on('destroy', this._empty, this);
+    // Assume an attached view is already in the region for pre-existing DOM
+    if (!view._isAttached) {
```
scott-w: Should we document this assumption?

paulfalgout: I think pre-existing DOM stuff needs its own documentation: https://github.com/marionettejs/backbone.marionette/issues/3128#issuecomment-258726575  and this is only a small portion of what isn't fully documented.

Issue #2926, opened 2016-03-11
Reconsider stopPropagation in view triggers


JSteunou:
View triggers actually set `stopPropagation` to `true` by default. I do not know why this was set at first, but from my pov it's a real pain. I have a lot of views inside region triggering events on click for parent of behavior to listen, and when that parent or another parent far above have to listen a click too, I fall into the usual trap: click is stopped by view triggers `stopPropagation`.

Can we re-discuss that behavior and maybe change it for v3 as a breaking change?

maniator:
Better idea: 

Instead of a "breaking change" just add some sort of options for the trigger to set stopProp to false for certain triggers as needed.

paulfalgout:
There are already options for `stopPropagation` and `preventDefault`, I think @JSteunou is suggesting changing the default.

@JSteunou are you suggesting leaving `preventDefault: true` but changing `stopPropagation: false'?

I'm not entirely aware of the caveats of changing the defaults.   Makes me a little nervous.

JSteunou:
It's clearly a call for a breaking change. `stopPropagation` was introduced way back when adding those default values to `triggers` because people wanted it for forms. But `preventDefault` is the only one needed to stop forms to be send. `stopPropagation` is rarely use and do more harm from my pov.

If some people relied on that feature to stop an event to bubble up to a parent view, they only be concerned if they listened that event on the parent view and want the event specifically come from that parent layout and not from the child. But I do think that case is minor compared to normal use case, and so default value should represent majority.

I'm really interested to know people usage of this feature and see if I'm the only one seeing that or if I'm totally wrong. My PoV is obviously tight to the apps I'm writing.

JSteunou:
@marionettejs/marionette-core could this be considered for v3? I am afraid it will not be after v3 go out because it introduces a breaking change.

Again, from my experience, `triggers` default options introduce not expected behavior. When using `events` to handle `click`, we have to manually call to `preventDefault` or `stopPropagation` when we need to, but when using `triggers` we are always surprised it is set by default.

I had the case again today with a co-worker, not understanding why its link `<a href=""></a>` did not work anymore. It was because he added a `triggers` on click on it.

paulfalgout:
[...]

Issue #2975, opened 2016-04-06
Trigger View Handler should return event object


frankcortes:
Hello everyone, I was reading [Marionette's docs about Trigger Handler arguments](http://marionettejs.com/docs/v2.4.4/marionette.view.html#trigger-handler-arguments) and it's weird for me the fact that I cannot retrieve the original `event` object. 
Now, if I want get the original event and execute a trigger I have to do this:

``` javascript
events: { 'click .my-class': 'onAwesomeClick' },
//...some code
onAwesomeClick: function(event) {
   this.trigger('real:awesome:click', event);
},
onRealAwesomeClick: function(event) {
   //Do everything you want
}
```

This looks like me like it's a kind of proxy, because if I could do this

``` javascript
triggers: { 'click .my-class': 'real:awesome:click'  },
//...some code
onRealAwesomeClick: function(args) { //where args.event contains the original event
   //Do everything you want
},
```

It would rather use it better, and it would remove a lot of boilerplate. What are your thoughts about this?

paulfalgout:
This is a dupe of https://github.com/marionettejs/backbone.marionette/pull/2690

paulfalgout:
We might consider this in v3 as a 2nd argument for `triggers`? The only caveat here is that we should document it so that we strongly recommend against using the event outside of the event's own view.  Using the `onTriggeredMethod` calls inside a View is a pretty regular pattern and it is awkward to _have_ to use events on the View if the original event is needed.

But now that there's no longer an odd trigger argument object, I'm ok with it.  I'm not pushing for it, but I'm not against it either.

Thoughts? @marionettejs/marionette-core

JSteunou:
Solved by #3261

PR #3261, opened 2016-11-16
Feature/trigger method event args

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2926 and https://github.com/marionettejs/backbone.marionette/issues/2975

This was two birds with one stone as they're related, but if one is contentious I can pull it.  I doubt the feature flag would cause any concern, but in the past people haven't wanted to add the originating DOM event to the triggers argument.  Previously in `v2` triggers returnd a pretty odd object containing the `view`, `model`, and `collection` and adding `event` to that seemed rather bad, but now appending it as the 2nd argument seems very reasonable to me.

Again we want to suggest using it only for the `onEvent` methods within the view, but if someone wanted to get around this, using events and triggering an event was drop dead simple anyways.. so not including it isn't enforcing anything.  But if `triggers` does contain the event, at work I can tell other coders to _always_ use `triggers` now instead of "always use `triggers` except when you need the DOM event, then you have to use `events`"

- [ ] TODO: If this is merged the live examples should be updated to reflect the small changes

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -181,7 +181,7 @@ var MyView = Mn.View.extend({
     console.log('Show the modal');
   },
 
-  onDataEntered: function() {
+  onDataEntered: function(view, event) {
     console.log('Data was entered');
   }
 });
--- a/docs/marionette.features.md
+++ b/docs/marionette.features.md
@@ -5,3 +5,10 @@ Marionette Features are opt-in functionality. That you can enable by setting `Ma
 ##### Goals:
 + make it possible to add breaking changes in a minor release
 + give community members a chance to provide feedback for new functionality
+
+## `triggersStopPropagating`
+
+This flag is set to `true` by default.
+
+It indicates the whether or not [`View.triggers` will call `event.stopPropagating()`](./marionette.view.md#view-triggers-event-object) if not explicitly defined by the trigger.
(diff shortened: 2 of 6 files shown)


rafde:
events as 2nd arguments? 👍 That's awesome. Events should have always been accessible and
```js
triggers: {
  'click a': {
    event: 'link:clicked',
    preventDefault: true, // this param is optional and will default to true
   stopPropagation: false
  }
}
```
should get deprecated after this is merged and remove after in Mn4

paulfalgout:
@rafde so you mean `triggers` should do anything to prevent or stop propagation of events?

I'm not 100% sure.  I think it's the common case for me to preventDefault.  And I _just_ realized why `stopPropagation` matters: https://github.com/marionettejs/backbone.marionette/issues/2926#issuecomment-261040027

JSteunou:
I agree with @rafde I prefer Mn to be unopinionated on this but offer tools / options so it can be easy to have this feature anyway.

paulfalgout:
Ok so I do think there's further conversation to have related to how we'll handle this in the future.

But regardless of how that turns out I think this is the next step and isn't a big jump.

I _did_ go ahead and add a feature flag for `preventDefault` as well since that also came up in conversation.

PR #3262, opened 2016-11-16
Fix region clean up, when view destroy

ftdebugger:
Region with "replaceElement" setting don't restore region element, when view was destroyed by self. Because view element remove from DOM before `destroy` event fired. Listen `before:destroy` fix this problem.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -72,6 +72,10 @@ const Region = MarionetteObject.extend({
 
     this._attachView(view, options);
 
+    if (this._isReplaced) {
+      view.on('before:destroy', this._empty, this);
+    }
+
     this.triggerMethod('show', this, view, options);
     return this;
   },
@@ -225,6 +229,10 @@ const Region = MarionetteObject.extend({
   },
 
   _empty(view, shouldDestroy) {
+    if (this._isReplaced) {
+      view.off('before:destroy', this._empty, this);
+    }
+
     view.off('destroy', this._empty, this);
     this.triggerMethod('before:empty', this, view);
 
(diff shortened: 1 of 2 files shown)


src/region.js
```diff
@@ -72,6 +72,10 @@ const Region = MarionetteObject.extend({
 
     this._attachView(view, options);
 
+    if (this._isReplaced) {
+      view.on('before:destroy', this._empty, this);
```
rafde: hmm... should https://github.com/marionettejs/backbone.marionette/pull/3262/files#diff-b8da5292cb07ec757ac64e78ff5af5c1R67
be updated to 'before:destroy'?

test/unit/region.spec.js
```diff
@@ -337,6 +337,22 @@ describe('region', function() {
               expect(this.$parentEl).to.contain.$html('<div id="region"></div>');
             });
           });
+
+          describe('and restore, when view was self destroyed', function() {
+            beforeEach(function() {
+              this.view.trigger('before:destroy', this.view);
+              this.view._removeElement();
```
rafde: I think `this.view.destroy();` should do the event emitting instead in order to match a real scenario.

paulfalgout:
👏 Good find.  

@rafde I don't think `before:destroy` will work everywhere because `empty` checks if a view is destroyed.

However I think the easiest cheapest solution to this is to add `view.on('before:destroy', this._restoreEl);` to `_replaceEl`

paulfalgout:
@ftdebugger any interest on revising this a bit?

Issue #2317, opened 2015-02-11
Provide APIs for rendering templates directly


jmeas:
Not every bit of HTML needs complex event handling. Region, and other APIs, should be able to render templates directly.

jasonLaster:
I think we can do better here as well. This is what we've got now:

```js
this.showChildView('footer', new Marionette.View({
  template: footerView,
  model: footerModel
}));
```

I think we could do something like:

```js
this.showContent('footer', footerTemplate, footerModel)
```

In this case, something like region events is nice for binding event handlers and what not...

samccone:
what about...

```js
this.renderInto('footer', {model: foo, template: Bar})
```

jasonLaster:
yea, I like that @samccone

thejameskyle:
If we add this I would prefer it just cast a template into a view. It's not expensive, it keeps the region api clean, and it ensures that there are no incompatibilities.

samccone:
"cast a template into a view" not sure whatcha mean?

thejameskyle:
```js
renderInto(region, options) {
  var view = new View(options);
  this.getRegion(region).show(view);
}
```

samccone:
yep exactly

denar90:
What if instead of adding new API method just extend `showChildView` functionality. My proposal is pass View class instead of instance
```
showChildView(name, View = Mn.View, ...args) {
    const region = this.getRegion(name);
    return region.show(new View(args));
 }
```
But it look like breaking change.

Issue #2290, opened 2015-02-05
Templates and views


jmeas:
In Ember, you can render templates directly. This might seem strange, but it makes sense. Firstly, you don't always need data to render a view. Imagine a root template for your application, or an "about" page. It's easy to think of a situation where either of these templates would not have dynamic, model-backed content.

Additionally, Ember provides powerful Handlebars helpers to make their templates super powerful. You can, for instance, create the equivalent of a Region directly in your template.

Your next question might be...so how do I access that region? Well, in Ember, all regions are accessible via Routes. Because Routes are nested objects, and so are views and templates, it works out really nicely. They use a resolver pattern to find the nearest 'region' in the tree with the name that you've provided.

From the perspective of our code, there's no difference between a rendered template and a rendered view. Ultimately, we attach HTML directly into a DOM element. It really doesn't matter where that comes from.

This is just something to consider moving forward. This is especially relevant in light of #2279 and #2280. The application's root element could have a template rendered into it, and that would serve as the root template for the route tree.

Possible actionable items:
- Add a region templateHelper (Handlebars and Underscore) to allow regions to be created within the template directly. Minionette provides something like this if I remember correctly //cc @jridgewell.
- Link up our regions to the router
- :dollar:

PR #3266, opened 2016-11-18
Allow showing a template or static string in a region

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2317 and https://github.com/marionettejs/backbone.marionette/issues/2290

I think this feature will be better when we can create regions on the fly from selectors.  For instance imagine this API:

```js
myApp.showView(`
  <div class="header">Hello ${ user.get('name') }</div>
  <div class="main-region main"></div>
  <div class="sidebar-region sidebar"></div>
`);

myApp.getView().showChildView('.main-region', new MainView({ model }));

myApp.getView().showChildView('.sidebar-region', `
  <h2>User Info</h2>
  <div>Select an option to see more info...</div>
`);
```

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -23,6 +23,7 @@ Regions maintain the [View's lifecycle](./viewlifecycle.md#regions-and-the-view-
   * [Using Regions on a view](#using-regions-on-a-view)
 * [Showing a View](#showing-a-view)
   * [Checking whether a region is showing a view](#checking-whether-a-region-is-showing-a-view)
+* [Showing a Template](#showing-a-template)
 * [Emptying a Region](#emptying-a-region)
   * [Preserving Existing Views](#preserving-existing-views)
   * [Detaching Existing Views](#detaching-existing-views)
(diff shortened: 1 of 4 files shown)


paulfalgout:
So initially I made it such that you could pass `show(template, viewOptions)`

This was awkward for two reasons.  
1. `viewOptions` could contain `template` in which case the 1st argument is useless.
2. The 2nd argument is also used for show options.  I initially thought these would be removed in the future, but `replaceElement` is likely to stay a show option.

So I've added a commit changing this to:
```js
myRegion.show({
  template: _.template('Hello <%= place %>!'),
  model: new Backbone.Model({ place: 'World' })
});
```

docs/marionette.region.md
```diff
@@ -261,6 +262,26 @@ mainRegion.hasView() // true
...
 [remove the existing View](#emptying-a-region) before showing the new one.
 
+## Showing a Template
+
+You can show a template or a string directly into a region. Additionally you can pass an object literal containing a template and any other view options. Under the hood a `Marionette.View` is instantiated using the template.
+
+```javascript
+var myView = new MyView();
+
+var template = _.template('This is the <%= section %> page');
```
scott-w: I'd prefer to use `<%- %>` here for documentation purposes to promote escaping where possible.

paulfalgout: oh good catch.. I reformat my underscore template to use `{{}}` and `{{{}}}` so I don't have any clue what the normal is :-D

paulfalgout:
I updated the docs line for @scott-w and updated the same issue in the specs.. so I'm going to go ahead and get this merged in.

Issue #3279, opened 2016-12-14
Build tests are failing


paulfalgout:
> Hi, the latest "next" branch appears to not be building. It's throwing a bunch of mocha errors with the Behavior mixin. Is it just me?

I believe this is not a code issue, but related to the spec refactoring.. now that the specs are importing only the units they're testing.. testing the full build may not be feasible.  Thoughts?

cc: @denar90

denar90:
@paulfalgout agree we are not able to test all build right now... If we have some broken functionality after build (some config/rollup/babel etc problems) how will we detect them? So I have doubts, should we get back to previous approach? Thoughts?

paulfalgout:
I think while running all of the unit tests on the build was convienent, most of the tests weren't useful..  the issue with a broken build is an integration one.. all the unit tests would still pass.  I think we should can the build test for now and either rely on the upcoming performance tests as a safety net or maybe add some cypress testing or something and possibly move all testing that isn't strictly unit testing to a different system that uses the build?  I dunno.  Breaking up the tests is correct, and I don't want to hamper that because we have a cheap solution to test the build.

denar90:
@paulfalgout I think do it every release will be annoying. I'll look at it, maybe we can do smth smart 💡

paulfalgout:
Yep, but the build tests are a recent addition so not a huge loss.

denar90:
@paulfalgout would you like rather concentrate on tests refactoring then on fixing this issue?

paulfalgout:
I think so.. I mean we can't build at all without doing something, but I think unless there is an easy solution, we should just disable it for now..  I don't think it should get in the way of better unit tests.

PR #3281, opened 2016-12-15
remove build tests

denar90:
### Proposed changes
 - Remove build tests

Link to the issue:
#3279

### DIFF
--- a/gulp/build.js
+++ b/gulp/build.js
@@ -71,5 +71,5 @@ gulp.task('build-lib', ['lint-src'], function(){
 });
 
 gulp.task('build', function(done) {
-  runSequence('build-lib', 'test-build', done);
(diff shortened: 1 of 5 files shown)


denar90:
@scott-w It's run when we are building all stuff, I think it's not needed anymore since It can't work with new way of writing tests. Do we have to keep it?

scott-w:
I'm not sure, I was just asking if you felt it was worth keeping the task. If we're not using it at all and don't intend to fix it (as opposed to completely rewriting it) then feel free to merge this :+1:

paulfalgout:
It'll still be in the repo.. we'll just have to dig.. I think it's worth removing it for now since it's broken.. and probably can't be fixed and instead needs replacement.

Issue #3278, opened 2016-12-12
CollectionView._getRemovedViews and modelless child view


ardi-n:
When collection reset() is called the view rendering it is updated. But _getRemovedViews finds removed views by removed models. When an arbitrary child view without a model is added to the view that method throws an error.

Here's a proof: https://jsfiddle.net/d5shn4nu/

paulfalgout:
This is fixed in the next CollectionView https://github.com/marionettejs/backbone.marionette/pull/3244 but this probably needs to get fixed in the current one.

PR #3283, opened 2016-12-16
Destroy children by view and not model

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3278

When destroy all children, some view may not have a model.
This also removes the need for the `checkEmpty` flag and subsequent tests around that flag.

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -136,9 +136,7 @@ const CollectionView = Backbone.View.extend({
   // in the collection in order to keep the children in sync with the collection.
   // "models" is an array of models and the corresponding views
   // will be removed and destroyed from the CollectionView
-  _removeChildModels(models, {checkEmpty} = {}) {
-    const shouldCheckEmpty = checkEmpty !== false;
-
+  _removeChildModels(models) {
     // Used to determine where to update the remaining
     // sibling view indices after these views are removed.
     const removedViews = this._getRemovedViews(models);
@@ -152,8 +150,8 @@ const CollectionView = Backbone.View.extend({
     // decrement the index of views after this one
     this._updateIndices(removedViews, false);
 
-    if (shouldCheckEmpty) {
-      this._checkEmpty();
+    if (this.isEmpty()) {
+      this._showEmptyView();
     }
   },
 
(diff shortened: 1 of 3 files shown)


paulfalgout:
@rafde destroy was inadvertently giving us coverage.. but now that it isn't removing by models.. I am not quite sure how to hit the missing line.  thoughts?

```js
_findGreatestIndexedView(views) {
    return _.reduce(views, (greatestIndexedView, view) => {
      // Even if the index is `undefined`, a view will get returned.
      Eif (!greatestIndexedView || greatestIndexedView._index < view._index) {
        return view;
      }
 
      return greatestIndexedView;  // not covered :-/
    }, undefined);
  },
```

rafde:
I'll take a look when I get the chance.

rafde:
@paulfalgout 
maybe

```js
  _updateIndices(views, increment) {
    if (!this.sort) {
      return;
    }

    const view = _.isArray(views) ? _.max(views, '_index') : views;

    if (_.isObject(view)) {
      this.children.each((laterView) => {
          laterView._index += increment ? 1 : -1;
        }
      });
    }
  },
```
and delete `_findGreatestIndexedView`

src/collection-view.js
```diff
@@ -742,8 +733,8 @@ const CollectionView = Backbone.View.extend({
     }
 
     this.triggerMethod('before:destroy:children', this);
-    const childModels = this.children.map('model');
-    this._removeChildModels(childModels, options);
+    this.children.each(_.bind(this._removeChildView, this));
```
rafde: nice

rafde:
Need help adding a test
I am thinking of adding it [here](https://github.com/marionettejs/backbone.marionette/blob/5af5d0fd93beca32cc59aea7a702073047211b9d/test/unit/collection-view.reset.spec.js#L25-L25)

```js
beforeEach(function() {
  this.collectionView.render();
  this.addChildView(
    new Marionette.View({template: function(){ return 'END'}), 
    this.collection.length
  );
  this.collection.reset([{foo: 'bar'}, {foo: 'baz'}]);
});
```
might cover it.

paulfalgout:
I like the change to`_updateIndices` a lot.  Then I tried the test..  at first it didn't work and I jacked with it a bit..  before noticing we were stubbing the destroy function 😲   Anyhow.. it's a tiny bit different now, but it works.  I think this is GTG.

blikblum:
LGTM

Issue #3282, opened 2016-12-16
Test the build


paulfalgout:
Related: https://github.com/marionettejs/backbone.marionette/issues/3279

Improvements to the test files made testing the build less than trivial.  So at the moment it is untested. 😿

denar90:
@paulfalgout what do you think about this idea? If we will have some layer where classes/modules will be exported like:

`layer.js` (have to think about better name)
```js
import Error from '../../src/error';
import Behavior from '../../src/behavior';
...

exports {
  Error: Marionette.Error | Error
  Behavior: Marionette.Behavior | Behavior
  ...
}
```

They we will imported in `behaviors.spec.js`

```js
import {Error} from 'layer';
import {Behavior} from 'layer';
...
```

But using this way we will have problems with Mixins and other stuff, they are not in `Marionette`. But if we can't import them, aren't they kinda private stuff. Shouldn't we test only public stuff which is part of `Marionette`?

paulfalgout:
I do think that's the root of the issue..  when unit testing we want to test each module as it's own unit.. and we only want to test the public API of the _module_ regardless if that API is public API for Marionette itself.  When testing the library as a whole, we don't care about the unit tests.  We want to know if the units integrated together work. Right now we're writing more unit tests that are less integration tests..  regardless the only reason we ran the whole test suite before was because it was convenient and easy.  I am wondering if we should consider not doing _any_ integration tests in mocha and instead pulling them out into something like cypress.io.  Then we could run cypress against the build.  If we're testing if 2 modules work together, we should be able to test all of them together.  Plus Cypress would make for some interesting example code.  But we're a long way off.  All that said, I don't personally want to figure out a way to get the unit tests to work on the build the way were doing.

denar90:
Sa far as I know cypress.io is not available for everyone. We can get only early access (after using it I can say that it's awesome), but it will not work with CI now. Hence we have to wait...

paulfalgout:
I think it will work with ci now.  At least people are talking about the headless version now..  In any case, Brian is a previous core member, so we can probably get it setup.  We'd have to try it, but I _think_ the headless node version will run the tests for anyone without beta access and can just be setup as a devDependency.  I think right now it is just the gui that is closed to the beta users.. We can probably make a quick test and PR something to see if travis'd run it, but I _think_ it will now.

denar90:
Cool if we can do this!

[...]

PR #3287, opened 2016-12-19
[WIP] Try Cypress CI

paulfalgout:
Relates to: https://github.com/marionettejs/backbone.marionette/issues/3282

This is just running the Cypress example.  But it seems to work on Travis.

I also modified the travis.yml such that we just depend on the tests in coverage running.. then the test-browser..  then cypress.  I'm on the fence though.. do we want it to fail faster but take longer overall, or fail with more detail and succeed faster?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,19 @@ sudo: false
 language: node_js
 node_js:
   - "6"
+cache:
+  directories:
+  - /home/travis/.cypress/Cypress
 env: MAINRUN=true
 before_install:
   - npm config set ca ""
+  - npm install -g cypress-cli
 install:
   - npm install && npm install travis-ci
+script:
+  - npm run-script coveralls
+  - npm run-script test-cross-browser
+  - cypress ci $CYPRESS
 after_success:
-  - npm run-script coveralls && npm run-script test-cross-browser
   - if [[ $TRAVIS_BRANCH = "master" ]] && [[ $TRAVIS_PULL_REQUEST = "false" ]]; then node trigger-deploy-mn-com.js; fi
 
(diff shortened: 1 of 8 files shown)


paulfalgout:
Closing.. just for internal review

denar90:
Nice 👍  

> I also modified the travis.yml such that we just depend on the tests in coverage running.. then the test-browser.. then cypress. I'm on the fence though.. do we want it to fail faster but take longer overall, or fail with more detail and succeed faster?

I think nightly builds might solve problem about long build. Before having it we can use way it was before (when we run build we run cypress).

For writing this kind of tests we have to create some app/apps regarding to each integration test file, haven't we?
So we have to think about:
 - should it be either one or several apps ? 
 - how to structure this app/apps ?

Is it a problem to have `CYPRESS_CI_KEY`, not to fail the build - https://travis-ci.org/marionettejs/backbone.marionette/builds/185153939#L1309 ?

paulfalgout:
I'll have to try the cypress key again.. with the exact same setup it worked here: https://travis-ci.org/paulfalgout/backbone.marionette#L1295

I must've messed up the key..  I'm going to try it again using `CYPRESS_CI_KEY`. I didn't know that was an option.

I think we'll be able to setup some very simplistic apps for test examples for testing very specific scenarios, vs a larger app.  We should have some generic fixtures that can be pulled from and an easy css framework for _some_ sort of style... maybe bootstrap if we think that'd be the most widely used.

paulfalgout:
Hmm I don't know why it is failing on the main repo and not my fork... only thing that is really different is sauce failed on my box because I don't have the username/password on my travis.. anyhow.. check out travis on my link above.. We can get that to work.. I think as a POC this should be ok..

denar90:
I'll try to play with it basing on `Behavior` tests which we started refactoring.

Issue #3276, opened 2016-12-07
Clarify templateContext


scott-w:
Oversaw a discussion on the Gitter channel where the documentation of `templateContext` and how it works with functions wasn't as clear as it could be. It was assumed the following wouldn't work:

```js
templateContext: function() {
  return {
    foo: this.foo
  };
}
```

When in fact it would because the binding to the serialised data happens after this function is called.

JSteunou:
I think it also could depend on the template engine you are using.

scott-w:
True, I think in this instance it's best to make sure the documentation covers the standard template engine first. Do we know of any popular alternative engines?

PR #3291, opened 2016-12-24
This Binding

scott-w:
I've added a small example to try and demonstrate the different ways `this` is bound on `templateContext`.

Link to the issue: #3276


scott-w:
Updated it now @denar90 - cheers 👍

Issue #3229, opened 2016-10-11
Support yarn


paulfalgout:
https://code.facebook.com/posts/1840075619545360

Probably should support the new shiny if its easy..

denar90:
Here my results of last Marionette.js v3.1. 
`npm -v` - 2.15.9
`yarn version` - v0.15.1

### Clean cache
```
npm cache clean
time npm i
```
vs
```
yarn cache clean
time yarn
```

|      |    Npm    |    Yarn   |
|------|:---------:|:---------:|
| real |  5m2.069s | 1m40.364s |
| user | 3m18.796s | 1m24.774s |
| sys  | 1m23.261s | 0m47.342s |


### With cache

`time npm i` vs `time yarn`

|      |    Npm    |    Yarn   |
|------|:---------:|:---------:|
| real | 2m54.259s | 1m13.979s |
| user | 2m43.205s |  1m2.717s |
| sys  |  1m0.757s | 0m39.589s |

Pretty good results as for me.

rafde:
I am not sure what there is to support. Yarn is a CLI client that works with NPM registry and bower. It doesn't look like MarionetteJS needs to register with yarn's registry.

paulfalgout:
Yeah I think this is safe to close.  Put it up moments after the post, when everyone was having nervous shakes and giggle like school children.

JSteunou:
I'm very surprised by your results with npm clean install. I understand better now all the fuzz with Yarn.

At work I get pretty much the same results between Yarn & npm but that's because we have a local private repository with cache on our network.

blikblum:
We should consider storing yarn.lock into SCM: https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all

paulfalgout:
Agreed

paulfalgout:
Also travis caching: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Caching-with-yarn

PR #3292, opened 2016-12-26
Switch to yarn for travis

paulfalgout:
Adds the lock file. Resolves https://github.com/marionettejs/backbone.marionette/issues/3229
Removes unnecessary config and env vars.
Adds env vars so that the two tests can run in parallel.

Testing is much faster now.  Both because of yarn and the parallel testing.


blikblum:
#3293 also commits yarn.lock file. Will not lead to a conflict? 
Otherwise 👍

Issue #2940, opened 2016-03-16
Add lodash v4 test to matrix


paulfalgout:
We'll also have to:
- [x] Update dependencies to be lodash 4 compat (babysitter, radio)
- [x] Update our tests to be lodash 4 compat (as the tests use the same underscore that gets lodashed in travis via the matrix)

rafde:
So address
https://github.com/marionettejs/backbone.radio/issues/244

https://github.com/marionettejs/backbone.babysitter/issues/66 which https://github.com/marionettejs/backbone.babysitter/pull/67 is addressing

and replace https://github.com/marionettejs/backbone.marionette/blob/master/package.json#L44 with radio?

paulfalgout:
Ah! so this'd be off the `next` branch, so Radio should already be a dependency there for the package.json

rafde:
I think the first check is complete.
Unfortunately, I have no idea how to resolve the 2nd check.

PR #3293, opened 2016-12-26
Test lodash

paulfalgout:
Note that this includes: https://github.com/marionettejs/backbone.marionette/pull/3292
I didn't want to introduce a slower CI or cause merge issues.

This adds the ability to test underscore and/or lodash.  It also would allow us to reinstate a test matrix.. though I don't know that we need to do that at the moment.

Resolves https://github.com/marionettejs/backbone.marionette/issues/2940

Testing that this actually works was moderately simple.  If you remove `_.invokeMap ||` from https://github.com/marionettejs/backbone.marionette/blob/master/src/utils/invoke.js#L5 then `yarn run test` still works, but `yarn run test-lodash` lights up big time.


blikblum:
Good work. Has my 👍

denar90:
@paulfalgout do you want to merge this one into `next` instead of `master` ?

paulfalgout:
Well I figured at the moment `next` is getting all of the code changes so we don't really need this on master, but we do need it on `next`.  I suppose we could put it on master and then pull it over, but to me at this point it seems like it would be fine either way as long as `next` gets it.

denar90:
Got it. Merging this. We will just have to pull master into next, to solve possible merge conflicts asap.

paulfalgout:
Ah so the 1st commit is the commit that went into master.. I just cherry-picked it here.. so there shouldn't be any merge conflicts.

PR #3295, opened 2016-12-31
Add cypress simple setup

denar90:
### Proposed changes
 - Added simple cypress setup for future testing

I wanna move step by step instead of creating large PR's. It will be easy for review and development.
Here just simple setup and structure with cypress, but I'm open to your comments.

PR #3296, opened 2016-12-31
Add cypress testing

paulfalgout:
Many thanks to @denar90 for getting this started.  I’ve been going back and forth with how cypress would be effective with Marionette.  The primary purpose of Cypress is really to test an app and not necessarily a library.  So all of the documentation is around testing an app.

At first I started by reviewing and modifying #3295 but after some time with it I think this approach was too different.

- I think since we are not running our own app we can get away with not running a local server.
- This also builds the library and tests from the built files.
- Then it allows setup of minimal Marionette examples for specific testing within the tests themselves.

I went ahead and setup a fixture example, though I doubt fixture files will be all that necessary.. we may want to have a couple of large dumps from something for funzies.. and we should likely have a simple css framework that makes the tests look clear.  Nothing fancy, but something.

Structure-wise I think we’d add files to `test/integration` and I _think_ we may want to make integration tests that follow the documentation?  I’m not sure on the structure exactly.. unit tests are easy.. one test file per src file.  But with integration we are testing concepts.  In any case we should come up with a way to organize it that is standard… but I think a companion to the docs might be easy to follow.. I just don’t know if that is good enough to cover everything.  Ideally though after scanning the integration tests, anyone could make a decent educated guess as to where something should be added.


paulfalgout:
Regarding "allows setup of minimal Marionette examples for specific testing"  https://github.com/marionettejs/backbone.marionette/pull/3296/files#diff-f785ed312c172b1397d4f59c98ad3882R9

denar90:
@paulfalgout don't you mind to merge it to move forward?

paulfalgout:
okie

PR #3301, opened 2017-01-19
Delete yarn.lock

denar90:
I think having this file in scope of repo is not needed. 
First of all it makes diffs ugly and second minus one file in repo.


blikblum:
There are benefits: https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all

Remember that yarn changes only when updating dependencies which is sporadic

paulfalgout:
Yeah I think the `yarn.lock` file is a plus.  This will make sure that until the lock file changes, nothing will break.  👎

denar90:
Yeap, make sense guys.

PR #3302, opened 2017-01-20
Add Marionette.Application integration tests

denar90:
### Proposed changes
 - Cypress application test

### DIFF
--- a/test/integration/application.spec.js
+++ b/test/integration/application.spec.js
@@ -1,0 +1,38 @@
+describe('Application', function() {
+  let app;
+  let fooView;
+  let barView;
+
+  beforeEach(function() {
+    cy.server();
+
+    cy.visitMn(function({_, Backbone, Marionette}) {
+      const Application = Marionette.Application.extend({
+        region: '#app-hook',
+      });
+
+      fooView = new Marionette.View({
+        template: _.template('Application region content')
+      });
+      barView = new Marionette.View({
+        template: _.template('<h1 class="title">Application region content</h1>')
+      });
+
+      app = new Application();
+      app.start();
+    });
+  });
+
+  it('should show a view without additional arguments', function() {
+    app.showView(fooView);
+    cy.get('#app-hook').should('contain', 'Application region content');
+  });
+
+  it('should show a view with additional arguments', function() {
+    app.showView(barView, {
+      selector: '.title',
+      replaceElement: true
+    });
+    cy.get('.title').should('contain', 'Application region content');
+  });
+});
--- a/test/integration/application.spec.js
+++ b/test/integration/application.spec.js
@@ -1,22 +1,18 @@
+const _ = Cypress._;
+
 describe('Application', function() {
   let app;
-  let fooView;
-  let barView;
+  let FooView;
 
   beforeEach(function() {
     cy.server();
 
-    cy.visitMn(function({_, Backbone, Marionette}) {
+    cy.visitMn(function({Marionette}) {
       const Application = Marionette.Application.extend({
         region: '#app-hook',
       });
 
-      fooView = new Marionette.View({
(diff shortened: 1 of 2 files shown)


test/integration/application.spec.js
```diff
@@ -0,0 +1,38 @@
...
+        template: _.template('<h1 class="title">Application region content</h1>')
+      });
+
+      app = new Application();
+      app.start();
+    });
+  });
+
+  it('should show a view without additional arguments', function() {
+    app.showView(fooView);
```
denar90: @paulfalgout can we do it like ?
```js
const view = new Marionette.View({
  template: _.template('Application region content')
});
app.showView(view);
```
to reduce logic in `cy.visitMn`

test/integration/application.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    cy.get('#app-hook').should('contain', 'Application region content');
+  });
+
+  it('should show a view with additional arguments', function() {
+    app.showView(barView, {
+      selector: '.title',
```
paulfalgout: What is `selector`?  just filler?  It seems wrong.. if it's region related `selector` was deprecated to `el`, but in this case I don't think it is doing anything?

test/integration/application.spec.js
paulfalgout: We can do that.. but in some instances it may not be possible because the library instances are only available to that function and we shouldn't let them bleed outside I don't think.  But yeah it seems super reasonable in this case.. otherwise we'd have soooo many `visitMn`s

test/integration/application.spec.js
denar90: yeah, it has to be `el` 😄

test/integration/application.spec.js
denar90: Should we leave it as it is for now?

test/integration/application.spec.js
paulfalgout: I don't really know what it's doing though..  it seems like `.title` is inside `barView`.. but `replaceElement` is being passed to the `region.show`.. `el` or `selector` isn't doing anything at all right?

test/integration/application.spec.js
paulfalgout: could be something like `FooView = Marionette.View;`  then in the tests do `new FooView({`?  I dunno.

denar90: If so, we will have to do `_ = _`, because in scope of `it` underscore also is not visible...

paulfalgout: `Cypress._` would be fine for templates. if ya want.. At work at the top of most files we do a `const _ = Cypress._;` just so we can use it like normal.

denar90: Can we do it the same for `Marionette`?
Hence 
```
const _ = Cypress._;
const Mn = Cypress.Marionette;

describe('Application', function() {
....
```

paulfalgout: Nope.  Cypress has it's own copy of `_` and `$`, but they're not the same at all. For instance if you were to `Backbone.$ = Cypress.$` you'd have big problems.  `Cypress._` is not the same copy of `_` as what's available to `window`, but particularly in the case of a `template` it doesn't matter.

denar90: Oh, thanks, got it. I thought it somehow related to 
```
runApp({
        $: win.$,
        _: win._,
        Backbone: win.Backbone,
        Radio: win.Radio,
        Marionette: win.Marionette,
        window: win
      });
```

now it's clear for me. Thanks @paulfalgout !!!

test/integration/application.spec.js
denar90: It's me screwing up. `el` is useless here...

PR #3306, opened 2017-01-30
replaceEl should noop if the els are already the same

paulfalgout:
To my surprise, `parentEl.replaceChild(childEl, childEl)` is far slower than returning if they’re the same element.

https://jsperf.com/replacechild-of-same-child/1

I don’t love the test exactly, but it follows suit with what is currently there.


blikblum:
LGTM

JSteunou:
:+1:

PR #3308, opened 2017-02-04
Fix code sample for modelEvents in View documentation

TDu:
There is an error In the modelEvents code sample, the callback function arguments are in the wrong order.

The model is passed first and then the value.


paulfalgout:
thanks! 👏

PR #3310, opened 2017-02-10
WIP add cypress collection view tests

denar90:
Cypress tests for CollectionView.

### DIFF
--- a/test/integration/collection-view.spec.js
+++ b/test/integration/collection-view.spec.js
@@ -1,0 +1,139 @@
+const _ = Cypress._;
+const $ = Cypress.$;
+
+describe('CollectionView', () => {
+  let BbModel;
+  let BbCollection;
+  let CollectionView;
+  let MnView;
+
+  beforeEach(() => {
+    cy.server();
+
+    cy.visitMn(function({Marionette, Backbone}) {
+      BbModel = Backbone.Model;
+      BbCollection = Backbone.Collection;
+      MnView = Marionette.View;
+
+      const FooView = Marionette.View.extend({
+        template: _.template('Foo template content')
+      });
+
+      const BarView = Marionette.View.extend({
+        template: _.template('Bar template content')
+      });
+
+      CollectionView = Marionette.CollectionView.extend({
+        className: 'collection-view-container',
+        collection: new BbCollection(),
+        childView(item) {
+          if (item.get('isFoo')) {
+            return FooView;
+          } else {
+            return BarView;
+          }
+        }
+      });
+
+    });
+  });
+
+  describe('childView', () => {
+    let collectionView;
+    let fooModel;
+    let barModel;
+
+    beforeEach(() => {
+      const Model = BbModel.extend({
+        defaults: {
+          isFoo: false
+        }
+      });
+
+      fooModel = new Model({
+        isFoo: true
+      });
+
+      barModel = new Model({
+        isFoo: false
+      });
+
+      collectionView = new CollectionView();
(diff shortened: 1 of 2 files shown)


test/integration/collection-view.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      const FooView = Marionette.View.extend({
+        template: _.template('Foo template content')
+      });
+
+      const BarView = Marionette.View.extend({
+        template: _.template('Bar template content')
+      });
+
+      CollectionView = Marionette.CollectionView.extend({
+        className: 'collection-view-container',
```
rafde: this might be better as `id: 'collection-view-container',` since there should be only one instance.

test/integration/collection-view.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      });
+
+      collectionView = new CollectionView();
+      collectionView.render();
+
+      $('#app-hook').html(collectionView.$el);
+    });
+
+    it('should be without childView\'s', function() {
+      cy.get('.collection-view-container').should('be.empty');
```
rafde: would
```js
expect(collectionView.el).to.be.empty;
```
work better? Of does this need to check if the `collectionView.el` exist in the DOM?

test/integration/collection-view.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    it('should show default childView', function() {
+      collectionView.collection.add(fooModel);
+
+      cy.get('#app-hook').should('contain', 'Foo template content');
```
rafde: I think `#app-hook` should be set as an alias
[`cy.get("#app-hook").as("app-hook")`](https://docs.cypress.io/docs/using-aliases#section-aliasing-dom-elements) for `cy.get("@app-hook")` usage instead of re-querying the document. todo for now

test/integration/collection-view.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    it('should have empty view element', () => {
+      cy.get('.collection-view-container').should($container => {
+        expect($container.find('.empty-view')).to.have.length(1);
+      });
+    });
+
+    it('should have empty view content', () => {
+      cy.get('.collection-view-container').should('contain', 'Empty view content');
```
paulfalgout: also I think after writing more complicated cypress tests, it's not possible to reasonably keep single assertions per `it`.  For performance and complexity reasons I think this test and the previous should be
`cy.get('.collection-view-container').find('.empty-view').should('contain', 'Empty view content');`
So many things include "implied asserts" such as `.find`, and Cypress will treat them fairly separately when running.

test/integration/collection-view.spec.js
rafde: is there a need for more?

denar90:
Guys, thanks for the review 👍  I've added small fixes. Later, I'll add some examples with large collections, sorting, filtering and stuff.

rafde:
Should we focus our effort on NextCollectionView using cypress and keep current CollectionView in mocha since it will get replaced.

paulfalgout:
In theory it'll be easy to switch these over to the new one which might give us some insight into if it works as expected and maybe a little bit of data on speed.

Also related.. I'm down to one spec file on `NextCollectionView`  Might be this week!

denar90:
Sorry guys, can't find time to finish this one.

paulfalgout:
This'll be a bit different now.. both for cypress and with collectionview

PR #3311, opened 2017-02-11
Fix: Collection _updateIndices - decrementing

### LINKED ISSUE #3274: CollectionView "filter" not respected when adding/removing models from collection
### Description:

The 'filter' method allow 3 parameters: model, index and collection.
The issue is in the index parameter.

I created a sample to reproduce the issue:
There's a button where the user can add new elements in a collection. but there's a rule that the view should render just 1 element of this collection.
Regardless of how many elements are in the collection, just one element should be rendered.

When the 'model' parameter is used, the logic works fine. But with the 'index' parameter the method doesn't have the behavior that I expect.
The 'index' is being iterated for all element of the collection. When the iteration ends, the filter is executed again with index equals '-1'.

Demo: http://codepen.io/anon/pen/jVYYjM

### Expected behavior:
 
When the user click the "Add" button, a new element should be added, but the HTML page should render just 1 element.

### Actual behavior:

All the elements of the collection are shown in the HTML page.

### Environment

Marionette version: v2.4.7
Note: This issue also happen in the v3.1.0

### LINKED ISSUE #3298: Collection _updateIndices - decrementing
### Description

The problem occurs after the update to 3. Namely, when I want to extend the collection with additional elements it okey, but when I want to roll back the additional elements the indexes are badly decremented.
Removing items from the collection updates indices, current indexes collection of elements should be reduced by the length of additional elements, not one.

### Expected behavior

```js
_updateIndices: function _updateIndices(views, increment) {
  if (!this.sort) {
    return;
  }

  var view = _.isArray(views) ? this._findGreatestIndexedView(views) : views;

  this.children.each(function (laterView) {
      laterView._index += increment ? 1 : -views.length;  // here
    }
  });
},
```

### Actual behavior

```js
_updateIndices: function _updateIndices(views, increment) {
  if (!this.sort) {
    return;
  }

  var view = _.isArray(views) ? this._findGreatestIndexedView(views) : views;

  this.children.each(function (laterView) {
      laterView._index += increment ? 1 : -1;  // here
    }
  });
},
```

### Environment

1. Marionette version: 3.1.0
2. Backbone version: 1.3.3

_Edited to improve code format for better visibility: @paulfalgout_

rafde:
### Proposed changes
 - when removing views, sort children views and re-index
 - include test to prove fix
 - added test for checking that filter is working with index value passed to filter.

Link to the issue:

fixes #3298 

closes #3274

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -572,13 +572,21 @@ const CollectionView = Backbone.View.extend({
       return;
     }
 
+    if (!increment) {
+      _.each(_.sortBy(this.children._views, '_index'), (view, index) => {
+        view._index = index;
+      });
+
+      return;
+    }
+
     const view = _.isArray(views) ? _.max(views, '_index') : views;
 
     if (_.isObject(view)) {
       // update the indexes of views after this one
       this.children.each((laterView) => {
         if (laterView._index >= view._index) {
-          laterView._index += increment ? 1 : -1;
+          laterView._index += 1;
         }
       });
     }
--- a/test/unit/collection-view.filter.spec.js
+++ b/test/unit/collection-view.filter.spec.js
@@ -362,6 +362,20 @@ describe('collection view - filter', function() {
         expect(this.collectionView.$el).to.contain.$text('24');
       });
     });
+
+    describe('when the filter changes based on the view\'s index', function() {
+      beforeEach(function() {
+        this.collectionView.setFilter(function(view, index) {
+          return index < 1;
+        });
+        this.collectionView.render();
+        this.collection.add({foo: true, bar: 5});
+      });
+
+      it('should only have 1 view', function() {
+        expect(this.collectionView.children).to.have.property('length', 1);
+      });
+    });
   });
 
   describe('combined with a reorder', function() {
--- a/test/unit/sorted-views.spec.js
+++ b/test/unit/sorted-views.spec.js
@@ -119,6 +119,17 @@ describe('collection/composite view sorting', function() {
           expect(this.collectionView.resortView).to.have.been.calledOnce;
           expect(this.compositeView.resortView).to.have.been.calledOnce;
         });
+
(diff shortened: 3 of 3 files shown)


src/collection-view.js
```diff
@@ -207,11 +207,13 @@ const CollectionView = Backbone.View.extend({
   // Render children views. Override this method to provide your own implementation of a
   // render function for the collection view.
   render() {
-    if (this._isDestroyed) { return this; }
-    this.triggerMethod('before:render', this);
-    this._renderChildren();
-    this._isRendered = true;
-    this.triggerMethod('render', this);
+    if (!this._isDestroyed) {
```
rafde: for code coverage

paulfalgout: I'm not totally sure what this is doing?  Did we not have code coverage before?

scott-w: The only way I can think we were missing code coverage on this is `this._isDestroyed` was always `false` in the tests. Would this change simply be giving us a false sense of testing security when we should be covering the `_isDestroyed` case too?

rafde: This part came up when I ran the code coverage test. It listed collectionview at 99.98%.
I can look into it again in a separate issue. Just thought I'd bring it up.

rafde: @paulfalgout @scott-w was this the only issue? I can separate this out to let the rest get merged.

paulfalgout: yeah the rest looks fine to me.

As far as coverage is concerned we have line and function coverage at 100% now.. this issue is a statement and branch issue, but this change in the code doesn't resolve the branch coverage.. for full coverage we'd need to test the if and else of the condition.. changing the `if` wrapping doesn't really fix coverage.

rafde: looks like it was a [false alarm ](https://coveralls.io/builds/10139813/source?filename=src%2Fcollection-view.js)

PR #3319, opened 2017-02-20
Bump and build v3.2.0

denar90:
Bump and build v3.2


lib/backbone.marionette.js
```diff
@@ -12,3319 +12,3477 @@
...
-	  // Enable getting options from this or this.options by name.
-	  getOption: getOption,
-
-	  // Enable binding view's events from another entity.
-	  bindEvents: bindEvents,
-
-	  // Enable unbinding view's events from another entity.
-	  unbindEvents: unbindEvents
-	};
+}(this, (function (Backbone,_,Radio) { 'use strict';
```
denar90: Strange... With new build `(`  was added additional which breaks diffs at all.

denar90:
Are we ok to leave extra `$1`, `$2` etc as methods prefix for now?

blikblum:
> Are we ok to leave extra $1, $2 etc as methods prefix for now?

Yes. Sometime ago i played with various alternatives in source code reorganization and could not get rid of them, specially `triggerMethod$1`. Seems a limitation in rollup. The better would isolate the "bug" and report in rollup but i could not create an self contained example

BTW: `ClassOptions*` can be easily overcome, just change `ClassOptions` to `[className]ClassOptions` but i think is not worth at this time

Issue #3312, opened 2017-02-14
Regions should better support animation


paulfalgout:
https://github.com/marionettejs/backbone.marionette/pull/3156

Personally I'd like to see an implementation that does not introduce async library dependencies but where it easy to introduce your own.

blikblum:
I think that we should look at prior work in order to accommodate most of use cases. Those are the solutions i found

v3 compatible:
- https://github.com/maratfakhreev/marionette-animated-region (uses velocity)
- http://codepen.io/blikblum/pen/oZxWjQ (uses CSS + matrix manipulation)

based on older Mn:
- https://github.com/jpdesigndev/MarionetteTransition/blob/master/src/scripts/MarionetteTransition.js (uses TweenLite)
- https://github.com/jmeas/marionette.transition-region/blob/master/example/script.js (animation type agnostic, examples use jQuery animation)
- https://github.com/marcinkrysiak1979/marionette.showAnimated (uses TweenMax)
- https://gist.github.com/jasonLaster/9794836 (uses CSS)

Supporting such animations https://tympanus.net/Development/PageTransitions/ is also desired

I will start to create demos with the various possibilities so we can design an abstraction

JSteunou:
This one is quite old but might also help you https://github.com/JSteunou/marionette.csstransition-region ;)

blikblum:
I just got working a proof of concept of animated Region using velocity, that can be see here: http://codepen.io/blikblum/pen/EWyaMv?editors=1010

Is still necessary to test the combination of options and create a version with CSS and jQuery animation but posting here initial notes to not forget

- Is necessary to override `attachHtml` and `empty`
- The overriden `attachHtml` is straight forward and is how is supposed to be used
- OTOH, in `empty` is necessary to copy Marionette implementation, still with some issues
  -  If there's a hide animation, at the time `_empty` is called the currentView points to the new view, so the need to restore it since `_empty` deletes it
  - For the same reason above, when replaceElement is used, `_restoreEl` called inside `_empty` can be problematic.

blikblum:
I created examples of how to implement animated Region with current version. Instead of overriding `_empty`, i override `_removeView` simplifying a bit:

* animate.css: http://codepen.io/blikblum/pen/PppGmG
* velocity: http://codepen.io/blikblum/pen/gmmwzg
* CSS Page transitions: http://codepen.io/blikblum/pen/peEqqM based on https://tympanus.net/Development/PageTransitions/
[...]

PR #3323, opened 2017-03-18
Refactor region to simplify implementation of animated regions

blikblum:
### Proposed changes
 - Rename _removeView to removeView
 - Add destroyView method and use it in removeView
 - Move shouldDestroy check and detachView code out of removeView
 - Add swappingView flag, to allow to check if there is already a currentView that is being swapped by another

Link to the issue:
#3312 

This change allows to implement animated region as follows:

```
class AnimatedRegion extends Marionette.Region {

  attachHtml(view) {    
    view.$el
      .css({display: 'none'})
      .appendTo(this.$el);      
    if (!this.swappingView) view.$el.fadeIn('slow')
  }
  
  removeView(view) {
    view.$el.fadeOut('slow', () => {
        this.destroyView(view) // or view.remove if sure that view is a Mn.View
        this.currentView.$el.fadeIn('slow')
    })    
  }  
}
```

CSS animation could be implemented with similar code as demonstrated by http://codepen.io/blikblum/pen/PppGmG

`destroyView` is optional since is possible to just call view.destroy for Mn.View or call the overridden method using ES6 super to take care of BB.View. But this would not fit the current API that AFAIK does not recommend / require calling super methods

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -443,8 +443,6 @@ var EmptyMsgRegion = Mn.Region.extend({
 ```
 [Live example](https://jsfiddle.net/marionettejs/c1nacq0c/1/)
 
-> `hasView` will not work reliably inside `onEmpty` because will always evaluate to false
-
 ## Set How View's `el` Is Attached
 
 Override the region's `attachHtml` method to change how the view is attached
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -30,7 +30,9 @@ Regions maintain the [View's lifecycle](./viewlifecycle.md#regions-and-the-view-
   * [Preserving Existing Views](#preserving-existing-views)
   * [Detaching Existing Views](#detaching-existing-views)
 * [`reset` A Region](#reset-a-region)
+* [Check If View Is Being Swapped By Another](#check-if-view-is-being-swapped-by-another)
 * [Set How View's `el` Is Attached](#set-how-views-el-is-attached)
+* [Configure How To Remove View](#configure-how-to-remove-view)
 
 ## Defining the Application Region
 
@@ -421,6 +423,28 @@ myRegion.reset();
 
 This can be useful in unit testing your views.
 
+## Check If View Is Being Swapped By Another
+
+The `isSwappingView` method returns if a view is being swapped by another one. It's useful
+inside region lifecycle events / methods.
+
+The example will show an message when the region is empty  
+
+```javascript
+var Mn = require('backbone.marionette');
+
+var EmptyMsgRegion = Mn.Region.extend({
+  onEmpty() {
+    if (!this.isSwappingView()) {
+      this.$el.append('Empty Region');
+    }    
+  }  
+});
+```
+[Live example](https://jsfiddle.net/marionettejs/c1nacq0c/1/)
+
+> `hasView` will not work reliably inside `onEmpty` because will always evaluate to false
+
 ## Set How View's `el` Is Attached
 
 Override the region's `attachHtml` method to change how the view is attached
@@ -463,3 +487,60 @@ var MyView = Mn.View.extend({
   }
 });
 ```
+
+## Configure How To Remove View
+
+Override the region's `removeView` method to change how and when the view is destroyed / removed
+from the DOM. This method receives one parameter - the view to remove.
+
+The default implementation of `removeView` is:
+
+```javascript
+var Mn = require('backbone.marionette');
+
+Mn.Region.prototype.removeView = function(view){
+  this.destroyView(view);
+}
+```
+
+> `destroyView` method destroys the view taking into consideration if is
+> a Marionette.View descendant or vanilla Backbone view. It can be replaced
+> by a `view.destroy()` call if is ensured that view descends from Marionette.View   
+
+This example will animate with a fade effect showing and hiding the view:
+
+```javascript
+var Mn = require('backbone.marionette');
+
+var AnimatedRegion = Mn.Region.extend({
+  attachHtml(view) {    
+    view.$el
+      .css({display: 'none'})
+      .appendTo(this.$el);      
+    if (!this.isSwappingView()) view.$el.fadeIn('slow');
+  },
+  
+  removeView(view) {
+    view.$el.fadeOut('slow', () => {
+      this.destroyView(view); // or view.destroy() for Marionette.View
+      this.currentView.$el.fadeIn('slow');
+    })    
+  }  
+});
+
+var MyView = Mn.View.extend({
+  regions: {
+    animatedRegion: {
+      regionClass: AnimatedRegion,
+      el: '#animated-region'
+    }
+  }
+});
+```
+
+[Live example](https://jsfiddle.net/marionettejs/c1nacq0c/2/)
+
+Using a similar approach is possible to create a region animated with CSS:
+
+[Live example](https://jsfiddle.net/marionettejs/9ys4d57x/1/)
+ 
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -521,11 +521,12 @@ var AnimatedRegion = Mn.Region.extend({
   },
   
   removeView(view) {
-    view.$el.fadeOut('slow', () => {
-      this.destroyView(view); // or view.destroy() for Marionette.View
-      this.currentView.$el.fadeIn('slow');
+    var self = this;
+    view.$el.fadeOut('slow', function() {
+      self.destroyView(view);
+      if (self.currentView) self.currentView.$el.fadeIn('slow');
     })    
-  }  
+  }   
 });
 
 var MyView = Mn.View.extend({
@@ -538,9 +539,9 @@ var MyView = Mn.View.extend({
 });
 ```
 
-[Live example](https://jsfiddle.net/marionettejs/c1nacq0c/2/)
+[Live example](https://jsfiddle.net/marionettejs/c1nacq0c/3/)
 
 Using a similar approach is possible to create a region animated with CSS:
 
-[Live example](https://jsfiddle.net/marionettejs/9ys4d57x/1/)
+[Live example](https://jsfiddle.net/marionettejs/9ys4d57x/2/)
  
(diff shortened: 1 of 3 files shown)


src/region.js
```diff
@@ -269,7 +275,11 @@ const Region = MarionetteObject.extend({
     delete this.currentView;
 
     if (!view._isDestroyed) {
-      this._removeView(view, shouldDestroy);
+      if (shouldDestroy) {
+        this.removeView(view);
```
paulfalgout: why not just call `destroyView` here, or put the content from `destroyView` inside `removeView` that way there's only one new public method

src/region.js
```diff
@@ -58,6 +59,8 @@ const Region = MarionetteObject.extend({
 
     if (view === this.currentView) { return this; }
 
+    this.swappingView = !!this.currentView;
```
paulfalgout: For consistency we should use an `_is` prefixed variable for state flags with an `is____()` accessor.

src/region.js
```diff
@@ -269,7 +275,11 @@ const Region = MarionetteObject.extend({
     delete this.currentView;
 
     if (!view._isDestroyed) {
-      this._removeView(view, shouldDestroy);
+      if (shouldDestroy) {
+        this.removeView(view);
+      } else {
+        this._detachView(view);
```
paulfalgout: ok so if you're animating a view you _can't_ detach the view.  That's probably ok, but there's quite a few features of region that can't be used with this.

src/region.js
blikblum: To make easier to override, without `destroyView`, with its content inside `removeView` would be necessary to do something like below to accommodate standalone BB.View:
```
removeView(view) {
    view.$el.fadeOut('slow', () => {
          if (view.destroy) {
            view.destroy();
           } else {
             destroyBackboneView(view);
            }
        this.currentView.$el.fadeIn('slow')
    })    
  } 
```
The problem is that `destroyBackboneView` is not exposed

or using ES6 super

```
removeView(view) {
    view.$el.fadeOut('slow', () => {
        this.super.removeView(view)
        this.currentView.$el.fadeIn('slow')
    })    
  } 
```

Anyway, as i said the destroyView is optional and can be removed

src/region.js
blikblum: AFAIK it can since `detachView` will call empty with shouldDestroy false (undefined)

blikblum: There's no change at all in how it was behaving before

src/region.js
blikblum: I'm fine with that

src/region.js
paulfalgout: I suppose what I mean is that calling `myRegion.empty()` might animate out the view, but calling `myRegion.detachView()` would not.

src/region.js
paulfalgout: I suppose `destroyView` works fine on it's own and complements `detachView`.  fair enough.

src/region.js
blikblum: Its the opposite. As is today, using `_removeView` as is done by http://codepen.io/blikblum/pen/PppGmG, calling `detachView` will animate. With this PR it will not animate which i think is the correct behavior

paulfalgout: so in the codepen `mainView.showChildView('content', view, { preventDestroy: true });` messes things up a bit.  Would that same thing be true with this implementation?

blikblum: AFAIK no, because `removeView` will not be called.

src/region.js
blikblum: Done

src/region.js
blikblum: I added tests, checking whether `removeView` is called depending on `preventDestroy` and after `detachView`

paulfalgout:
Need to get the coverage back to 100%.. otherwise @marionettejs/marionette-core  ?

test/unit/region.spec.js
```diff
@@ -581,6 +592,10 @@ describe('region', function() {
       it('should not have a parent', function() {
         expect(this.detachedView).to.not.have.property('_parent');
       });
+
+      it('should not call removeView', function() {
+        expect(this.region.removeView).not.to.have.been.called;
+      })
```
rafde: Hmm... I am surprise eslint didn't fix this.

rafde:
Should this have a mock animation test using setTimeout?

test/unit/region.spec.js
paulfalgout: I think the linter on our tests is really lax..

test/unit/region.spec.js
```diff
@@ -56,6 +56,10 @@ describe('region', function() {
         (Backbone.Marionette.Region.extend({el: $('the-ghost-of-lechuck')[0]}))();
       }).to.throw;
     });
+
+    it('should not been swapping view', function() {
```
scott-w: `not be`

test/unit/region.spec.js
```diff
@@ -230,6 +237,10 @@ describe('region', function() {
       expect(this.region.onBeforeShow).to.have.been.calledWith(this.region, this.view, this.showOptions);
     });
 
+    it('should not been swapping view', function() {
```
scott-w: `not be`

test/unit/region.spec.js
```diff
@@ -268,6 +279,10 @@ describe('region', function() {
       it('should still have a view', function() {
         expect(this.region.hasView()).to.equal(true);
       });
+
+      it('should been swapping view', function() {
```
scott-w: `should be`

test/unit/region.spec.js
```diff
@@ -915,6 +933,10 @@ describe('region', function() {
     it('should not have a view', function() {
       expect(this.region.hasView()).to.equal(false);
     });
+
+    it('should not been swapping view', function() {
```
scott-w: `not be`

blikblum:
I can write the docs. In which branch? master or next?

paulfalgout:
docs can go directly on this branch if you want.

docs for existing functionality would go on master.. but docs for upcoming functionality would be on next

src/region.js
```diff
@@ -284,19 +299,18 @@ const Region = MarionetteObject.extend({
     this._parentView.stopListening(view);
   },
 
-  _removeView(view, shouldDestroy) {
-    if (!shouldDestroy) {
-      this._detachView(view);
-      return;
-    }
-
+  destroyView(view) {
```
paulfalgout:
```js
if (view._isDestroyed) {
  return;
}
```
I just started testing the related part of the next collection view and it has a near identical function.. https://github.com/marionettejs/backbone.marionette/pull/3244/files#diff-c4fcc4edadcafa174aa6de6124ecbeacR652

this can probably move to the mixin for v4

paulfalgout: This function should also `return this;`

blikblum: done

docs/marionette.region.md
```diff
@@ -463,3 +487,60 @@ var MyView = Mn.View.extend({
...
+var AnimatedRegion = Mn.Region.extend({
+  attachHtml(view) {    
+    view.$el
+      .css({display: 'none'})
+      .appendTo(this.$el);      
+    if (!this.isSwappingView()) view.$el.fadeIn('slow');
+  },
+  
+  removeView(view) {
+    view.$el.fadeOut('slow', () => {
```
scott-w: Can you change this to ES5 functions please?

```js
var self = this;
view.$el.fadeOut('slow', function() {
  self.destroyView(view);
  self.currentView.$el.fadeIn('slow');
});
```

I prefer the `() =>` syntax as well but this will prevent us having to explain the syntax to newer developers.

blikblum: done

docs/marionette.region.md
```diff
@@ -421,6 +423,28 @@ myRegion.reset();
...
+  onEmpty() {
+    if (!this.isSwappingView()) {
+      this.$el.append('Empty Region');
+    }    
+  }  
+});
+```
+[Live example](https://jsfiddle.net/marionettejs/c1nacq0c/1/)
+
+> `hasView` will not work reliably inside `onEmpty` because will always evaluate to false
```
paulfalgout: I'm not sure I understand this?  `hasView` _should_ always be false in `onEmpty`

scott-w: I understood it to mean "don't depend on `hasView` inside `onEmpty`"

paulfalgout: when you show a view.. it first empties the region.  `hasView` is entirely dependable in returning `false` inside of `onEmpty`  It should always return `false` inside of `onEmpty`.  While someone may assume that when a view is being swapped `hasView` would be true.. that is a misconception.  The region does not yet have the view until `onShow` _after_ the `show` in completed.  There is never a use case for `hasView` showing anything other than `false` inside of `onEmpty`  If `hasView` showed true inside of `onEmpty` that would effectively mean that the empty failed.

blikblum: There's no problem with hasView.
The note is to show that for the purpose of checking if a view is about to be rendered hasView is not an option inside onEmpty.

It can be better worded

blikblum: Or can be removed

blikblum: I removed, since does not add and can create confusion

paulfalgout:
I missed this, but this refactor was done on `master` not `next`..  Not sure how we want to proceed.

paulfalgout:
I cherry-picked this over on to `next`.  It's a little odd that it's in `master` now, but it should be ok

blikblum:
Let me know if needs to do other actions

paulfalgout:
I think it's ok now.. though check out https://github.com/marionettejs/backbone.marionette/issues/3357

blikblum:
I was puzzled by why this happened since i was sure that i branched this PR from next instead of master

Now, after creating a new PR i noticed that github offers, by default, the master branch to be used as base instead of next regardless of where the PR branch is forked. 

Not sure if we could do something about it

paulfalgout:
I think you can change the default, but it effectively just makes a different branch master which I don't think we want either

JSteunou:
Very nice works guys, this will definitely be useful and help reduce the number of loc in previous AnimatedRegion :) :clap:

Issue #2164, opened 2014-12-26
Global templateContext


jamesplease:
Template helpers are just extra data added to the data available in your template. There should be an easy way to configure global template helpers.

@jakobdamjensen implemented this recently in his app by modifying the templateHelpers prop on the View prototype. This isn't ideal because future views would need to `_.extend(View.prototype.templateHelpers, myTemplateHelpers)`, which should not be necessary.

There should just be a way to specify global template helpers. I'm thinking a possibility would be adding them to the Renderer object.

``` js
Mn.Renderer = {
  // global template helpers
  templateHelpers: {}
};
```

Then we would mix that in [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/view.js#L46-L53).

//cc @jakobdamjensen

bgaillard:
@jmeas  :+1: :+1: :+1:  This feature is definitly a must have, we need it in almost all our Marionette projects.

Because it "seems to be very simple" to implement and does not break the APIs it could be great to have it in a 2.5 version of Marionette (is this version scheduled ?). 

I know their is a discussion to rename `templateHelpers` into `templateContext` https://github.com/marionettejs/backbone.marionette/issues/2166 for Mn 3.X.

So for version 2.5 we could keep `templateHelpers` and change this name into `templateContext` for version 3.X ? 

```javascript
// Version 2.5
Mn.Renderer.templateHelpers = { ... };
// In Marionette View code
mixinTemplateHelpers = function(target) {
    target = target || {};
    var templateHelpers = this.getOption('templateHelpers');
    templateHelpers = Mn._getValue(templateHelpers, this);
    target = _.extend(target, Marionette.Renderer.templateHelpers); // Only one line added
    return _.extend(target, templateHelpers);
};

// Version 3.X
Mn.Renderer.templateContext = { ... };
// In Marionette View code
mixinTemplateContext = function(target) {
    target = target || {};
    var templateContext = this.getOption('templateContext');
    templateContext = Mn._getValue(templateContext, this);
    target = _.extend(target, Marionette.Renderer.templateContext);
    return _.extend(target, templateContext);
};
```

Temporarily in Mn 2.4.X we add this in our projects to make it work.
```javascript
/**
 * Function used to overwrite the "standard" Marionette `Marionette.View.mixinTemplateHelpers(target)` method to be 
 * able to use global template helpers.
 * 
 * @param {Object} target The data to mixin (i.e where to attach additional template helper methods).
 */
[...]

PR #3331, opened 2017-03-29
Support custom renderer

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/2474

Hopefully this allows for something like:
```js
import IDomRender from 'marionette-idom';
Marionette.View.setRenderer(IDomRender);
```

Todo:
- [x] Tests
- [x] Docs

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -146,8 +146,13 @@ const View = Backbone.View.extend({
     const data = this.mixinTemplateContext(this.serializeData());
 
     // Render and add to el
-    const html = Renderer.render(template, data, this);
+    const html = this._renderHtml(template, data);
     this.attachElContent(html);
+  },
+
+  // Renders the data into the template
+  _renderHtml(template, data) {
+    return Renderer.render(template, data, this);
   },
 
   // Get the template for this view
@@ -196,6 +201,11 @@ const View = Backbone.View.extend({
       .map('currentView')
       .compact()
       .value();
+  }
+}, {
+  // Sets the renderer for the Marionette.View class
+  setRenderer(renderer) {
+    this.prototype._renderHtml = renderer;
   }
 });
 
--- a/docs/marionette.renderer.md
+++ b/docs/marionette.renderer.md
@@ -1,15 +1,68 @@
-# Marionette.Renderer
+# Marionette Renderer
+
+The renderer that Marionette uses to render data into a template can now be set
+per `Marionette.View` class via the [`Marionette.View.setRenderer` function](#setting-a-renderer).
+This feature supersedes modifying the external [`Marionette.Renderer`](#deprecated-marionette-renderer).
+The default renderer is still the `Marionette.Renderer.render` method so that
+documentation still applies.
+
+## Documentation Index
+
+* [Deprecated Marionette.Renderer](#deprecated-marionette-renderer)
+* [Setting a Renderer](#setting-a-renderer)
+* [Basic Usage](#basic-usage)
+* [Pre-compiled Templates](#pre-compiled-templates)
+* [Custom Template Selection And Rendering](#custom-template-selection-and-rendering)
+* [Using Pre-compiled Templates](#using-pre-compiled-templates)
+
+## Deprecated Marionette.Renderer
+
+Customizing the `Renderer` object has been deprecated in favor of the [`setRenderer` method](#setting-a-renderer).
 
 The `Renderer` object was extracted from the `View` rendering process, in order
 to create a consistent and re-usable method of rendering a template with or
 without data.
 
-## Documentation Index
+## Setting a Renderer
 
-* [Basic Usage](#basic-usage)
-* [Pre-compiled Templates](#pre-compiled-templates)
-* [Custom Template Selection And Rendering](#custom-template-selection-and-rendering)
-* [Using Pre-compiled Templates](#using-pre-compiled-templates)
+You can set the renderer for a View Class by using the class method `setRenderer`.
+The renderer accepts two arguments. The first is the template passed to the view,
+and the second argument is the data to be rendered into the template. The renderer
+should return a string containing the result of applying the data to the template.
+
+```javascript
+Marionette.View.setRenderer(function(template, data) {
+  return _.template(template)(data);
+});
+
+var myView = new Marionette.View({
+  template: 'Hello <%- name %>!',
+  model: new Backbone.Model({ name: 'World' })
+});
+
+myView.render();
+
+myView.el === '<div>Hello World!</div>';
+```
+
+The renderer can also be customized separately on any extended View.
+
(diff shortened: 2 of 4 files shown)


denar90:
I like it, so small change bring powerful API.

scott-w:
Is there an advantage to doing it this way instead of allowing:

```javascript
import {Renderer} from 'backbone.marionette';
import IDomRenderer from 'marionette-idom';

Renderer.setRenderer(IDomRenderer);
```

blikblum:
The `renderHtml` method is a nice addition allowing to more granular control 

> Is there an advantage to doing it this way instead of allowing:

Or as is today:

```
import {Renderer} from 'backbone.marionette';
import IDomRenderer from 'marionette-idom';

Renderer.render = IDomRenderer;
```

`Marionette.View.setRenderer(IDomRender);` does the same as using `Renderer.render`  but IMO, `View.setRenderer` is a better abstraction. We must consider deprecating Renderer.render, and remove Renderer altogether if this is applied

paulfalgout:
Previous conversation https://github.com/marionettejs/backbone.marionette/pull/2911#issuecomment-197673522

Having the Render object is an extra function call, and I've hoping to potentially introduce other configurations via class functions:

`Marionette.View.setRenderer`
[`Marionette.View.setTemplateContext`]()
[`Marionette.View.setDomMixin`](https://github.com/marionettejs/backbone.marionette/issues/2164#issuecomment-197537666)(https://github.com/marionettejs/backbone.marionette/issues/3145#issuecomment-243033085)

Also looking back I realized this wasn't exactly correct.. I updated this so it is more granular on whichever view class you want to apply it to.

This makes me think that `renderHtml` should actually be `_renderHtml` and we would use the static functions for overriding it.

And that's mainly to keep people from having to really grasp prototypes.. setting `Marionette.View.prototype.renderHtml = function() { }` as on of the first things they might have to do when learning Marionette.

paulfalgout:
I agree about deprecating the other renderer..  Can't figure out how to add the dep warning in the code though since it would have to know if you overrode it..  

Honestly I thought we might want to even extract the current renderer into a plugin if we could figure out optional dependencies.. it includes the templatecache.. which if you're overriding the renderer is a chunk of code that'll never be used.

blikblum:
> This makes me think that renderHtml should actually be _renderHtml and we would use the static functions for overriding it.

IMO making `renderHtml` private is a step back.

Say i have a default renderer and want to customize a specific View.
With `renderHtml` public i would do:

```
MyView = Mn.View.extend({
  renderHtml: customRenderer
})
```

with `renderHtml` private:

```
MyView = Mn.View.extend({ })

MyView.setRenderer(customRenderer)
```
Both will have the same effect, but the former is consistent with current way of extending Mn classes and the later needs a second statement / call

Also, if the idea is to add also `Marionette.View.setTemplateContext`, to be consistent with the API setting a private method, `templateContext` would have to be made private also.

paulfalgout:
I'm a little conflicted on this.  I made this a 2nd commit so it could be easily reverted, but I'm just not sure there's much use case for setting the renderer from view to view that often.  I'm personally still leaning towards having one way to do something with a smaller API.

I also don't think this is quite the same as how `setTemplateContext` will hopefully work.
I'm hoping `setTemplateContext` will allow for:

```js
Marionette.View.setTemplateContext({ BASE_URL: '/' });
//...
MyFormView.setTemplateContext({ current_user_name: currentUser.get('name') });
//...
var myBasicsForm = new MyFormView({
  template: _.template(`
       <form action="<%- BASE_URL %>basics">
        Hello <%- current_user_name %>,
        <input type="text" value="<%- phone_number_display %>">
      </form>
  `),
    templateContext: {
       phone_number_display() {
          return formatNumber(phone)
       }
    }
});
```

So part of the nicety of these setters is that you don't have to know what it's extending or how it works.  And you'll never have to ask which method to use.

Also going back to original example, it'd be nice if most of the renderer were like plugins.
```js
import IDomRender from 'marionette-idom';
Marionette.View.setRenderer(IDomRender);
```
Asking the user to do that requires 0 knowledge of the renderer, but asking the user to override the `renderHtml` function brings up one added layer of complexity.

I'm definitely not super sold on `_renderHtml`, but it'd be great to get a concensus @marionettejs/marionette-core

scott-w:
I'd agree with your second commit and err on the side of making it private at first. We can always make it public in the next version if people want to use it.

blikblum:
Thinking more, having only `setRenderer` as an API is a better option than exposing `renderHtml` method. So has my OK.

Just take a look at the comment about the implementation

paulfalgout:
> comment about the implementation
@blikblum what comment specifically?

blikblum:
> Why not this.protoype._renderHtml = renderer ?

src/view.js
```diff
@@ -195,6 +200,11 @@ const View = Backbone.View.extend({
       .compact()
       .value();
   }
+}, {
+  // Sets the renderer for the Marionette.View class
+  setRenderer(renderer) {
+    _.extend(this.prototype, { _renderHtml: renderer });
```
paulfalgout: There it is :-)

paulfalgout:
@marionettejs/marionette-core 👀

paulfalgout:
@scott-w I'm nervous about modifying `Marionette.View.prototype` in the tests as there's only the one copy.
I'm not sure I 100% follow the list either.  And I wasn't sure about the standard renderer since that's currently tested in the renderer itself.. I suppose I can test that the standard rendered gets called if it isn't overridden...

scott-w:
@paulfalgout Markdown didn't put my list together correctly. I meant to do the following:

```
I'm thinking we should get some tests in to cover a few potential edge-cases:

1. `View` has custom renderer, inherit from `View`:
  a. `View` uses custom renderer (covered)
  b. Inherited view uses custom renderer
2. Inherit from `View` without custom renderer:
  a. Inherited view uses custom renderer
  b. `View` uses standard renderer
```

I think you got the gist of it anyway - basically just making sure the right renderer gets used in the different cases.

paulfalgout:
By 1b. do you mean the inherited view uses the same custom renderer as 1a?

scott-w:
Yeah, I'm assuming that extending a view should share the same renderer?

paulfalgout:
Ok so I have a `ViewClass` inherited from `Marionette.View` and a `SubViewClass` inherited from `ViewClass`.
I'm now checking to make sure:
`ViewClass` can have a custom renderer and if so that it doesn't affect `Marionette.View`, but it does affect the `SubViewClass`.
I'm also checking that the `SubViewClass` can have it's own custom renderer different than the class it inherited from.
Pretty sure that covers all of the scenarios.

JSteunou:
:+1: for me too

PR #3332, opened 2017-03-30
Add initialize event to behaviors

paulfalgout:
Resolves: https://github.com/marionettejs/backbone.marionette/issues/1579

The use-case here is that often times you want to initialize a behavior with things relative to the initialized view, but the behavior itself is initialized prior to the view's initialization so that it's events can be delegated.

This new event allows for developers to access view properties once the view is ready

```js
var MyBehavior = Marionette.Behavior.extend({
  onInitializeView: function(view) {
    view.$el.addClass('my-behavior');
  }
});
```

Todo:
- [x] Tests
- [x] Docs
- [x] Apply this to https://github.com/marionettejs/backbone.marionette/pull/3244 upon merge of one or the other

### DIFF
--- a/docs/marionette.behavior.md
+++ b/docs/marionette.behavior.md
@@ -332,9 +332,11 @@ The View + Behavior initialize process is as follows:
 2. Behavior is constructed
 3. Behavior is initialized with view property set
 4. View is initialized
+5. View triggers an `initialize` event on the behavior.
 
 This means that the behavior can access the view during its own `initialize` method.
 The view `initialize` is called later with the information eventually injected by the behavior.
+The `initialize` event is triggered on the behavior indicating that the view is fully initialized.
 
 [Live example](https://jsfiddle.net/marionettejs/qb9go1y3/)
 
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -64,6 +64,8 @@ const CollectionView = Backbone.View.extend({
     Backbone.View.prototype.constructor.apply(this, args);
 
     this.delegateEntityEvents();
+
+    this._triggerEventOnBehaviors('initialize', this);
   },
 
   // Instead of inserting elements one by one into the page, it's much more performant to insert
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -60,6 +60,8 @@ const CollectionView = Backbone.View.extend({
     this._initEmptyRegion();
 
     this.delegateEntityEvents();
+
+    this._triggerEventOnBehaviors('initialize', this);
   },
 
   // Internal method to set up the `children` object for storing all of the child views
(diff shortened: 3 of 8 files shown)


scott-w:
Why `initialize:view` instead of just `initialize`?

paulfalgout:
Hmm I dunno.. good question...  there will still be an `initialize` function so would that be a little redundant?

like:

```js
var MyBehavior = Marionette.Behavior.extend({
  initialize: function() {
    this.on('initialize', function() {
      console.log('didn't this already happen?');
    });
  }
});
```

scott-w:
Ahh, sorry I wasn't clear. I just meant why did you choose the name `initialize:view` instead of just `initialize` for the event? So the behaviour is:

```js
const MyBehavior = Behavior.extend({
  onInitialize() {
    console.log('View has initialized');
  }
});
```

paulfalgout:
it isn't the behavior initializing.. it is the view.
```js
var MyBehavior = Marionette.Behavior.extend({
  initialize: function() {
    this.on('initialize', function() {
      console.log('didn\'t this already happen?');
    });
   console.log('is this different than the other 2?');
  },
  // or
  onInitialize: function () {
    console.log(How is this different from `initialize`)?
  }
});
```
vs
```js
var MyBehavior = Marionette.Behavior.extend({
  initialize: function() {
    this.on('initialize:view', function() {
      console.log('View has initialized');
    });
  },
  onInitializeView: function () {
    console.log('View has initialized');
  }
});
```

paulfalgout:
I would say this event is analogous to something like `add:region` or 'add:child'

blikblum:
Nice addition.

As @scott-w pointed ':view' suffix does not match current event naming. Is also redundant and not consistent, all other view proxyed events, like `onRender`, do not use ':view' suffix being implicit that means "on view render".

It's not easy though, the only name i could find is `onAfterInitialize`

paulfalgout:
It is fair that the other proxied events don't get the `:view` suffix.  Although I'm not sure that's a good thing either.. what does `onDestroy` handle?

I'd be open to just `initialize` because of that.  all events are "after" so I dunno that I want to enter that arena.  we'll just need to heavily define that the `initialize` event has nothing to do with the behavior's `initialize` and only has to do with the view's.

@marionettejs/marionette-core maybe we could get more 👀  before proceeding?

blikblum:
In fact i agree `initialize` would be the most consistent with current naming.

JSteunou:
I would have been for `initialize` if not already taken by the behavior own initialize.

And indeed, all event from the view are proxied with the same name (`onRender`, `onAttach`), so `onViewInitialize` or `onInitializeView` would be a premier, but I'm in favour in order to not introduce breaking change neither confusion with actual behavior `initialize`.

Documentation will be very important.

paulfalgout:
@marionettejs/marionette-core I think this is GTG

PR #3333, opened 2017-04-02
Update babel to latest version (6.24)

blikblum:
### Proposed changes
 - Update babel to latest version (6.24)

With new babel version, the generated code for `CollectionView.reorder` is optimized being smaller and faster since it does not create an closure

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,12 +50,12 @@
     "underscore": "~1.8.3"
   },
   "devDependencies": {
-    "babel-core": "^6.24.0",
+    "babel-core": "6.24.0",
     "babel-eslint": "7.0.0",
-    "babel-plugin-transform-regenerator": "^6.22.0",
-    "babel-polyfill": "^6.23.0",
-    "babel-preset-es2015": "^6.24.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,12 +50,12 @@
     "underscore": "~1.8.3"
   },
   "devDependencies": {
-    "babel-core": "6.17.0",
+    "babel-core": "^6.24.0",
```
rafde: `"babel-core": "6.24.0",`
better to keep exact version instead of feature. In the past, testing and builds were inconsistent from feature updates.
same for the other changes

blikblum: Done

paulfalgout: I think this is fine, though now that we're using yarn the `^` just indicates how `yarn upgrade` works. The lock file should keep it from changing versions.  As long as everyone uses yarn that is.

blikblum:
This has the same problem as #3323 . It was pulled into master

paulfalgout:
I cherry picked everything on master that wasn't in next

PR #3334, opened 2017-04-03
Do not bind render method in View classes

### LINKED ISSUE #2372: Stop binding render to the view
At view construction, we bind render to the view.
https://github.com/marionettejs/backbone.marionette/blob/master/src/view.js#L10

This is 
1. Generally unnecessary
2. Unexpected by new users of Marionette
3. Kind of weird
4. A problem when testing

``` js
  initialize: function () {
      // yes, this could use modelEvents, but it's a general problem
      this.listenTo(this.model, 'change', this.render);
  }
});

describe('MyView', function () {
  it('calls render when the model changes', function () {
    var model = new Backbone.Model();
    sinon.stub(MyView.prototype, 'render');
    var view = new MyView({ model: model });
    model.set('foo', 'bar');
    expect(view.render).to.have.been.called;  // error! view.render is not a spy (it's a bound function that calls a spy) 
    expect(MyView.prototype.render).to.have.been.called; // the way we currently have to test this
  });
});
```

In v3, we should just remove the binding.

blikblum:
### Proposed changes
 - Do not bind render method in View and CollectionView
 - Adjust tests accordingly

Link to the issue:
Fixes #2372

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -47,8 +47,6 @@ const CollectionView = Backbone.View.extend({
   // option to pass `{viewComparator: compFunction()}` to allow the `CollectionView`
   // to use a custom sort order for the collection.
   constructor(options) {
-    this.render = _.bind(this.render, this);
-
     this._setOptions(options);
 
     this.mergeOptions(options, ClassOptions);
--- a/src/view.js
+++ b/src/view.js
@@ -32,8 +32,6 @@ const ClassOptions = [
 const View = Backbone.View.extend({
 
   constructor(options) {
-    this.render = _.bind(this.render, this);
-
     this._setOptions(options);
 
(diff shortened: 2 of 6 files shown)


paulfalgout:
This is technically breaking.  Should probably wait until `v4`

scott-w:
Looks good. I think it's worth adding a test from #2372 to test the behaviour is as we expect.

paulfalgout:
I'm not sure a test is necessary TBH. This is a pretty oddball line to add.  We don't test that any other function isn't bound.  However this function like most do require that the function be called with the context.  I guess we could test that `myView.render === View.prototype.render`, but again that'd be a test we could apply to every function.

blikblum:
Done

Issue #2538, opened 2015-04-29
Allow behaviors to be added/removed after instantiation


adorsk:
It would be handy to add/remove behaviors dynamically to views, after instantiation.

For example, 

```
var view = new MyView({...});
view.addBehavior('fooBehavior', {fooBehaviorOpt1: 'blah'});
view.removeBehavior('fooBehavior');
```

However, implementing this would require some clarification and refactoring of the behaviors system.

The clarification would involve defining how behaviors should cascade. See this comment: https://github.com/marionettejs/backbone.marionette/issues/2386#issuecomment-97202548

Also, for deletion cascades, if a behavior is removed, should its nested behaviors also be removed?

jasonLaster:
Interesting idea, I'm not opposed but fear it'd add some complexity. Let's see how many ppl want this.

paulfalgout:
I'm not for or against this at the moment. but I'm having a hard time coming up with a scenario in which I could use this effectively.  Can you provide some practical example?

adorsk:
Here are a few examples I had mind:

## Example 1: Dynamically making something draggable .

Say we are making a game with blocks that we can drag. Some of the blocks can be dragged right away, but others start out 'locked'. When we 'unlock' blocks, we want to add a 'draggable' behavior to them, but without blowing away the original view for the block.

Dynamic behaviors would allow us to do this. Perhaps something like this:
````
block.on('unlock', function() {
  block.addBehavior('draggable', {});
});
````

## Example 2: Dynamically creating components through a domain-specific language
Let's say we are creating components through a custom DSL. Our DSL has very low-level commands, like 'createWidget', 'addRegion', etc. 

We would like to have an 'addBehavior' command, so that we don't have to put behavior config in the 'createWidget' commands. This would make the DSL simpler to use.

For example, something like...
````
var commands = [
{cmd: 'createWidget', id: 'widget1', type: 'MyView', widgetOpts: {...}},
{cmd: 'addBehavior', target: 'widget1', behavior: 'draggable', behaviorOpts: {}}
]
````

Do these examples help?

cmaher:
I really want this (at least dynamically adding behaviors).  It would also provide a way to create behaviors at initialize, as opposed to now where they get created before

ianmstew:
Hey all, any further thoughts on this?

paulfalgout:
See https://github.com/marionettejs/backbone.marionette/issues/2386#issuecomment-290418709

PR #3335, opened 2017-04-03
Remove behavior or region from a view when they are destroyed

paulfalgout:
Kind of resolves https://github.com/marionettejs/backbone.marionette/issues/2538

@Haraldson on gitter brought up a scenario in which he'd like to destroy a behavior on a view.

behaviors can be destroyed in which they'll stop reacting to events, but the view's `triggerMethod` ignores the state of the behavior and calls the `onMethod` handler anyhow.  In researching this, it looked like regions should probably be handled the same way.  Ironically `childView`s in `NextCollectionView` will remove themselves on `destroy` as well.

This PR is mostly to get the conversation started.. we can split up the behavior/region stuff if necessary.

The changes to the tests were a result of less-than-ideal tests, not breaking behavior.

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -59,6 +59,8 @@ const Behavior = MarionetteObject.extend({
   destroy() {
     this.stopListening();
 
+    this.view._removeBehavior(this);
+
     return this;
   },
 
--- a/src/mixins/behaviors.js
+++ b/src/mixins/behaviors.js
@@ -89,6 +89,13 @@ export default {
     _invoke(this._behaviors, 'destroy', ...args);
   },
 
+  // Remove a behavior
+  _removeBehavior(behavior) {
+    // Don't worry about the clean up if the view is destroyed
+    if (this._isDestroyed) { return; }
+    this._behaviors = _.without(this._behaviors, behavior);
+  },
+
   _bindBehaviorUIElements() {
     _invoke(this._behaviors, 'bindUIElements');
   },
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -69,6 +69,7 @@ export default {
     this.triggerMethod('before:add:region', this, name, region);
 
     region._parentView = this;
+    region._name = name;
 
     this._regions[name] = region;
 
@@ -102,6 +103,12 @@ export default {
     delete this._regions[name];
 
     this.triggerMethod('remove:region', this, name, region);
+  },
+
+  // Called in a region's destroy
+  _removeReferences(name) {
+    delete this.regions[name];
+    delete this._regions[name];
   },
 
   // Empty all regions in the region manager, but
--- a/src/mixins/view.js
+++ b/src/mixins/view.js
@@ -136,10 +136,12 @@ const ViewMixin = {
     // remove children after the remove to prevent extra paints
     this._removeChildren();
 
-    this._destroyBehaviors(args);
-
     this._isDestroyed = true;
     this._isRendered = false;
+
+    // Destory behaviors after _isDestroyed flag
+    this._destroyBehaviors(args);
+
     this.triggerMethod('destroy', this, ...args);
 
     this.stopListening();
--- a/src/region.js
+++ b/src/region.js
@@ -357,6 +357,12 @@ const Region = MarionetteObject.extend({
 
   destroy(options) {
     this.reset(options);
+
+    if (this._name) {
+      this._parentView._removeReferences(this._name);
+    }
+    delete this._parentView;
+
     return MarionetteObject.prototype.destroy.apply(this, arguments);
   }
 });
--- a/test/unit/behaviors.spec.js
+++ b/test/unit/behaviors.spec.js
@@ -1142,7 +1142,7 @@ describe('Behaviors', function() {
     let behavior;
 
     beforeEach(function() {
-      behavior = new Behavior();
+      behavior = new Behavior({}, new View());
       this.sinon.spy(behavior, 'destroy');
     });
 
--- a/test/unit/mixins/behaviors.spec.js
+++ b/test/unit/mixins/behaviors.spec.js
@@ -411,6 +411,44 @@ describe('Behaviors Mixin', function() {
     });
   });
 
+  describe('#_removeBehavior', function() {
+    let behaviorsInstance;
+    let FooBehavior;
+    let BarBehavior;
+
+    beforeEach(function() {
+      behaviorsInstance = new Behaviors();
+      FooBehavior = Behavior.extend({});
+      BarBehavior = Behavior.extend({});
+
+      behaviorsInstance.behaviors = {foo: FooBehavior, bar: BarBehavior};
+      behaviorsInstance._initBehaviors();
+    });
+
+    it('should remove the behavior from the view\'s behaviors', function() {
+      const behaviorInstance = behaviorsInstance._behaviors[0];
+
+      behaviorsInstance._removeBehavior(behaviorInstance);
+
+      expect(behaviorsInstance._behaviors).not.to.include(behaviorInstance);
+      expect(behaviorsInstance._behaviors.length).to.equal(1);
+    });
+
+    describe('when the view is destroyed', function() {
+      it('should not remove the behavior', function() {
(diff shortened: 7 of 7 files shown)


Haraldson:
Thanks for creating the PR/issue – after the discussion I didn’t feel like I had the right idea about how to solve my problem, but I guess making sure a Behavior gets cleaned up properly upon destruction has its own value. 👍🏼

blikblum:
I've been looking into the performance of Mn in https://github.com/krausest/js-framework-benchmark and found that setting events is the most expensive operation after Dom manipulation by jQuery. The absolute  time is not high though.

On the other side, in terms of memory allocation, the events are by far the king. Backbone `onApi` allocates the double of jQuery `add`.  `listenTo` comes in a honored fourth place.

So i would avoid adding unconditionally another event listener to such edge cases. I believe is possible to implement without listening to destroy event

paulfalgout:
How is `onApi` increasing memory allocation though?  I could see how an error here would be using an anonymous function.  we should really only have one instance, but beyond that if it is just storing references to already allocated things, it shouldn't really increase the memory profile..

paulfalgout:
After digging into this a bit more.. if an element is listening to something for the first time or if an element is being listened to for the first time, string ids and event hashes are created.  As far as I can tell a brand new `listenTo` seems to take up about 300 bytes in memory.  A bit more if a new anonymous function is the handler.  

However if an object is already listening the memory hit is much less.  If the object being listened to is already being listened to it is also less.  

The 2nd listener between a listening object on a listened to object appears to add about 30 bytes.

I the case of regions we know the view listener will most likely already be listening.. though perhaps nothing else is listening to the region itself.  For behaviors it is less likely but still probably that the view is listening to something.

I'm not sure it'd be a huge hit to add this, still it's worth investigating further.

blikblum:
> The 2nd listener between a listening object on a listened to object appears to add about 30 bytes.

This for storing the event info

There's also the memory taken by the listeners that are not simple anonymous functions but closures. Also for each listener a second closure is created by Backbone through `_.bind`. Last time i checked each function / closure take about 70 bytes.

paulfalgout:
So I attempted removing the `onApi` from 5 events per view in the `monitor-view-events` module here: https://github.com/paulfalgout/backbone.marionette/commit/69c61952025f3b88f43277281e0e62377384bbee#diff-1e2bac97bb88a4999324b1687d82349eR68

I assumed I'd see a decent improvement running thousands of views, but it doesn't seem that drastic.  Granted `listenTo` is marginally bigger than a plain `on`, but am I missing something?  

screen shot 2017-04-04 at 11 38 24 pm

paulfalgout:
> Also for each listener a second closure is created by Backbone through _.bind

I don't think this is correct.  It looks like it goes to great lengths to pass the necessary context around so that it can be used in `.call(ctx...` on trigger.
And it doesn't seem to add a closure.. if you pass it an existing closure it looks like it only maintains the reference to it the entire time, so as far as I can tell no additional function is created _unless_ you use an anonymous handler.

blikblum:
> I don't think this is correct. It looks like it goes to great lengths to pass the necessary context around so that it can be used in .call(ctx... on trigger.

Sorry, you are right `listenTo` does not calls `_.bind`. I got confused by `delegateEvents` that calls `_.bind` at line:
```
this.delegate(match[1], match[2], _.bind(method, this));
```

> I assumed I'd see a decent improvement running thousands of views, but it doesn't seem that drastic. Granted listenTo is marginally bigger than a plain on, but am I missing something?

As i said the absolute time is not high compared to jQuery DOM manipulation that takes 10x more time than setting events, so changes in it will not have big impact in timing. 

The impact is bigger in memory consumption.

As you pointed setting an additional event is takes small amount of memory, but the created listener closure is not negligible.

To see how much memory a closure can take run the JS Allocation profile in Chrome and Create 1000 rows. `buildViewTrigger` which at each call creates a closure, allocates 114Kb (57 bytes per closure). Additionally, although not related to `listenTo`, `delegateEvents` add more overhead: 135kb with `_.bind`  410Kb with jQuery `add`. Disabling triggers in ChildView the `buildViewTrigger` obviously does not allocate memory, `_.bind` goes down to 32Kb and jQuery `add` also does not allocate.

src/mixins/regions.js
```diff
@@ -72,6 +72,11 @@ export default {
 
     this._regions[name] = region;
 
+    this.listenTo(region, 'destroy', () => {
```
rafde: `this.listenToOnce`? I wonder if this matters.

paulfalgout: I think `listenToOnce` is more complicated even though that's certainly what will happen here.

paulfalgout:
I'm personally still leaning on both of these getting in, though I do think avoiding the anonymous functions is a good idea. I'm not fully convinced that the memory allocation is an issue here and that most of our perf issues are wrapped up in mostly jquery-land

blikblum:
Aside from memory usage, does behavior triggers destroy event?

Since behavior already override destroy method i suggest the following (much less invasive):

 ```
// Stops the behavior from listening to events.
  // Overrides Object#destroy to prevent additional events from being triggered.
  destroy() {
    this.stopListening();

    if (this.view) this.view._unregisterBehavior(this);

    return this;
}
```

**Comment by paulfalgout** [comment]
Ha! you're right about that.  Ugh.. that.. makes this less appealing in general..

Let's say I destroy a behavior on a view.. the behavior may not get fully cleaned up until the view is destroyed anyhow.. since it doesn't trigger any events.  I understand why it doesn't trigger events.. due to the handler naming convention.. but maybe it should..  Still though, seems reasonable to do the `_unregisterBehavior` thing I think.. behavior knows about the view already.

**Comment by blikblum** [comment]
Another even less intrusive alternative.
Add a public `removeBehavior` method to view that would be called by user only when necessary.

**Comment by paulfalgout** [comment]
I'd lean towards making `destroy` seem correct vs adding view API.  A `removeBehavior` would suggest an `addBehavior` which is going to be a bad idea due to the re-`delegateEvents` needed.
Plus that would essentially be
```
removeBehavior(behavior) {
  this._unregisterBehavior(behavior);
  behavior.destroy();
}
```
I like your first suggestion.

**Comment by paulfalgout** [comment]
Ok went in a different direction that doesn't use listeners.  Tried to keep this as minimal as possible as far as perf hits.  @marionettejs/marionette-core 👀

File: src/mixins/behaviors.js [hunk 2]
```diff
@@ -89,6 +89,15 @@ export default {
     _invoke(this._behaviors, 'destroy', ...args);
   },
 
+  // Remove a behavior
+  _removeBehavior(behavior) {
+    // Don't worry about the clean up if the view is destroyed
+    if(this.view._isDestroyed) {
```
blikblum: `this._isDestroyed`

src/behavior.js
```diff
@@ -59,6 +59,10 @@ const Behavior = MarionetteObject.extend({
   destroy() {
     this.stopListening();
 
+    if (this.view) {
```
blikblum: I think is not necessary this check since view property is assigned unconditionally and in other methods we do not check

src/mixins/behaviors.js
paulfalgout: Yep sorry I had that fix.. just hadn't pushed

src/behavior.js
paulfalgout: Yeah.. tests fail without it, but I think you're right and tests aren't using a view or something correctly..  looking into it now.

src/mixins/behaviors.js
blikblum: How about:
```
  if (!this._isDestroyed) {
    this._behaviors = _.without(this._behaviors, behavior);
  }  
```

paulfalgout: Yeah I dunno.. I much [prefer guards](http://there4.io/2015/06/10/refactoring-with-guard-clauses-php-javascript/) myself..  and it's not uncommon in Mn to do so.. however in every case it's just `{ return; }` which I changed this to.. though.. if people feel strongly it's not a big deal in this function.

paulfalgout:
@rafde a bit of changes made since your last 👍 
Is this good to merge?

PR #3337, opened 2017-04-07
Make a childViewEventPrefix feature flag

paulfalgout:
This flag allow the user to disable automatic event bubbling across the app.

This can be a performance improvement and only helps to increase code organization when the use explicitly has to set the prefix.

I would recommend in v4 that this is `false` by default.

I'm also hoping to work on a blog post that will help explain this further.

### DIFF
--- a/docs/marionette.features.md
+++ b/docs/marionette.features.md
@@ -1,10 +1,29 @@
 # Features
 
 Marionette Features are opt-in functionality. That you can enable by setting `Marionette.FEATURES` in your app.
+It is a good practice to set these flags only once prior to instantiating any Marionette class.
+
+```javascript
+Marionette.setEnabled('fooFlag', true);
+
+var myApp = new MyApp({
+  region: '#app-hook'
+});
+
+myApp.start();
+```
 
 ##### Goals:
 + make it possible to add breaking changes in a minor release
 + give community members a chance to provide feedback for new functionality
+
+## `childViewEventPrefix`
+
+This flag is set to `true` by default.
+
+This flag indicates whether [`childViewEventPrefix`](./events.md#a-child-views-event-prefix)
+for all views will return the default value of `'childview'` or if it will return `false`
+disabling [automatic event bubbling](./events.md#event-bubbling).
 
 ## `triggersPreventDefault`
 
--- a/src/config/features.js
+++ b/src/config/features.js
@@ -1,6 +1,7 @@
 // Add Feature flags here
 // e.g. 'class' => false
 const FEATURES = {
+  childViewEventPrefix: true,
   triggersStopPropagation: true,
   triggersPreventDefault: true
 };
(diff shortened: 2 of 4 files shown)


docs/marionette.features.md
```diff
@@ -6,6 +6,14 @@ Marionette Features are opt-in functionality. That you can enable by setting `Ma
 + make it possible to add breaking changes in a minor release
 + give community members a chance to provide feedback for new functionality
 
+## `childViewEventPrefix`
+
+This flag is set to `true` by default.
+
+This flag indicates whether [`childViewEventPrefix`](./events.md#a-child-views-event-prefix)
+for all views will return the default value of `'childview'` or if it will return `false`
+disabling [automatic event bubbling](./events.md#event-bubbling).
```
scott-w: Possibly stating the obvious, but should this point out that the flag needs to be set _before_ a view is instantiated?

paulfalgout: for better or worse, `childViewEventPrefix` is currently calculated per proxied event so you can flip this switch on-and-off all day long and it'll respect what it is set to at the moment of a `triggerMethod`.

I think this'd be a good thing to set during view instantiation.  both `events` and `triggers` are setting during `delegateEvents`

I think the `childViewEventPrefix` `_.result` should move into `_buildEventProxies` along with the `normalizeMethods` [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/mixins/view.js#L194).  And we should default `_childViewEvents` and  `_childViewTriggers` to `{}` so we can remove the first checks

Honestly I think we should calculate those hashes such that the handler should be able to be as simple as:
```js
_childViewEventHandler(eventName, ...args) {
    if (this._childViewEvents[eventName]) {
      this._childViewEvents[eventName].apply(this, args);
    }

    if (this._childViewTriggers[eventName])) {
      this.triggerMethod(this._childViewTriggers[eventName], ...args);
    }

    if (this._childViewEventPrefix !== false) {
      const childEventName = this._childViewEventPrefix + ':' + eventName;

      this.triggerMethod(childEventName, ...args);
    }
  }
};
```

scott-w: So if I understand correctly, if we do:

```js
Marionette.setEnabled('childViewEventPrefix', false);

const myView = new Marionette.View();
myView.childViewEventPrefix() === false;

Marionenette.setEnabled('childViewEventPrefix', true);
myView.childViewEventPrefix() === true;
```

I'm just wondering if, when I'm building an app, I'd be tripped up by that flag being changed somewhere I didn't expect (granted it's something that you shouldn't be changing willy-nilly).

paulfalgout: I'm not sure how else we'd use these flags... right now they're all essentially
```js
if (flag) {
  doFlagThing();
} else  {
  doOtherThing();
}
```
So if you change the flag at some point after the start of your app, you're changing the behavior.

paulfalgout:
@marionettejs/marionette-core :eyes:

blikblum:
I would put a note highlighting that is a good practice to set this flag only once and at app start

paulfalgout:
Ok I added a note about how to use feature flags

blikblum:
Has my OK

Issue #3294, opened 2016-12-26
getRegion should render an unrendered view


paulfalgout:
You can certainly call render, but I don't know the value of `getRegion` erroring if the view hasn't been rendered.  I don't _think_ I would consider this breaking.

```js
const myApp = Mn.Application({
  region: '#app-hook',
  onStart() {
    const rootLayout = new MyLayout();

    rootLayout.isRendered(); // false
    rootLayout.showChildView('header', new MyHeader());
    rootLayout.isRendered(); // true
    rootLayout.showChildView('side-bar', new MySide());
    rootLayout.showChildView('main', new MyPreloader());

    // by pre-rendering and showing children this is a single html attach
    this.showView(rootLayout);  
  }
});

myApp.start();
```

https://jsfiddle.net/adhv48ky/4/

scott-w:
I like this - it should make Marionette easier for newcomers as well. A common error I see with people new to Mn (and one I made myself) is showing views in a region whose parent hasn't been rendered yet.

PR #3339, opened 2017-04-07
getRegion will render an unrendered view

erikahlswede:
### Proposed changes
 - getRegion will render the view if it hasn't been rendered yet
 - Updated tests
 - Added test

It feels a little wrong for `getRegion` to be responsible for this.  I tried moving the logic down to the `Region._ensureElement`, but that also didn't work well.

Is there a more pure spot to put this?

Link to the issue:
#3294

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -263,6 +263,7 @@ methods to help utilize regions.
 - `getRegion(name)` - Request a region from a view by name.
  - Note: If the view hasn't been rendered at this point, it will be.
 - `getRegions()` - Returns an object literal of all regions on the view organized by name.
+ - Note: If the view hasn't been rendered at this point, it will be.
 - `hasRegion(name)` - Check if a view has a region.
 - `emptyRegions()` - Empty all of the regions on a view.
 
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -261,6 +261,7 @@ In addition to adding and removing regions there are a few
 methods to help utilize regions.
 
 - `getRegion(name)` - Request a region from a view by name.
+ - Note: If the view hasn't been rendered at this point, it will be.
 - `getRegions()` - Returns an object literal of all regions on the view organized by name.
 - `hasRegion(name)` - Check if a view has a region.
 - `emptyRegions()` - Empty all of the regions on a view.
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -123,6 +123,9 @@ export default {
   // Accepts the region name
   // getRegion('main')
   getRegion(name) {
+    if (!this.isRendered()) {
+      this.render();
+    }
     return this._regions[name];
   },
 
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -123,7 +123,7 @@ export default {
   // Accepts the region name
   // getRegion('main')
   getRegion(name) {
-    if (!this.isRendered()) {
+    if (!this._isRendered) {
       this.render();
     }
     return this._regions[name];
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -86,7 +86,7 @@ export default {
 
   // Remove all regions from the View
   removeRegions() {
-    const regions = this.getRegions();
+    const regions = this._getRegions();
 
     _.each(this._regions, _.bind(this._removeRegion, this));
 
@@ -130,8 +130,15 @@ export default {
   },
 
   // Get all regions
-  getRegions() {
+  _getRegions() {
     return _.clone(this._regions);
+  },
+
+  getRegions() {
+    if (!this._isRendered) {
+      this.render();
+    }
+    return this._getRegions();
   },
 
   showChildView(name, view, ...args) {
--- a/src/view.js
+++ b/src/view.js
@@ -190,7 +190,7 @@ const View = Backbone.View.extend({
   },
 
   _getImmediateChildren() {
-    return _.chain(this.getRegions())
+    return _.chain(this._getRegions())
       .map('currentView')
       .compact()
       .value();
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -597,10 +597,8 @@ describe('region', function() {
           subRegion: '.sub-region'
         },
 
-        render: function() {
-          $(this.el).html('<div class="sub-region"></div><div>some content</div>');
-          this._isRendered = true;
-          this.triggerMethod('render');
+        template: function() {
+          return '<div class="sub-region"></div><div>some content</div>';
         },
 
         onRender: function() {
@@ -1035,9 +1033,6 @@ describe('region', function() {
   describe('when getting a region', function() {
     beforeEach(function() {
       this.itemView = new Backbone.Marionette.View();
-      this.itemView.template = function() {
-        return 'content';
-      };
       this.itemView.render = sinon.stub();
       this.itemView.addRegions({
         MyRegion: '#region',
@@ -1053,7 +1048,7 @@ describe('region', function() {
 
     it('should call render if getRegion is called without being rendered', function() {
       this.itemView.getRegion('whoCares');
-      expect(this.itemView.render.called).to.be.true;
+      expect(this.itemView.render).to.be.calledOnce;
     });
   });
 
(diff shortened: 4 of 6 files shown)


erikahlswede:
also, first contribution, so bear with me :)

src/mixins/regions.js
```diff
@@ -123,6 +123,9 @@ export default {
   // Accepts the region name
   // getRegion('main')
   getRegion(name) {
+    if (!this.isRendered()) {
```
paulfalgout: I think we need to use `_isRendered` instead of the method to support non-Marionette Views.

test/unit/region.spec.js
```diff
@@ -599,6 +599,7 @@ describe('region', function() {
 
         render: function() {
```
paulfalgout: wow.. this test was just bad to begin with..
I think this should be
```js
template: _.template('<div class="sub-region"></div><div>some content</div>'),
onRender:
```
then no need to override the `render` function..
sorry you had to mess with that 😞

test/unit/region.spec.js
```diff
@@ -1031,6 +1035,10 @@ describe('region', function() {
   describe('when getting a region', function() {
     beforeEach(function() {
       this.itemView = new Backbone.Marionette.View();
+      this.itemView.template = function() {
```
paulfalgout: if you're stubbing the `render` function it doesn't seem like the `template needs to be set.

test/unit/region.spec.js
```diff
@@ -1031,6 +1035,10 @@ describe('region', function() {
...
+        return 'content';
+      };
+      this.itemView.render = sinon.stub();
```
paulfalgout: there's a sinon attached to `this` that auto cleans up, so we tend to use it `this.sinon.stub()`

test/unit/region.spec.js
```diff
@@ -1042,6 +1050,11 @@ describe('region', function() {
     it('should return the region', function() {
       expect(this.itemView.getRegion('MyRegion')).to.equal(this.region);
     });
+
+    it('should call render if getRegion is called without being rendered', function() {
+      this.itemView.getRegion('whoCares');
+      expect(this.itemView.render.called).to.be.true;
```
paulfalgout: `expect(this.itemView.render).to.be.calledOnce`  That's sinon-chai and will produce a better error than "expected `false` to be `true`"

paulfalgout:
Regarding where this should go.. I think this is fine for now.

I realize right now that this is just pulling from a regions array, but hopefully in the future regions won't be instantiated until the first `getRegion`.  And some day we could set this up so that `myView.showChildView('.some-selector', new MyChildView());`

test/unit/region.spec.js
```diff
@@ -1042,6 +1050,11 @@ describe('region', function() {
     it('should return the region', function() {
       expect(this.itemView.getRegion('MyRegion')).to.equal(this.region);
     });
+
+    it('should call render if getRegion is called without being rendered', function() {
+      this.itemView.getRegion('whoCares');
```
scott-w: Should this be a region that's defined on the view e.g. `MyRegion` or `anotherRegion`?

erikahlswede: I use an arbitrary region because it shouldn't matter whether the region exists or not.  It will render the view regardless.

test/unit/region.spec.js
erikahlswede: Didn't realize this. 👍

test/unit/region.spec.js
erikahlswede: I should have realized this and cleaned it up :rofl:

erikahlswede:
@paulfalgout, @scott-w - Thanks for the review.  I should have address all your comments.  Please let me know if I can make any more changes.

docs/marionette.region.md
```diff
@@ -261,6 +261,7 @@ In addition to adding and removing regions there are a few
 methods to help utilize regions.
 
 - `getRegion(name)` - Request a region from a view by name.
+ - Note: If the view hasn't been rendered at this point, it will be.
 - `getRegions()` - Returns an object literal of all regions on the view organized by name.
```
paulfalgout: @marionettejs/marionette-core @erikahlswede thoughts?

erikahlswede: Sure thing

paulfalgout: It'll require a code change too.  `getRegions` isn't calling `render` or `getRegion`

erikahlswede: Ah, I see.  Will take care of it

erikahlswede: @paulfalgout - part of view destruction includes `removeRegions`, which will then `getRegions` and re-render the view.  Consider this:

```js
const foo = new View({
   regions: {
      blah: '.foo'
   }
});
foo.destroy();
```

Do you see an issue with `destroy()` rendering the view AFTER the UI elements are unbound, and the view el is removed?  

```
destroy(...args) {
    const shouldTriggerDetach = !!this._isAttached;

    this.triggerMethod('before:destroy', this, ...args);
    if (shouldTriggerDetach) {
      this.triggerMethod('before:detach', this);
    }

    // unbind UI elements
    this.unbindUIElements();

    this.removeEl(this.el);

    if (shouldTriggerDetach) {
      this._isAttached = false;
      this.triggerMethod('detach', this);
    }

    // VIEW RENDERED HERE <-------------------------------------------
    this._removeChildren();

    this._destroyBehaviors(args);

    this._isDestroyed = true;
    this._isRendered = false;
    this.triggerMethod('destroy', this, ...args);

    this.stopListening();

    return this;
  },
```

paulfalgout: We can safely change `removeRegions` to not call `getRegions`:
```js
  removeRegions() {
    const regions = _.clone(this._regions);
```

paulfalgout:
@marionettejs/marionette-core  Interestingly.. and maybe for another issue, but `getUI` could really follow suit here..  It would actually make that function a bit more useful.

src/view.js
```diff
@@ -190,7 +190,7 @@ const View = Backbone.View.extend({
   },
 
   _getImmediateChildren() {
-    return _.chain(this.getRegions())
+    return _.chain(this._getRegions())
```
erikahlswede: @paulfalgout do you see an issue with this?  Maintains functionality before auto-render was introduced.

src/mixins/regions.js
```diff
@@ -123,14 +123,24 @@ export default {
...
-  getRegions() {
+  _getRegions() {
     return _.clone(this._regions);
   },
 
+  getRegions() {
+    if (!this._isRendered) {
+      this.render();
+    }
+    return this._getRegions();
```
erikahlswede: Do you see an issue with the `_.clone` happening here?

docs/marionette.region.md
erikahlswede: @paulfalgout taken care of.  Can you comment on my questions?
https://github.com/marionettejs/backbone.marionette/pull/3339/files#r110733890
https://github.com/marionettejs/backbone.marionette/pull/3339/files#r110733839

src/mixins/regions.js
paulfalgout: The `_.clone` is good.

src/view.js
paulfalgout: Seems fine to me.

erikahlswede:
@scott-w what do you think?

blikblum:
Has my approval. Just wondering if is a breaking change since theoretically a view can be rendered in different order

paulfalgout:
Yeah I dunno.. doesn't seem like it would be..  I guess if the user is already rendering the view to get this to work, this'll introduce an extra render..  But I don't know that anything will break?  But maybe we can come up with a use case where it would be an issue?

scott-w:
It's possible an extra render could cause odd behaviour in apps such as fields with input already in.

I think the best way to handle that is to make it clear in the Release Notes for v3.3

paulfalgout:
I don't _think_ that is an issue as this only affects views that are not already rendered.. So this might only introduce an extra render before the view is shown.

scott-w:
You're right, I'll get myself another coffee :coffee: 

In which case, that's probably the only thing we'll need to mention in the release notes in case anyone is doing anything funky with views and rendering.

paulfalgout:
In that case I am merging ☺️

Issue #3320, opened 2017-03-06
View's region using el from child view, not own view


ambischof:
### Description
See this fiddle https://jsfiddle.net/hsrkbs8d/

If a view has more than one region and one region's view contains an element with the same selector as another region's selector and comes before it in the DOM and is rendered first, the element from the child view will be used.

I need it to happen like this because my view can be recursive and changing the order of rendering is not an option.

I think this is happening because the regions seem to be grabbing their elements on ensureElement, which may happen after other views have already rendered.

### Expected behavior
The elements for the regions should be picked only from the views own element, not child elements of other regions.

Seems like they could try to grab their element after the template is rendered and before any 'render' event handlers can be called.

### Actual behavior
Regions will find their element from childviews of other regions if the childview was rendered first and comes first in the DOM.

### Environment
See JSFiddle

paulfalgout:
I'm not entirely sure what a great fix for this is going to be.

An easy one is to make your selectors more specific.. such as this for the fiddle:
```js
regions: {
  c1: '> .child1',
  c2: '> .child2'
}
```

Otherwise, yes the view would need to loop through it's regions and set the `el` onRender before any child shows.

But Marionette is moving towards less of that, and more towards not instantiating the region at until a `getRegion` is called.  This will also allow for regions to be fairly hidden and created on the fly where a user could `this.showChildView('.some-region-selector', new MyView());`  Currently you can `addRegion` at any time, so developers still have to be aware of scope of children.

But this isn't a problem that is unique to regions.  `events` (and `triggers` kind-of) are scoped to any selector within the view tree.. so if you have:
```js
events: {
  'click button': 'onButtonClick'
}
```
on a recursive view structure then the top parent is going to handle the click for itself and all of it's children.

This is also true for `ui` and those same keys in `behavior`

Something that might be worth doing.. something similar to what @scott-w suggested before.. if a region is looking for an `el` and gets more than 1 results, we should probably throw at least a warning if not an error, to make this issue easier to understand.  But the same can't be done for `ui` or handlers where they could already be handling a number of elements.

As far as a bug is concerned I think this is a wontfix, but better documentation and ways of notifying the developer are probably needed.

scott-w:
I've added the `docs` label as I agree, if we're not going to fix this at this point then we'll need to let people know about the unexpected behaviour somehow.

ambischof:
[...]

PR #3341, opened 2017-04-09
throw error when region has more than one element

erikahlswede:
### Proposed changes
 - Throw an error if a region has more than one element

Link to the issue:
#3320

I played around with a warning mechanism...but at the end of the day, it was `console.warn`.  I like the idea of throwing an error in this situation.  This way, it is much more clear to the user, rather than having views show up in the wrong spots ☹️


paulfalgout:
@marionettejs/marionette-core feeling on whether or not this should be considered breaking?

erikahlswede:
When writing this, I figured it would be considered breaking. There may be cases where multiple elements are found for a region and the app functions perfectly fine (happen to be the first element in the list).

rafde:
@paulfalgout It's definitely a major and breaking change. This is best something for Mn 4.

scott-w:
I agree, this is a breaking change. While a developer _shouldn't_ add regions where selectors will hit multiple nodes, the fact is that they _can_ and, thus, probably have.

I'd happy :+1: this for Mn 4 as I think it's the right way to go.

JSteunou:
@erikahlswede could you rebase your branch ?

paulfalgout:
FWIW I'm no longer certain this should go in v4.  I think perhaps Region has some problems that could be resolved better than an error.

paulfalgout:
I think in the end the solution will be handled differently.  Too many caveats in the error.

PR #3345, opened 2017-04-12
Add onDomRemove event

paulfalgout:
This event is added to compliment `onDOMRefresh`.

`onAttach` and `onBeforeDetach` are great if you are concerned with the view's `$el` being available in the DOM.  `onDOMRefresh` is a combination of `onAttach` and subsquent renders which is useful if you are concerned with the view's _contents_ being in the DOM, but there is no equivalent `onBeforeDetach` for `onDOMRefresh`.  If you _must_ add something that is concerned with the view's contents being in the DOM the cleanup would look like:
```js
  template: MyTemplate,
  ui: {
    button: '.js-date-button'
  },
  onDomRefresh() {
    this.ui.button.showDatePickerPlugin();
  },
  onBeforeRender() {
     if (this.isRendered()) {
        this.ui.button.destroyDatePickerPlugin();
     }
  },
  onBeforeDetach() {
    this.ui.button.destroyDatePickerPlugin();
  }
});
```

While this event is technically a "before:" I think adding the prefix would complicate things.

Without this event `onDOMRefresh` is deceivingly useful and unsafe without a deeper understanding of the events.

With this event we can say:

|      method     |          use        |
| ------------- | ------------- |
| initialize         | Add children to pre-rendered view  |
| onRender | Add children to view (including re-rendering a prerendered view) |
| onAttach | Add plugins to the view's `el` |
| onBeforeDetach | Destroy plugins on the view's `el` |
| onDomRefresh | Add plugins to the view's content |
| onDomRemove | Destroy plugins on the view's content |

### DIFF
--- a/docs/viewlifecycle.md
+++ b/docs/viewlifecycle.md
@@ -66,8 +66,9 @@ part of the destruction lifecycle.
 |   1   |  `before:destroy` |       `view` - view being destroyed       |
 |       |                   | `...args` - arguments passed to `destroy` |
 |   2*  |  `before:detach`  |       `view` - view being detached        |
-|   3*  |  `detach`         |       `view` - view being detached        |
-|   4   |  `destroy`        |       `view` - view being destroyed       |
+|   3*  |  `dom:remove`     |       `view` - view being detached        |
+|   4*  |  `detach`         |       `view` - view being detached        |
+|   5   |  `destroy`        |       `view` - view being destroyed       |
 |       |                   | `...args` - arguments passed to `destroy` |
 
 The events marked with "\*" only fire if/when the view was attached to the DOM.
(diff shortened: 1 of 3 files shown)


docs/viewlifecycle.md
```diff
@@ -66,8 +66,9 @@ part of the destruction lifecycle.
 |   1   |  `before:destroy` |       `view` - view being destroyed       |
 |       |                   | `...args` - arguments passed to `destroy` |
 |   2*  |  `before:detach`  |       `view` - view being detached        |
-|   3*  |  `detach`         |       `view` - view being detached        |
-|   4   |  `destroy`        |       `view` - view being destroyed       |
+|   3*  |  `dom:remove`     |       `view` - view being detached        |
```
paulfalgout: I think what is getting confusing is storing the lifecycle events by creation and destruction.  The only reason that `dom:remove` is being triggered here is because the view is detaching.  `dom:refresh` docs say the view that is "rendering".    I'm not sure what it should say in the context of the view destruction..

scott-w: Yeah that's a fair point. Does `dom:remove` fire when you do `region.empty()`?

paulfalgout: it will fire on the view in the region only if the region is in the DOM

scott-w: I think I get it all now. We'll need a separate ticket to improve the documentation in this area to simplify it.

Issue #3354, opened 2017-04-18
compiled code of Array Destructuring Assignment  could not run on IE 9-


JounQin:
### Description

There are some destructuring assignment in source code, and it finally compiled into some codes which are not supported on IE 9- (`Array.isArray`, `Array.from`)
 
https://github.com/marionettejs/backbone.marionette/blob/master/src/mixins/behaviors.js#L61,L89

https://github.com/marionettejs/backbone.marionette/blob/master/lib/backbone.marionette.js#L700

### Expected behavior

do not use array destructuring assignment in source code.

### Actual behavior
should run well on IE9- also

### Environment

1. Marionette version: 3.2.0
2. Backbone version: 1.3.3
3. Additional build tools, etc: it is babel's wrong?

scott-w:
Good find @JounQin 

@marionettejs/marionette-core On the similar subject, have we documented the versions of IE that we support?

paulfalgout:
I believe ie9+ is documented on the readme. Can check later. This issue is not about destructuring. We should be able to transpile for ie9.

paulfalgout:
Nope just the saucelabs stuff, but that starts at IE10.  Honestly though I think that's a reasonable expectation.  IE9 support ended even for extended support April 11th 2017.  There will be no more security updates for that browser.  While this _is_ a problem for browsers we support.. I don't know that we should actively be concerned with IE9 and below.  To support those browsers we should require polyfills.

While we do destructuring all over the place.. it's only the 3 behavior instances that are causing the issue.  I think because we include the babel-polyfill in the test runner we're not seeing IE11 fail.. so I'm going to break it first.. then try to fix it.

paulfalgout:
Ok.. so _again_ further inspection #3355 does remove the `_toConsumableArray` issue.  However.. this problem is _only_ for IE8 and below as `Array.isArray` works in IE9, _and_ what we're passing to `_toConsumableArray` is _always_ an array and these are all private functions so `Array.from` will _never_ be an issue.

While I don't particularly care to support IE9.. I _really_ don't care to support IE8.  IE8 can polyfill `Array.isArray`
```js
if (!Array.isArray) {
  Array.isArray = function(arg) {
    return Object.prototype.toString.call(arg) === '[object Array]';
  };
}
```
[...]

PR #3355, opened 2017-04-18
Remove polyfill so the IE10/IE11 actually fail tests

paulfalgout:
WIP to resolve https://github.com/marionettejs/backbone.marionette/issues/3354

### DIFF
--- a/src/mixins/behaviors.js
+++ b/src/mixins/behaviors.js
@@ -58,12 +58,16 @@ export default {
 
   _getBehaviorTriggers() {
     const triggers = _invoke(this._behaviors, 'getTriggers');
-    return _.extend({}, ...triggers);
+    return _.reduce(triggers, function(memo, _triggers) {
+      return _.extend(memo, _triggers);
+    }, {});
   },
 
   _getBehaviorEvents() {
     const events = _invoke(this._behaviors, 'getEvents');
-    return _.extend({}, ...events);
+    return _.reduce(events, function(memo, _events) {
+      return _.extend(memo, _events);
+    }, {});
   },
 
   // proxy behavior $el to the view's $el.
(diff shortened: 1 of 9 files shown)


src/mixins/behaviors.js
```diff
@@ -81,7 +85,7 @@ export default {
     _invoke(this._behaviors, 'undelegateEntityEvents');
   },
 
-  _destroyBehaviors(args) {
+  _destroyBehaviors(...args) {
```
blikblum: Any specific reason why changing the method signature?

paulfalgout: destructuring function arguments does not cause IE to break because it can assume that `arguments` is already array-like.  But destructuring any variable requires the babel helper we can't use.  FWIW this function is private and only called in one location, so it shouldn't be an issue.
If we wanted to keep the arguments the same for some reason we'd have to create a `_.partial` function and then apply the arguments to it... that seems worse.

paulfalgout:
This does successfully remove the babel helper in the build causing the problem.. but the test suite doesn't really reflect that..  saucelabs is failing all the time without actually failing, so I can't really tell if it is doing anything at all.

src/mixins/behaviors.js
blikblum: Thanks for the info

paulfalgout:
I think this solves https://github.com/marionettejs/backbone.marionette/issues/3356
Saucelabs will always be a little flaky because of our concurrency issue, but this should definitely help improve that problem dramatically.  This also adds functional IE11 testing and fixes the previous IE11- issues on behaviors with babel destructuring.
@marionettejs/marionette-core 👀

blikblum:
With this we can remove dev dependency on babel-polyfill ?

blikblum:
Another candidate to be removed is babel-plugin-transform-regenerator . I could not find a yeld reference in source code

paulfalgout:
Good catch.. I don't think that `transform-regenerator` was ever used

Issue #3357, opened 2017-04-18
destroyView should return the destroyed view


paulfalgout:
In v3.3 (although already merged into master 😿 ) we're introducing `destroyView` to Region.  It returns `this` instead of the `view` being destroyed.  This is inconsistent with similar functions like `detachView` and `CollectionView`'s `addChildView`/`removeChildView` along with other functions that add relationships like `addRegion` and `removeRegion`  In every other case, ([other than `showChildView`](https://github.com/marionettejs/backbone.marionette/issues/3352)) the thing being added/removed is returned, and this function should do the same to be consistent.

It hasn't made a release yet so this would be nonbreaking to get it in now.

PR #3360, opened 2017-04-19
Returns destroyed view instead of this in destroyView

blikblum:
### Proposed changes
 - Returns destroyed view instead of this in destroyView

Link to the issue:
#3357


blikblum:
Fixed

PR #3361, opened 2017-04-20
Make the logo an absolute URL so npmjs.org isn't broken

paulfalgout:


blikblum:
CI failing?

paulfalgout:
This is master. Fix for CI is only in next

Issue #3322, opened 2017-03-15
region.show() not rendering Backgrid view


DavidUnzue:
### Environment
Marionette version: 3.2.0
Backbone version: 1.3.3
Backgrid version: 0.3.7
Build tool: browserify

### Description
I'm trying to integrate [Backgrid](http://backgridjs.com/) into a Marionette application. I want to show a Backgrid Grid, which is a view inheriting from Backbone.View, into a region of a Marionette view like this:
```javascript
var backgridView = new BackgridView({collection: myCollection});
mainRegion.show(backgridView);
```

### Expected behavior
The `.show()` function should trigger `render()` on `backgridView` and then attach it to the DOM.

### Actual behavior
`render()` is not being triggered on `backgridView`. The view will be attached to the DOM but won't be properly rendered before that, so that it won't show up on my application.

### Detailed issue description
In a nutshell: my `backgridView` is having the property `_isRendered` set to `true`, so that Marionette is preventing the view to be rendered (see [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L100)):
```javascript
_renderView(view) {
    if (view._isRendered) {
      return;
    }

    ...

    view.render();

     ...
 }
```

Now I was wondering, since a Backgrid view inherits from Backbone.View, why is it having the property `_isRendered` at all? And this is why:
When I do 
```javascript
mainRegion.show(backgridView);
```
then Marionette is doing this with my view (see [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L57)):
```javascript
view = this._getView(view, options);
```

Now the `_getView` function does the following (see [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L169-L171)):
```javascript
if (view instanceof Backbone.View) {
  return view;
}
```
So it checks if my View is just a plain Backbone view (a child of Backbone.View), and if that is the case, it leaves the view as it is. Otherwise, Marionette will do the following (see [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L173-L175)) and create a new Marionette view out of my view:
```javascript
return new View(viewOptions);
```

Now the thing is that my `backgridView` is actually an instance of Backgrid.Grid, which is again a child of Backbone.View. That means that `backgridView`, although being ultimately a backbone view, is not a direct child of Backbone.View. Because of that, the code:
```javascript
if (view instanceof Backbone.View) {
```
[...]

PR #3363, opened 2017-04-23
Some tweaks for #3322

scott-w:
I made a couple of changes to the documentation to cover necessary improvements for Backgrid and a general case for Backbone Views.

### Proposed changes
 - Add documentation for regular `Backbone.View`.
 - Add documentation for Backgrid view.
 - Link View documentation directly to Region documentation.

Link to the issue: #3322

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -25,6 +25,8 @@ Regions maintain the [View's lifecycle](./viewlifecycle.md#regions-and-the-view-
 * [Using Regions on a view](#using-regions-on-a-view)
 * [Showing a View](#showing-a-view)
   * [Checking whether a region is showing a view](#checking-whether-a-region-is-showing-a-view)
+  * [Non-Marionette Views](#showing-non-marionette-views)
+    * [Showing Backgrid Views](#showing-backgrid-views)
 * [Showing a Template](#showing-a-template)
 * [Emptying a Region](#emptying-a-region)
   * [Preserving Existing Views](#preserving-existing-views)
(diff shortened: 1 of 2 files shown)


docs/marionette.region.md
```diff
@@ -314,6 +316,60 @@ mainRegion.hasView() // true
...
+
+  template: _.template('<div class="child-view"></div>'),
+
+  onRender: function() {
+    this.showChildView('child', new MyChildView());
+  }
+});
+```
+
+#### Showing Backgrid Views
```
paulfalgout: I dunno that this issue is specifically related to backgrid.

I don't think there's an issue for Marionette to fix here as things preference a Marionette.View or a typical flow for a Backbone.View

However any view that puts more than a wrapper `el` into the view's `el` and then attempts to render _inside_ that `el` will see this same issue.
ie:
```js
var View = Mn.View.extend({
  el: $('<div><p>')
});

var view = new View();
view.isRendered(); // true
```

The reason this seems _ok_ is that the typical case for a view is that view creates the `el` as a single element for rendering within _or_ the view is pre-rendered.  So if the view is _partially_ pre-rendered then the user will have to know to `render` it again before or after showing it in the region.  And that'll be true for backgrid or any other case where this is a thing.

docs/marionette.region.md
```diff
@@ -314,6 +316,60 @@ mainRegion.hasView() // true
...
+display instances of regular [`Backbone.View`](http://backbonejs.org/#View).
+To do this, ensure your view defines a `render()` method and just treat it like
+a regular Marionette View:
+
+```javascript
+var Bb = require('backbone');
+var Mn = require('backbone.marionette');
+var _ = require('underscore');
+
+var MyChildView = Bb.View.extend({
```
paulfalgout: Do we want to show here that this view will also get events like `onRender`?

docs/marionette.region.md
```diff
@@ -503,28 +559,28 @@ Mn.Region.prototype.removeView = function(view){
...
-  
+
   removeView(view) {
     var self = this;
     view.$el.fadeOut('slow', function() {
       self.destroyView(view);
       if (self.currentView) self.currentView.$el.fadeIn('slow');
-    })    
-  }   
+    })
```
paulfalgout: seeing as there is a change on this line `;` wouldn't hurt

docs/marionette.region.md
scott-w: That makes sense, though I think it's important to reference specific libraries that do this here. My aim is to ensure users of libraries like Backgrid can just see straight away what they need to do without digging into the internals of how both Backgrid and Marionette work.

I'll come up with a better way of wording it for the docs.

docs/marionette.region.md
scott-w: Done 👍

blikblum:
Very good

PR #3367, opened 2017-05-01
Changed the order of the deprecation message, fixes #3366

### LINKED ISSUE #3366: Behavior `defaults` deprecation message is always triggered
### Description

The deprecation message regarding the Behavior `defaults` deprecation is always visible when DEV_MODE = true;  

### Expected behavior

When no behaviour is using the `defaults` property the message should not be displayed.

### Actual behavior

The deprecation message is always displayed as the `defaults` property gets set just above the line where it is checked:

```
this.defaults = _.clone(_.result(this, 'defaults', {}));

if (this.defaults) {
    deprecate('Behavior defaults are deprecated. For similar functionality set options on the Behavior class.');
}
```

https://github.com/marionettejs/backbone.marionette/blob/0186a48f1c72422f49731811f59e036ebca9f87e/src/behavior.js#L34-L38

### Environment

Doesn't matter.

wvankuipers:
### Proposed changes
Fix for #3366, changed order of check & set for behavior defaults property

Link to the issue: #3366

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -31,11 +31,12 @@ const Behavior = MarionetteObject.extend({
     // wants to directly talk up the chain
     // to the view.
     this.view = view;
-    this.defaults = _.clone(_.result(this, 'defaults', {}));
 
     if (this.defaults) {
       deprecate('Behavior defaults are deprecated. For similar functionality set options on the Behavior class.');
     }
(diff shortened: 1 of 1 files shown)


paulfalgout:
Good find 👏

PR #3368, opened 2017-05-01
Prevent additional region destroys from errorring on removeReferences

paulfalgout:
In the edge case where a parent view destroys itself when a child view destroys, a 2nd `region.destroy` is called during the `this.reset()`.  As such the destroy cannot bail out successfully and will try and re-`_removeReference` even though `parentView` was already deleted.
I believe the easiest solution is to also remove the `_name` references so that the removal is no longer in question.

Related `_removeReference` removes the need for `_removeRegion` to remove its own references as it will happen in the `region.destroy`

Failing example: https://jsfiddle.net/2py5oqr4/

### DIFF
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -1168,6 +1168,20 @@ describe('region', function() {
     it('should return the region', function() {
       expect(this.region.destroy).to.have.returned(this.region);
     });
+
+    describe('when the region is already destroyed', function() {
+      it('should not reset the region', function() {
+        this.region.reset.reset();
+        this.region.destroy();
+        expect(this.region.reset).to.not.have.been.called;
+      });
+
+      it('should return the region', function() {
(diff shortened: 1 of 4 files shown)


test/unit/region.spec.js
```diff
@@ -1168,6 +1168,20 @@ describe('region', function() {
     it('should return the region', function() {
       expect(this.region.destroy).to.have.returned(this.region);
     });
+
+    describe('when the region is already destroyed', function() {
+      it('should not reset the region', function() {
+        this.region.reset.reset();
```
rafde: I am probably not looking hard enough but the extra reset in `this.region.reset.reset()` looks off. I can't find ref to the extra `reset`. I'll take a better look at this when I get the chance.

paulfalgout: it's resetting the sinon spy

PR #3369, opened 2017-05-02
V3.3.1

paulfalgout:


paulfalgout:
@marionettejs/marionette-core I'll wait on the blog post and announcements of v3.3 until this gets pushed out since it's close anyhow.

PR #3370, opened 2017-05-02
Improve emulate collection perf

paulfalgout:
This improvement will really only affect `NextCollectionView` by moving the `_.values` loop out to the only childview container that needs it.

Additionally this replaces the `_.toArray` with the slice which is barely a perf improvement.. but is more consistent with how we've done this elsewhere.

Then I removed any internal use of the emulation as we can safely avoid it since the child view container is internal and gain even more perf

It's also worth noting that this is probably a slight perf decrease for the current collectionview emulated functions since it's adding the extra function call..  but that should be slight considering the other improvements

This should probably wait on v3.3.1 to be done before merging.

### DIFF
--- a/src/child-view-container.js
+++ b/src/child-view-container.js
@@ -12,12 +12,16 @@ const Container = function(views) {
   _.each(views, _.bind(this.add, this));
 };
 
-emulateCollection(Container.prototype, '_views');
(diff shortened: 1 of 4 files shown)


paulfalgout:
@marionettejs/marionette-core I think this is an easy win.

Issue #842, opened 2014-01-07
note ability to change this.$el.html to use thie.el.innerHTML for perf


samccone:
Was working on a pretty large app today, and noticed a weird bottleneck in render speeds

example
https://github.com/marionettejs/backbone.marionette/blob/e9b13519168b142f07aedf90438bb011c903f23d/src/marionette.itemview.js#L51

perf
http://jsperf.com/jquery-html-vs-innerhtml-the-better-way/3

cc @brian-mann @cobbweb 
## Update

So it seems we should not hard swap to using innerHTML
http://robflaherty.github.io/jquery-annotated-source/docs/10-manipulation.html

jquery does quite a bit of polyfill work for us as well as cleaning up nodes to prevent memory leaks.
This being said I think there are cases where you would want to use innerHTML instead. Perhaps we can leave a note in the source code exposing this perf boost ability.

PR #3372, opened 2017-05-06
Abstract DOM Mixin

paulfalgout:
The backbone way of using `instanceof` to check if the `el` is already a instance of `$` is 10x faster than re-wrapping the jquery.
https://jsperf.com/re-jquery-vs-instanceof/1

At first I intended to make this public, but it conflicts with the region function by the same name.  Rather than try and sort through that conflict I though the simplest approach was to get this in, and we can hash out naming in subsequent PRs.


blikblum:
This has a great potential to help significantly the performance. Aside from the raw performance, it should help also with reducing GC pressure since reduces temporary jQuery objects.

I just would implement as a local function instead of a method. Its used only in this DOM mixin (effectively an implementation detail) and not useful for other DOM mixins.

paulfalgout:
@blikblum I agree in spirit.. however I discovered this while working on some improvements on Region. I believe we can make use of `getEl` for Region directly for v4, and the contextual jquery won't be necessary as I think the view's themselves can do a better job of passing the correct `el`

Mind keeping it attached for future internal use?

blikblum:
> Mind keeping it attached for future internal use?

Being for internal use why not implement as a shared function being imported as necessary? When the time comes to be used elsewhere just import it 

As a method it will be attached to the View prototype polluting the namespace. Also other DOM mixin not using jQuery it wont use it. IMO in the mixin should go only the methods that necessarily should be overriden

paulfalgout:
Well I meant that it would eventual replace `findEls` in `Region` and will need to be public then.  I guess I can pull it out now and put it back soon.

paulfalgout:
I think I have a better solution.. one moment

paulfalgout:
Ok so I'm hijacking this PR for a bigger change:

Ideally dom mixins could be an external 3rd party plugins that could be enacted similarly to `setRenderer` by:
```js
// Everywhere
Marionette.setDomMixin(someMixin);
// or per class
Marionette.MyView.setDomMixin(someMixin);
```

This PR removes the mixin from being mixed into each class which greatly increased the API for each class.  Furthermore most classes already and still have ways other methods for overriding on a per class basis and ideally a mixin should be seen as a whole solution and not something where you want to modify how something gets attached in one particular instance.

This change allows the mixin to act more as a package and abstracts what might be difficult about modifying various prototypes while continuing the interface introduced with `setRenderer`.  It provides a very clear and easy path to implementing a mixin, even if the end user doesn't fully understand what it is doing under the hood.

The Dom "mixin" is exposed to each class so that users can reuse methods on the mixin as they choose.

Note that I moved the `beforeEl` to the `CollectionView` since we won't likely continue to use that interface for mixins after `CollectionView` is deprecated for `NextCollectionView`

Additionally since this is an experimental API I took the opportunity to make some of the names more similar.  We documented this specifically so we could make this kind of change without a major version bump.

TODO:
- [ ] Move and possible rename `DomMixin` to something more fitting?
- [ ] Fix test coverage
- [ ] Update docs

src/mixins/dom.js
```diff
@@ -3,17 +3,28 @@
 
 import Backbone from 'backbone';
 
+function getEl(el, context) {
+  return el instanceof Backbone.$ ? el : Backbone.$(el, context);
```
rafde:
```js
if (context) {
  return context instanceof Backbone.$ ? context.find(el) : Backbone.$(el, context);
}

return el instanceof Backbone.$ ? el : Backbone.$(el);
```
this code looks goofy, but it covers the `context` case.

rafde:
this is definitely nice and more flexible. `DomMixin` could just be `domAPI` or however to case.
I'll see if can help in this PR with docs and test.
Due to being stuck in Mn 2.4.x, I can't give any experience in dom mixin usage.

src/mixins/dom.js
paulfalgout: could be
```js
if (context) {
  return getEl(context).find(el);
} 
```

paulfalgout: though TBH I think we will benefit from _removing_ the context portion.  hrmm

paulfalgout: I dunno maybe we'll find it is still useful..  I _would_ like to remove the `parentEl` stuff from Region however.. and the only other contextual `getEl` call is in `CompositeView`, but I could also see utilizing the `DomApi` to replace the `view.$()` implementation..

rafde: eh... now that you pointed that out they should be separated out to `getEl(el)` and `findEl(context, el)` where `findEl will use `return getEl(context).find(el)`. Lumping them, currently, looks to be producing goofy fallbacks.

blikblum:
> This has a great potential to help significantly the performance

The above sentence was assuming that the el parameter was in majority of cases a jQuery object. But running an app i noticed that is always a HtmlElement thus a new jQuery object will be created at each call with no performance gain at all

paulfalgout:
In general the best way to handle the dom mixin is with plain dom elements, but views can be optimized for jquery by calling the dom api differently

blikblum:
I saw that a jquery object is passed to the domMixin in nextcollectionview . Maybe we can pass the $el in other places.

Also se should consider using native methods by default in trivial functions like append / before

blikblum:
After close look, jquery object is passed in CompositeView and not nextcollectionview

src/mixins/dom.js
paulfalgout: Eh I know why there's a fallback.. context is optional

src/mixins/dom.js
```diff
@@ -13,21 +13,27 @@ export function setDomMixin(mixin) {
 }
 
 export default {
-  getEl,
+  getEl({el, context}) {
+    if (context) {
+      return Backbone.$(el, context);
```
paulfalgout: This _cannot_ be `$context.find(el)` because it is breaking.  For some reason we have a test that account for `Backbone.$(someEl, notTheParent);` and expects it to work.. which I guess jquery handles?  super weird.

Anyhow.. I think the `parentEl` interface of Regions should go away.... so this can be more optimized soon.

paulfalgout:
OK.. soo.. I think we can just send as much info as we have to the the DomApi so that we can optimize for jQuery by default, but other api plugins can just use the vanilla elements if they want.

src/next-collection-view.js
```diff
@@ -558,20 +562,28 @@ const CollectionView = Backbone.View.extend({
...
     });
 
     return elBuffer;
   },
 
   // Override this method to do something other than `.append`.
   // You can attach any HTML at this point including the els.
-  attachHtml(collectionView, els) {
-    this.appendChildren(collectionView.el, els);
+  attachHtml(els) {
```
paulfalgout: Also worth mentioning.. small breaking change to `NextCollectionView`.  I really can't find a use case for passing the collectionview like this.  we're not doing it in `detachHtml` or anywhere else..  seems like now is a good time to make this very small break while it is experimental.

rafde: I think that part was meant for `CompositeView`. So it shouldn't really matter for `NextCollectionView`

rafde:
It was a lot simpler before but now the direction of this is very different now. Now I am seeing 2 sets of objects sent but they are both almost the same, except on is wrapped in jQuery. This looks confusing to me. It creates a "what should I send" moment.

paulfalgout:
We can send just vanilla in which case we have to rewrap jquery which is quite a bit slower. 
We can send just jquery and other plugins can pull the vanilla off of it, but then we are strengthening our reliance on jquery
We can send both.  I think _only_ the vanilla should be publicly documented.. Send jquery is only for internal use. Guess we could send those as a 2nd arg instead of a single arg?

blikblum:
+1 second (or last) arg. IMO no need to send the wrapped content just $el

paulfalgout:
I can recreate the jsperf, but the wrapped content was also quite a bit faster

blikblum:
My only concern is having a method signature with 4 arguments.

BTW `attachContents` is one of the methods that could be implemented natively by default. 

Since we pass a element or array of element we could write something like (quick and dirty)
```
if (_.isArray(contents)) {
   contents.forEach(el.appendChild, el)
} else {
   el.appendChild(contents)
}
```

I bet is faster than any jQuery variant

paulfalgout:
I don't think we can abandon `$append` while it's nearly `appendChild` it handles some special cases regard table rendering that would be breaking.  I think moving away from jquery is a good use-case for a DOM plugin.  

New attempt.  This keeps the same args except that it adds one additional that can be left undocumented.  It is entirely optional and uses `_`d property names to help indicate it's a private element.

blikblum:
By my side is better now and almost read to go. The only thing that i would do, aside the `before` thing (see review), is pass $el as a direct argument for almost all methods, leaving the hash argument only to `appendContents`. This saves some code related to destructuring.

src/collection-view.js
```diff
@@ -711,9 +712,14 @@ const CollectionView = Backbone.View.extend({
     return false;
   },
 
+  // Override to handle DOM inserting differently
+  beforeEl(el, siblings) {
+    this.Dom.getEl(el).before(siblings);
```
paulfalgout: So you would prefer that `beforeEl` go back on to the DOM API for the purpose of supporting the older `CollectionView`?

paulfalgout:
aside from [`beforeEl`](https://github.com/marionettejs/backbone.marionette/pull/3372#discussion_r115138861) I've renamed things and started documentation changes.  There is more to document and fix as well as some additional tests that need to be written.  Otherwise this is getting close.

src/collection-view.js
blikblum: As is now a native implementation for getEl would break this part of the code. I'd be happy with any solution that allows a native implementation

src/collection-view.js
```diff
@@ -690,9 +691,14 @@ const CollectionView = Backbone.View.extend({
     return false;
   },
 
+  // Override to handle DOM inserting differently
+  beforeEl(el, siblings) {
+    this.$(el).before(siblings);
```
paulfalgout: @blikblum how's this?

blikblum: This is good.

Although will not work with [Native.View](https://github.com/akre54/Backbone.NativeView/blob/master/backbone.nativeview.js#L69) but is something to be thought later

docs/dom.api.md
```diff
@@ -19,52 +19,56 @@ API. You will, however, [need to also handle Backbone's jQuery integration](#bac
...
 other DOM functions.
 
-### `appendChildren(el, children)`
+### `getEl(selector, context)`
 
-Takes the DOM node `el` and appends the rendered `children` to the end of the
-element's contents.
+Lookup the `selector` string within the DOM node for `context`. The optional
+`context` argument will come in as a DOM Node reference to run the `selector`
+search. If `context` hasn't been set, then `getEls` should search the entire
```
paulfalgout: True.  There's a handful of changes to make to the docs too.  I only modified them slightly so far.

paulfalgout:
Ok docs are updated.  This just needs a bunch of testing

paulfalgout:
So in trying to make an actual external DOM API I'm noticing 2 problems.  `getEl` or previous `findEls` is doing two jobs and it probably shouldn't be.  It needs to be used for query the DOM, but we're also using it to `this.$el = getEl(this.el)` even if `this.el` is already a DOM element.  We really can't use the DOM Api for this job... so I'm going to look into that.
The other is related and also related to https://github.com/marionettejs/backbone.marionette/issues/3379
Essentially we support both `detachEl` and `removeEl`.  In the case of `removeEl` `$.remove` is detaching but also removing any jquery listener.  This is fine, but outside of jquery there's no real equivalent.  So the removeEl would _also_ need to be able to somehow call `view.undelegateEvents`

We _could_ potentially call `undelegateEvents` ourselves and then just detach... _but_ that's probably technically breaking for anyone setting up other DOM listeners externally depending on Mn to auto-clean them up.  However other than that I'm not sure how to create a DOM API that can safely remove jQuery DOM interactions.

paulfalgout:
> With no jQuery when the element is removed the event listeners are automatically removed, so no need to call view.undelegateEvents

I don't think this is true. The dom api will not affect jquery handlers being added via events or triggers. So all the views will be leaky with remove. Similarly native.view also removes their added event handlers. To my knowledge vanillajs does nothing to clean up any other handlers on remove which is why our detach works at all.

blikblum:
> I don't think this is true. The dom api will not affect jquery handlers being added via events or triggers. So all the views will be leaky with remove

Seems you are trying to use native DOM calls while keeping the default delegateEvents that uses jQuery event system. In this case is not possible to have a pure DOMAPI. 

The method to cleanup you are looking for this case is jQuery.cleanData, not undelegateEvents, that should be used also `setInnerContents` 

See https://github.com/marionettejs/backbone.marionette/issues/842 for reference

paulfalgout:
That was the original intention of the DOM mixin. To provide alternative ways to deal with the DOM but its goal was never specifically remove jquery as a dependency. Since we handle unbinding ui on render and since everything else is delegated to the el, innerHTML should not be an issue, but undelegateEvents is not called on destroy because it is already handled by $.remove. So not unbinding things we bind to the $el will be leaky

blikblum:
> Since we handle unbinding ui on render and since everything else is delegated to the el, innerHTML should not be an issue,

If you use $.data or a jquery plugin in the view contents it will leak.

> undelegateEvents is not called on destroy because it is already handled by $.remove. So not unbinding things we bind to the $el will be leaky

Yes seems that $.off call is also necessary in removeEl

> To provide alternative ways to deal with the DOM but its goal was never specifically remove jquery as a dependency. 

Think the other way: keeping ability to use jquery events / plugins prevents using only native methods to manipulate the DOM. This is not bad per si and is not specific to Marionette. Its more like a tradeoff that we should be aware.

paulfalgout:
Marionette does not use $.data so that should not be a concern if you switch to a different DOM API. There is no doubt that different APIs will produce different caveats, but the current DOM API cannot be anything but jquery because of our use of remove and getEl

paulfalgout:
@marionettejs/marionette-core alright.. I made some arguably big changes.  I'll need to fix the docs again and still needs test.. _but_ I'm pretty confident that this is a clearer path to a jQuery DOM mixin.. it still doesn't work out of the box to remove jQuery, but I think with this you can abstract jQuery from Marionette's API... not accounting for jQuery reliance in backbone.

I also think that improvements to region will make this simpler in the future.

Notably I'm adding a feature flag for using the DOM API on view's `removeEl`  I don't see another way to make this work reasonably without being breaking.

I'm sure there'll be some stuff to hash out.  Bring it.

src/mixins/view.js
```diff
@@ -129,7 +129,7 @@ const ViewMixin = {
     this.unbindUIElements();
 
     // remove the view from the DOM
-    this.removeEl();
```
paulfalgout: @blikblum I just realized this was introducing new API to view that was only there before because of the DOM Mixin.  https://github.com/marionettejs/backbone.marionette/issues/3379

blikblum:
I think that the last changes regarding the element removal complicates the things:

1) Instead of removing the reliance on jQuery, it increase by calling `$el.remove` directly in `_removeElement`. We could arg that with the flag 'view$Remove' will not be called but than we stumble in the next issue

2) The feature flag 'view$Remove' is global making  impossible have different view classes in same app with different DomAPI

3) `_removeElement` is called twice, the second one being protected by _isDestroyed check. This introduces a level of indirection making the code hard to reason about. Also, anyone overriding `_removeElement` will need to do the check itself

IMO the reason to find a clear solution be so hard boils down to the Marionette need to separate the moment of removing the element and undelegating the events thus not calling `View.remove`. 

The Backbone.View lifecycle defines that the events are delegated through `delegate` and removed in `_removeElement` providing the hooks to customize the event delegation system. 

Currently in Marionette the undelegation occurs inside removeEl DOM API method because under jQuery the el and events removal occurs in the same method (`$.remove`) but it should not be responsibility of `removeEl` DOM API method to undelegate the events

My proposal is to revert this last change in remove element code (reintroducing dom.removeEl / stopListening calls). Later in another PR (this PR is already too big) we can expose the `_removeElement` method by calling `View.remove` so it can be overridden by third party mixins or View subclasses. So we would have two clear layers to customize DOM manipulation: the direct DOM manipulation done by domAPI interface and event delegation through `delegate` and `_removeElement` methods.

I have in mind a way that can achieve the following interface in a cleaner way with less code and no breaking change.

paulfalgout:
What I guess I am missing is outside of jquery dom.removeEl and dom.detachEl are redundant. In fact it is impossible to write a removeEl that does not rely on jquery. So how can we keep it?

paulfalgout:
Ok I reverted the `dom.removeEl` stuff along with the feature flag... it does seem though that a clear solution for `removelEl` should be considered here.  I also don't think we should expand the view API unless we have to.  Certainly doesn't have to solve the `view.remove` issue, but `removeEl` still makes a vanillajs solution impossible as far as I can tell.  @blikblum thoughts?

what about:
```diff
- this.Dom.removeEl();
+ this.unDelegateEvents();
+ this.Dom.detachEl(this.el, this.$el, true);
```
then:
```js
detachEl(el, _$el = getEl(el), _useRemove) {
  if (_useRemove) {
    _$el.remove();
  } else {
    _$el.detach();
  }
}
```

After reverting, this does keep removing the DOM API from templateCache and the CompositeView container.  I think those are both the correct option as templateCache is essentially it's own thing (that could reasonably be extracted from Marionette) and already has a method for changing loading templates.  And CompositeView was only using the DOM API for a single case and was using `this.ui` for others.. so `this.$` is the lowest dependency.

blikblum:
>  In fact it is impossible to write a removeEl that does not rely on jquery

Yes. The only way to write a native removeEl is when events are cleaned somewhere else 

```
detachEl(el, _$el = getEl(el), _useRemove) {
  if (_useRemove) {
    _$el.remove();
  } else {
    _$el.detach();
  }
}
```
Pretty good!

> After reverting, this does keep removing the DOM API from templateCache and the CompositeView container

Yes. Those are not a concern

paulfalgout:
OK one more thought..  related to https://github.com/marionettejs/backbone.marionette/pull/3381

`view.remove` is a public API for calling and is documented by Backbone, but we'll be ensuring that `view.remove` is only `view.stopListening` and it won't actually remove the view from the DOM.

While I don't know what use `remove` has in Marionette besides as a hook for some other plugins, it does seem like `_.noop`ing `_removeElement` is a breaking change.. particularly since right now `view.remove` does work as described even though it isn't called by `view.destroy`.  Someone _could_ be using it legitimately.

So it brings me back to wanting to utilize `_removeElement` internally, make it the same as nativeview essentially, but bailing if it is already destroyed.

```js
_removeElement() {
  if (this._isDestroyed) return;
  this.undelegateEvents();
  this.Dom.detachEl(this.el, this.$el, true);
}
```

If we don't do this and the user _does_ override `_removeElement` without adding the `isDestroyed()` check they will _still_ be calling `undelegateEvents` twice, and they'll be checking if there's a parentNode again.  It seems the same to me either way.

blikblum:
It's not a breaking change because is not currently being called.

> If we don't do this and the user does override _removeElement without adding the isDestroyed() check they will still be calling undelegateEvents twice

It will still be needed  the check with your proposed change the difference is that _removeElement will be called twice

Also your proposed change cleanup the events twice. One with the direct call to undelegateEvents and another through $.remove making less efficient for our default case (jquery)

Finally anyone overriding _removeElement in a Marionette view should be aware that Mn does not use It to remove El, this is responsibility of domapi. Something for the docs

paulfalgout:
`view.remove` is not being called internally, but it is a valid API for anyone using any `Marionette.View`.  We extend `Backbone.View` so I don't see why anyone would think `view.remove` should never be used (though granted you probably shouldn't, someone posted in gitter today an example where they were calling it)  I don't think we can safely change the behavior of `view.remove` in a minor release.

In any case, not considering #3381, since we're extracting the DOM API from being mixed in, a logic way to handle this is to put it back where it was, which was calling `_removeElement` directly.

The issue is that `_removeElement` calls `$.remove` and simply overridding it with a DOM API call that also calls `$.remove` makes cleaning up the view properly impossible via the DOM API alone.

So in order to make this non-breaking, but still allow for the DOM API to be modified without additional modification to the view, `$.remove` must be broken up.  So I added `undelegateAll` which cleans up the element in the same way `$.remove` does so that simply `$.detach` can be called via the DOM API.

It may seem as if perf would be an issue here, but in testing it is rather indiscernible such that the fastest case switches between all of the options that do not call `undelegateEvents` directly.  I tested this in chrome, canary, chrome mobile, and firefox with very similar results.

blikblum:
I think this is a good solution. There's the issue of the double `_removeElement` call if / when remove is called but as i said this is something to another PR.

Just one question is necessary `undelegateAll` method? Why not just call `off/removeData` directly in `_removeElement`?

src/mixins/view.js
```diff
@@ -150,6 +152,17 @@ const ViewMixin = {
...
+  _removeElement() {
+    this.undelegateAll();
+    this.Dom.detachEl(this.el, this.$el);
+  },
+
+  // Equates to $.remove clean-up
+  undelegateAll() {
+    this.$el.off();
+    this.$el.removeData();
+  },
```
blikblum: Is there any problem with `this.$el.off().removeData()` ?

paulfalgout: oh yeah good call

paulfalgout:
I'm ok with removing `undelegateAll`.. particularly for now.. it's not necessary for the dom api work Though `undelegateAll` (or something similar) is a friendlier method for overriding than `_removeElement` and having to recall the dom api function.

src/config/dom.js
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!parent) {
+      return;
+    }
+
+    parent.replaceChild(newEl, oldEl);
+  },
+
+  // Replace the contents of `el` with the HTML string of `html`
+  setContents(el, html = '', _$el = getEl(el)) {
```
blikblum: Setting html by default to an empty string is a breaking change.

Currently when a template function returns undefined `$.html` returns the innerHTML instead of modifying. With the change, will empty the element contents

Affects the ability to `_.noop` replace `template: false`

paulfalgout: Ah interesting.. Yeah I added this because `this.el.innerHTML = undefinedVar;` puts "undefined" in the html.  But we should probably just check if html is ! undefined before acting.

blikblum: > But we should probably just check if html is ! undefined before acting

Exactly what i was thinking, just don't Know if is something to another PR.

IMO we need a test to check that when template function returns undefined it does not try to set the contents. This is useful both for `_.noop` as replacement to template: false and for renderer not string based

paulfalgout:
I'm also done testing this :tada:.  I ran into one bug and one improvement that seems valuable at this time.  I don't like adding yet another thing, but it does seem like it greatly simplifies things for people making 3rd party DomApi for `appendContents` to not support arrays.  This requires the use of documentFragment, which in [perf testing](https://jsperf.com/mn-append-with-node-only/1) wasn't any different from what I could tell.  It's also worth noting that this only affects `CollectionView` and not `NextCollectionView`

@marionettejs/marionette-core @blikblum 👀

blikblum:
>  I don't like adding yet another thing, but it does seem like it greatly simplifies things for people making 3rd party DomApi for appendContents to not support arrays

The array support is pretty simple. Just need a check for array and a simple loop.

>  This requires the use of documentFragment, which in perf testing wasn't any different from what I could tell

Without modifications there's a 10% perf hit in my side.

Also the test is not effectively running the actual code path. The line to call `appendContents` should be modified to

`_.each(elsToReorder, function(el){ $(bufferEls).append(el); });`

Since document fragment is not an jQuery instance, one will be created each time appendContents is called

Other thing to consider is to test inside a table where each call to `append` or `html` has an extra overhead in jQuery. 

So i propose to revert this part.

BTW i tried to edit the jsperf without luck. Below is the relevant code plus the cited above:

```
<div id="root">
<table id="table">

</table>
</div>

var ChildView = Marionette.View.extend({
  tagName: 'tr',
  template: _.template('<td><%- id %></td>')
});

var myCollectionView = new Marionette.CollectionView({
  tagName: 'tbody',
  childView: ChildView,
  reorderOnSort: true,
  collection: myCollection
});

myRegion.show(myCollectionView);

myCollectionView.viewComparator = 'order';

var myCollectionView = new BufferCollection({
  tagName: 'tbody',
  childView: ChildView,
  reorderOnSort: true,
  collection: myCollection
});

myRegion.show(myCollectionView);

```

paulfalgout:
I'm not seeing anywhere near 10% perf.  Here's a better test also including the native variety: https://jsperf.com/mn-append-with-node
It's worth noting that jQuery is regularly faster than the native with loop and single appends.  By far the fastest case is using the buffer with native.

In any case, I'm going to suggest harder that adding 2 responsibilities for `appendChildren` for the DOM API is not a good idea.  For one it adds an `_.isArray` check to many parts of the app where it will never be necessary (in fact if you're only using Mn.View and Mn.NextCollectionView you will never need it)  Adding this requirement means needing a breaking change to remove it later, and we should be more careful here as this may include 3rd party plugins.

But it might be easier to suggest it's value (or lack there of) by explaining why we need it
.
Currently `appendChild` needs to support an array as well as a node for the children because:  If you're using `CollectionView` (not `NextCollectionView`) _and_ you have set `reorderOnSort` to true (which is not the default) _and_ you have to resort the collection _and_ you're not adding any models to the collection at the time, _only then_ we can _maybe_ get 5% perf improvement on the resort (it's consistently less in my testing) if you're using jQuery and not any other DomApi.

src/template-cache.js
```diff
@@ -77,7 +78,7 @@ _.extend(TemplateCache.prototype, DomMixin, {
   // using a template-loader plugin as described here:
   // https://github.com/marionettejs/backbone.marionette/wiki/Using-marionette-with-requirejs
   loadTemplate(templateId, options) {
-    const $template = this.findEls(templateId);
+    const $template = Backbone.$(templateId);
```
scott-w: So, as I understand it, we're constraining the DOM API to just operate on views - and removing it from `CompositeView` due to it being deprecated.

Do you see the `TemplateCache` implementation being removed in future or is it just generally something that people who aren't using jQuery tend not to need?

blikblum:
> It's worth noting that jQuery is regularly faster than the native with loop and single appends. By far the fastest case is using the buffer with native.

Under chrome i get the two versions basically the same. With Firefox the buffered is about 10% slower

> In any case, I'm going to suggest harder that adding 2 responsibilities for appendChildren for the DOM API is not a good idea.

IMO we should not worry about having a state of the art API for dom. AFAICS there will be only the native implementation with a reference implementation which as i pointed is easy enough to create (i've implemented one in 30min).

With this last change, in a collection with 100 rows, instead of calling the DOMApi (jQuery by default) one time will cal 101 times each one creating a new jQuery instance. And the original reason to this PR was reduce the number of new allocated jQuery instances!

This is my opinion. I dismissed my review and will let remaining core members to decide.

src/template-cache.js
paulfalgout: If `TemplateCache` solutions were popular they should have their own plugin that isn't tied to how the views use the DOM.  Like you should be able to change your `TemplateCache` solution and your `DomApi` solution independently of each other.

FWIW `TemplateCache` already has a method for determining how to load templates that is overridable and there's really only that single use, if it is even in use.  While the default implementation does use the DOM I don't think we'd suggest that it is the only place to store templates.  It's really doing a different job than views and regions.  The only thing currently using `getEl` is `Region` and I think a better implementation of Region would avoid Regions doing their own look-ups.  So I would highly bet that we'd have DomApi that would be specifically for `TemplateCache` and to me it is already redundant.

PR #3373, opened 2017-05-09
Documentation for 'triggerMethod' was stale

ianShoenberger:
'triggerMethod' documentation under 'EventBubbling' implies that the child view is passed automatically to the parent view's handler, however that is not the case in the marionette 3.x. I corrected the example so that 'triggerMethod' explicitly passes the view as an argument and, added a note after the Event Bubbling section and removed lines that explicitly stated that the view was passed as an argument.

### Proposed changes
 - updated Event Bubbling code example
 - added note about child view being passed explicitly 
 - corrected descriptions

Link to the issue:

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -247,7 +247,7 @@ var MyView = Mn.View.extend({
   },
 
   doSomething: function() {
-    this.triggerMethod('did:something');
+    this.triggerMethod('did:something', this);
   }
 });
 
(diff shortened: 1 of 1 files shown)


paulfalgout:
@marionettejs/marionette-core we need to update the example too.. it's doing some wild stuff to get around the issue this PR addresses.

PR #3348, opened 2017-04-13
Deprecate Behavior defaults

paulfalgout:
Resolve #3347 

This is actually an untested feature..  I added a test-ish

PR #3374, opened 2017-05-10
Do not inject merged options into View constructor arguments

blikblum:
### Proposed changes
 - Do not inject merged options into View constructor arguments

Currently, the Marionette View classes changes the constructor arguments to pass the merged options down to Backbone View constructor and thus to initialize.

The reasoning for this PR is:

1) Is not consistent with Backbone View and other Marionette classes that pass the options/arguments as is down to class hierarchy
2) The arguments manipulation prevents the code to be optimized by the JIT compiler 

If the developer needs to access the merged options in initialize it can use `getOption`, feature available in all Marionette classes.

Properties like `className` and `attributes` that are used by Backbone View can still be declared in the standard way which is more flexible than passing though options.

This change aligns with #3348 as it remove a redundant feature

It is a breaking change

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -13,6 +13,7 @@ patterns etc.
 * [Setting Options](#setting-options)
   * [The `getOption` Method](#the-getoption-method)
   * [The `mergeOptions` Method](#the-mergeoptions-method)
+  * [The `options` Property](#the-options-property)
 
 ## Class-based Inheritance
 
@@ -254,3 +255,35 @@ In this example, `model` and `something` are directly available on the
 `MyObject` instance, while `another` must be accessed via `getOption`. This is
 handy when you want to add extra keys that will be used heavily throughout the
 defined class.
+
+### The `options` Property
+
+Is possible to define an `options` property in the class definition that will
+be used as default values. The `options` argument passed at class instantiation
+has precedence over the property.
+
+> The `options` argument passed in `initialize` method is equal to the passed at 
+> class instantiation. To get the option inside initialize considering the 
+> `options` property is necessary to use `getOption`  
+
+```javascript
+var Bb = require('backbone');
+var Mn = require('backbone.marionette');
+
+var MyObject = Mn.Object.extend({
+  options: {
+    foo: 'bar',
+    another: 'thing'
+  },
+  
+  initialize: function(options) {
+    console.log(options.foo) // undefined
+    console.log(this.getOption('foo')) // 'bar'
+    console.log(this.getOption('another')) // 'value'            
+  }
+});
+
+var myObject = new MyObject({    
+  another: 'value'
+});
+```
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -258,9 +258,9 @@ defined class.
 
 ### The `options` Property
 
-Is possible to define an `options` property in the class definition that will
-be used as default values. The `options` argument passed at class instantiation
-has precedence over the property.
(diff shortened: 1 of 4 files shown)


paulfalgout:
I think I'm good with this.  Though do we even care to keep the `options` as something you can declare on the instance?  If we keep it we need to clearly define that backbone view options cannot be defined on the options hash defined on the view.
Although this _could_ be a non-breaking change I think if we added the view options to our `ClassOptions` for merging.  What's faster double merging or optimizing for the JIT?  Then we can get the JIT now and remove the `ClassOptions` later?

blikblum:
> Although this could be a non-breaking change I think if we added the view options to our ClassOptions for merging

It would still be breaking since the options argument passed in initialize will not be the merged one

I think we can wait, the perf gains is pretty small. The main reason for it is a saner API

paulfalgout:
This same thing should happen for all of the views. https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L62

Could be argued that we can make this change to `NextCollectionView` now

blikblum:
This PR already does change View, CollectionView and NextCollectionView

paulfalgout:
oh heh.. sure enough

paulfalgout:
Needs a rebase.. maybe doc updates too

blikblum:
Done with docs. I will write the upgrade notes in a different PR

docs/basics.md
```diff
@@ -254,3 +255,35 @@ In this example, `model` and `something` are directly available on the
 `MyObject` instance, while `another` must be accessed via `getOption`. This is
 handy when you want to add extra keys that will be used heavily throughout the
 defined class.
+
+### The `options` Property
+
+Is possible to define an `options` property in the class definition that will
+be used as default values. The `options` argument passed at class instantiation
+has precedence over the property.
```
scott-w: Reading the code example below, I'm taking this to mean:

> Options passed in will be merged with this `options` property.
> The arguments passed in at instantiation will override properties set on the `options` attribute.

blikblum: Updated

docs/basics.md
```diff
@@ -254,3 +255,35 @@ In this example, `model` and `something` are directly available on the
...
+
+### The `options` Property
+
+The Marionette classes accept an `options` property in the class definition
+which is merged with the `options` argument passed at instantiation. The 
+values from the passed in `options` overrides the property values.
+
+> The `options` argument passed in `initialize` method is equal to the passed at 
+> class instantiation. To get the option inside initialize considering the 
+> `options` property is necessary to use `getOption`
```
scott-w: Do you need the `>` here?

blikblum: The idea is to highlight the behavior which is being changed by this PR

paulfalgout: I think the general docs should really be present timelessly and not indicate what is new

blikblum: This excerpt does not have a reference of the change between versions or even that had a different behavior so is time lees. The objective is to highlight an aspect that could be wrongly interpreted by reading the text above where it says that the options are merged, so the user could think that this is the case for the options passed in initialize.

Anyway if there's objection i can remove the remark

paulfalgout: works for me

paulfalgout:
It's unlikely the view constructors will change now, so we should be able rebase this and get it in

blikblum:
> It's unlikely the view constructors will change now, so we should be able rebase this and get it in

Done

PR #3377, opened 2017-05-15
Improve Object's internal extendability

paulfalgout:
Internally we extend `Region`, `Behavior`, and `Application` from `Object`.
For each of these cases we call `_setOptions` in the various classes, but then it is also called again by `Object`.  This adds a minor perf improvement by not re-calculating `this.options`.
Similarly while looking into `Region` refactors I found it would be useful to know the `cid` in order to throw an error, but it wasn't available to the constructor.  This would allow us to set it once reliably.

### DIFF
--- a/src/mixins/common.js
+++ b/src/mixins/common.js
@@ -11,19 +11,19 @@ export default {
 
   // Imports the "normalizeMethods" to transform hashes of
   // events=>function references/names to a hash of events=>function references
-  normalizeMethods: normalizeMethods,
+  normalizeMethods,
 
-  _setOptions: _setOptions,
+  _setOptions,
 
   // A handy way to merge passed-in options onto the instance
-  mergeOptions: mergeOptions,
+  mergeOptions,
 
   // Enable getting options from this or this.options by name.
-  getOption: getOption,
+  getOption,
 
   // Enable binding view's events from another entity.
-  bindEvents: bindEvents,
+  bindEvents,
 
   // Enable unbinding view's events from another entity.
-  unbindEvents: unbindEvents
+  unbindEvents
 };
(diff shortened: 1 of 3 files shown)


blikblum:
I'm not fan of this change. 

First i doubt will have any measurable impact in performance in a Marionette app. I've been profiling a few and there's no trace of _setOptions being significant even before it was optimized by #3376.

Also is pretty hacky, mutating the object instance and changing the method behavior on the fly.

blikblum:
BTW if is a big deal, the same behavior can be achieved with (in Object constructor):

```
if (!this.hasOwnProperty('options')) this._setOptions(options);
```

The same for cid

paulfalgout:
It cannot as part of `_setOptions` is merging `this.options` into `this.options`

We could set a flag I guess.  Backbone recommends using `_.once` but that seems excessive.

blikblum:
> It cannot as part of _setOptions is merging this.options into this.options

It merges this.protoype.options into this.options. At the time Object constructor is called this will already be done in Region / Application / Behavior constructors.

The PR makes a noop the second call to _setOptions, my proposal avoids the second _setOptions  call which AFAICS has the same effect.

paulfalgout:
ok fair enough.  I went ahead and made logic guards.  I agree this is a pretty limited perf if any and am ok with removing the changes to `setOptions` entirely.  thoughts?
If that is the case the change to setting cid can be moved to the PR where it would need it.

blikblum:
I'm ok with keeping or removing the `_setOptions` changes. There's a secondary potential benefit in keeping it: the first object stored in this.options looses all references at the time is overwritten by the second call being a candidate for garbage collection. This is hard to measure. Tried to find a tool to show where the objects that triggers GC come without luck.

I would just do a change: instead of checking for `this.hasOwnProperty('options')` in the function wrapper i would do it inside MarionetteObject constructor. This saves a function call and also the check will not be executed when called by  Marionette.View that calls `_setOptions` only once

paulfalgout:
OK.  This should be good.  Best part of this PR was probably refactoring the object tests.. there was one test that was just testing listening to an event on a model and wasn't spying on the correct handler.

blikblum:
Very good!

Issue #3379, opened 2017-05-16
Backbone.View.remove is not called by Marionette.View.destroy


blikblum:
### Description

Starting from https://github.com/marionettejs/backbone.marionette/commit/1459e746ebea3f57d6ab2f7d7e6a14880becf83c#diff-9e0fc915cc771da1abb32dc1d433cf7a `Marionette.View.destroy` does not calls `Backbone.View.remove` anymore.

Functionality wise there's no loss but can affects libraries that monkey patch `Backbone.View.remove` like stickit

Re-adding call to `remove` will require a non trivial refactoring

### Expected behavior

If doable, call `Backbone.View.remove` otherwise document this change

### Actual behavior

`Marionette.View.destroy` does not calls `Backbone.View.remove`

### Environment

1. Marionette version: 3.x
2. Backbone version: 1.3.3

paulfalgout:
calling `remove` will `stopListening` prior to things that are potentially listening.

paulfalgout:
specifically https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/utils/view.spec.js#L24
or
https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/view.child-views.spec.js#L241

depending on where `remove` would be called.

blikblum:
One alternative is overriding `_removeElement` with a noop so basically we could call in place of stopListening

blikblum:
Closing as no relevant anymore

PR #3381, opened 2017-05-19
Call Backbone.View.remove inside Marionette.View.destroy

blikblum:
Link to the issue:
#3379 

Related to #3372 

Tests are WIP

### DIFF
--- a/test/unit/region.spec.js
+++ b/test/unit/region.spec.js
@@ -111,9 +111,9 @@ describe('region', function() {
         });
 
         it('should not render the view', function() {
-          this.sinon.spy(this.region, '_renderView');
+          this.sinon.spy(this.myView, 'render');
           this.region.show(this.myView);
-          expect(this.region._renderView).not.to.have.been.called;
+          expect(this.myView.render).not.to.have.been.called;
         });
       });
     });
@@ -749,7 +749,6 @@ describe('region', function() {
 
       this.sinon.spy(this.view, 'destroy');
       this.sinon.spy(this.region, 'attachHtml');
-      this.sinon.spy(this.region, '_renderView');
       this.sinon.spy(this.view, 'render');
       this.region.show(this.view);
     });
@@ -762,8 +761,8 @@ describe('region', function() {
       expect(this.region.attachHtml).not.to.have.been.calledWith(this.view);
     });
 
-    it('should not call "_renderView"', function() {
-      expect(this.region._renderView).not.to.have.been.called;
+    it('should not call "render" on the view', function() {
+      expect(this.view.render).not.to.have.been.called;
     });
   });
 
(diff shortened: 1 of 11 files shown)


test/unit/region.spec.js
```diff
@@ -901,18 +901,15 @@ describe('region', function() {
         })
       });
 
-      this.MyView = Backbone.View.extend({
-        render: function() {
-          $(this.el).html('some content');
-        },
-
-        destroy: function() {}
+      this.MyView = Backbone.Marionette.View.extend({
```
paulfalgout: these tests were/are awkward. 
I think the point of them is to check that region doesn't double destroy a view.  For Marionette.View this is impossible since `destroy` guards against this, but for a `Backbone.View` it is possible to double remove.  I don't know that changing this to a `Marionette.View` is helpful

paulfalgout:
In #3372 you mentioned:
> Also, anyone overriding _removeElement will need to do the check itself

Wouldn't that also be the case here?  What is the purpose of `_removeElement` in Marionette?

If it is intended that the DOM API could be disabled for view, but `view.remove` could be used, it is much less performant to remove the children prior to removing the parent from the DOM.  _and_ that would require suggesting overriding an `_`'d method, which is counter to Mn... though I do think that Backbone exposes it for this purpose... so maybe that's a grey area.

blikblum:
> Wouldn't that also be the case here?

With this PR _removeElement is called only once and e.g. backbone.nativeview with a native dom api will work out of the box. The check that is already done is if el has parentNode, not if is destroyed a Marionette specific thing 

>  What is the purpose of _removeElement in Marionette?

_removeElement is a public API provided by backbone, despite the low dash in name. In MN since we do its job in other place is noop by default

> If it is intended that the DOM API could be disabled for view, but view.remove could be used, it is much less performant to remove the children prior to removing the parent from the DOM.

By disable DOM api do you mean doing dom.removeel a noop? Besides the validity of such use case, there no performance concern because the element would not be removed at all, as intended by who wrote the dom api

test/unit/region.spec.js
blikblum: Tests are wip. We should test BB.view as before, but should  also test MN.view were we are not calling remove

paulfalgout:
I do not believe there is a single use-case in Marionette for overriding `_removeElement`.  It should not be overridden and remain a `_.noop` and should be documented as such.  If it happens to get overridden it's just should be entirely to realize that the el has already been removed and do nothing.
If that's the case I can get behind this, otherwise I'm going to need some descriptions of why it could ever be useful.

blikblum:
> If that's the case I can get behind this, otherwise I'm going to need some descriptions of why it could ever be useful.

backbone.nativeview override it

In fact we cannot control not being called because is a Backbone.View thing

paulfalgout:
nativeview overrides it, but nativeview isn't a marionette.view so it shouldn't be an issue.  I can't see a use-case for overriding it in marionette or even suggesting that it should ever be overridden on a marionette.view

test/unit/region.spec.js
paulfalgout: I don't think region needs to worry about what things might be called within `View.destroy` if we want Region's unit tests to make sure we don't call `View.destroy` multiple times that makes sense to me, but what happens after that points seems likes something that should be tested on the view

blikblum:
Yes. I think we should wait for #3372 or even considering not calling `Backbone.View.remove` at all

blikblum:
This can be closed since the domApi refactor made it unecessary

thatsmeta:
Just to be sure: Is `view.remove` called now? Or do I need to call it myself, e.g. when I use Backbone.Stickit? I do not understand why the domApi refactor made it unnecessary. Will I have to call it manually, if I implement my own domApi?

paulfalgout:
In v4 `view.remove` is called for non-marionette views https://github.com/marionettejs/backbone.marionette/blob/next/src/common/view.js#L37

We should have a v4 release candidate soon

thatsmeta:
So if I use Marionette 3.5.1, a `Marionette.View` + `Backbone.Stickit`, I will have to call `unstickit` myself, otherwise the stickit listeners will not be properly removed when the view is destroyed?

paulfalgout:
Oh actually I misread this...  `_removeElement` not `remove` is called now in v3.5.1 and v4 which isn't what happened before.. however this does not call `view.remove`  you will need to call unstickit as part of `before:destroy`.

thatsmeta:
Ok, that's good to know, thanks for your help.

Issue #3340, opened 2017-04-07
Consider publishing a ES6 module build through pkg.module


blikblum:
### Description

 1. Add a  ES6 module build through pkg.module
 
### Expected behavior

Distributes an optional build that leverages benefits of ES6 modules like tree shaking. See https://github.com/rollup/rollup/wiki/pkg.module  

### Actual behavior

Only umd module is published

pglewis:
PR for consideration: #3384

PR #3384, opened 2017-05-30
Feature/3340 es6 pkg.module

pglewis:
### Proposed changes
Provide an ES6 module build when building the main 'lib/'

Link to the issue:
#3340


package.json
```diff
@@ -4,6 +4,7 @@
   "version": "3.3.1",
   "homepage": "https://marionettejs.com/",
   "main": "lib/backbone.marionette.js",
+  "module": "lib/backbone.marionette.es.js",
```
rafde: `esm` ?

pglewis: Is anything using esm yet?  There is also a lot of discussion on .mjs.  My leaning is: there currently seems to be no benefit to changing the file extension but it does potentially introduce complications with IDEs and syntax highlighting.  That will probably change in 8 months but for _today_ I don't see any benefit to a non '.js' extension.

pglewis: Maybe '.esm.js'?  Probably more clear than '.es.js'.

rafde: yes, that's what I meant. `lib/backbone.marionette.esm.js`
that would match https://github.com/rollup/rollup/wiki/pkg.module recommendation
sorry for the vagueness.

paulfalgout:
This looks fine to me.. but before merging I think we should test the built file in a jsfiddle or something.

blikblum:
Many thanks. This looks good to me.
Just note that to take advantage of tree shaking is necessary to change how classes are exported: instead of attaching it to Marionette class, use export keyword. This can be done in another PR

pglewis:
> Just note that to take advantage of tree shaking is necessary to change how classes are exported

I still get benefit from it now just by having all my dependencies in the same format.  But, yeah, `export`ing the individual classes is where the real benefit would kick in with tree-shaking goodness.

blikblum:
It can be done with `outro` option. Do you want to try do in this PR or let to another one?

pglewis:
I'll probably have to leave that to someone else right now.

pglewis:
Also: from experimenting, I think the build can be quite a bit more efficient if it can be done through Rollup from end-to-end.  It quickly became clear that approach would fundamentally change how the existing build is assembled, however, so I opted to punt on that in favor of an approach that shouldn't change the existing build at all in theory.  It just wedges the ESM generation into what's known to be working.  

I think it's potentially worth pursuing an approach that moves as much out of the gulp pipeline and directly into Rollup's, I just didn't want to introduce that much uncertainty for step 1 of producing an es module.

blikblum:
I plugged the esm build into one of my apps and worked fine

pglewis:
Roger that, I'll do a revert on it.

paulfalgout:
LGTM:  @marionettejs/marionette-core 2nd thumb make sure to squash and merge

just-boris:
After we have updated our app to version 3.4, it has blown up.

We used the following style of imports

```js
import {View} from 'backbone.marionette';
```

Previously it worked, because of CommonJS <-> ES Modules interop. 3.4 version of Marionette is ships es6-modules-friendly version, that has only default export, but no named.

Is there any chance to make it fixed soon? There is a PR https://github.com/marionettejs/backbone.marionette/pull/3393 but it is stuck for now.

paulfalgout:
We'll have v3.4.1 out shortly which keeps the new build but removes the `module` key from `package.json`  Bummer, but there's no way around a breaking change for this.  Anyone wanting to use the es6 module will need to alias or import directly the particular build.

just-boris:
Yes, while ES6-friendly bundle will remain invisible, that would work for me. We can try this out as soon as it will be ready for us.

Thank you!

Issue #3352, opened 2017-04-13
Application showView should return the view


paulfalgout:
This is more for consistency than usefulness. Most functions return `this` for chaining purposes..

Right now it's returning the region that the view is shown in... which isn't consistent with really anything.. or useful..  we should either return the application or _maybe_ the view instance.

In the latter case you'd have
```js
var myView = this.showView(new MyView());
```

Thoughts?

paulfalgout:
I'm kind of leaning towards `view`..  would match: 
```js
var myChildView = myView.showChildView('regionName', new MyChildView());
```

paulfalgout:
Yikes! I assumed incorrectly https://github.com/marionettejs/backbone.marionette/blob/master/src/mixins/regions.js#L136

`showChildView` also returns the region.

I don't think it should either.

JSteunou:
This big question is do we want to *chain* with the actual view or the child view. If returning the child view chaining would be weird to read, wouldn't be?

```js
this
  .showChildView('foo', new MyChildView())
  .showChildView('bar', new AnotherChildView()) // are you showing view inside `this` or 1st child view?
;
```

paulfalgout:
I'm not a big fan of chaining..  but I suppose we could support it.. no one has complained about not being able to chain yet though.

However the more I look into it I think returning the thing that is added is correct.  Both `addChildView and `removeChildView` on `CollectionView` return the `view`.  `addRegion` returns the added region.  Region's `detachView` returns the view.  It seems like consistently we return `this` if it is doing an action on itself, but if it is related to adding or removing a relationship to another element, that element is returned.  The new exception that has yet to be released is the new `destroyView` on Region.

JSteunou:
Ho I'm not a big fan either and I do not think this is often used in Marionette, it's just I though you care about chaining because you said 

> Most functions return this for chaining purposes..

PR #3386, opened 2017-06-15
Changed Application showView and Region showChildView to return the v…

siliconjohn:
…iew (#3352)

### Proposed changes
 - returns the view from Application showView 
 - returns the view from  Region showChildView 

Link to the issue:
https://github.com/marionettejs/backbone.marionette/issues/3352


paulfalgout:
Rebased 👏  here goes to v4.

PR #3387, opened 2017-06-18
Add viewComparator: false to NextCollectionView

paulfalgout:
Currently if you need to avoid sorting for any reason, you would have to add a _.noop to the view comparator, but the views would still be looped through.

This short-circuits the sorting if it isn't wanted.

### DIFF
--- a/docs/marionette.nextcollectionviewadvanced.md
+++ b/docs/marionette.nextcollectionviewadvanced.md
@@ -400,13 +400,14 @@ cv.removeFilter({ preventRender: true });
 The `sort` method will loop through the `NextCollectionView` `children`
 and sort them with the [`viewComparator`](#nextcollectionviews-viewcomparator).
 By default, if a `viewComparator` is not set, the `NextCollectionView` will sort
-the views by the order of the models in the collection.
+the views by the order of the models in the collection. If set to `false` view
+sorting will be disabled.
 This method is also triggered internally when rendering and `before:sort` and
 `sort` events will be triggered before and after sorting.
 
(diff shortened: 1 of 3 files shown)


blikblum:
Just wondering if this should not be the default behavior

rafde:
I think it should default to false for V4. I think most don't know that it will be sorted by collection and again by collection view.

paulfalgout:
The sortWithCollection flag is the trigger. It answers whether or not the sort occurs when the collection sorts. By default the viewComparator sorts by the collection order. Setting viewComparator to false will disable all sorting entirely.

paulfalgout:
Just to clarify:

The `viewComparator` is just that, it sorts the views, which don't necessarily _have_ to be tied to models in a collection.  So let's say you've injected additional views to the CollectionView and you need them sorted relative to the data.  You may very well want to sort the CollectionView when the collection sorts with a custom `viewComparator`.

I would imagine the default should likely be that the collectionView sorts along with the collection to match.. which is the current default and adding this option wouldn't change this.  While that may not be the most performant option, it is likely the sensible default for this library.  If `viewComparator: false` was the default people would have to add their own comparator to keep the collectionView in sync with a collection.

Now perhaps better than adding `viewComparator: false` is to say that if `viewComparator` is falsy and `sortWithCollection` is also false, then it should do nothing.

rafde:
@paulfalgout since this is a 3.4 change, should this merge into master?

paulfalgout:
@rafde Everything currently has been docs into master and everything else into `next` and we've just been sitting on breaking PRs

PR #3390, opened 2017-06-20
WIP View Class templateContext

paulfalgout:
Initial code concept for #2164

Needs :eyes: primarily and then tests/docs

@marionettejs/marionette-core 

This would allow for a dev to set the `templateContext` on a view class and any inheriting ViewClass would get those as well as any `templateContext` that was added to that ViewClass.

For simplicity and to keep the cost of walking up the prototypes to a minimum the ViewClass will "inherit" the `templateContext` set on it's parent _at the time of setting_ and won't get any changes to the `templateContext` of its ancestors after setting its own.


src/view.js
```diff
@@ -206,6 +206,30 @@ const View = Backbone.View.extend({
...
+  // the current templateContext of its prototypes
+  setTemplateContext(context) {
+    const templateContext = {};
+    let ViewClass = this.prototype;
+
+    while (ViewClass.prototype !== Backbone.View) {
+      ViewClass = ViewClass.prototype;
+      if (ViewClass._templateContext) {
+        _.extend(templateContext, ViewClass._templateContext);
+      }
```
blikblum: It should extend from child to parent?

blikblum:
Do you plan to support templateContext as function?

IMO allowing only static objects as templateContext will make a limited feature.

I just stumbled with a use case for it in Mn Inspector. Relevant commits:

https://github.com/marionettejs/marionette.inspector/pull/344/commits/15e553b5fee3ec91cea2fc0fa765b725c000b795

https://github.com/marionettejs/marionette.inspector/pull/344/commits/a5c3f598625e2c36f381cc5246d4367221e0c851

src/view.js
paulfalgout: oh right..
guess this should be `templateContext = _.extend({}, ViewClass._templateContext, templateContext);`

paulfalgout:
The issue with templateContext as a function is that at the moment when they're all merged together there's no view instance.  so while templateContext on the view itself can be a function, I don't know what it'd buy anyone on the set... if we somehow supported functions these would be equivalent:
```js
const templateContext = function() {
  return {
   foo: 'bar'
  };
};

// Since the set would have to reconcile the merge
// these statements would ultimately be equivalent
ViewClass.setTemplateContext(templateContext);
ViewClass.setTemplateContext(templateContext());
```

blikblum:
> The issue with templateContext as a function is that at the moment when they're all merged together there's no view instance

Yes. To support templateContext as a function it should walk the prototype at the time `mixinTemplateContext` is called

src/view.js
blikblum: The prototype should walk in reverse order also

paulfalgout: why?

paulfalgout:
I don't really understand the use case for these needing to be functions.  For one it'll make the feature quite a bit more complicated, and certainly much more expensive for rendering.

src/view.js
blikblum: Nevermind. By changing the order of extend does not need to reverse the walk

blikblum:
To support variables that change at runtime or we dont know upfront. Like logged user.

paulfalgout:
Still seems like that could be done by making the templateContext items themselves functions:
```
ViewClass.setTemplateContext({
  foo() {
    Radio.request('user');
  },
  bar(someInstance) {
    someInstance.get('baz');
  }
});
```

Otherwise you're writing a helper that if used assumes the view has the context.. and I don't know how to write global helpers that understand fully the context of every view they're applied to..

It still seems like the use-case is limits when considering we have to loop the prototypes and extend along the way on ever View render.

blikblum:
I agree It would make implementation harder. And is a thing that should not be encouraged anyway.

blikblum:
> Still seems like that could be done by making the templateContext items themselves functions:

This would require to walk each context key that has its own perf hit. So i think should be kept simples as is now

paulfalgout:
> This would require to walk each context key that has its own perf hit. So i think should be kept simples as is now

Ahh.. I think my use-case is spilling into my thoughts for this

Maybe some code would help clarify.  If I'm using something like handlebars I can use templateContext as functions now.
```
// foo.hbs
<div>{{ foo }}</div>
<div>{{ bar }}</div>
<div>{{ toUpperCase 'hello' }}</div>
```
```js
  template: fooHBS,
  serializeData() {
    return {  
      baz: 'I will be all uppercase'
    };
  }
  templateContext: {
     foo: 'static',
     bar() {
       return this.toUpperCase(this.baz);
     },
     toUpperCase(str) {
       return String(str).toUpperCase();
     }
  }
});
```

So then I could definitely see `toUpperCase` moving to the viewclass helpers.. 

_But_ for templates like underscore that don't utilize functions..  I can see how it would be much less useful.. what if we did something like..
```
templateContext(classContext) {
  return {
    foo: 'static',
    bar: classContext.toUppercase(this.model.get('baz'))
  }
}
```
where `classContext` (or `helpers` or something.. I really don't know what to name it) is the ViewClass `templateContext`  (maybe we'd even rename `setTemplateContext` for that reason)
and it'd get passed to `templateContext` as an obj _and_ merged with it... is that odd.. or any other ideas?

paulfalgout:
> This would require to walk each context key that has its own perf hit. So i think should be kept simples as is now

I'm not sure I agree with this now
```

PR #3392, opened 2017-06-23
Make monitorViewEvents opt-out

paulfalgout:
The monitorViewEvents object make `attach`/`detach` events and the `dom:refresh` / `dom:remove` events possible.

Disabling this for a view will make these events unreliable for a view and all of this view's descendants.

That said, for edge-cases where performance is paramount, disabling the monitor can help speed up rendering.

Between this, https://github.com/marionettejs/backbone.marionette/pull/3387 and some other reasonable modifications I'm able to get the js-framework-benchmark down to being amongst the competition.

### DIFF
--- a/docs/viewlifecycle.md
+++ b/docs/viewlifecycle.md
@@ -31,6 +31,7 @@ their children.
 * [`Region`s and the View Lifecycle](#regions-and-the-view-lifecycle)
   * [Show View Events](#show-view-events)
   * [Empty Region Events](#empty-region-events)
+* [Advanced Event Settings](#advanced-event-settings)
 
 ## `View` Lifecycle
 Marionette views define a number of events during the creation and destruction
@@ -896,3 +897,21 @@ destroyed.
 
 Fired after the entire destruction process is complete. At this point, the view
 has been removed from the DOM completely.
+
+
+## Advanced Event Settings
+Marionette is able to trigger `attach`/`detach` events down the view tree along with
+triggering the `dom:refresh`/`dom:remove` events because of the view event monitor.
+This monitor starts when a view is created or shown in a region (to handle non-Marionette views).
+
+In some cases it may be a useful performance improvement to disable this functionality.
+Doing so is as easy as setting `monitorViewEvents: false` on the view class.
+
+```javascript
+const NonMonitoredView = Mn.View.extend({
+  monitorViewEvents: false
+});
+```
+
+**Important Note**: Disabling the view monitor will break the monitor generated
+events for this view _and all child views_ of this view. Disabling should be done carefully.
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -76,7 +76,7 @@ const CollectionView = Backbone.View.extend({
   },
 
   _endBuffering() {
-    const shouldTriggerAttach = !!this._isAttached;
+    const shouldTriggerAttach = this._isAttached && this.monitorViewEvents !== false;
     const triggerOnChildren = shouldTriggerAttach ? this._getImmediateChildren() : [];
 
     this._isBuffering = false;
@@ -182,6 +182,7 @@ const CollectionView = Backbone.View.extend({
     this.triggerMethod('before:remove:child', this, view);
 
     this.children._remove(view);
+    view._shouldDisableEvents = this.monitorViewEvents === false;
     destroyView(view);
 
     this.stopListening(view);
@@ -600,7 +601,7 @@ const CollectionView = Backbone.View.extend({
   _attachView(view, index) {
     // Only trigger attach if already attached and not buffering,
     // otherwise _endBuffering() or Region#show() handles this.
-    const shouldTriggerAttach = !view._isAttached && !this._isBuffering && this._isAttached;
+    const shouldTriggerAttach = !view._isAttached && !this._isBuffering && this._isAttached && this.monitorViewEvents !== false;
 
     if (shouldTriggerAttach) {
       triggerMethodOn(view, 'before:attach', view);
--- a/src/common/monitor-view-events.js
+++ b/src/common/monitor-view-events.js
@@ -74,7 +74,7 @@ function handleRender() {
 // Monitor a view's state, propagating attach/detach events to children and firing dom:refresh
 // whenever a rendered view is attached or an attached view is rendered.
 function monitorViewEvents(view) {
-  if (view._areViewEventsMonitored) { return; }
+  if (view._areViewEventsMonitored || view.monitorViewEvents === false) { return; }
 
   view._areViewEventsMonitored = true;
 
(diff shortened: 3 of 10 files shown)


src/view.js
```diff
@@ -19,6 +19,7 @@ const ClassOptions = [
   'collectionEvents',
   'events',
   'modelEvents',
+  'monitorViewEvents',
```
blikblum: Do we want this option to be set through options?

I think is safer to keep it only in the class definition to avoid using in a class that was created expecting those events to trigger.

paulfalgout: Yeah, good idea

paulfalgout:
I think this one is GTG @marionettejs/marionette-core

docs/viewlifecycle.md
```diff
@@ -896,3 +897,21 @@ destroyed.
...
+Doing so is as easy as setting `monitorViewEvents: false` on the view class.
+
+```javascript
+const NonMonitoredView = Mn.View.extend({
+  monitorViewEvents: false
+});
+```
+
+**Important Note**: Disabling the view monitor will effectually break the monitor generated
+events for this view _and all children_ of this view. Disabling should be done carefully.
```
scott-w: By "all children" do you mean the result of `NonMonitoredView.extend()` or all child views? It's not clear from this documentation to me.

docs/viewlifecycle.md
```diff
@@ -896,3 +897,21 @@ destroyed.
...
+Doing so is as easy as setting `monitorViewEvents: false` on the view class.
+
+```javascript
+const NonMonitoredView = Mn.View.extend({
+  monitorViewEvents: false
+});
+```
+
+**Important Note**: Disabling the view monitor will effectually break the monitor generated
+events for this view _and all child views_ of this view. Disabling should be done carefully.
```
paulfalgout: @scott-w

paulfalgout:
Before this gets committed:  @marionettejs/marionette-core 

The CollectionViews and Region will trigger `attach` on their children if they are attached themselves at the moment the child gets attached.  Should we _also_ disable this?  Otherwise it will work inconsistently.  The children will not get the `attach` events if they're attached when the parent is _not_ already attached.  (because the view monitor does this job)

While I think disabling this is useful on its own, I also think it leaves inconsistent functionality that can't be trusted..  maybe the setting needs to be something like `supportsDomLifecycle` to match `supportsDestroyLifecycle` and `supportsRenderLifecycle`  However those are slightly different.. if those two are true it means the view is _already_ triggering these events.

Thoughts?

docs/viewlifecycle.md
scott-w: So this is all views shown by `this.getRegion('myRegion').show(new ChildView());` as part of this view.

JSteunou:
I'm on the idea of consistency. I would not people though

> hey even if I disable the monitor I still have attach event on collection child view, nice

and finally having its app being buggy because in fact it does not have the event in any case and should not rely on it when disabling the monitor. I am afraid even well documented, this case could happen.

paulfalgout:
Ok I've yet to test this and it will need clarification in docs (which are currently outdated in this PR) with warnings.. _but_ I think I have a consistent implementation where if you do `myView.monitorViewEvents` any descendant will not get events: `attach`, `detach`,`dom:refresh` or `dom:remove` under any circumstance.  Needs 👀

src/collection-view.js
```diff
@@ -76,7 +76,7 @@ const CollectionView = Backbone.View.extend({
   },
 
   _endBuffering() {
-    const shouldTriggerAttach = !!this._isAttached;
+    const shouldTriggerAttach = !!this._isAttached && this.monitorViewEvents !== false;
```
JSteunou: You could drop the `!!` I think

src/common/view.js
```diff
@@ -27,7 +27,7 @@ export function destroyView(view) {
     triggerMethodOn(view, 'before:destroy', view);
   }
 
-  const shouldTriggerDetach = !!view._isAttached;
+  const shouldTriggerDetach = !!view._isAttached && !view._shouldDisableEvents;
```
JSteunou: same

src/mixins/view.js
```diff
@@ -118,7 +118,7 @@ const ViewMixin = {
   // Handle destroying the view and its children.
   destroy(...args) {
     if (this._isDestroyed) { return this; }
-    const shouldTriggerDetach = !!this._isAttached;
+    const shouldTriggerDetach = !!this._isAttached && !this._shouldDisableEvents;
```
JSteunou: idem

src/next-collection-view.js
```diff
@@ -505,7 +505,7 @@ const CollectionView = Backbone.View.extend({
   },
 
   _detachChildView(view) {
-    const shouldTriggerDetach = !!view._isAttached;
+    const shouldTriggerDetach = !!view._isAttached && this.monitorViewEvents !== false;
```
JSteunou: idem

src/next-collection-view.js
```diff
@@ -541,7 +541,7 @@ const CollectionView = Backbone.View.extend({
   },
 
   _attachChildren(els, views) {
-    const shouldTriggerAttach = !!this._isAttached;
+    const shouldTriggerAttach = !!this._isAttached && this.monitorViewEvents !== false;
```
JSteunou: idem

src/region.js
```diff
@@ -319,7 +325,7 @@ const Region = MarionetteObject.extend({
   },
 
   _detachView(view) {
-    const shouldTriggerDetach = !!view._isAttached;
+    const shouldTriggerDetach = !!view._isAttached && !this._shouldDisableMonitoring();;
```
JSteunou: idem

src/collection-view.js
paulfalgout: I suppose so, but I didn't add them.  There are a lot of other cases of `!!` casting in similar places, perhaps a separate PR that investigated and changed that would be better than just in the instances?

JSteunou: My point was rather `!!foo` => `!!foo && bool` could be written `foo && bool` because of the `&&` making the cast not because `foo` might be already a boolean

paulfalgout:
@JSteunou fixed.
@marionettejs/marionette-core would really like :eyes: on the later changes here so I know if I should flush out the docs/tests.

src/common/monitor-view-events.js
```diff
@@ -74,7 +74,7 @@ function handleRender() {
 // Monitor a view's state, propagating attach/detach events to children and firing dom:refresh
 // whenever a rendered view is attached or an attached view is rendered.
 function monitorViewEvents(view) {
-  if (view._areViewEventsMonitored) { return; }
+  if (view.monitorViewEvents === false || view._areViewEventsMonitored) { return; }
```
blikblum: I would swap the check order since areViewEventsMonitored will be the most common flag set

paulfalgout:
OK docs and specs updated.  I think this is GTG @marionettejs/marionette-core

PR #3393, opened 2017-06-23
Implementation of named exports

blikblum:
### Proposed changes
 - Exports classes and functions as named exports instead of attaching to Marionette instance. The main benefit is to allow tree shaking of unused code. Also helps IDE to auto complete.
 
This is a proof of concept just to trigger discussion, missing to declare some classes and functions. 

It's breaking change.

umd or CommonJS users should not be affected but those using ES modules will not be able to use class definitions attached to Marionette instance

The code below

```
import Mn from 'backbone.marionette'

export default Mn.View.extend({})
```

must be converted to 

```
import {View} from 'backbone.marionette'

export default View.extend({})
```

Other issues
 * how to export Object and Error classes?
 * Backbone.Marionette will not be set to exported Marionette/global.Marionette even in umd build

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -1,4 +1,3 @@
-import Backbone from 'backbone';
 import {version} from '../package.json';
 
 import proxy from './utils/proxy';
@@ -33,17 +32,6 @@ import {
   isEnabled,
   setEnabled
 } from './config/features';
-
-const previousMarionette = Backbone.Marionette;
-
-// This allows you to run multiple instances of Marionette on the same
-// webapp. After loading the new version, call `noConflict()` to
-// get a reference to it. At the same time the old version will be
-// returned to Backbone.Marionette.
-export const noConflict = function() {
-  Backbone.Marionette = previousMarionette;
-  return this;
-};
 
 // Utilities
 export const bindEvents = proxy(_bindEvents);
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -3,31 +3,29 @@ import {version} from '../package.json';
 
 import proxy from './utils/proxy';
 import extend from './utils/extend';
-import deprecate from './utils/deprecate';
 
 import {
-  bindEvents,
-  unbindEvents
+  bindEvents as _bindEvents,
+  unbindEvents as _unbindEvents
 } from './common/bind-events';
 import {
-  bindRequests,
-  unbindRequests
+  bindRequests as _bindRequests,
+  unbindRequests as _unbindRequests
 } from './common/bind-requests';
-import getOption from './common/get-option';
-import mergeOptions from './common/merge-options';
+import _getOption from './common/get-option';
+import _mergeOptions from './common/merge-options';
 import monitorViewEvents from './common/monitor-view-events';
-import normalizeMethods from './common/normalize-methods';
-import triggerMethod from './common/trigger-method';
+import _normalizeMethods from './common/normalize-methods';
+import _triggerMethod from './common/trigger-method';
 
 import BackboneViewMixin from './mixins/backboneview';
 
-import MarionetteObject from './object';
+import MnObject from './object';
 import View from './view';
 import CollectionView from './collection-view';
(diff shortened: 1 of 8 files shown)


paulfalgout:
Can we add the exports now without detaching them from Mn?  I realize that's not really that big of a difference, but I assume it might help with IDE static analysis, or no?

blikblum:
> Can we add the exports now without detaching them from Mn?

Yes

RobertCZ:
There's also possibility of importing everything - just needs to be documented:

`import * as Mn from 'backbone.marionette'`

scott-w:
There's probably an argument that, for v4, we should rename `Object` to stop it conflicting with the reserved `Object` in JS?

blikblum:
Updated the PR:

- export remaining named exports
- do not export deprecate and Error since they are a lot generic
- until further definition, renamed Object to MnObject

With it, only DEV_MODE is attached to default export

Do we need to export FEATURES ?

TBD: Tests needs to be updated

paulfalgout:
I don't think we need `FEATURES`

paulfalgout:
I think this is fine.  Made a vote issue for the renaming of Object just for more consensus.

paulfalgout:
So is there any benefit to putting `DEV_MODE` as a named export and having no default?  I _think_ a lot of the loaders will do the following:
`MyLib = 'default' in MyLib ? MyLib.default : MyLib;`
Sooo I think if we have no default export that might be friendlier to older loaders? It may be a less of a refactor for people not currently utilizing named imports as well.

blikblum:
> So is there any benefit to putting DEV_MODE as a named export and having no default?

The best way is to remove the default export and keep only the named exports.

The problem is that doing so will prevent configuring DEV_MODE

Maybe change how DEV_MODE is handled?

paulfalgout:
Ah yes. Good point. It could be a feature flag.

paulfalgout:
I think if `DEV_MODE` moves to a feature flag then most of the tests will work again as well.

It also looks like `MnObject` is getting the majority votes, so 🎉

blikblum:
I will wait until other changes are merged so i can do a new rebase

I Will also open a PR to move devmode to a feature flag

src/backbone.marionette.js
```diff
@@ -39,56 +37,52 @@ import {
...
+  Application,
+  AppRouter,
+  isEnabled,
+  setEnabled,
+  monitorViewEvents,
+  BackboneViewMixin,
+  extend,
+  DomApi,
+  version as VERSION,
+  FEATURES
```
paulfalgout: I think we can drop this export.  It isn't documented and there's a clear API for how to enable it.  I assume this is here because originally things were attached to the global `Marionette` a lot pre ES6 modules

src/backbone.marionette.js
```diff
@@ -3,32 +3,30 @@ import {version} from '../package.json';
...
 import BackboneViewMixin from './mixins/backboneview';
 
-import MarionetteObject from './object';
+import MnObject from './object';
 import View from './view';
 import CollectionView from './collection-view';
 import Behavior from './behavior';
 import Region from './region';
 import Application from './application';
 import AppRouter from './app-router';
```
paulfalgout: AppRouter 👋

blikblum:
I got confused in the middle of a rebase and pushed before was done.

Already fixed locally. Will push soon

blikblum:
This is basically done.

Only one issue remains: Backbone.Marionette is not set even in umd build. So basically noConflict is basically useless

One possibility is to set Backbone.Marionette outside of the build as is done with Mn. Other is removing this feature (attach Marionette instance to Backbone instance)

paulfalgout:
I'd be for removing Marionette from Backbone.  Honestly `noConflict` is necessary in backbone because it aggressively attaches itself to the window.  Marionette doesn't do this so it's not as big of a deal.

JSteunou:
Good job @blikblum !! I did not review all the PR yet but seems nice so far :+1:

PR #3395, opened 2017-06-24
Partial test cleanup

paulfalgout:
Related: https://github.com/marionettejs/backbone.marionette/issues/3248

Adds clean tests per utils file.
This doesn't test the files that have no logic.

Updates deprecate tests and ignores a difficult to test trivial line to reach 100% branching coverage.

This also moves as many files as were easy into the same file structure as the code.


paulfalgout:
This doesn't touch production code, and has a thumb, I'm going to merge it and we can fix anything anyone has an issue with later.

PR #3397, opened 2017-06-26
Add hasContents to DomApi and fix view isRendered bug

paulfalgout:
When we are determining if a view is previously rendered we were using `$el` which caused issues when trying to implement backbone.nativeview.

Not only that, but I believe the current implementation should be considered a bug.  A view is rendered if a template is render inside the view's _el_.  The current check is just seeing if the view already has an _el_ and not whether it contained any value.  I believe this is a broken behavior.  We should consider a view rendered only if the view's el contains content.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -96,7 +96,7 @@ const View = Backbone.View.extend({
     Backbone.View.prototype.setElement.apply(this, arguments);
 
     if (hasEl) {
-      this._isRendered = !!this.$el.length;
+      this._isRendered = this.Dom.hasContents(this.el);
       this._isAttached = isNodeAttached(this.el);
     }
 
--- a/docs/dom.api.md
+++ b/docs/dom.api.md
@@ -46,6 +46,10 @@ functions, this only takes a literal string for its second argument.
 Takes the DOM node `el` and appends the DOM node `contents` to the end of the
 element's contents.
 
+### `hasContents(el)`
+
+Returns a boolean indicating if the `el` has child nodes.
+
 ### `detachContents(el)`
 
 Remove the inner contents of `el` from the DOM while leaving `el` itself in the
--- a/src/config/dom.js
+++ b/src/config/dom.js
@@ -64,6 +64,11 @@ export default {
     _$el.append(_$contents);
   },
 
+  // Does the el have child nodes
+  hasContents(el) {
+    return el.hasChildNodes();
+  },
+
   // Remove the inner contents of `el` from the DOM while leaving
   // `el` itself in the DOM.
   detachContents(el, _$el = getEl(el)) {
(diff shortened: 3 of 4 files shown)


src/view.js
```diff
@@ -96,7 +96,7 @@ const View = Backbone.View.extend({
     Backbone.View.prototype.setElement.apply(this, arguments);
 
     if (hasEl) {
-      this._isRendered = !!this.$el.length;
+      this._isRendered = this.Dom.hasContents(this.el);
```
blikblum: Why not `this._isRendered = this.el.hasChildNodes()`?

I dont see a different implementation warranting an abstraction and is much simpler

paulfalgout:
I would argue that this is fixing a bug. isRendered is being set too aggressively. When a view is rendered the view's el should have children. Any view you instantiate passes `!!the.$el.length` prior to rendering.

src/view.js
paulfalgout: Yeah I was on the fence about it and could go the other way.  But I'm leaning towards it all being on the api at the moment xref: https://github.com/marionettejs/backbone.marionette/issues/3398 

I suppose it comes down to if the DomApi and "utils" like `isNodeAttached` are useful directly outside of the internal uses or if we should let users do it on their own.

JSteunou:
What about Views without template? Use case: collection view with tiny child view without backbone.view el with classes, attributes...

paulfalgout:
@JSteunou rendering is still the process of rendering a template and putting it inside the view's el and has nothing to do with the `el` itself.  The `el` including `className` and `attributes` are created at instantiation and rendering does not affect them at all.  

Honestly maybe deprecating `template: false` was the wrong call..  _however_ it was being abused.  Essentially if the `el` of a view has content, it's rendered.. if it does not.. it is not assumed to be rendered.. if a view is then `render()`ed but no template string is rendered inside of the el, the view is still understood at that point to be rendered.  (if the template for instance was _.noop)  

A view should only be rendered if it has a template it is intending to put inside the view. Neither `template: false` or `template: _.noop` should be used as a way to trigger `onRender` for a view with no template.  In the case of a collectionview, the view will attempt to render either way.  The only case in which it would not render would be if the user overrode `buildChildView` and passed it an `el` _with_ inner contents..

However let's say that in the DOM pre-rendered was:
```
<ul>
  <li id="view1"></li>
  <li id="view2"></li>
  <li id="view3"></li>
</ul>
```
And the user wanted to use a collectionview to render children into the pre-defined `li`s in the DOM.  Currently this would be pretty much impossible as simply passing `$('#view1')` to the childview would erroneously mark it as rendered even though nothing was rendered inside of the els yet.

This PR fixes that bug.

PR #3399, opened 2017-06-26
Append only added views if they're added at the end

paulfalgout:
Prevents a large re-append when not necessary.  This adds performance benefits when a number of views are added to a large collectionview at the end.

It is unlikely to cause perf issues when it can't be done as the `_.every` will bail at the first moment it can't be used.

It is possible that if you have a small collection and are adding a very large number of views that it isn't worth it, but it shouldn't be too costly even in those circumstances.

In the following screenshot this is how long it takes to create 10000 rows and how long it takes to add 100 rows to the 10000:
screen shot 2017-06-26 at 4 28 27 pm

### DIFF
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -122,7 +122,7 @@ const CollectionView = Backbone.View.extend({
     // Remove first since it'll be a shorter array lookup.
     const removedViews = this._removeChildModels(changes.removed);
 
-    this._addChildModels(changes.added);
+    this._addedViews = this._addChildModels(changes.added);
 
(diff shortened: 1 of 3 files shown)


scott-w:
I like the change. Are there any existing tests that cover this? I'm assuming so if coverage stays at 100%?

paulfalgout:
Yeah... I think a test is necessary.. I don't think this is all the way correct yet either.

paulfalgout:
Ok.  I think this is ready

paulfalgout:
@marionettejs/marionette-core I think this one is ready to go

PR #3402, opened 2017-06-29
Add swapChildViews to NextCollectionView

paulfalgout:
- Adds `DomApi.swapEl` similar to `DomApi.replaceEl`
- Adds `children.hasView` to `NextCollectionView`

When looking into `js-framework-benchmark` I found there was no real performant way to swap to views in a collectionview.

I'm not sure how I feel about this change, but it does seem rather difficult to achieve outside of such a solution.

There are a couple of things to consider here.  Not all collectionViews must have a collection... while you could still store data on the views and use the `viewComparator` somehow, that is a lot of work if all you want to do is swap them.

If the collectionView does have a collection and is sorting based on the collection _or_ other `viewComparator` even, then this API becomes decidedly less useful as the state of the view order is also managed elsewhere.. however the same could be said for `addChildView`

Before I do any testing or documentation, I thought I'd ask for :eyes:

### DIFF
--- a/src/config/dom.js
+++ b/src/config/dom.js
@@ -33,6 +33,11 @@ export default {
     return _$el.find(selector);
   },
 
+  // Returns true if the el contains the node childEl
+  hasEl(el, childEl) {
+    return el.contains(childEl && childEl.parentNode);
+  },
+
   // Detach `el` from the DOM without removing listeners
   detachEl(el, _$el = getEl(el)) {
     _$el.detach();
--- a/src/config/dom.js
+++ b/src/config/dom.js
@@ -53,6 +53,26 @@ export default {
     parent.replaceChild(newEl, oldEl);
   },
 
+  // Swaps the location of `el1` and `el2` in the DOM
+  swapEl(el1, el2) {
+    if (el1 === el2) {
+      return;
+    }
+
+    const parent1 = el1.parentNode;
+    const parent2 = el2.parentNode;
+
+    if (!parent1 || !parent2) {
+      return;
+    }
+
+    const next1 = el1.nextSibling;
+    const next2 = el2.nextSibling;
+
+    parent1.insertBefore(el2, next1);
+    parent2.insertBefore(el1, next2);
+  },
+
   // Replace the contents of `el` with the HTML string of `html`
   setContents(el, html, _$el = getEl(el)) {
     _$el.html(html);
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -588,6 +588,11 @@ const CollectionView = Backbone.View.extend({
     this.children._swap(view1, view2);
     this.Dom.swapEl(view1.el, view2.el);
 
+    // If the views are not filtered the same, refilter
+    if (this.Dom.hasEl(this.el, view1.el) !== this.Dom.hasEl(this.el, view2.el)) {
+      this.filter();
+    }
+
     return this;
   },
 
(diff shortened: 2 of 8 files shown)


src/next-collection-view.js
```diff
@@ -577,6 +577,17 @@ const CollectionView = Backbone.View.extend({
     this.Dom.appendContents(this.el, els, {_$el: this.$el});
   },
 
+  swapChildViews(view1, view2) {
+    if (!this.children.hasView(view1) || !this.children.hasView(view1)) {
```
rafde: This is nice for drag and drop view swapping or something similar.
Does this need something like a DOM parent check for both views to make sure they are being swapped from one dom parent to another? 
Or can this support something like, view1 has no parent but will swap with view2 which has a parent. Not sure what will happen with view2. I guess the person using the function will have to figure how to handle that.

paulfalgout: The Dom code actually assumes that the first view has a parent..  they both have to be children of the collection, but technically I guess at the moment the 2nd could be detached?  that gets a little weird.  Guess I could check they're both `_isAttached`..

rafde: I'd put more trust in `el2.parentNode` than `_isAttached`

src/config/dom.js
```diff
@@ -53,6 +53,24 @@ export default {
...
+      return;
+    }
+
+    const parent = el1.parentNode;
+
+    if (!parent) {
+      return;
+    }
+
+    const next1 = el1.nextSibling;
```
rafde: what if you are swapping an only sibling?

paulfalgout: if `nextSibling` is `null` `insertBefore` will put it at the end

src/next-collection-view.js
paulfalgout: maybe so.. but I think we'd have to check the `parentNode` in this function so that the children don't also swap.  We'd also have to document clearly that only attached children can be swapped, but I think that probably makes sense.

rafde: I think it'd be nice to be able to swap out views that don't belong to the same parent. But an issue I foresee is swapping views and the collection view has a filter that runs and it creates some type of unintended side affect.

paulfalgout: yeah I think you'll only be able to swap attached views.

we can handle two parents in `swapEl`, but that'll be a bit weird for a single collectionview to have multiple parents.. but I guess if you've overridden attachHtml it is possible

paulfalgout: if you're "swapping" views between 2 collectionviews, what you're actually doing is removing a view from one collection and adding it to another.. and that's what you should do in that case.

blikblum: I'm ok with the feature but allowing to swap between to different collection views will complicate the things

rafde:
This pr looks good. Just need test coverage and it'll be good

paulfalgout:
@marionettejs/marionette-core ok so I've finished the original concept with code coverage, however I've added a 2nd commit that covers an edge case that I have not tested.  Additionally I still need to do documentation.  However I need thumbs one way or the other on the edge case.

paulfalgout:
Yep, I think the docs may depend a little on the consensus regarding refiltering... either by documenting the behavior or the lack their of for the edge case

paulfalgout:
@marionettejs/marionette-core could really use some eyes on the 2nd commit.  If it gets :thumb:s I'll flush out the docs and tests.

JSteunou:
:+1: but I'm still missing the use case

rafde:
👍

paulfalgout:
Ok just working on docs and then this should be ready.

Issue #3409, opened 2017-08-02
Release v3.4


paulfalgout:

blikblum:
I'd like to update the sourcemap dependency to see if fix the current issue

paulfalgout:
We can try it, but this is a very old problem and according to the other issues hasn't been resolved

PR #3411, opened 2017-08-02
V3.4

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3409


JSteunou:
Is using DomAPI instead of DOM mixin a breaking change?

paulfalgout:
DOM Mixin is listed as an experimental API so that such changes can be made without requiring a major release.

JSteunou:
Then :+1:  !

PR #3413, opened 2017-08-05
Actually detachContents for perf empty

paulfalgout:
Fortunately/unfortunately this bug doesn't come up for `$()`, but it clearly wasn't detaching anything.

Not going to cause a problem for anyone upgrading to 3.4 unless they use `monitorViewEvents: false`, but it's definitely a problem
😿


JSteunou:
I think this is good to go :+1:

scott-w:
I definitely need this because I've got a case for `monitorViewEvents: false` right now :+1:

paulfalgout:
This'll only bite you in v3.4 if you're replacing jquery.. jquery will fail silently I believe without this patch

Issue #3414, opened 2017-08-07
Behavior this.options is always empty


JSteunou:
When a view passes options to behaviors, those should be set in the options property of the behavior. This is no more the case in 3.4

3.3 https://jsfiddle.net/7z8wxrdo/2/
3.4 https://jsfiddle.net/7z8wxrdo/3/

JSteunou:
ok find why @paulfalgout 

https://github.com/marionettejs/backbone.marionette/commit/b2ad46f9c8930be98735387b30602ecfdd98bdd2#diff-e5931bcc4ee3beb1f1bcb2460133d657 introduce regression because at this loc we give 2 parameters https://github.com/marionettejs/backbone.marionette/blob/master/src/behavior.js#L41

paulfalgout:
Good find.  Unfortunate missing test case.  I think though rather than reverting we should do `this._setOptions(_.extend({}, this.defaults, options))` so that only behaviors get a slightly perf hit instead of every class.  Thoughts?  Want to open a PR @JSteunou

JSteunou:
Yep, have a little time and I like your idea

PR #3415, opened 2017-08-07
Fix Behavior _setOptions call 

JSteunou:
Clone & merge `this.defaults` with `options` before calling `_setOptions` has it takes only one parameter

Fixes issue #3414

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -38,7 +38,7 @@ const Behavior = MarionetteObject.extend({
 
     this.defaults = _.clone(_.result(this, 'defaults', {}));
 
-    this._setOptions(this.defaults, options);
+    this._setOptions(_.extend({}, this.defaults, options));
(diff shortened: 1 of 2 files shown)


paulfalgout:
Saucelabs is broken unfortunately.. but the tests are passing.  However it is probably worth adding a test that prevents this regression in the future... super surprised there wasn't one already TBH.

JSteunou:
Yep I'm searching but dont find where I should put one

paulfalgout:
So I've been working as part of some issue we have to make the test files equate to specific files in the src rather than concepts.. sometimes the tests are broken up by concept within a class dir... so that it is clear on where it goes.. I have no idea where the behavior stuff is.. I'd just suggest adding `/test/unit/behavior.spec.js` for now

JSteunou:
Just added a test. Run locally, it works with the fix, it fails without. It's bare minimal but I dont have this much time :(

PR #3418, opened 2017-08-08
Remove pkg.module

paulfalgout:
This is breaking in some cases https://github.com/marionettejs/backbone.marionette/pull/3384#issuecomment-320964876

We'll still build the package so people can alias it or import it directly if they really want this, and we'll fix it for everyone for v4.


JSteunou:
:+1: GTG

PR #3420, opened 2017-08-08
Use source map generated by rollup directly instead of loading through gulp-sourcemaps

### LINKED ISSUE #3328: Browser debuggers do not work correctly with Marionette using source map
### Description

 1. Trying to debug Marionette code in Chrome or Firefox debugger do not work when using source map

### Expected behavior

Debugger should work with source map

### Actual behavior

Without source map, debugging works fine

When using source map the place where breakpoint is set and the highlighted line differs from the actual code being run. As example load a simple marionette app like http://codepen.io/blikblum/pen/QpQxZb (to avoid confusion factor from CodePen shell, i extracted the code from the debug view into a single HTML file and ran locally with a fresh build from next branch).

In chrome debugger, put a break point at src/region.js line 202 (`this._restoreEl()`). Try to step inside function. It will go inside `Events.on` method instead of going into `_restoreEl`.

Now put the breakpoint at line 204, over `view.on('before:destroy', this._restoreEl, this)`. Look at elements tab. It should have 'div#hook' element. Try to step over `view.on` call. Look at the elements: 'div#hook' is now replaced by 'div "Foo"' meaning that `this.replaceEl` call is already done although, by the line being highlighted it should be called only after the next step over.

### Environment

1. Marionette version: 3.2 (next)
2. Backbone version: 1.3.3

blikblum:
In current setup the source map generated by rollup is temporarily appended to the generated source as a data uri. Than the source map is extracted by gulp-sourcemaps and written to file or passed to gulp-uglify

### Proposed changes
 - Save the source map directly to file and pass the generated source content and map directly to uglify
 
This has some advantages: faster since there's no append / extract step, less dependencies 

It does not fix _ #3328 but it makes easier to track the root cause since there are less steps and tools involved

### DIFF
--- a/gulp/_banner.js
+++ b/gulp/_banner.js
@@ -3,11 +3,14 @@ import {version} from '../package.json';
 const now = new Date();
 const year = now.getFullYear();
 
-export default `// MarionetteJS (Backbone.Marionette)
-// ----------------------------------
-// v${version}
-//
-// Copyright (c)${year} Derick Bailey, Muted Solutions, LLC.
-// Distributed under MIT license
-//
-// http://marionettejs.com\n\n`;
+export default `/**
+* @license
+* MarionetteJS (Backbone.Marionette)
+* ----------------------------------
+* v${version}
+*
+* Copyright (c)${year} Derick Bailey, Muted Solutions, LLC.
+* Distributed under MIT license
+*
+* http://marionettejs.com
+*/\n\n`;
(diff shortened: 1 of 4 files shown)


blikblum:
Yes, they exhibit the same behavior, not surprising since the generated source map is the same. In chrome 61, both versions are better handled but there still a issue with setting breakpoint

blikblum:
I added the configuration to preserve license info on minified source

blikblum:
This PR went into master. My bad

PR #3422, opened 2017-08-09
Do not use region $el, prefer el refs #3421

### LINKED ISSUE #3421: Region appendView use $el instead of el
Tricky regression from v3.4

In some context, a View could have a region with a selector like `[data-region="body"]` and another region with a childview with also `[data-region="body"]` in its own DOM

Before v3.4 the append logic [casted the region.el into jquery object](https://github.com/marionettejs/backbone.marionette/blob/v3.3.1/src/mixins/dom.js#L12) to put the view content in it. The region.el is always the region.$el **1st element**. See [here](https://github.com/marionettejs/backbone.marionette/blob/v3.4.1/src/region.js#L39) & [here](https://github.com/marionettejs/backbone.marionette/blob/v3.4.1/src/region.js#L135)

Since v3.4 the append logic uses the region.$el that could contains more than one element, thus the append will put the view inside the **last element**. See [here](https://github.com/marionettejs/backbone.marionette/blob/v3.4.1/src/region.js#L245) & [here](https://github.com/marionettejs/backbone.marionette/blob/v3.4.1/src/config/dom.js#L88)

JSteunou:
Force DomApi to use region el instead of cached $el because this one can contains multiple nodes.

Should fix #3421

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -242,7 +242,7 @@ const Region = MarionetteObject.extend({
   // Override this method to change how the new view is appended to the `$el` that the
   // region is managing
   attachHtml(view) {
-    this.Dom.appendContents(this.el, view.el, {_$el: this.$el, _$contents: view.$el});
+    this.Dom.appendContents(this.el, view.el, {_$contents: view.$el});
   },
 
   // Destroy the current view, if there is one. If there is no current view, it does


paulfalgout:
we really should have included a test here

Issue #3242, opened 2016-10-25
Don't return this for functions meant to be overridden


paulfalgout:
Example `View#attachElContent`

This function is for overriding if you want to attach html differently.  It is not meant for calling directly.  As such it has a `return this` in it which is misleading and confusing and isn't used internally.  When overriding the user should not be worried about chaining.. particularly when you shouldn't use this function this way to begin with.

https://github.com/marionettejs/backbone.marionette/blob/v3.1.0/src/view.js#L184

rafde:
Sounds good since what's considered public which returns is unsafe to rely on when used internally.

siliconjohn:
Are there other functions besides the above example that are meant to be overridden? I can change all of them but the example is the only one I have found.

paulfalgout:
I have not really looked. It is possible it is the only example.

PR #3423, opened 2017-08-11
Allow chaining after setRenderer

JSteunou:
Because one day I get bitten by `export default Mn.View.extend(...).setRenderer(...)`

### DIFF
--- a/test/unit/config/dom.js
+++ b/test/unit/config/dom.js
@@ -1,6 +1,6 @@
 import $ from 'jquery';
 import _ from 'underscore';
-import DomApi from '../../../src/config/dom';
+import DomApi, { setDomApi } from '../../../src/config/dom';
 
 // Copied from https://github.com/jashkenas/underscore/blob/1.8.3/underscore.js#L137
 const MAX_ARRAY_INDEX = Math.pow(2, 53) - 1;
@@ -22,6 +22,14 @@ chai.use(function(_chai, utils) {
 
 
 describe('DomApi', function() {
+  describe('#setDomApi', function() {
+    it('should return the current class', function() {
+      const MyObject = function() {};
+      MyObject.setDomApi = setDomApi;
+      expect(MyObject.setDomApi()).to.be.eq(MyObject);
+    });
+  });
+
   describe('#createBuffer', function() {
     it('should return an appendable node', function() {
       expect(DomApi.createBuffer().appendChild).to.be.a('function');
--- a/src/config/dom.js
+++ b/src/config/dom.js
@@ -11,6 +11,7 @@ function getEl(el) {
 // Static setter
 export function setDomApi(mixin) {
   this.prototype.Dom = _.extend({}, this.prototype.Dom, mixin);
+  return this;
 }
 
 export default {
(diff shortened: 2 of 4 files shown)


paulfalgout:
bleh chaining..  I guess it is a convention we should follow.. however in that case `setDomApi` should do the same.  Also.. as silly as it is, we should test this

JSteunou:
So maybe I should include `setDomApi` chaining in this PR along with testing ?

rafde:
@paulfalgout @JSteunou this conflicts with https://github.com/marionettejs/backbone.marionette/issues/3242

JSteunou:
arrgh :(

paulfalgout:
I don't see how it conflicts?  The API for setting the function isn't the same as being a function meant to be overridden.  I don't think anyone should overwrite `setRenderer` or `setDomApi`

rafde:
What is meant to be overridden it vague. The time I know something isn't meant to be overridden is when property/function name has an `_` in front of it. All others are safe to override.
Clarification for https://github.com/marionettejs/backbone.marionette/issues/3242 is probably needed, then.

paulfalgout:
I think what is meant to be overridden is documented.  There will never be a case of a `_`'d function that returns `this`.  Only public API returns `this`, and those public API that do return `this` should not be overridden.  This is because those function call internally private API and could not be reproduced otherwise. Otherwise the original issue should have said "never return `this` because all public methods are overriddable" but I do not think that is the case at all.  
The `_` only distinguishes between methods you should call directly or override without fear of breaking for minor versions, and methods that you should _not_ call _or_ override because their interface and even existence is not guaranteed within a major version.

blikblum:
setRenderer and setDomApi are class methods unlike thebl cited in other issue

scott-w:
I agree with @paulfalgout that we should add a test for the `return this` and formalise the definition.

I assume this goes in a 3.5 release?

paulfalgout:
Anyone up for saying "`setRenderer` and `setDomApi` were not returning the prototype as expected" and add this as a bugfix for `3.4.2`?

test/unit/config/dom.js
```diff
@@ -22,6 +23,13 @@ chai.use(function(_chai, utils) {
 
 
 describe('DomApi', function() {
+  describe('#setDomApi', function() {
+    it('should return the current class', function() {
+      const MyObject = Backbone.Marionette.Object.extend({}, {setDomApi: setDomApi});
```
paulfalgout: it'd be nice to do unit tests without having to do a build in the future.. so it'd be nice not to use the global `Backbone` or `Marionette`.  In this case though I think you could just 
```js
const MyObject = function() { };
MyObject.setDomApi = setDomApi;
```

test/unit/config/dom.js
```diff
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import _ from 'underscore';
 import DomApi from '../../../src/config/dom';
+import { setDomApi } from '../../../src/config/dom';
```
paulfalgout: `import DomApi, { setDomApi } from '../../../src/config/dom';`

JSteunou:
Good call @paulfalgout, I pushed the changes

Issue #3417, opened 2017-08-08
Rethink the way regions are created in View


JSteunou:
From #1971 & #3389 it seems that a good clean up for view's regions would be to create those only when needed / accessed.

What about stop create regions from within View constructor, but rather create each one when being accessed, from the getter? Also the `region.el` should be created from the view el, and not being looked up from some `parentEl` option, unless the `el` is already a node or a region without parent view. This would solve the missing `region.$el` case.

paulfalgout:
I completely agree with this.  I don't think regions should do _any_ selector work.  I'm not convinced we'll want this in v4... but maybe.

JSteunou:
To much breaking change?

paulfalgout:
I think our major releases shouldn't require much app refactoring.. I think hopefully v4 can be `NextCollectionView` replacing `CollectionView`.. whatever we do with `CompositeView`.. and possibly extracting `AppRouter` among a few other small breaking changes.. perhaps there's a small breaking change for a region we could add that opens the door to doing some of this in v4.x.. but I'm a bit nervous to change too much.  it made v2->v3 too difficult

paulovieira:
Somehow related to this issue, here are some ideas to simplify the way regions are declared (not completely new I guess). 

### 1) "anonymous" regions

In this scenario we would use a boolean attribute in the template to mark the elements that will be used as the regions' containers:

```html
<div>
	...
	<nav data-mr-region></nav>
	...
	<div data-mr-region></div>
</div>
```

There would be no need to use 'regions' option anymore (at least for the anonymous regions):
 
```js
var ViewA = Mn.View.Extend({
	template: ...
	
	//regions: {   // <-- not needed!
	//	...
	//}
})
```

The api would be same as in v3, except that the regions don't have a name anymore. We would use the respective index number instead:

```js

// code inside ViewA

this.getRegion(0).showView(new ViewB)
this.getRegion(1).showView(new ViewC)

// or 

this.showChildView(0, new ViewB)
this.showChildView(1, new ViewC)
```

This would be valid only for standard regions. For custom regions we would use the standard approach.



The main use case for this new way of declaring regions would be for view with only 1 or 2 regions (which for me is quite common). Having to think of a region name everytime I need new one takes effort (plus adding it to the 'regions' option). 

------

### 2) "automatic named" regions

It's similar to the above, but for those cases where we actually want the region to have a name. I've seen other people proposing this.

```html
<div>
	...
[...]

PR #3426, opened 2017-08-14
Make sure $el contains only the el

paulfalgout:
Alternative solution to https://github.com/marionettejs/backbone.marionette/issues/3421

This is a WIP and needs tests, but I think doing this check once on `_ensureElement` is better than within `attachHtml`  And it makes `this.$el` more consistent and matches what happens in the region constructor

This still needs a test

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -133,6 +133,8 @@ const Region = MarionetteObject.extend({
     if (!_.isObject(this.el)) {
       this.$el = this.getEl(this.el);
       this.el = this.$el[0];
+      // Make sure the $el contains only the el
+      this.$el = this.Dom.getEl(this.el);
     }
 
     if (!this.$el || this.$el.length === 0) {
@@ -241,7 +243,7 @@ const Region = MarionetteObject.extend({
   // Override this method to change how the new view is appended to the `$el` that the
   // region is managing
   attachHtml(view) {
-    this.Dom.appendContents(this.el, view.el, {_$contents: view.$el});
+    this.Dom.appendContents(this.el, view.el, {_$el: this.$el, _$contents: view.$el});
   },
 
   // Destroy the current view, if there is one. If there is no current view, it does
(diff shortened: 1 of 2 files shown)


JSteunou:
Calling `getEl` twice eh... Kinda ironic. Wouldnt be better if the 1st call always return only one element. I dont see where we use multiple elements from `getEl`

paulfalgout:
I can't find a way in which returning a single element from region.getEl isn't far more complicated. 
FWIW all I did was move a getEl from attach which is called on every show up into the show which will only be called the first time. We could maybe just call this.Dom.getEl directly.. That might be better here.

scott-w:
Would jQuery `.first()` not work to resolve that issue? https://api.jquery.com/first/

paulfalgout:
We shouldn't require jQuery here at all.

scott-w:
I meant inside the DOM API somewhere.

paulfalgout:
I would consider that a breaking change

paulfalgout:
I suppose it is still an experimental API.  still though..  I'm not sure what the concern is about this change.  it is 100% the same code that is currently ready for v3.4.2 it's just that the `Dom.getEl` is now called once for a region instead of every show.

scott-w:
I'm fine with it I just wanted to play devils advocate :smiling_imp:

JSteunou:
Same... The solution just seems clunky even if right, I cannot stop thinking it must be a more elegant way

paulfalgout:
I think the more elegant way comes with https://github.com/marionettejs/backbone.marionette/issues/3417 and removing any sort of `getEl` from regions altogether.

JSteunou:
What about [here](https://github.com/marionettejs/backbone.marionette/blob/master/src/region.js#L197) ? We could return the 1st element directly.

JSteunou:
Which conflict with #3425 

Ok no more ideas :D 

:+1:

scott-w:
I'm happy with this, we should add a test, however.

paulfalgout:
updated with a test

Issue #3427, opened 2017-08-16
Prevent reInit regions on render


JSteunou:
Now Marionette has a way to set multiple Renderer logic, it should also add a way to prevent regions to be re-initialized, because some renderer allow morphing the view by preserving actual regions content.

Either `_reInitRegions` shoud become public, or a new flag should be added into View.

what do you think @marionettejs/marionette-core ?

paulfalgout:
This is nearly the same as https://github.com/marionettejs/backbone.marionette/issues/1263
I think not only is reiniting an issue, but so is recommending childviews be shown in onRender. Merely not clearing the region is not enough, you also cannot reshow the children which would also clear the previous children. 
I would resubmit the ready event as the appropriate solution  https://github.com/marionettejs/backbone.marionette/issues/3243

paulfalgout:
Oh I also think this should likely be accomplished with a flag on the region much like replaceElement. 
Either way I think this will require the view to "set" the el of a region rather than for the region to re-query the selector.

JSteunou:
I had an idea for a quick solution that could be shipped into 3.5 before going all breaking for 4.0

I though we could either add a public `reInitRegions` which would called the private one or add a flag and put an `if` statement before calling the reinit.

Because changing the actual Render logic and adding a Ready event in the view lifecycle might need a major version and I think we should be able to propose this feature without breaking.

blikblum:
> Oh I also think this should likely be accomplished with a flag on the region much like replaceElement.

The flag should be in view not in region since the proposed change would affect all regions in a view.

Since the behavior is linked to the renderer an option is to pass as an parameter to `setRenderer`

`setRenderer(myRenderer, {destructive: false})`

or change the signature of renderer, instead a function an object with a `render` function and more options/functions. Some of the work like reinitRegions could be delegated to the renderer

> I would resubmit the ready event as the appropriate solution #3243

It should work but it should not match 1:1 render event. For non destructive render would fire only in the first render

The only concern i have  is that Marionette.View has already a lot of events. I would try to replace one of the events by the new one, or change the behavior of a existing

paulfalgout:
I think a renderer wanting nondestructive regions is a single use case for nondestructive regions, but a renderer is not the only use case so it should be per region. Particularly since the current behavior is the opposite. 
[...]

PR #3428, opened 2017-08-16
Add isFirstRender flag during rendering phase

JSteunou:
Call on `showChildView` to set children views is usually set during `onRender`, but as regions & children are not always erased on re-rendering (it depends on the Renderer the user set) I think you should offer another way to know more easily if this call should be done or not.

Today we have this check

```js
if (!this.getRegion('foo').hasView()) {
  this.showChildView('foo', new myView());
}
```

or user can set its own flag

```js
onRender() {
  this._firstRender = this._firstRender === undefined;
  if (this._firstRender) {
    this.showChildView('foo', new myView());
  }
}
```

I would like Marionette offers the second logic to all Views, because it will be more versatile 

This will be a great feature along with #3427

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -117,11 +117,13 @@ const View = Backbone.View.extend({
   render() {
     if (this._isDestroyed) { return this; }
 
-    this.triggerMethod('before:render', this);
+    const isFirstRender = !this._isRendered;
+
+    this.triggerMethod('before:render', this, isFirstRender);
 
     // If this is not the first render call, then we need to
     // re-initialize the `el` for each region
-    if (this._isRendered) {
+    if (!isFirstRender) {
       this._reInitRegions();
     }
 
@@ -129,7 +131,7 @@ const View = Backbone.View.extend({
     this.bindUIElements();
 
     this._isRendered = true;
-    this.triggerMethod('render', this);
+    this.triggerMethod('render', this, isFirstRender);
 
     return this;
   },
--- a/test/unit/view.spec.js
+++ b/test/unit/view.spec.js
@@ -189,8 +189,26 @@ describe('item view', function() {
       expect(this.onBeforeRenderStub).to.have.been.calledOnce;
     });
 
+    it('should call a "onBeforeRender" with parameter on true if this is the 1st render', function() {
+      expect(this.onBeforeRenderStub).to.have.been.calledWith(this.view, true);
+    });
+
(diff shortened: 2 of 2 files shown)


blikblum:
> Today we have this check
```
if (!this.getRegion('foo').hasView()) {
  this.showChildView('foo', new myView());
}
```
Currently this does not work in `onRender` because hasView is always false because the regions are always re initialized

an alternative is to add a method ensureChildView method that would only show the child view if not exists

```
this.ensureChildView('myregion', () => new MyView({}))
```
It could be used safely inside onRender

Anyway, any solution would require an option to not reinit the regions

blikblum:
Yet another alternative is to make the show child view declarative in the region definition

```
var MyView = View.extend({
  regions: {
     myregion: {
        el: '.my-region',
        view: function () { return new MyChildView ()},
        showOnlyOnce: true
     }
  }
})
```

or


```
var MyView = View.extend({
  regions: {
     myregion: {
        el: '.my-region',
        viewClass: MyChildView,
        viewOptions: function() {return {model: this.model}}, 
        showOnlyOnce: true
     }
  }
})
```

The advantage is that the implementation of how the child view is show would not be exposed to user. We could change it safely

JSteunou:
> Currently this does not work in `onRender` because hasView is always false because the regions are always re initialized

True, I was already ahead of #3427 

I dont know if I like the declarative way with the `showOnlyOnce`, but if there is a way to make it use `ensureChildview` under the hood I'm in

paulfalgout:
Again I think this further complicates the view
```js
  initialize() {
    if(this.isRendered()) {
      // preexisting DOM
      this.showChildView('view1', new View1());
    }
  },
  onRender(view, firstRender) {
    // Will only be called if the view is re-rendered
    if(firstRender) { 
      console.log('If view had preexisting DOM I will never occur');
     }
    if(this.isRendered()) { console.log('If view had preexisting DOM  I will always occur'); }
    
  }
});
```
vs

JSteunou:
Good point I like it better !

blikblum:
> // onReady Will be called whether or not the view has preexisting el but only once  
  
onReady must be called after every render with a destructive renderer

JSteunou:
@blikblum I though we could introduced the `ready` event only for non destructive view at first (this would make the change not breaking) then on a major release make all view non destructive by default. In this major release, if a user want the destructive behavior he could call `showChildView` at each render. 
This would keep `ready` all its *way* and make it not double with `render`

Issue #3431, opened 2017-08-22
Calling render in NextCollectionView initialize can fail on empty


paulfalgout:
Currently since the empty region is initialize after `initialize` (for good reason it needs the `el`) this means that `render`ing inside of `initialize` can be breaking.

For this reason uses of the emptyRegion should utilize a method that initializes the region if it doesn't already exist.  This method could even be publicize so that the `emptyRegion` does not need to be exposed, and it would allow people to access the region inside of `initialize` which is currently impossible.

PR #3433, opened 2017-08-26
POC NextCompositeView

paulfalgout:
While I'm still not convinced of the merits of including this in Marionette, it does seem like it can be accomplished with very minimal code.

I'll make some additional notes on the breaking tests, but the known caveats..
- If the collection is reset then the compositeview template re-renders.. it didn't used to..  This seems like no big deal to me
- Because this isn't modifying `NextCollectionView` it is using `this.on('before:render'` to handle the template rendering.. I think this'll be fine in general.. however `onBeforeRender` on the `compositeView` would come prior to the template render.. but `foo.listenTo(myCompositeView, 'before:render'` would come after.. we could move most of what is between the events in `NextCollectionView` to a `_render` that could be called.. but to me that'd be an unfortunate change... however it could account for both of the caveats if we did so.

**edit** known caveats covered below

### DIFF
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -18,8 +18,6 @@ const CompositeView = CollectionView.extend({
   constructor(options) {
     this.mergeOptions(options, ClassOptions);
 
-    this.on('before:render', this._onBeforeRender);
-
     CollectionView.prototype.constructor.apply(this, arguments);
   },
 
@@ -36,10 +34,11 @@ const CompositeView = CollectionView.extend({
     this.Dom.appendContents(this.$container[0], els, {_$el: this.$container});
   },
 
-  _onBeforeRender() {
+  _render() {
     this._renderTemplate();
     this.bindUIElements();
     this._getChildViewContainer();
+    this._showChildren();
   },
 
   _getChildViewContainer() {
--- a/src/composite-view.js
+++ b/src/composite-view.js
@@ -2,9 +2,8 @@
 // --------------
 
 import _ from 'underscore';
-import deprecate from './utils/deprecate';
 import MarionetteError from './error';
-import CollectionView from './collection-view';
+import CollectionView from './next-collection-view';
 import View from './view';
 
 const ClassOptions = [
@@ -15,67 +14,17 @@ const ClassOptions = [
 
 // Used for rendering a branch-leaf, hierarchical structure.
 // Extends directly from CollectionView
-// @deprecated
 const CompositeView = CollectionView.extend({
-
-  // Setting up the inheritance chain which allows changes to
-  // Marionette.CollectionView.prototype.constructor which allows overriding
-  // option to pass '{sort: false}' to prevent the CompositeView from
-  // maintaining the sorted order of the collection.
-  // This will fallback onto appending childView's to the end.
   constructor(options) {
-    deprecate('CompositeView is deprecated. Convert to View at your earliest convenience');
-
     this.mergeOptions(options, ClassOptions);
+
+    this.on('before:render', this._onBeforeRender);
 
     CollectionView.prototype.constructor.apply(this, arguments);
   },
 
-  // Configured the initial events that the composite view
-  // binds to. Override this method to prevent the initial
-  // events, or to add your own initial events.
-  _initialEvents() {
-
-    // Bind only after composite view is rendered to avoid adding child views
-    // to nonexistent childViewContainer
-
-    if (this.collection) {
-      this.listenTo(this.collection, 'add', this._onCollectionAdd);
-      this.listenTo(this.collection, 'update', this._onCollectionUpdate);
-      this.listenTo(this.collection, 'reset', this.renderChildren);
-
-      if (this.sort) {
-        this.listenTo(this.collection, 'sort', this._sortViews);
-      }
-    }
-  },
-
-  // Retrieve the `childView` to be used when rendering each of
-  // the items in the collection. The default is to return
-  // `this.childView` or Marionette.CompositeView if no `childView`
-  // has been defined. As happens in CollectionView, `childView` can
-  // be a function (which should return a view class).
-  _getChildView(child) {
-    let childView = this.childView;
-
-    // for CompositeView, if `childView` is not specified, we'll get the same
-    // composite view class rendered for each child in the collection
-    // then check if the `childView` is a view class (the common case)
-    // finally check if it's a function (which we assume that returns a view class)
-    if (!childView) {
-      return this.constructor;
-    }
-
-    childView = this._getView(childView, child);
-
-    if (!childView) {
-      throw new MarionetteError({
-        name: 'InvalidChildViewError',
-        message: '"childView" must be a view class or a function that returns a view class'
-      });
-    }
-
-    return childView;
+  childView() {
+    return this.constructor;
   },
 
   // Return the serialized model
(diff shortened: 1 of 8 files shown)


src/composite-view.js
```diff
@@ -15,164 +14,48 @@ const ClassOptions = [
...
+        name: 'ChildViewContainerMissingError',
+        message: `The specified "childViewContainer" was not found: ${childViewContainer}`
+      });
     }
   }
 });
 
-// To prevent duplication but allow the best View organization
-// Certain View methods are mixed directly into the deprecated CompositeView
+// TODO: Extract to a TemplateRenderMixin
```
paulfalgout: If we go this way we should officially move these functions to a mixin that is also used by view.. I _kinda_ like that idea anyway

test/unit/composite-view/composite-view.child-view-container.spec.js
```diff
@@ -21,13 +21,9 @@ describe('composite view - childViewContainer', function() {
       this.CompositeView = Marionette.CompositeView.extend({
         childView: this.View,
         template: this.templateFn,
-        ui: {foo: '#foo'},
```
paulfalgout: this change to the test is actually related to this bug https://github.com/marionettejs/backbone.marionette/issues/3431

test/unit/composite-view/composite-view.child-view-container.spec.js
```diff
@@ -21,13 +21,9 @@ describe('composite view - childViewContainer', function() {
...
-        initialize: function() {
-          this.render();
-        }
+        ui: {foo: '#foo'}
       });
 
-      this.resetChildViewContainerSpy = this.sinon.spy(this.CompositeView.prototype, 'resetChildViewContainer');
```
paulfalgout: the method documented as "internal" despite not having a `_` is no longer used

test/unit/composite-view/composite-view.child-view-container.spec.js
```diff
@@ -99,7 +98,7 @@ describe('composite view - childViewContainer', function() {
     });
 
     it('should throw an error', function() {
-      expect(this.compositeView.render).to.throw('The specified "childViewContainer" was not found: #bar');
+      expect(_.bind(this.compositeView.render, this.compositeView)).to.throw('The specified "childViewContainer" was not found: #bar');
```
paulfalgout: render on `NextCollectionView` is not bound to the view

test/unit/composite-view/composite-view.child-view-container.spec.js
```diff
@@ -180,14 +179,13 @@ describe('composite view - childViewContainer', function() {
...
       this.compositeView = new this.CompositeView();
-      this.onCollectionAddSpy = this.sinon.spy(this.compositeView, '_onCollectionAdd');
     });
 
     it('should not raise any errors when item is added to collection', function() {
       expect(this.addModel).not.to.throw;
     });
 
-    it('should not call _onCollectionAdd when item is added to collection', function() {
+    it.skip('should not call _onCollectionAdd when item is added to collection', function() {
```
paulfalgout: this private function isn't on this class

test/unit/composite-view/composite-view.spec.js
```diff
@@ -47,7 +47,7 @@ describe('composite view', function() {
     });
   });
 
-  describe('when instantiating a composite view', function() {
+  describe.skip('when instantiating a composite view', function() {
```
paulfalgout: no longer deprecated

test/unit/composite-view/composite-view.spec.js
```diff
@@ -283,7 +283,6 @@ describe('composite view', function() {
       this.sinon.spy(this.compositeView, 'onRender');
       this.sinon.spy(this.compositeView, '_renderTemplate');
       this.sinon.spy(this.compositeView, 'bindUIElements');
-      this.sinon.spy(this.compositeView, 'renderChildren');
```
paulfalgout: renderChildren is not a function

test/unit/composite-view/composite-view.spec.js
```diff
@@ -521,11 +519,6 @@ describe('composite view', function() {
         expect(this.compositeView.$el).to.contain.$text('baz');
         expect(this.compositeView.$el).to.contain.$text('quux');
       });
-
-      it('shound send childView to getChildViewContainer', function() {
```
paulfalgout: this previously was documented as an internal function despite not being `_` and the API has changed anyhow.. no need to test

test/unit/composite-view/composite-view.spec.js
```diff
@@ -666,9 +659,7 @@ describe('composite view', function() {
         tagName: 'table',
         template: this.gridTemplateFn,
         childView: this.GridRow,
-        attachHtml: function(collectionView, itemView) {
```
paulfalgout: I don't know why these tests were written this way.. super odd... and not how `attachHtml` works in `NextCollectionView`

test/unit/sorted-views.spec.js
```diff
@@ -1,6 +1,6 @@
 import _ from 'underscore';
 
-describe('collection/composite view sorting', function() {
+describe.skip('collection/composite view sorting', function() {
```
paulfalgout: For this POC just commenting out this entire spec file as it is a mix between old collectionview tests and compositeview ones that no longer apply.

paulfalgout:
@marionettejs/marionette-core I think this in combination with the extracted oldcollectionview oldcompositeview should make a v4 update pretty painless.  Biggest change I think will be collectionview filters and comparators now use the view instead of the model.. but again anyone running into an issue could easily import the extracted oldcollectionview and use that.

test/unit/composite-view/composite-view.spec.js
```diff
@@ -212,7 +212,7 @@ describe('composite view', function() {
       region.show(this.compositeView);
     });
 
-    it('should call onAttach on its empty view', function() {
+    it.skip('should call onAttach on its empty view', function() {
```
paulfalgout: same issue as https://github.com/marionettejs/backbone.marionette/pull/3433/files#r135382303

test/unit/composite-view/composite-view.spec.js
```diff
@@ -370,7 +369,7 @@ describe('composite view', function() {
       expect(Backbone.Marionette.Renderer.render.callCount).to.equal(2);
     });
 
-    it('should destroy all of the child collection child views', function() {
+    it.skip('should destroy all of the child collection child views', function() {
```
paulfalgout: I believe this is an implementation detail of `NextCollectionView` always calling the `_destroyChildren` method inside render

test/unit/composite-view/composite-view.spec.js
```diff
@@ -495,7 +493,7 @@ describe('composite view', function() {
         this.collection.reset([this.m3, this.m4]);
       });
 
-      it('should not re-render the template with the model', function() {
+      it.skip('should not re-render the template with the model', function() {
```
paulfalgout: This is the main caveat.  in this version a `reset` on the collection _does_ re-render the template

src/next-collection-view.js
```diff
@@ -280,14 +287,19 @@ const CollectionView = Backbone.View.extend({
...
-    this._showChildren();
+    this._render();
 
     this._isRendered = true;
 
     this.triggerMethod('render', this);
     return this;
   },
 
+  // For overriding in CompositeView
```
paulfalgout: This was added so that composite view can define what occurs during the render.. it was actually important to do it this way so that children could be destroyed while still attached.  This clears up one of the caveats

src/next-collection-view.js
```diff
@@ -112,7 +112,14 @@ const CollectionView = Backbone.View.extend({
   },
 
   _onCollectionReset() {
-    this.render();
+    this._destroyChildren();
```
paulfalgout: Previously reset just called render.. it was convienent, but not _really_ what needed to happen.  This duplicates very little of the render, but _only_ does what needs to occur and doesn't mess with the CompositeView.  So this _is_ a breaking change as the `render` events will not be triggered... but the regular events related to children will be.  This will also make the reset consistent between collection and composite views which currently v3 it is not.

src/view.js
```diff
@@ -215,6 +119,6 @@ const View = Backbone.View.extend({
   setDomApi
 });
 
-_.extend(View.prototype, ViewMixin, RegionsMixin);
+_.extend(View.prototype, ViewMixin, RegionsMixin, TemplateRenderMixin);
```
paulfalgout: I went ahead a POC'd the TemplateRenderMixin.  I like it a lot.. we'll want to merge quite a few of the other open PRs before actually implementing this, but it looks really nice to me.

paulfalgout:
Closing in favor of #3485

PR #3435, opened 2017-08-26
Fix nextcollectionview render during init on empty

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3431

This seems like a big change, but this removed an undocumented attachment of the `emptyRegion` on a still "experimental" API.  `getEmptyRegion` also, to me, coincides better with `view.getRegion` rather than simply attaching it.  This also prevents having to call `_initEmptyRegion` for any edge case where it may be needed prior to the `el` existing.

This is however new API which should be tested and possibly documented?  I'm not sure this warrants v3.5... maybe, but I do think we should look into getting this in prior to v4 so those on v3 can attempt to switch to this collectionview prior to updating.

### DIFF
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -58,7 +58,8 @@ const CollectionView = Backbone.View.extend({
     args[0] = this.options;
     Backbone.View.prototype.constructor.apply(this, args);
 
-    this._initEmptyRegion();
+    // Init empty region
+    this.getEmptyRegion();
 
     this.delegateEntityEvents();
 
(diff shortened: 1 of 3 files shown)


src/next-collection-view.js
```diff
@@ -71,10 +72,16 @@ const CollectionView = Backbone.View.extend({
...
   // Create an region to show the emptyView
-  _initEmptyRegion() {
-    this.emptyRegion = new Region({ el: this.el, replaceElement: false });
+  getEmptyRegion() {
+    if (this._emptyRegion && !this._emptyRegion.isDestroyed()) {
+      return this._emptyRegion;
+    }
+
+    this._emptyRegion = new Region({ el: this.el, replaceElement: false });
+
```
blikblum: replaceElement is false by default, so does not need it here

paulfalgout: the reason it is here is because replaceElement: true would totally break the collectionview. It can never be true here. So if someone changes the default on the region prototype collectionview would break.

Issue #3429, opened 2017-08-17
Rename Marionette.Object to _____


paulfalgout:
For v4 we want to have named exports https://github.com/marionettejs/backbone.marionette/pull/3393

However this is complicated for both `Object` and `Error`.  `Error` seems like no big deal to rename to `MarionetteError` or something.  It's a lib that _maybe_ can be utilized by Marionette plugins, but isn't _really_ that useful to anyone else... I could also see just privatizing it.. easy enough for someone to make their own.

But for `Object`.. this has always been a naming issue.. originally called `Marionette.Controller` it was just meant to be a generic thing to acts as a mediator for things.. it wasn't a view.. it wasn't a model.. so it was named the "C" in MVC.. however that wasn't a great name.. because the view is also essentially the "C" and a `Marionette.Controller` wasn't strictly speaking the "C" either.
So the next best generic name we could come up with was `Object`.. which worked great except now that we want to export it, it's a reserved name.  bummer.

Since that time, the radio interface was added to `Object` giving it a _little_ more purpose. I believe the name "Service" is still very generic.. it likely describes what an Object is doing in most cases, and it highlights the radio interface.  There is precedence for this naming here https://github.com/benmccormick/marionette-service in the POC library for the radio interface that landed for v3.

**Bottomline**  We pretty much must rename `Object` to something. `Service` seems like the next best thing to me.  Thoughts?

blikblum:
Base?
MnObject?

JSteunou:
I am not in favor for Service. For me a *service* is like a singleton, a single app running alone... You can think of Radio being a Service but not the View classes.

I am thinking inheritance and more toward Base Root Origin...

rafde:
[MarionetteObject](https://github.com/marionettejs/backbone.marionette/blob/master/src/object.js#L19) since its already named that in the source.

paulfalgout:
@JSteunou I actually want to move away from inheritance where possible.  Right now `Object` is used as the "base" for `Application`, `AppRouter`, `Behavior`, and `Region`.
What does object provide?
It provides the radio interface, it mixes in `Backbone.Events` and the `CommonMixin`. The only code it really provides of use or any real logic is `destroy` 
[...]

PR #3437, opened 2017-08-27
Remove Marionette.Object Inheritance

paulfalgout:
~~Would resolve https://github.com/marionettejs/backbone.marionette/issues/3429~~

Before this is dismissed outright.. hear me out.. I believe there's a good case for this.

1. We know that we must rename `Object` for Mn v4 in order for exports to work correctly.  So users will _have_ to rename all instance of `Object` for v4.

2. Object was used as the base for only `Application`, `Region`, and `Behavior`.  It provided the `RadioMixin` interface, `CommonMixin` API, `triggerMethod`, `cid`, `initialize`, and `destroy`.
  - `Application` re-sets options in order for options to be available for `_initRegion`; otherwise it is a `Object`
  - `Behavior` re-sets options as well so they would be available for `ui`.  Additional it completely overrode the `destroy` function and did not implement `_isDestroyed` breaking `isDestroyed()`.  Additionally Behavior gets the full `RadioMixin` API which we do not give to Views because Radio requests cannot be duplicated.. so it does not follow that a behavior should allow for it.  It gets almost no benefit from being an `Object` if anything it causes more issues than it helps.
  - `Region` again re-sets options.  Additionally it added functionality within it's `destroy` that occurred unfortunately before `before:destroy` due to inheritence.  It also broke the destroy api by only passing `options` (though I think it may be reasonable to do that everywhere anyhow, I didn't make that change here)

So I think there is a case, whether or not we decide to remove Object that it should be removed from the inheritence chain of `Region` and `Behavior`.  So this would leave _only_ `Application`.

`Application` has historically been a single point of origin for the app and prior to v3 you could not have multiple instances of it.  However in v3 you can, making `Application` a slightly more useful `Object` with very little perf differences... if any..  I personally have nearly stopped using `Object` altogether using `Application` in its place and find it useful.

Regardless if anyone finds it useful, reducing the API and the inheritance is a good thing.  The inheritance is not buying us anything.  It definitely does not reduce much if any repetition..  Again users will have to rename `Object` as it is.. let's simply remove it.

For those wanting to keep `Object` by some other name.. why would `Application` not work for that use case?

For those wanting to keep `Object` as an inheritance.. why?

### DIFF
--- a/docs/marionette.application.md
+++ b/docs/marionette.application.md
@@ -198,6 +198,10 @@ app.start({
 The Marionette Application provides helper methods for managing its attached
 region.
 
+### `initialize`
+
+### `triggerMethod`
+
 ### `getRegion()`
 
 Return the attached [region object](./marionette.region.md) for the Application.
@@ -212,8 +216,6 @@ Return the view currently being displayed in the Application's attached
 `region`. If the Application is not currently displaying a view, this method
 returns `undefined`.
 
-### Marionette.Object Methods
+### `isDestroyed`
 
-`Marionette.Application` extends `Marionette.Object` and, as such, implements
-the same method interface. See the [`Object`](./marionette.object.md)
-reference for the full list.
+### `destroy`
--- a/docs/marionette.object.md
+++ b/docs/marionette.object.md
@@ -1,8 +1,6 @@
 # Marionette.Object
 
-A base class which other classes can extend from.
-Object incorporates many backbone conventions and utilities
-like `initialize` and `Backbone.Events`.
+Object incorporates backbone conventions `initialize` and `Backbone.Events`.
 Object has all of the [Common Marionette Functionality](./common.md).
 
 ## Documentation Index
--- a/src/mixins/destroy.js
+++ b/src/mixins/destroy.js
@@ -1,0 +1,18 @@
+export default {
+  _isDestroyed: false,
+
+  isDestroyed() {
+    return this._isDestroyed;
+  },
+
+  destroy(...args) {
+    if (this._isDestroyed) { return this; }
+
+    this.triggerMethod('before:destroy', this, ...args);
+    this._isDestroyed = true;
+    this.triggerMethod('destroy', this, ...args);
+    this.stopListening();
+
+    return this;
+  }
+};
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -7,8 +7,6 @@ identify where your region must be displayed.
 See the documentation for [laying out views](./marionette.view.md#laying-out-views---regions) for an introduction in
 managing regions throughout your application.
 
-A Region is a [`Marionette.Object`](./marionette.object.md) and has all of its functionality.
-
 Regions maintain the [View's lifecycle](./viewlifecycle.md#regions-and-the-view-lifecycle) while showing or emptying a view.
 
 ## Documentation Index
--- a/src/application.js
+++ b/src/application.js
@@ -1,26 +1,52 @@
 // Application
 // -----------
+
+import _ from 'underscore';
+import Backbone from 'backbone';
+import extend from './utils/extend';
 import buildRegion from './common/build-region';
-import MarionetteObject from './object';
+import triggerMethod from './common/trigger-method';
+import CommonMixin from './mixins/common';
+import DestroyMixin from './mixins/destroy';
+import RadioMixin from './mixins/radio';
 import Region from './region';
 
 const ClassOptions = [
+  'channelName',
+  'radioEvents',
+  'radioRequests',
   'region',
   'regionClass'
 ];
 
-// A container for a Marionette application.
-const Application = MarionetteObject.extend({
+const Application = function(options) {
+  this._setOptions(options);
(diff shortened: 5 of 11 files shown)


JSteunou:
I agree with all the exposed points but we will have to find a very good way to explain that little bizarre thing about Application being used like Object when not used like a true application.  This aside I found this solution the better for our Object issue.

blikblum:
> For those wanting to keep Object as an inheritance.. why?

I'm all for removing the inheritance, independent of removing Object.

> For those wanting to keep Object by some other name.. why would Application not work for that use case?

Application is a specific name. Using as a generic class will kill its semantic, e.g., in marionette.routing  [Route class](https://github.com/blikblum/marionette.routing/blob/master/src/route.js) extends from Object. I can't imagine a Route extending from an Application. In a project i would have an Application class used by its original purpose and various other Application classes with a different role.

Also it would kill Mn.Inspector that uses Application to detect the view tree.

BTW: the Marionette.Object code would still be there, but with a different name (Application)

In conclusion i would remove the inheritance for Region, Behavior. Keep Object with whatever name and extends Application from Object

I would remove also merging ClassOptions in Object level

bazineta:
I'm in general agreement with the above comments. In general when one arrives at looking at a suitable description for a base class (i.e., object, widget, component, etc.) it can be an indication that it's really just a bag of functions, i.e., lacks substitutability, but all metaphors break down somewhere. I guess something equally generic, such as 'Component' or similar generic term might be warranted, rather than the 'use Application as an Object' thing, which seems a bit obtuse.

paulfalgout:
I would argue that its original purpose was mostly to store a bunch of global stuff along with the wreqr object and module api. When all of those things are removed it seems reasonable that an app could be a starting point for a view at any level and does not need to be a singular instance for any reason other than supporting `Mn.Inspector`.  Mn users have regularly called mediator objects they use after the `Application` "sub-apps" and I'd hope now they'd reuse Application, as I suspect it provides a cleaner way to attach a layout to a "sub-app."  I certainly take that concept much further with `Marionette.Toolkit` so I suppose perhaps I have some bias.
I'd also argue that `Mn.Inspector` shouldn't rely on a single `Application` instance to detect the view tree.. particularly since `Application` can, and I think should, be reused.

Application adds 2 things to `Object`: an optional region, and a `start` method which is merely:
```
start() {
  this.triggerMethod('before:start');
  this.triggerMethod('start');
}
```
We could just as easily suggest that users, "take an Object and make a function called `start` and put the stuff you need in it"  For most users apps it would make absolutely no difference.

Despite the history of `Application` or how Inspector treats it, I find little reason why it can't be described as something generic enough to be extended into a router or any other thing.  Like the Radio API on Object I see nothing that suggests that you should only utilize `Object` if you intend to use the Radio API.  Likewise I don't suggest that using `Application` means you intend to attach a view or call `start`.

If removing `Object` made `Application` too complicated, or possibly if `Object` had any clear purpose apart from `Application`, I could see wanting two classes and two sets of documentation.. but `Application` including `Object` is under 100 lines.  And again we have to rename `Object` to some other generic term.. outside of the context of Mn history, "Application" is a pretty generic term, "...designed to perform a group of coordinated functions, tasks, or activities..." according to Wikipedia.  I would suggest that if `Application` wasn't already taken it'd be a fairly reasonable name for `Object`.

denar90:
This one sgtm, I just wanna leave comment about the way we've been using `Object` in our apps. For building components `Object` was kinda entery point (mediator) to them, so in this case, case of removing, it will be a bit paintfull transition. 
I'm also 👎  for using `Application` as `Object` replacement, but I dunno about new name for it due to @bazineta's 
Maybe drop it at all and leave it up to the users. Maybe they can use https://github.com/marionettejs/backbone-metal or just es6 class for their needs.

docs/marionette.application.md
```diff
@@ -198,6 +198,10 @@ app.start({
 The Marionette Application provides helper methods for managing its attached
 region.
 
+### `initialize`
```
paulfalgout: TODO: Need to figure out where this API should be documented as it'll be duplicated in other places, but it isn't common across everything either

docs/marionette.object.md
```diff
@@ -1,8 +1,6 @@
 # Marionette.Object
 
-A base class which other classes can extend from.
-Object incorporates many backbone conventions and utilities
-like `initialize` and `Backbone.Events`.
+Object incorporates backbone conventions `initialize` and `Backbone.Events`.
```
paulfalgout: I removed the "A base class which other classes can extend from." since it is no longer true, but now this class doesn't have a description.. what should it be?  My suggestion would be "Essentially and Application without a region that has no particular use case" ;-p  
Seriously though, perhaps the description can come with the rename, but I don't know what this is for.

src/mixins/destroy.js
```diff
@@ -0,0 +1,18 @@
+export default {
```
paulfalgout: This is essentially the shared functionality between Object and Application.. moved to a mixin

paulfalgout:
OK I reinstated Object as its own thing and nothing extends from it.  Here is what an Object is:

PR #3438, opened 2017-08-27
Replace CollectionView with NextCollectionView

paulfalgout:
Getting the ball rolling.
This also removes CompositeView.
It can be re-added in another PR.


paulfalgout:
@marionettejs/marionette-core I believe this is good to go.  It _does_ remove `CompositeView` (mostly because it has to), and also the upcoming `CompositeView` will be pretty different.

JSteunou:
:+1:  but conflict must be solved

paulfalgout:
resolved

PR #3442, opened 2017-09-06
Clarify wording for NextCollectionView.isEmpty docs

wuservices:
Previous wording made it sound like the method was called once, then 2 more times for a total of 3 times instead of just twice overall.


paulfalgout:
👏

PR #3446, opened 2017-09-14
Remove triggerMethodOn

blikblum:
### Proposed changes
 - Removes `triggerMethodOn`
 - Add `BackboneViewMixin` to support vanilla `Backbone.View`

`triggerMethodOn` is an util function used to support View classes that do not implement `triggerMethod` method and is called in various places. Its implementation adds overhead by adding an extra function call and manipulating the arguments which can impact performance specially in `CollectionView`. Currently this performance penalty affects all Marionette users regardless using vanilla `Backbone.View` or not

With this change, if the developer needs to use vanilla `Backbone.View` with `Marionette` it should apply the `BackboneViewMixin` once, in the start of the app:

```
_.extend(Backbone.View.prototype, Marionette.BackboneViewMixin);
```

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -11,6 +11,8 @@ import getOption from './common/get-option';
 import normalizeMethods from './common/normalize-methods';
 import monitorViewEvents from './common/monitor-view-events';
 
+import BackboneViewMixin from './mixins/backboneview';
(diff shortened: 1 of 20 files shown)


blikblum:
I've started documenting at #3447

PR #3447, opened 2017-09-16
Upgrade guide v3 - v4

blikblum:
### Proposed changes
 - Start documenting v4 changes
 - The proposed changes to ease the migration are grouped together instead of at each v3 change. The reasonably is that the developer will be looking for these changes at the same stage. Personally, when i was migrating some apps i found the current way confusing
 - For each change document old and new behavior, reason and remedy

Not in the PR but i propose to separate v2 to v3 and v3 to v4 in its own documents


scott-w:
I agree with your proposal. v2 to v3 is big enough that it deserves its own documentation - in my view - so readers don't get confused.

scott-w:
That said, I'm having a quick look now and I'm not sure it's worth spending time re-writing the v2 to v3 guide unless people are having major issues with it. From experience, I know that v3 to v4 will take up  more time than it seems initially 😉

blikblum:
Just updated the guide. Splitted v3-v4 changes into own docs.

Also splitted v2-v3 but as @scott-w pointed i think is not worth completing that

blikblum:
Just updated the migrate guide, already accounting for the possible change to restore default export and Marionette.Object

By my side, this is done, although may be necessary to expand on the CollectionView upgrade. Since i dont use it i dont have knowledge

paulfalgout:
This is a good start but there's quite a bit more.. to name a few.. router was extracted, templateCache was extracted, render is no longer bound to the view instance, behavior no longer has `defaults` `childViewEventPrefix` is now `false` by default, Region no longer has a `selector`.. there are more..

paulfalgout:
Here's a list of the commits I'm looking to make sure are documented for the upgrade.
Some are pretty minor and some are just feature additions but should be mentioned.
Some are already taken care of on this PR.  Anyhow.. I'll try to have these this week in the format proposed here.  I'm also currently thinking we should rename the current `upgrade.md` to `upgrade-v2-v3.md` rather than introducing or rewriting it as mentioned above.

Changed Application showView and Region showChildView to return the v…,
Remove clones from serialization but extend empty on mixin,
Do not bind render method in View classes,
Add Marionette.Events for Backbone.View
Remove triggerMethodOn,
Remove isNodeAttached, use this.Dom.hasEl instead,
Remove defaults property from Behavior (#3459),
Remove selector Region option,
Do not call attachElContent when template returns undefined (#3469),
Remove support for preventDestroy Region option (#3464),
Remove behavior lookup (#3462),
Remove renderer / TemplateCache (#3482),
Replace CollectionView with NextCollectionView,
Do not inject merged options into View constructor arguments (#3374),  - options on the view declaration are not passed to the bb.view constuctor
set childViewEventPrefix flag to false (#3487),
reimplement template: false to prevent render events (#3490),
implement bindRequests on views (#3492),
Make CollectionView a CompositeView,
Make DEV_MODE a feature flag (#3494),
Remove AppRouter (#3493),
Replace `_.bind` with Function.prototype.bind (#3514),
Remove noConflict,
Remove Marionette.Object inheritance,
Make MarionetteError a private util,
Propagate only the first argument (options) passed to destroy,
Propagate only the third argument (options) passed to showChildView,
Support event arguments with View#triggers,
Add jsnext:main field to package.json,
Allow for use of Underscore 1.9.x.,

paulfalgout:
This is your PR @blikblum so I can't add you as a reviewer, but checkout the latest commit.  I _think_ I've covered everything other than new features which can be in the change log with the bump and build and maybe in the blog post.  Alright.. I think v4 is very close now

blikblum:
It's all good by me.

blikblum:
Regarding v2-v3 upgrade guide, the old guide IMO is really confusing. Talk by own experience when i was upgrading from v2 to v3. I can try to finish the migration to the new format if deems appropriate

paulfalgout:
Sounds fine to me, but maybe we can add it as a different PR so it doesn't slow down anything since it can be changed at any time?

blikblum:
Of course to a different PR

PR #3449, opened 2017-09-17
Avoid crash when instantiating a view with a non existing el

### LINKED ISSUE #3448: An exception occurs if el is set to an non existent selector
### Description

 1. Trying to pass a non existent selector to `el` gives a TypeError inside `Dom.hasContents`. [Example](https://jsfiddle.net/b2yz38gj/77/)

### Expected behavior

Should not crash

### Actual behavior

It crashes

### Environment

1. Marionette version: 3.4.3
2. Backbone version: 1.3.3

This bug was exposed by #3397 but probably existed before because `hasEl` variable was misleading. changing the check in setElement to: 

```    
    Backbone.View.prototype.setElement.apply(this, arguments);
    const hasEl = !!this.el;
    this._isRendered = hasEl && this.Dom.hasContents(this.el);
    this._isAttached = hasEl && isNodeAttached(this.el);
```

fixes the issue.

Another bug that fixes is when passing $el directly when isAttached and isRendered return misleading values. [Example](https://jsfiddle.net/b2yz38gj/78/)

blikblum:
Fixes #3448

### DIFF
--- a/src/config/dom.js
+++ b/src/config/dom.js
@@ -92,7 +92,7 @@ export default {
 
   // Does the el have child nodes
   hasContents(el) {
-    return el.hasChildNodes();
+    return !!el && el.hasChildNodes();
   },
 
   // Remove the inner contents of `el` from the DOM while leaving
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -193,13 +193,9 @@ const CollectionView = Backbone.View.extend({
   // if an el was previously defined. If so, the view might be
   // attached on setElement.
   setElement() {
-    const hasEl = !!this.el;
-
     Backbone.View.prototype.setElement.apply(this, arguments);
 
-    if (hasEl) {
-      this._isAttached = isNodeAttached(this.el);
-    }
+    this._isAttached = isNodeAttached(this.el);
 
     return this;
   },
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -253,13 +253,9 @@ const CollectionView = Backbone.View.extend({
   // if an el was previously defined. If so, the view might be
   // attached on setElement.
   setElement() {
-    const hasEl = !!this.el;
-
     Backbone.View.prototype.setElement.apply(this, arguments);
 
-    if (hasEl) {
-      this._isAttached = isNodeAttached(this.el);
-    }
+    this._isAttached = isNodeAttached(this.el);
 
     return this;
   },
--- a/src/view.js
+++ b/src/view.js
@@ -91,14 +91,10 @@ const View = Backbone.View.extend({
   // if an el was previously defined. If so, the view might be
   // rendered or attached on setElement.
   setElement() {
-    const hasEl = !!this.el;
-
     Backbone.View.prototype.setElement.apply(this, arguments);
 
-    if (hasEl) {
-      this._isRendered = this.Dom.hasContents(this.el);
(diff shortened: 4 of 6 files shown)


src/config/dom.js
```diff
@@ -92,7 +92,7 @@ export default {
 
   // Does the el have child nodes
   hasContents(el) {
-    return el.hasChildNodes();
+    return !!el && el.hasChildNodes();
```
paulfalgout: new behavior should be tested here: https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/config/dom.js#L265

blikblum:
This PR restore the behavior of before adding `Dom.hasContents`. IMO changing the behavior like throwing an error or warning should be done in a separate issue

paulfalgout:
I don't disagree.  The request for change was for the test in the DOM unit test file.  I was merely asking if there's a use case I'm missing for this I'm missing.

paulfalgout:
Can you rebase this off master?  Since `next` is 4.0 we'll need to handle bugfixes off `master`

blikblum:
> Can you rebase this off master? Since next is 4.0 we'll need to handle bugfixes off master

No problem, but should not be also on next?

paulfalgout:
We can keep `next` up to date with changes to master on releases.

blikblum:
> We can keep next up to date with changes to master on releases.

Yes, but if we decide to change the behavior throwing an error will be conflicts. It's a shortcoming of the branch layout.

I'll rebase anyway

src/config/dom.js
blikblum: Done

blikblum:
Added a discussion about the possibility of changing the behavior at #3451

paulfalgout:
😊 sooo now this will also have to be pulled against next as it might conflict with the removal of `isNodeAttached`  Similarly if we're getting rid of the `hasEl` check, I think maybe we should also do that for `NextCollectionView` https://github.com/marionettejs/backbone.marionette/blob/master/src/next-collection-view.js#L260

JSteunou:
I'm used to a kind of git flow system with better *support* branch for in production releases. This works very well 

master / develop / support/X.Y / feature/younameit

paulfalgout:
I don't see how gitflow solves a change need to a divergent master and develop branch.  This seems like renaming next to develop and adding additional work for rolling patch releases

blikblum:
Code wise is there something else to be done in this PR?

I will need to open another PR on next anyway so better settle the changes here

paulfalgout:
I think the collectionviews only have the `if (hasEl) {` in their `setElement` because `View` had the same pattern, but with this change it doesn't seem necessary and should probably match this change

blikblum:
CollectionView will not be affected by the bug because `isNodeAttached` will always return false (and not crash). Anyway i think is a good thing to make the same change

NextCollectionView crashes when passing a non existing selector in getEmptyRegion with a NoElError: An "el" must be specified for a region, regardless of changing `setElement`

Any idea how to fix?

blikblum:
I updated CollectionViews setElement. The NextCollectionView issue is just an example of problems with an view with empty el and should, IMO, should be handled at #3451.

src/config/dom.js
blikblum: @paulfalgout can you dismiss this request?

Issue #1871, opened 2014-09-12
Consider creating a short-name alias for the Marionette object


paulovieira:
Instead of typing `Marionette` all the time (how many characters is that? 25?), why not use simply `M`? It would go nicely with the logo and save a lot of typing!

Before (long and ugly):

``` javascript
var myLV = Marionette.LayoutView.extend({ ... })

var MyCV = Marionette.CollectionView.extend({ ... })
```

After (short and simple):

``` javascript
var myLV = M.LayoutView.extend({ ... })

var MyCV = M.CollectionView.extend({ ... })
```

This might seem a bit disturbing at first as the code will have a different look, but it makes a lot of sense once you get used to it.

If you think about it, pretty much all the libraries have a short-name alias available, and not necessarily some funny character like `$` or `_`. For instance, the Leaflet mapping library uses `L`:

``` javascript
var myMap = new L.Map('myMap')
L.marker([51.5, -0.09]).addTo(myMap)
```

See http://leafletjs.com/

PR #3452, opened 2017-09-17
Remove Mn alias

blikblum:
### Proposed changes
 - Remove Mn alias that is set to global object (window) in umd build. This does not align with current best practices and can be easily set for those that use it: `window.Mn = window.Marionette`

### DIFF
--- a/gulp/build.js
+++ b/gulp/build.js
@@ -36,7 +36,6 @@ function generateBundle(bundle) {
     sourceMap: true,
     sourceMapFile: 'backbone.marionette.js',
     banner: banner,
-    footer: 'this && this.Marionette && (this.Mn = this.Marionette);',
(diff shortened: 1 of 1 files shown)


paulfalgout:
:-1: I don't really have any issue with the alias, and find it somewhat useful.  As we move towards es6 exports, it will hold far less importance and thus it's best purpose will be highlighted which I believe is for quick sandboxing.
The original winning argument for adding the alias is still true: https://github.com/marionettejs/backbone.marionette/issues/1871#issuecomment-55438026

So I'm not sure where this does not align with "current best practices"

You'll also notice that in the original argument it was suggested that anyone can easily create an alias, and yet it was added anyways.

blikblum:
Ok. It's not really a big deal. Closing

PR #3457, opened 2017-09-19
Add space-delimited list of events support for triggerMethod

nlhuykhang:
### Proposed changes
 - `triggerMethod` supports space-delimited list of events

Link to the issue:
#2937


paulfalgout:
Yeah I agree that, in the very least, any change like this would need to be perf scrutinized to the nth degree.  I think the help wanted tag might be misplaced.  I think this issue is entirely dependent on the possible replacement of `Backbone.trigger`  if/when that happens.. if/when the Bb API needs to be upheld.

nlhuykhang:
Close this as it is not needed

Issue #3347, opened 2017-04-12
Deprecate behavior defaults 


paulfalgout:
defaults on behaviors are already redundant to how `options` work.

Setting defaults now:
```js
var MyBehavior = Marionette.Behavior.extend({
  defaults: {
    foo: 'bar'
  }
});
```
Also setting defaults now:
```js
var MyBehavior = Marionette.Behavior.extend({
  options: {
    foo: 'bar'
  }
});
```

blikblum:
Has my approval. I suggest deprecate in 3.3, remove in v4

rafde:
I had no idea this existed.

PR #3459, opened 2017-09-21
Remove defaults property from Behavior

blikblum:
### Proposed changes
 - Followup to #3347

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -32,13 +32,7 @@ const Behavior = MarionetteObject.extend({
     // to the view.
     this.view = view;
 
-    if (this.defaults) {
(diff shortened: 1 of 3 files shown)


src/behavior.js
```diff
@@ -32,13 +32,7 @@ const Behavior = MarionetteObject.extend({
     // to the view.
     this.view = view;
 
-    if (this.defaults) {
-      deprecate('Behavior defaults are deprecated. For similar functionality set options on the Behavior class.');
```
paulfalgout: this might be the only use of `deprecate` so the import could be removed.

blikblum: Done

PR #3463, opened 2017-09-24
Remove reference to triggerMethodOn on main file

blikblum:
### Proposed changes
 - Followup to #3446 
 - Fixes build

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -23,8 +23,7 @@ import {
 } from './common/bind-requests';
 
 import {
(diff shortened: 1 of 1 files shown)


paulfalgout:
I don't know if this is the best place for this or not, but most of the `common` files have only a default export.  `triggerMethod` might as well be the default export

blikblum:
I can do that. I think a separate PR i better in case we regret.

paulfalgout:
@marionettejs/marionette-core an easy 👀

PR #3465, opened 2017-09-24
Remove selector option in Region definition

blikblum:
### Proposed changes
 - Using selector in region definition is deprecated. Should use el instead

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -83,7 +83,7 @@ var MyView = Mn.View.extend({
 ### Additional Options
 
 You can define regions with an object literal. Object literal definitions expect
-an `el` property - the CSS selector string to hook the region into. With this
+an `el` property - the selector string to hook the region into. With this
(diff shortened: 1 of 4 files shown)


docs/marionette.region.md
```diff
@@ -83,9 +83,9 @@ var MyView = Mn.View.extend({
 ### Additional Options
 
 You can define regions with an object literal. Object literal definitions expect
-an `el` property - the jQuery selector string to hook the region into. The
-object literal is the most common way to define whether showing the region
-overwrites the `el` or just overwrites the content (the default behavior).
+an `el` property - the CSS selector string to hook the region into. With this
```
paulfalgout: I think I'd prefer just "the selector"  For one it's not necessarily a selector used in a style sheet, but 2 there are certainly possible jQuery selectors that aren't possible in CSS.

blikblum: done

PR #3466, opened 2017-09-24
Simplify buildRegionFromDefinition (assign property directly instead of using _.extend)

blikblum:
### Proposed changes
 - Using _.extend is overkill at those places


paulfalgout:
no way this is controversial.  merging

Issue #3265, opened 2016-11-17
Add lang linter for docs


denar90:
Sometimes errors in spelling might appear, it will be nice to get to know about them without disturbing code reviewer, other then that reviewers are people too.
I propose to use either https://github.com/btford/write-good or another tool as additional checking.

nlhuykhang:
I made a Gulp plugin for write-good [here](https://github.com/nlhuykhang/gulp-write-good).

Have a look at the plugin. If the plugin is good, I can make a PR to use it in the docs linting process.

PR #3468, opened 2017-09-26
Add gulp-write-good plugin to lint docs

nlhuykhang:
### Proposed changes
 - Add [gulp-write-good](https://www.npmjs.com/package/gulp-write-good) plugin to docs linting process

Link to the issue:
#3265


paulfalgout:
oh also `yarn` should be run so that the lock file is updated

Issue #3307, opened 2017-01-30
Deprecate template: false


paulfalgout:
`template: false` is a terrible API.  It is a hack that in v2 short-circuited the rendering process so that people could attach views to existing DOM, but could kick-start the view's lifecycle by calling render.

In v3 you can attach views to existing DOM, but you should _not_ need to call render to get the view setup.  Calling render on an already rendered view doesn't make sense.  You should only call `render` if you intend to render a template into the view's el and we shouldn't hack solutions for other purposes.

Today, `template: false` has very little value other than confusing people and perpetuating the hack.  I use it occasionally when I'm lazy in a js fiddle.. but even then it is entirely unnecessary.  Simply:
```js
var myView = new Mn.View({ template: _.noop });
myView.render(); // doesn't crash / doesn't render anything
```

Of course anyone using `template: false` to attach a view to existing DOM would have to switch to passing the view the DOM element.. so it could certainly be a less-than-ideal breaking change for some people..  Hence the deprecation... no need to remove it right away.

scott-w:
I think that's a good call - it'll let us remove the special case for checking `false`. We could start with the documentation for 3.2 or 3.3 and add a simple warning in the code itself.

paulfalgout:
Good point.. I removed the `major` label since it's certainly not breaking just to deprecate it.

JSteunou:
If you do not use `template: false` on an existing DOM element, do you still have to call `render` ?

JSteunou:
Nice! So indeed, what are the use cases for a `template: false` anymore?

paulfalgout:
_Maybe_
```js
const MyEmptyTextarea = Mn.View.extend({
  tagName: 'textarea',
  template: false
});
```
but even then.. I dunno if it is worth keeping around as 
```js
const MyEmptyTextarea = Mn.View.extend({
  tagName: 'textarea',
  template: _.noop
});
```

bazineta:
To answer @JSteunou, at least in my own experience it's useful when a View is being used as a proxy for a jquery plugin. For example, it can be handy to pass a Collection to a View that's proxying a bootstrap-select, informing the select of the items it should display. The view really just deals with this.$el, so there's no template, and one gains the benefit of being able to automatically destroy the select plugin in onBeforeDetach for leak avoidance.

blikblum:
@bazineta does `template: _.noop` works in this use case?

bazineta:
[...]

PR #3470, opened 2017-09-29
Remove support for passing false to template option

blikblum:
### Proposed changes
 - Followup to #3307


src/view.js
```diff
@@ -136,12 +135,6 @@ const View = Backbone.View.extend({
...
     const template = this.getTemplate();
 
-    // Allow template-less views
-    if (template === false) {
-      deprecate('template:false is deprecated.  Use _.noop.');
-      return;
-    }
-
     // Add in entity data and template context
     const data = this.mixinTemplateContext(this.serializeData());
```
rafde: having this line still execute is the odd part in the code.
the template context and serializeData will do work that might not get used.
I can't figure out what to do about that needless work, but template false did solve it.

paulfalgout: I do think that there will be some less than ideal scenarios for when users are wanting views that do not contain renderable content.  Granted I would also wonder what the use case is for `templateContext` and to some degree model data when no rendering is to be done..  So while, yes, there's some extra work being done, I don't see the use case where it is _really_ making that much of a difference.

On the contrary `template: false` has been predominately used as a method for attaching children within `onRender` when the view itself _should not be rendered_  That is a terrible terrible use of this API and we should do whatever we can to avoid it being the prominent case.  Instead I believe we should provide a better hook for when child views should be added as `onRender` is not good enough.

So I would argue that keeping this functionality is far worse than removing.

paulfalgout:
this moved to the template-render mixin and composite view is currently gone.

src/view.js
paulfalgout: In light of https://github.com/marionettejs/backbone.marionette/issues/3484 can we push this through @rafde ?

blikblum:
Close in favor of #3490

Issue #3467, opened 2017-09-25
NextCollectionView add at end does not account for an unsorted collection


paulfalgout:
By default if items are added to a backbone collection and all of the view are added to the end of the collection it only renders the new views.  This is a problem _if_ a collection is _not_ currently sorted to its `comparator` (as in it is changed after anything is added).  `NextCollectionView` will follow the sort of the collection as-is.. when the item is added backbone will resort the collection, but marionette assumes that the backbone collection is already in order and adds the new view to the end.

So for instance this will not work:
```js
// assuming the collection isn't currently sorted by foo
myCollectionView.collection.comparator = 'foo';
// assuming this view is added to the end
myCollectionView.collection.add({ id: 10, foo: 'zzbar' }); 
```

paulfalgout:
ok after some pretty indepth searching I think the add at end perf gain can only happen if `sortWithCollection` is false.  The perf gain comes from assuming the rest of the collection hasn't changed, but this can't be done with `sortWithCollection` without adding a non-trivial check to prove consistency.  Plus if you're super worried about perf then `sortWithCollection` isn't the best option next to `viewComparator`

PR #3471, opened 2017-10-02
Use add at end perf only when sortWithCollection is false

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3467

Because a collection can be modified without the full understanding on the collectionview, the optimistic adding that reduces perf is not ideal when `sortWithCollection: true`  This opens the door to all kinds of edge cases.  So we should only apply it when the collectionview itself is maintaining the sort.

However this brought up a related issue.  If a collection was `sortWithCollection: false` the _default_ `viewComparator` was still sorting with collection on add or update which mean that setting it to false only prevented a resort if `collection.sort()` was called directly, but not for any other circumstance.  If the collectionview is not maintaining the collection's sort, this prevents the sort from occurring at all.

In `getComparator` I arranged the `return false` check based on the order I suspect matters most when tweaking perf.  It is unlikely that `sortWithCollection` is true in extreme perf situations so I went with that one first.  and if `!collection.length` you likely don't really have perf concerns so I left that one to last.  Granted it's so cheap it probably doesn't matter.  Similarly though I'm returning `false` as this function is meant to be overridden. It isn't necessary, but it might be more clear for the user.

### DIFF
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -358,11 +358,11 @@ const CollectionView = Backbone.View.extend({
 
   // Sorts views by viewComparator and sets the children to the new order
   _sortChildren() {
-    if (this.viewComparator === false) { return; }
+    let viewComparator = this.getComparator();
+
+    if (!viewComparator) { return; }
 
     this.triggerMethod('before:sort', this);
-
-    let viewComparator = this.getComparator();
 
     if (_.isFunction(viewComparator)) {
       // Must use native bind to preserve length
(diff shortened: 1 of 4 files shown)


src/next-collection-view.js
```diff
@@ -433,7 +438,7 @@ const CollectionView = Backbone.View.extend({
     delete this._addedViews;
 
     if (!viewFilter) {
-      if (addedViews && _.every(addedViews, _.bind(this._isAddedAtEnd, this))) {
+      if (!this.sortWithCollection && addedViews && _.every(addedViews, _.bind(this._isAddedAtEnd, this))) {
```
rafde: maybe start switching `_.bind` to `this._isAddedAtEnd.bind(this)`?

paulfalgout: I think to keep consistency we should make stylistic or strategic changes in singular PRs so we don't end up with multiple ways of doing the same task

PR #3478, opened 2017-10-03
Fix test-browser

paulfalgout:
BBView was being created outside of the test scope which is problematic when relying on globals instead of imports for browser testing.

I thought saucelabs was just being flaky as it definitely was on and off for a while, but turns out the tests were just broken.


paulfalgout:
Since this doesn't touch anything in `src/` and because this'll fix tests for every other `next` branch I'm going to merge it on 1 👍

PR #3480, opened 2017-10-04
Add module field to package.json

blikblum:
### Proposed changes
 - Add module field to package.json so bundlers can enable tree shaking: https://github.com/rollup/rollup/wiki/pkg.module

It was added and removed in v3 due to backward compatibility (#3418). With v4 we can do the breaking change

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "3.4.3",
   "homepage": "https://marionettejs.com/",
   "main": "lib/backbone.marionette.js",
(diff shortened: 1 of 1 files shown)


paulfalgout:
I suppose technically we shouldn't do this unless https://github.com/marionettejs/backbone.marionette/pull/3393 will get in, but I think we can commit to that making it

blikblum:
I think we can wait

Issue #3477, opened 2017-10-03
setRenderer should be configurable from Marionette


paulfalgout:
`Marionette.setDomApi` will set the DOM API for every class that has the option in Marionette.

We should do the same for `setRenderer` which currently is only available on `Marionette.View`, but if `CompositeView` is rebuild with `NextCollectionView` that will also need the renderer API.

Setting at the top is consistent.

Issue #3440, opened 2017-09-02
Extract Template Cache Renderer for v4


paulfalgout:
The TemplateCache renderer is dated.  Mainly used for caching templates pulled from `<script>` tags in the DOM.  It never gets updated.  We should pull this out into its own repo so that the default renderer can simply be `return Template(data);`

We will have an easy interface in v4 for importing the old render and applying it to whatever views need it.  It's a small perf improvement for most people.. it'll make the lib a tiny bit smaller.. but more importantly it reduces the amount of documentation and API to read about when learning the library.

PR #3482, opened 2017-10-04
Remove Renderer & TemplateCache

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3477 https://github.com/marionettejs/backbone.marionette/issues/3440

`TemplateCache` should be made as an external renderer and does not need to be the default.
Additionally this moves the view rendering to a mixin which is I think nice organization and in anticipations of the next `CompositeView` work: https://github.com/marionettejs/backbone.marionette/pull/3433

This also adds a top level `setRenderer`

I'm also working on the abstracted `TemplateCache` Renderer now


blikblum:
Any idea how to migrate current demos (jsfiddle/codepen) that uses `template: '#template-id'`?

paulfalgout:
yep so when any of those are upgraded they'll need to add the external `TemplateCache` renderer and `Mn.setRenderer(TemplateCache.render)`.

paulfalgout:
Probably not in this PR, but I want to move all of the tests related to the TemplateRender mixin to a unit test for this mixin specifically.. but they're in all sorts of states and places right now..  might wait until other things settle.

Issue #3461, opened 2017-09-21
v4 Changes Documentation


scott-w:
Following #3454 I've created this ticket to track all the major changes to the API from v3 to v4. Add items here and tick off items as the documentation gets completed:

* [ ] `Behavior#defaults` removed in favour of `Behavior#options`
* [ ] Removal of `behaviorLookup`
* [ ] Removal of `triggerMethodOn` https://github.com/marionettejs/backbone.marionette/issues/3454
* [ ] childViewEventPrefix is now `false` by default
* [ ] TemplateCache was removed by default (remove template: '#some-selector')
* [ ] View related docs should be addressed regarding the render-template mixin being available to both (all) views

paulfalgout:
whoever picks this up should double check the milestone https://github.com/marionettejs/backbone.marionette/milestone/33

paulfalgout:
We also need to go through the inline comments.  I've seen a few referring to the Marionette.Renderer and there's probably others.

paulfalgout:
we should address this https://github.com/marionettejs/backbone.marionette/issues/3562

paulfalgout:
@marionettejs/marionette-core thoughts on code samples for v4 docs.  Are we keeping `var Mn = require('backbone.marionette');` or should we start moving to ES6 `const { MnObject } from 'backbone.marionette';`?

JSteunou:
I would prefer `import from` statement

bazineta:
I would also prefer import, if only for the public service of raising awareness that there's now better ways to do things among those who would not otherwise know.

PR #3485, opened 2017-10-06
Make CollectionView a CompositeView

paulfalgout:
Alternative to https://github.com/marionettejs/backbone.marionette/pull/3433

This PR attempts to make the CollectionView support most of the CompositeView API by default.

This also renders the collectionview if `addChildView` is called on a non-rendered collection.  I _think_ this should be implemented either way as if you add a child view and then show the collection in a region, the region will render any view that isn't flagged as rendered and will remove the added child.

Pros:
No additional class
No inheritance
A lot of functionality for little cost

Cons:
Adding to an already large API
Does add cost even if little

The cost for a regular collectionview:
- 3 additional merged options which would affect initialization
- `this.getTemplate()` and a check against the result
- `_.result(this, 'childViewContainer');` and a check against the result
- a check `!this.$container.length`

the last 3 affecting only the initial render

TODO:
- [x] additional code commenting
- [x] documentation
- [x] tests

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -87,6 +87,7 @@ Marionette.setDomApi = function(mixin) {
   View.setDomApi(mixin);
 };
 Marionette.setRenderer = function(renderer) {
+  CollectionView.setRenderer(renderer);
   View.setRenderer(renderer);
 };
 
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -10,10 +10,12 @@ import MarionetteError from './error';
 import Region from './region';
 import ViewMixin from './mixins/view';
 import { setDomApi } from './config/dom';
+import { setRenderer } from './config/renderer';
 
 const ClassOptions = [
   'behaviors',
   'childView',
+  'childViewContainer',
   'childViewEventPrefix',
   'childViewEvents',
   'childViewOptions',
@@ -24,6 +26,8 @@ const ClassOptions = [
   'events',
   'modelEvents',
   'sortWithCollection',
+  'template',
+  'templateContext',
   'triggers',
   'ui',
   'viewComparator',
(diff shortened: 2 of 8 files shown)


blikblum:
Great work. I'm all for it.

>    3 additional merged options which would affect initialization

This should be negligible and the initialization cost we see with large collections come from the child views not the collection view itself. Anyway does make sense passing `templateContext` as an option? It is executed with the context (this) of the class so sharing between different classes is at best asking for trouble 
 
>    this.getTemplate() and a check against the result
>    _.result(this, 'childViewContainer'); and a check against the result
>    a check !this.$container.length

Since is in the initial render no big issue at all. Again the costs we see come from rendering children

paulfalgout:
Yeah regarding `templateContext` that's just a matter of consistency as that's what `View` does as well.  I do think there are a number of options that probably don't need to be available, but I also know there are people that use some of the options I would not have expected.. so I'd want to remove them carefully.

paulfalgout:
Oh I should also mention one feature of `CompositeView` not present here (maybe one of the only ones) is that by default the childview is not the `CollectionView`.. that seems like a bad idea for most use cases that will lead to difficult errors, but it is easy to implement it for the particular situations in which it is needed.

JSteunou:
Very nice, I'm all in :+1:

scott-w:
:+1: on the concept, it fits with my original thought process of how Mn CollectionView worked way back in v1.

paulfalgout:
Ok I think this is ready.  I do think there is much more documentation to do, but I think it is very intertwined with other documentation work that probably shouldn't be done here.  Mainly because we now have many common functionality that should be documented outside of collectionview directly.  That can be handled https://github.com/marionettejs/backbone.marionette/issues/3461

paulfalgout:
I think it makes sense to wait for https://github.com/marionettejs/backbone.marionette/pull/3490/files#r143854326 to get in to re-utilize the `_renderTemplate(template)` syntax

paulfalgout:
ok @marionettejs/marionette-core

Issue #3476, opened 2017-10-03
Set childViewEventPrefix feature flag to false


paulfalgout:
`childViewEventPrefix: false` should be the default case.  It is much more performant.  People can always switch it back to `true`

PR #3487, opened 2017-10-09
set childViewEventPrefix flag to false

wesmangum:
### Proposed changes
 - Set `childViewEventPrefix: false` to be the default case for performance improvements

Link to the issue: https://github.com/marionettejs/backbone.marionette/issues/3476

### DIFF
--- a/test/unit/mixins/view.spec.js
+++ b/test/unit/mixins/view.spec.js
@@ -425,7 +425,7 @@ describe('view mixin', function() {
         Marionette.setEnabled('childViewEventPrefix', false);
       });
 
-      it('should set childViewEventPrefix to false', function() {
+      it('should set childViewEventPrefix to "childview"', function() {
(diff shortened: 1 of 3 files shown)


blikblum:
Updated title to make it correct

test/unit/mixins/view.spec.js
```diff
@@ -410,5 +412,22 @@ describe('view mixin', function() {
...
+      beforeEach(function() {
+        Marionette.setEnabled('childViewEventPrefix', true);
+        myView = new Marionette.View();
+      });
+
+      afterEach(function() {
+        Marionette.setEnabled('childViewEventPrefix', false);
+      });
+
+      it('should set childViewEventPrefix to false', function() {
```
paulfalgout: test is correct.  The `it` statement is not

Issue #3484, opened 2017-10-05
Reimplement template:false


paulfalgout:
`template:false` was deprecated mostly because of its misuse.  It was initially added so that view's that were pre-rendered could have `onRender` events for attaching children without wiping the pre-existing html.  However in v3 views supported being rendered upon instantiation and thus users should not render a view if they do not want it rendered.  There is a use-case however when the only thing you want from the view is the `el` and there is nothing to render into the view itself.  This is by far the best use for `template: false`.  However I personally think that keeping it around is problematic as it support a bad pattern for setting up children in views.

I would support `template: false` if it prevented `render` events from occurring.  If there is no template, no rendering is occurring.  `template: false` should effectively no-op the render method.  If there is no template then there's no need to re-bind ui either, nor is there any use for render events, so this would in fact support a more performant template-less view better than the current implementation _and_ it would prevent the bad pattern.

@marionettejs/marionette-core thoughts?

Related: https://github.com/marionettejs/backbone.marionette/pull/3470#discussion_r141997852

bazineta:
Only the render event? Looking through my own stuff, there are many cases that use `template: _.noop` but still depend on the render event; presumably these could be refactored to the attach event.

In general, our use of `template: _.noop` is either as a wrapper for a jQuery plugin, where we're depending on things like `onBeforeDetach` for leak elimination, or it's in places where there really is no template, e.g., where `className` is used, and there's actually no template required.

paulfalgout:
`attach`/`detach` would be the correct events in those circumstances as they deal with when the `el` is in the DOM.  `render` is specifically about when content is put in the `el` which in the cases you describe there isn't.   Granted `_.noop` would still work as you've described.

JSteunou:
I am in favor of template : false or either undefined. _.noop is such a pain to write and to explain why... My use cases are mostly for tiny views with just tagName & className so having to write that _.noop thing each time...  Frustrating

bazineta:
I agree; gladly would I trade having to visit all my code again if it meant returning to `false` instead of `_.noop`.

bazineta:
Figuring this was worthy of real data, in my codebase there were 31 uses of `template: _.noop` in 27 files. Of these, 12 files were using either the `render` event or `onRender` function; the rest were all using `attach`.

In practically every case, the only interaction was with the `el`, and thus everything being done in the render could be pushed to the constructor (we're a CS shop).

[...]

PR #3490, opened 2017-10-10
reimplement template: false to prevent render events

wesmangum:
### Proposed changes
 - Reimplement `template: false` value.
 - Check in view's `render` for falsy template value and will prevent view bindings and any view event triggers.

Link to the issue: https://github.com/marionettejs/backbone.marionette/issues/3484

### DIFF
--- a/src/mixins/template-render.js
+++ b/src/mixins/template-render.js
@@ -1,5 +1,4 @@
 import _ from 'underscore';
-import deprecate from '../utils/deprecate';
 
 // MixinOptions
 // - template
@@ -11,12 +10,6 @@ export default {
   // and template context
   _renderTemplate() {
     const template = this.getTemplate();
-
-    // Allow template-less views
-    if (template === false) {
-      deprecate('template:false is deprecated.  Use _.noop.');
-      return;
-    }
 
     // Add in entity data and template context
     const data = this.mixinTemplateContext(this.serializeData());
--- a/src/mixins/template-render.js
+++ b/src/mixins/template-render.js
@@ -8,9 +8,7 @@ export default {
 
   // Internal method to render the template with the serialized data
   // and template context
-  _renderTemplate() {
-    const template = this.getTemplate();
-
+  _renderTemplate(template) {
     // Add in entity data and template context
     const data = this.mixinTemplateContext(this.serializeData());
 
--- a/src/view.js
+++ b/src/view.js
@@ -71,7 +71,9 @@ const View = Backbone.View.extend({
   // Subsequent renders after the first will re-render all nested
   // views.
   render() {
-    if (this._isDestroyed) { return this; }
+    const template = this.getTemplate();
+
+    if (this._isDestroyed || template === false) { return this; }
 
     this.triggerMethod('before:render', this);
 
--- a/src/view.js
+++ b/src/view.js
@@ -73,7 +73,7 @@ const View = Backbone.View.extend({
   render() {
     const template = this.getTemplate();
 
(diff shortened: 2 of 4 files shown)


src/view.js
```diff
@@ -71,7 +71,9 @@ const View = Backbone.View.extend({
   // Subsequent renders after the first will re-render all nested
   // views.
   render() {
-    if (this._isDestroyed) { return this; }
+    const template = this.getTemplate();
```
blikblum: `getTemplate` is being called twice

src/mixins/template-render.js
```diff
@@ -12,12 +11,6 @@ export default {
   _renderTemplate() {
```
blikblum: I suggest passing template var as a `_renderTemplate` argument

src/mixins/template-render.js
```diff
@@ -12,12 +11,6 @@ export default {
...
     const template = this.getTemplate();
```
blikblum: and remove this call

test/unit/view.spec.js
```diff
@@ -151,8 +131,8 @@ describe('item view', function() {
       expect(this.attachElContentSpy).to.not.have.been.called;
     });
 
-    it('should claim isRendered', function() {
-      expect(this.view.isRendered()).to.be.true;
+    it('should not claim isRendered', function() {
```
blikblum: I think we should test also a template: false with a prerendered content. AFAIK isRendered() should be true when the view is created on an el with content

src/view.js
```diff
@@ -71,7 +71,9 @@ const View = Backbone.View.extend({
...
+
+    if (this._isDestroyed || template === false) { return this; }
```
paulfalgout: as insignificant as this is, swapping these checks will be slightly more performant for the hotter route.

test/unit/view.spec.js
```diff
@@ -94,48 +94,28 @@ describe('item view', function() {
...
 
-    it('should call bindUIElements', function() {
-      expect(this.bindUIElementsSpy).to.have.been.calledOnce;
+    it('should not call bindUIElements', function() {
+      expect(this.bindUIElementsSpy).to.not.have.been.called;
     });
 
-    it('should trigger a before:render event', function() {
-      expect(this.triggerSpy).to.have.been.calledWith('before:render', this.view);
+    it('should not trigger a before:render event', function() {
```
paulfalgout: this test can just be removed.  if the other 2 tests pass this is impossible

test/unit/view.spec.js
```diff
@@ -94,48 +94,28 @@ describe('item view', function() {
...
+      expect(this.triggerSpy).to.not.have.been.calledWith('before:render', this.view);
     });
 
-    it('should trigger a rendered event', function() {
-      expect(this.triggerSpy).to.have.been.calledWith('render', this.view);
+    it('should not trigger a rendered event', function() {
```
paulfalgout: this test is also not needed with the previous onRender test I think..

test/unit/view.spec.js
paulfalgout: Agreed.  Here's an example of that test setup: https://github.com/marionettejs/backbone.marionette/blob/master/test/unit/view.spec.js#L14
You can probably do this without too much extra setup by passing an `el` at instantiation like:
`this.elView = new this.View({ el: '#foo' });`
you just have to make sure `#foo` is in the fixture with content

paulfalgout:
so this would replace https://github.com/marionettejs/backbone.marionette/pull/3470

Issue #2956, opened 2016-03-25
Implementing RadioMixin in View


scott-w:
I've been going through the code and realised `Marionette.View` doesn't use the `RadioMixin`. Is there any particular reason for this? Perhaps something to do with Views being more commonly created/destroyed?

paulfalgout:
Yeah.. I'm not sure.. only one object can handle any particular request on a channel at one time.  _maybe_ it would be useful on views too.. but it could promote some bad patterns as well. 

For events it isn't hard to implement this on your own too:

paulfalgout:
So after sitting on this a bit I do think this is best left as a a recipe rather than a feature...  radio requests can only have 1 replier and there are many instances where you have multiple copies of the same view.  I think this just opens us up to dealing with issues around people not understanding radio patterns.

scott-w:
I think that's fair enough, it'll only end up confusing developers needlessly. We can put a ticket in for advanced documentation to show users how to do this.

gdavis2:
Thanks for the followup @scott-w, and consideration @paulfalgout.
Upon further thought, a agree that it's best left as a recipe. I was just playing around with the 3.0 features and mistakenly assumed Mn.View extended Mn.Object. Interestingly, I have a case where I would like my view to respond to more than one radio channel so the standard pattern would have to be worked around if View acted like Object anyway.

rafde:
How about allowing for this scenario
```
const MyView = Marionette.View.extend({
  radioEvents: {
    myChannel:{
      myChannelEvent: 'onMyChannelEvent',
      myChannelOtherEvent: 'onMyChannelOtherEvent',
    },
    mySpanishChannel:{
      mySpanishChannelEvent: 'onMySpanishChannelEvent',
      mySpanishChannelOtherEvent: 'onMySpanishChannelOtherEvent'
    },
  },
  onMyChannelEvent() { console.log('hey'); },
  onMyChannelOtherEvent() { console.log('hello'); },
  onMySpanishChannelEvent() { console.log('ey'); },
  onMySpanishChannelOtherEvent() { console.log('hola'); }
});
```

bazineta:
bindEvents, rather than bindEntityEvents?

rafde:
reply/request are one-to-one. So if you bind a request to a view, and you have another view of the same class, then the request is bound to the new view.

I don't think binding request is viable.

Having View's listen to channels is do-able since they are limited to one-to-one like reply/request

[...]

PR #3492, opened 2017-10-11
implement bindRequests on views

wesmangum:
### Proposed changes
 - implementation of `bindRequests` on views for potential solution to "sinking" events down to child views. 
- Provides consistency for the 2-arg vs. 3-arg nature of `bindRequests` and `bindEvents`.

Link to the issue: https://github.com/marionettejs/backbone.marionette/issues/2956

### DIFF
--- a/src/mixins/common.js
+++ b/src/mixins/common.js
@@ -6,6 +6,10 @@ import {
   bindEvents,
   unbindEvents
 } from '../common/bind-events';
+import {
+  bindRequests,
+  unbindRequests
+} from '../common/bind-requests';
 
 export default {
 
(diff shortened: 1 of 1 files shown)


paulfalgout:
I think this is good.  Looking into it a bit further I think that all four bind imports can be removed from https://github.com/marionettejs/backbone.marionette/blob/master/src/mixins/radio.js since they're now on the common mixin and we don't export the radio mixin for other use.  Doesn't change much really, but we don't really benefit from having it duplicated.

bazineta:
Entirely possible that I'm missing something, but it's not evident to me that there's not a leak here. 

That is, it seems that one would call this.bindRequests(), expecting that there'd be automatic unbinding in the `destroy` as in this.bindEvents(), but I don't see where that would be the case.

bazineta:
A fiddle against the current Mn to demonstrate my confusion: https://jsfiddle.net/orj1sckr/5/

JSteunou:
I think you're right, if the other bindings get unbind automatically, this one should also be unbind, or have a different name.

paulfalgout:
It is impossible to prevent events from being unbound on destroy.

Issue #3169, opened 2016-09-07
Remove AppRouter and make it an external lib


paulfalgout:
`AppRouter` is entirely optional. Many users don't use it, and there are many great alternatives. Personally I do not believe Marionette needs a single sanctioned router, but whether or not another router is added, `AppRouter` can be removed.

There may be some concern that Marionette is incomplete as a library without a router, but I think backbone's router is just as good of a starting place as `AppRouter`.  Once someone needs to solve additional problems in routing, they can explore the many alternative options, and hopefully we can provide food recipes for some.

JSteunou:
I have split feelings about this. On the one hand I totally agree with you, on the other hand appRouter brings bindings with Controller and `onRoute` event that are useful and a good no brain starter for beginners. There is a lot to learn when starting with a fw/lib (even if Mn is far away from the hardest to learn) and adding more stuff to learn / setup could be a drag. But you also can argue that learning about appRouter & Controller is more stuff to learn... Having beginners PoV could help, mine might be thwart from being use to use appRouter.

paulfalgout:
My biggest thought behind this is that many people may not use this router.  And I suspect if we made it an external lib that less people would use it.  So while this router might be better than Backbone's, it may not be the best router for someone's app, but by including it in the lib it seems like the opinion of Marionette is that you should use this router.  By making it external we can still support it (to some degree router issues become easier to support) and we can offer other options as well in Mn documentation.

denar90:
I'm 👍  for deprecating it in v4. Moving it in separate repo might give it second wind. First version probably will be the same as it is right now but later we can make a lot of improvements. @jmeas made good list https://github.com/jmeas/router-comparison about what was missed in Backbone router, so our one can fill empty gaps.

paulfalgout:
As part of this work wee need to consider:

https://github.com/marionettejs/backbone.marionette/issues/2280
https://github.com/marionettejs/backbone.marionette/issues/2092

PR #3493, opened 2017-10-12
Remove AppRouter

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3169

The abstracted lib for AppRouter is currently a private repo on the Mn org.  It will need some minor modification to be released with v4.

Again the main reasons for removing it:
- It reduces cognitive load for first adopting Mn and understanding its purpose.
- It removes any undue bias towards AppRouter as being the preferred router
- It makes the library smaller without needing tree-shaking. This isn't a _huge_ deal, but it certainly doesn't hurt

We might be concerned that this makes yet another lib to maintain.  I'd argue that no changes have been made to AppRouter in a very long time with only a single issue opened that has never been addressed.


JSteunou:
In order to ease the v4 adoption, I'm in favor as long as app-router still exist as a separate lib

scott-w:
Should we not do a slower migration process here first? Make new instances of using `Marionette.AppRouter` warn the developer with notes on the migration requirements?

paulfalgout:
https://github.com/marionettejs/marionette.approuter/
Yep the AppRouter should be available here and we can make a blog post soon about some of the more major changes coming up.
This change will require either:
```diff
<script src="backbone.marionette.js"></script>
+ <script src="marionette.approuter.js"></script>

<script>console.log(Marionette.AppRouter)</script>
```
or
```diff
- import { AppRouter } from 'backbone.marionette';
+ import AppRouter from 'marionette.approuter';

console.log(AppRouter);
```

Issue #3500, opened 2017-10-16
Do model/collection need to be cloned (when serialized) to render?


nlhuykhang:
### Description

I was having performance issue with my project, so I spent time looking into Marionette code hoping to find out some ways to optimize performance. I found out that by default a view will [shallow clone models during serialization before rendering](https://github.com/marionettejs/backbone.marionette/blob/v3.3.1/src/view.js#L56-L80). My concern is that if these models are only used for rendering, why do we need to clone them? given that clone is an expensive operation

### Expected behavior

For future version, Should we just return `model.attributes` without cloning? If application developers need to modify models, they can clone themselves.

### Environment

1. Marionette version: 3.3.1
2. Backbone version: 1.3.3

paulfalgout:
Good catch. Interestingly enough they will not be cloned for v4. https://github.com/marionettejs/backbone.marionette/blob/next/src/mixins/template-render.js#L60

I think this was somewhat legacy regarding how `templateHelpers` used to be attached directly to the return of the `serializeData`.  Now we're extending onto an empty object, which doesn't remove the cost entirely, but it should be much cheaper..  however I'm looking into simply returning the serialized data if `templateContext` isn't defined which would be an additional improvement.

This was a breaking change if even minor so it had to wait for v4, but I would imagine in a majority of cases, back-porting whats on the `next` branch will not be breaking.

PR #3501, opened 2017-10-16
Return data without extend if either data isn't defined.

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3500

The `_.extend` is expensive if it is not needed where-as the simple `if`s are cheap.

For those being extremely perf sensitive this will be the ideal scenario with minor change to the typical case where both are defined.

Additionally this is still far cheaper than what is happening in v3.

I also renamed the `target` variable to make this function more explicit in its use.

### DIFF
--- a/src/mixins/template-render.js
+++ b/src/mixins/template-render.js
@@ -10,7 +10,7 @@ export default {
   // and template context
   _renderTemplate(template) {
     // Add in entity data and template context
-    const data = this.mixinTemplateContext(this.serializeData());
+    const data = this.mixinTemplateContext(this.serializeData()) || {};
 
     // Render and add to el
     const html = this._renderHtml(template, data);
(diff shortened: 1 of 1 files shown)


src/mixins/template-render.js
```diff
@@ -32,9 +32,11 @@ export default {
   // object literal, or a function that returns an object
   // literal. All methods and attributes from this object
   // are copies to the object passed in.
-  mixinTemplateContext(target) {
+  mixinTemplateContext(serializedData) {
     const templateContext = _.result(this, 'templateContext');
-    return _.extend({}, target, templateContext);
+    if (!serializedData) { return templateContext; };
+    if (!templateContext) { return serializedData; };
```
blikblum: I would swar the check order since the most common situation is `!templateContext`

paulfalgout:
yep good idea.  switched

PR #3502, opened 2017-10-16
Refactor view _getImmediateChildren

paulfalgout:
A reduce is much faster

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -25,6 +25,15 @@ const ClassOptions = [
   'ui'
 ];
 
+// Used by _getImmediateChildren
+function childReducer(children, region) {
+  if (region.currentView) {
+    children.push(region.currentView);
+  }
+
+  return children;
+}
+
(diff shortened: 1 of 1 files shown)


src/view.js
```diff
@@ -97,10 +97,15 @@ const View = Backbone.View.extend({
...
 
   _getImmediateChildren() {
-    return _.chain(this._getRegions())
-      .map('currentView')
-      .compact()
-      .value();
+    return _.reduce(this._regions, this._childReducer, []);
+  },
+
+  _childReducer(children, region) {
```
blikblum: I would make `_childReducer` an standalone function instead of a method. There are two advantages: minimizes better and does not clutter view class namespace. Otherwise all good

paulfalgout: I moved it.  I don't love the distance it puts between the methods, but agreed that it's better in production.  There's probably some other instances where anonymous functions or private methods that don't use `this` could be extracted.

PR #3504, opened 2017-10-16
Refactor build region

paulfalgout:
Removes an extend and an omit.
Keeps original data unmutated.


paulfalgout:
@blikblum jsperf was down earlier, but for the number of defaults we have, to my surprise, it looks like `_.extend({},` will be faster.  Makes the code a bit smaller as well.

blikblum:
Still good

Issue #3506, opened 2017-10-20
Destroyed behaviors keep view events and triggers


paulfalgout:
Right now if you destroy a behavior it will remove itself from the view's proxy so view events won't trigger on the behavior, but handlers setup on the behaviors `events` and `triggers` will still handle

Since the events and triggers are namespaces with a unique namespace it would be hard to extract these.  

I could see fixing this a couple of ways.  One we can give the namespaces a predictable name based perhaps on the behavior's `cid`.  This doesn't seem hard and easy to implement removal _however_ we'd be calling `undelegate` I assume with simply a namespace to remove the events and I don't know if the native variety of Mn could handle that.  @blikblum 

So the other option?  if a behavior is destroyed on the live view we could run re-run `delegateEvents` (which will undelegate prior to re-delegating by default)  This is probably much less performant, but it's simple.

Also this feels bugish and we do have another v3.4.x to release it looks like.. 

Thoughts?

blikblum:
> Right now if you destroy a behavior it will remove itself from the view's proxy so view events won't trigger on the behavior, but handlers setup on the behaviors events and triggers will still handle

Any example of this bug?

> I could see fixing this a couple of ways. One we can give the namespaces a predictable name based perhaps on the behavior's cid.

I think is doable and something i was looking into more as a performance than a bug fix.

BTW: The only purpose of namespace is avoiding key collision when building the events hash. It could be removed at all.

> I don't know if the native variety of Mn could handle that

I have to double check. But last time i looked should be handled by Mn.Native out of box

paulfalgout:
> The only purpose of namespace is avoiding key collision when building the events hash.

I think collision avoidance is necessary for behaviors, but this would introduce a new use case using namespacing for removing only the destroyed behavior events/triggers

https://jsfiddle.net/3p35cbL8/1/

Issue #3496, opened 2017-10-15
Behavior getEvents reduce is breaking if missing handler


paulfalgout:
https://github.com/marionettejs/backbone.marionette/blob/next/src/behavior.js#L113

This line should return the memo `events` otherwise it short circuits previous results.

This work needs a test.

PR #3508, opened 2017-10-24
Fix behavior events / triggers

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3496
Resolves https://github.com/marionettejs/backbone.marionette/issues/3506

This changes from uniquely delegated events to specified namespaced event such that
- view events: regular backbone event namespace (this is true for all of the options)
- view triggers: `.trig{ view.cid }`
- behavior events: `.{ behavior.cid }`
- behavior triggers: `.trig{ behavior.cid }`

At first I looked at namespacing such that a single `undelegate` could remove both events and triggers, but it unnecessarily complicated the shared trigger logic and `behavior.destroy` The two `undelegate`s is arguably slower than if we had an additional unique namespace, but I'm not sure if it is worth the code changes.  Behavior destroying is probably not something done often (clearly since this hasn't been an issue)

- [ ] Needs tests


src/behavior.js
```diff
@@ -66,6 +66,9 @@ const Behavior = MarionetteObject.extend({
   destroy() {
     this.stopListening();
 
+    // Remove behavior-only triggers and events
```
blikblum: A `!this.view._isDestroyed` check here will prevent the call in the most common case (behavior being destroyed together with the view)

paulfalgout: good catch.  I also noticed this check is already happening inside `_removeBehavior` so I'm going to move the `undelegate`s inside that method as well.  I think this makes sense anyhow as if you're removing the behavior for any reason you'll want these `undelegate`s to occur (if for some reason we reuse `_removeBehavior` in the future)

Issue #3503, opened 2017-10-16
Deprecate multiple handlers


paulfalgout:
Marionette allows users to set handlers as string names of methods which I think is useful as in:
```js
  modelEvents: {
    'change': 'render'
  }
```
But it also allows multiple handlers such as:
```js
  modelEvents: {
    'change': 'render onChange onModelChange'
  }
```
This to me seems like an unnecessary shortcut that keeps us from further simplifying event creation.

JSteunou:
Always found this a bad practice. I prefer having one callback like `onChangeFoo` calling multiple methods if needed.

PR #3511, opened 2017-10-30
Deprecate multiple event handlers

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3503


paulfalgout:
Maybe?  It's a bit sneaky maybe to do it so soon.. but I figured with a v3.5 release this is as early of a notification as possible regardless of if we remove it in v4.

blikblum:
I think deprecating in 3.5 is enough to remove in v4.

JSteunou:
:+1:

paulfalgout:
@JSteunou are you approving this PR or just @blikblum 's comment?

JSteunou:
Sorry, the comment

Issue #3412, opened 2017-08-04
viewFilter function for lodash users


scott-w:
### Description

For `NextCollectionView`, the behavior of `viewFilter` differs depending on whether you have lodash or underscore because of the difference in the `partition` API.

The lodash `partition` function is incompatible with Underscore's by not passing `index`/`key` or `collection` in as its 2nd third arguments. This means `viewFilter(view, index, children)` when using Underscore is `viewFilter(view)` when using lodash.

To fix this, we should put a note in the docs for lodash users where the workaround is to do `this.children.findIndexByChild(view)` and `this.children` is still accessible anyway.

### Environment

1. Marionette version: v3.3.1
2. Backbone version: v1.3.3
3. Additional build tools, etc: lodash

paulfalgout:
Note that the use-case for index in filtering is probably limited as well.  Most of the time someone will probably filter off of data on the model, or maybe even the view.

scott-w:
I don't think it's a common thing, it's just something I encountered where the docs didn't reflect what was happening =)

bazineta:
Glancing through the documentation for both libraries, it seems to me that if one wanted a common works-anywhere solution for this, it'd have to be `_.reduce`.

paulfalgout:
hmm I think `_.reduce` would be a fairly unfortunate change.  The benefit of `partition` in both libraries is that there is a single loop that requires the user to return a truthy/falsey result as you would expect with a filter handler that then returns both the views to filter in and out.  With reduce we'd be requiring the user to build some other data structure and it would be much less like a filter.

bazineta:
Just basically musing on the intersections between the two libraries; in short, no matter what happens in lodash, all paths lead to `_.reduce` eventually. In this case, the memo object passed in would be an object or array tuple, one for the keepers, one for the others, such that it'd be one loop only.

It'd require some effort to deal with the predicates, but it seems as if it'd be doable such that the user would still see exactly the same predicate API that `viewFilter` has now. If interested, I could take a crack at it, which would avoid having to dual-path the documentation if it works. If you're ok with dual-path documentation, that's ok with me as well.

paulfalgout:
if the `viewFilter` api does not change for the user and there is no performance hit on large lists, I'm ok with a change

bazineta:
@rafde Given what you're keeping an eye on in Lodash 5, do you thing this is still a good path to go down?

bazineta:
[...]

PR #3512, opened 2017-10-31
Replace viewFilter _.partition

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3412

Removing `_.partition` actually made a decent perf improvement since we are handling the predicate and since we could also remove the bind.  This also makes a small feature improvement of adding the partition results to the filter event.

Inspired by removing the bind, I also made a small change to the view comparator where comparator binding isn't done until calling `_sort` this is all within internal code and should be entirely non-breaking.  All it _really_ does is remove the `_.isFunction` check.  It also removes the `.bind()` in v3..  certainly not a big deal.. we're probably shortly to move in the opposite direction, but now everything in v3 is consistent.

If need-be I can move the comparator change to a separate PR.

### DIFF
--- a/src/next-child-view-container.js
+++ b/src/next-child-view-container.js
@@ -48,17 +48,17 @@ _.extend(Container.prototype, {
   },
 
   // Sort (mutate) and return the array of the child views.
-  _sort(comparator) {
+  _sort(comparator, context) {
     if (typeof comparator === 'string') {
       comparator = _.partial(stringComparator, comparator);
       return this._sortBy(comparator);
     }
 
     if (comparator.length === 1) {
-      return this._sortBy(comparator);
+      return this._sortBy(_.bind(comparator, context));
     }
 
-    return this._views.sort(comparator);
+    return this._views.sort(_.bind(comparator, context));
   },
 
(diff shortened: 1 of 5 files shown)


blikblum:
I'm fine with the changes. Any ideas why travis is failing?

paulfalgout:
Travis is having issues with our server type. I will rerun it later

JSteunou:
Better to get merged before the removal of _.bind begins :D

bazineta:
Your concern for the hapless users of IE8 is inspiring, @JSteunou.

paulfalgout:
```

bazineta:
Well...in that my company is in the business of selling retail products, for which we expect to be compensated, typically in US dollars, that's a solved problem for us; if they can't update past IE8, then they can't afford our stuff, either...

And let me say, mmmm, that's a nice commit there; much better than my minimally-invasive approach was.

JSteunou:
Hu... I was just thinking about #3513 and PR #3514 which will be pending until this is not merged, or will have to be opened again.

Issue #3513, opened 2017-10-31
Replace `_.bind` with native bind


paulfalgout:
Related to but certainly not covering anywhere close to all of https://github.com/marionettejs/backbone.marionette/issues/3391 we could remove `_.bind` for v4.

PR #3514, opened 2017-10-31
Replace `_.bind` with Function.prototype.bind

PaulCapron:
This closes issue #3513.


JSteunou:
We should wait for #3512 to land into next because it brings other `_.bind`

paulfalgout:
Honestly there's going to be work merging v3.5 into v4 regardless such that I think we can safely merge this.  We'll just need to be careful bring in the v3.5 changes to v4, which would be true anyways.

Issue #3507, opened 2017-10-20
unbindEvents and unbindRequests shouldn't require the bindings hash


paulfalgout:
Currently in order to effectively unbind events or requests you must reproduce an identical bindings hash.  This can lead to leaks such as:
```js
myRequests() {
  return {
    'foo'() {
        console.log('I am an anon handler that will never be cleaned up');
    }
  };
},
initialize(options) {
  this.channel = options.channel;
  this.bindRequest(this.channel, _.result(this, 'myRequests'));
},
onBeforeDestroy() {
  this.unbindRequest(this.channel, _.result(this, 'myRequests'));
}
```
outside of calling `options.channel.stopReplying(null, null, this);`
there is no way to clean up the bound requests.
the unbind functions should allow for `this.unbindRequest(this.channel);`
This would allow for a fail safe way to unbind everything on the view.

Also I think having the bind functions call the unbind is a false sense of security and could probably be avoided in v4.

This doesn't allow for automatic clean up of bound requests when an element is destroyed.  We should add special documentation here as events _are_ automatically cleaned up by nature of a regular destroy without needing to unbind.

blikblum:
> the unbind functions should allow for this.unbindRequest(this.channel);

The way i see this feature  implemented is storing some info in the instance (`this`) at bind call and retrieving later at unbind call. 
It introduces complexity in a feature that at minimum, its usage is questionable (requests at views).

The way i see implemented is something like
```
this.boundRequestInfo = this.bindRequest(this.channel, _.result(this, 'myRequests'));
```
later

```
this.unbindRequest(this.channel, this.boundRequestInfo);
or
this.boundRequestInfo() //a destroy function
```

>  as requests are automatically cleaned up by nature of a regular destroy without needing to unbind.

Do you mean events?

paulfalgout:
> Do you mean events?

yep edited

returning a specific unbind is interesting.  I like it better as a destroy function as I think there is value in keeping the `unbindRequest` arguments the same.  I also think that `this.unbindRequests(this.channel);` should also remove everything from the instance,  but that's much more dramatic than `this.boundRequestInfo();` which is precise.

It is notable to mention that this issue will be true for _every_ class in Mn, though probably the most problematic use case will be views.

blikblum:
>  I also think that this.unbindRequests(this.channel); should also remove everything from the instance, but that's much more dramatic than this.boundRequestInfo(); which is precise.

[...]

PR #3515, opened 2017-11-01
allow unbindEvents and unbindRequests to work without bindings hash

wesmangum:
### Proposed changes
 - These changes allow for `unbindEvents` and `unbindRequests` to be called without the bindings hash. Doing so will unbind all events from the entity and all requests, respectively.

- [ ] Todo docs about needing to unbind when destroying

Link to the issue: https://github.com/marionettejs/backbone.marionette/issues/3507

Note that this fix only resolves one part of #3507

### DIFF
--- a/src/common/bind-events.js
+++ b/src/common/bind-events.js
@@ -38,8 +38,6 @@ function bindFromStrings(target, entity, evt, methods, actionName) {
 
 // generic looping function
 function iterateEvents(target, entity, bindings, actionName) {
-  if (!entity || !bindings) { return; }
-
   // type-check bindings
   if (!_.isObject(bindings)) {
     throw new MarionetteError({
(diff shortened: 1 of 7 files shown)


paulfalgout:
I think this LGTM for v3.5  @marionettejs/marionette-core ?

src/common/bind-events.js
```diff
@@ -57,11 +55,20 @@ function iterateEvents(target, entity, bindings, actionName) {
 }
 
 function bindEvents(entity, bindings) {
+  if (!entity || !bindings) { return this; }
+
```
blikblum: It's just style (although slightly smaller) but how about:
```
if (entity && bindings) {
  iterateEvents(this, entity, bindings, 'listenTo');
}
return this;
```

paulfalgout: meh.. I typically promote guards vs conditional logic if possible and this is more pattern-like with the `bindRequests` function.  FWIW I think much of this will be short-lived as it can be heavily refactored for v4 since we're removing the multi-handlers.. events and requests can share more of the logic

Issue #3489, opened 2017-10-10
CollectionView should not show only added views if viewComparator is false


paulfalgout:
The added view perf should also respect the viewComparator false flag
https://github.com/marionettejs/backbone.marionette/blob/next/src/collection-view.js#L443

We'll now have this check in 3 places:
`!this.sortWithCollection || this.viewComparator === false`

Think it warrants a method like: `_shouldSort` or something?

paulfalgout:
so negative.. not the same check as the others it turns out

PR #3516, opened 2017-11-02
Add at end perf improvement can only occur if no sorting occurred

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3489

Sorting requires reappending of all children.  Rather than handling all of the various flag states to determine this, we can simply remove the added views cache if a sort occurs.

### DIFF
--- a/src/next-collection-view.js
+++ b/src/next-collection-view.js
@@ -368,6 +368,9 @@ const CollectionView = Backbone.View.extend({
 
     if (!viewComparator) { return; }
 
+    // If children are sorted prevent added to end perf
+    delete this._addedViews;
+
     this.triggerMethod('before:sort', this);
 
     if (_.isFunction(viewComparator)) {
@@ -432,11 +435,6 @@ const CollectionView = Backbone.View.extend({
     return this;
   },
 
-  _isAddedAtEnd(addedView, index, addedViews) {
-    const viewIndex = this.children._views.length - addedViews.length + index;
-    return addedView === this.children._views[viewIndex];
-  },
-
   _filterChildren() {
     const viewFilter = this._getFilter();
     const addedViews = this._addedViews;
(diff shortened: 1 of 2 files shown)


JSteunou:
Does this should be the inverse?

If there is no comparator, then no sort, then obviously added at then, then no need to check with `every this._isAddedAtEnd`

With this change, in case of no comparator the check in `filter` is always called, whereas in case of a comparator the check is not called but added views could have been sorted.

paulfalgout:
Well it's not necessarily obvious that the views are added at the end as views can be added at any index with `addChildView`

It does make this perf improvement a bit less impressive as sorting will probably be common, but whether or not the views are added to the end is unimportant if any of the view prior to them have changed locations.  And that check would be more costly than it is worth I think.  Rather than make a super smart sorting / rendering algorithm anyone would get far better performance at this point from utilizing a dom diffing renderer.

Really a renderer coupled with a modified DOM API could be written to compare the in DOM nodes with the buffer and do whatever calculations to insert dom nodes as needed.  I just do not think it would be a good idea to do that work here.

paulfalgout:
although you're right.. this can be simpler..

JSteunou:
What bothers me, is that there is either a check on all added views to check if there are at end, or a render / detach / attach on all views.

I have no solution but it would be nice to just *know* when all added views are added at the end in order to not have to check in it during the filter.

paulfalgout:
yep.. when the perf improvement was added it was intended to account for sorting, but since we cannot rely on previously rendered children to be in their sorted state when something is added, there's no reason to be concerned about the sort order at all.. so as you suggested, the better solution is to just determine upon add if it is at the end, cache the views, and wait to see if no sorting or filtering occurs and simply append them only.

paulfalgout:
Ok @marionettejs/marionette-core I think this is ready.  v3.5 is almost ready 🎉

JSteunou:
Nice job @paulfalgout :+1:

PR #3524, opened 2017-11-05
Refactor parseBehaviors

paulfalgout:
Uses `_.reduce` which removes `concat` and `flatten`

This is technically non-breaking because it only changes private vars


JSteunou:
I would consider it breaking because it changes the return signature of some public API.

paulfalgout:
I mean it doesn't really matter if it is break since it's going into v4, but I'm not sure what public API it changes

JSteunou:
You're right my bad, `getBehaviorClass` is not exported

PR #3526, opened 2017-11-07
Added live example

CharlyJazz:
Added live example in the Region Documentation

### Proposed changes
 - Added live example in the section _Define the application region_ in the Region Documentation
 - The example shows the simple way to show different views in the main region of Marionette.Application

Link to the issue: [Issue #3209](https://github.com/marionettejs/backbone.marionette/issues/3209)

### DIFF
--- a/docs/marionette.region.md
+++ b/docs/marionette.region.md
@@ -56,6 +56,8 @@ var App = Mn.Application.extend({
 });
 ```
 
+[Live example](http://jsfiddle.net/marionettejs/9fburmb8/)
+
 For more information, see the
 [Application docs](./marionette.application.md#root-layout).
 


docs/marionette.region.md
```diff
@@ -56,6 +56,8 @@ var App = Mn.Application.extend({
 });
 ```
 
+[Live example](http://jsfiddle.net/charlyjazz/gze5ykov/6/)
```
paulfalgout: this looks good.  would you be up for swapping the router to Backbone.Router?
AppRouter will be available as a separate lib for v4, so rather than do this now and update it in a few weeks, using Backbone's router would keep this example solid for v4 as well.

CharlyJazz: I will work on that and I will notify you

CharlyJazz: Ready. I have updated the fiddle. Check it

paulfalgout:
👏 thanks!

PR #3529, opened 2017-11-08
Remove Multiple events

paulfalgout:
Removes deprecated multiple events.  Need to update docs and tests.  Many of the test are testing the implementation a bit heavy which is why there's so many breaking tests.  This should _only_ be removing the multiple events handler and cleaning up the code some.


paulfalgout:
Ok @marionettejs/marionette-core this is GTG but needs 👀 
Changes to make `unbindEvents` made it necessary to also touch the delegate entity events mixin.  I think it is an improvement.

Currently I have no plans for other code changes and would suggest against breaking region changes or new region features for v4.0 as we've touched a lot already.  I think after this PR, baring other perf updates, we should focus on improving the tests and docs and possibly rolling a alpha RC with a blog post.

paulfalgout:
I am pretty sure the coverage is the same or better (the report didn't change, but beyond that I think the use cases are better now)   the unbind tests were moved into the bind files so that the test files match the file they're testing.  the view-events test file was renamed to the mixin tests.  
So for instance this file https://github.com/marionettejs/backbone.marionette/pull/3529/files#diff-dc84d1de8b539f0c0ff7f8a80f571d31L3 is now represented here: https://github.com/marionettejs/backbone.marionette/pull/3529/files#diff-ea025ad50f1a23f3706bfd965d3181a6R86
and this file https://github.com/marionettejs/backbone.marionette/pull/3529/files#diff-d3fe180ef1839c619e6768b8c87dabc7L1 is https://github.com/marionettejs/backbone.marionette/pull/3529/files#diff-0e4b726c7a050b5b1eff915b5794a721R1

Issue #3389, opened 2017-06-20
Region: $el set in constructor might be resolved wrongly


ardi-n:
### Description

1. Having a server-generated html structure of a collection I want to initialize it using a CollectionView with any needed nested views.
2. Because child views are identical they have identical region definitions.
3. Regions are initialized *before* el/$el are resolved for a view.
4. Thus region's $el is resolved wrongly in constructor - parentEl is not yet defined so region's $el is being searched in a whole document resulting in a jQuery collection of as many elements as there are in a list.
5. I can't use region's $el to i.e. search within it for existing html that should be initialized with nested views.
6. Naturally region's $el is fixed after show() is called.

### Expected behavior

Region's $el should be usable if it is set in constructor. Regions could be initialized *after* a view calls setElement.

### Environment

1. Marionette version: 3.3.1
2. Backbone version: 1.3.3

paulfalgout:
I'm pretty sure I follow this, but a fiddle would help.

This would require regions be setup after `initialize` as well.

Personally I think that the ideal fix would be for regions to never be initialized unless `getRegion` is called, so they wouldn't be in the constructor at all.

Then I also think that we should remove the DOM lookup from the region altogether.  I think ideally the view should pass the `el` to the region rather than the whole `parentEl` business.

We might be able to make the `getRegion` a lazy evaluation without being breaking.

paulfalgout:
A pretty poor work around for the moment would be to:
```js
regions() {
   this.setElement(this.el);

  return {
    region: '.selector'
  }
}
```

ardi-n:
I'll try to deliver a fiddle tomorrow. But this workaround should indeed work.

ardi-n:
Here's a demo: https://jsfiddle.net/9732v9cx/
The suggested workaround does not work. Only if you call _ensureElement() on a region its $el is fixed.

paulfalgout:
ah the problem in the work around is that `this.el` isn't set until the backbone.view constructor call.. but if you change it to `this.getOption('el')` it should work.

JSteunou:
> Personally I think that the ideal fix would be for regions to never be initialized unless getRegion is called, so they wouldn't be in the constructor at all.

@paulfalgout that's kind of what I did in my autoregion plugin, and indeed that's way better https://github.com/JSteunou/marionette.autoregion

paulfalgout:
solved in v5

Issue #1972, opened 2014-10-06
Regions: There and Back Again


jamesplease:
###### Work in progress

This issue is to document my proposal Regions 3.0.
#### The Motivation

Regions work _awesomely_ in the most common use cases. But there are still a set of features that you would expect to work, but don't. There are three known issues with Regions:
1. When a LayoutView is created from a `tagName`, its Regions are not set up until (1) you call show on the region, or (2) you render the layoutView twice. I have no qualms with number one, but number two is just odd. Regions should be available on the first render. In the past we've hand-waved this away as a performance feature, but closer inspection revealed it was added as a means to fix an issue where re-rendering a Layout would not destroy the old views. (#1971)
2. LayoutViews do not scope the region lookup mechanism within their own element  (#1685)
3. When a LayoutView already exists, resetting its regions will mess up the selector. (#1673)
# 

From my research, none of these issues stem from a single source. Rather, it's the culmination of PRs upon PRs adding new features onto a system that was created from a shaky foundation.

This solution is about a few key changes to the system that will make the code simpler, easier to document and explain, and just as powerful. Do note that this is just an outline, though. Some of the details might need to be adjusted as we go into the implementation phase, but I think the underlying philosophy behind the system is solid.

The only possible negative to this solution is that it depends on the idea that we get rid of the RegionManager class. 
##### Why merge RegionManager into LayoutView?

LayoutViews are the only class that should use the RegionManager logic. They should be in the view, or otherwise a mixin, a la Backbone.Events, that we add to the prototype.
#### That's not true, though! Applications have a RegionManager!

That's true, but that is a bad feature. Applications should just be containers. Instead of `app.addRegion`, we should promote creating a LayoutView as the base of the view chain. This LayoutView can use an existing element on the page to begin the view tree from. This is better in _so many ways_. It reduces the complexity of explaining how things work. You no longer need to even think about how a RegionManager – or the code that it represents – even exists. Marionette will become views all the way down. It's a simple system that's so easy to explain, and much easier for new users to understand.
#### I'm still not convinced

I did my best :(
# 
### The solution

The system proposed is very simple. In a lot of ways, it doesn't change the expectations people have for how Regions work. It only changes the logic behind it, so that the behavior of the system matches up with our expectations better.

[...]

PR #3530, opened 2017-11-08
WIP: Add auto discover of region element

JSteunou:
### Proposed changes
 - add auto discover of region in view
 
Link to the issue: #3417 #3389 #3320 and maybe #1972 ?

### DIFF
--- a/src/mixins/regions.js
+++ b/src/mixins/regions.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
 import _invoke from '../utils/invoke';
 import buildRegion from '../common/build-region';
+import MarionetteError from '../error';
 import Region from '../region';
 
 // MixinOptions
@@ -126,11 +127,11 @@ export default {
   // Provides access to regions
   // Accepts the region name
   // getRegion('main')
-  getRegion(name) {
+  getRegion(name, options) {
     if (!this._isRendered) {
       this.render();
     }
-    return this._regions[name];
+    return this._regions[name] || this.autoAddRegion(name, options);
   },
 
   // Get all regions
@@ -146,7 +147,7 @@ export default {
   },
 
   showChildView(name, view, ...args) {
-    const region = this.getRegion(name);
+    const region = this.getRegion(name, ...args);
     region.show(view, ...args);
     return view;
   },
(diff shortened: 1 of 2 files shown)


JSteunou:
Some points I did not resolved around how far we want to go on this. I lot some open door for the user so he can

1. use it's own selector: I set `data-region` by default but it could be `data-mr-region`
2. use it's own logic on selection: I set a default `data-region="${name}"` but I'm currently using `data-region="${_.kebabCase(name)}"`
3. add it's own region options: should Marionette add some feature around it, like `data-region-replace` => `replaceElement: true` or be very light and unopinionated? I started with light.
4. should we raise an error when the element is not found?
5. ...

paulfalgout:
so I do like this in theory, but I'd be interest on waiting until the core of region can be readdressed, or at least until we know how to readdress it.

I think some more code examples would help understand the feature fully as well.

JSteunou:
So more v4.x or v5?

paulfalgout:
Yep I think v4.x - v5 should be focused on region like v3.x - 4 was on collectionView

taburetkin:
i don't know if it helps
but personally i have this head pain with region
1) i am always have to define template for container view
f.e.
```
<div class='actions-region'></div>
<div class='entries-region'></div>
<div class='something-region'></div>
```
2) i am always have to define regions in a container view definition
f.e.
```
regions:{
   'actions':'.actions-region',
   'entries':{el:'.entries-region', replaceElement:true},
   'something':'.something-region'
}
```
3) i am always have to provide `onRender` logic in every container view
wich is in 99% of cases is something like this
```
let View = getActionsViewClass();
let options = getActionsViewOptions();
let view = new View(options);
this.showChildView('actions',view);
```
4) and if there are few nested views in a container its looks pretty ugly

so, i ended with one mixin for a View and one helper

now my container view looks like
```
let Container = BaseContainerView.extend({
   template:_.template('')
   initialize(){
      this.createNested({
         name:'actions',
         viewClass: ActionsView,
         viewOptions: () => somehowBuildOptions(), //optional,
         regionTemplate:{ replaceElement: true }, //if i do not provide any selector information region will be created automaticaly
      })
   },
   onRender(){
       this.showAllNested();
   }
})
```
so, i do not care about container`s template any more
and i do not need to repeat nested views logic

paulfalgout:
Looks at regions on https://github.com/marionettejs/marionette

PR #3532, opened 2017-11-09
Tests/view destroy error get immediate children object

wesmangum:
### Proposed changes
- Reorganize tests to match order of codebase

Link to the issue: #3248

### DIFF
--- a/test/unit/destroying-views.spec.js
+++ b/test/unit/destroying-views.spec.js
@@ -1,45 +1,54 @@
+import View from '../../src/view';
+
+
 describe('destroying views', function() {
   'use strict';
 
   describe('when destroying a Marionette.View multiple times', function() {
+    let onDestroyStub;
+    let view;
+
     beforeEach(function() {
-      this.onDestroyStub = this.sinon.spy(function() {
+      onDestroyStub = this.sinon.spy(function() {
         return this.isRendered();
       });
 
-      this.view = new Marionette.View();
-      this.view.onDestroy = this.onDestroyStub;
+      view = new View();
+      view.onDestroy = onDestroyStub;
 
-      this.view.destroy();
-      this.view.destroy();
+      view.destroy();
+      view.destroy();
     });
 
     it('should only run the destroying code once', function() {
-      expect(this.onDestroyStub).to.have.been.calledOnce;
+      expect(onDestroyStub).to.have.been.calledOnce;
     });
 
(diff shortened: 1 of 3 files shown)


test/unit/destroying-views.spec.js
```diff
@@ -1,45 +1,53 @@
+import Marionette from '../../src/backbone.marionette';
```
paulfalgout: This file shouldn't import the entire library only the dependencies needed.

test/unit/get-immediate-children.spec.js
```diff
@@ -1,74 +1,85 @@
...
         .to.be.instanceof(Array)
         .and.to.have.length(0);
     });
 
     describe('without regions', function() {
+      let layoutView;
+
       beforeEach(function() {
-        this.layoutView = new Marionette.View({
+        layoutView = new Marionette.View({
```
paulfalgout: Ideally `View` is imported vs pulled off the global

test/unit/object.spec.js
```diff
@@ -1,4 +1,4 @@
-import MnObject from '../../src/object';
+import Object from '../../src/object';
```
blikblum: This works?

paulfalgout: Babel converts:
```js
import Object from '../../src/object';

new Object();
```
to:
```js
'use strict';

var _object = require('../../src/object');

var _object2 = _interopRequireDefault(_object);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

new _object2.default();
```

paulfalgout: so in light of this working for babel.. and working apparently in node..  do we rename it?

blikblum: But in export will work? In this case renaming is not possible

paulfalgout: actually exporting works as well.  we should really test this in node, but I think it's just _recommended_ not to export `Object` it looks like it will work.  I'm kind of thinking we would be fine renaming it back to `Object`

blikblum: Is asking for trouble

```
import Object from 'backbone.marionete'

export default Object.extend({
  initialize() {
    this.keys = Object.keys(this)
  }
})
```

paulfalgout: for now @wesmangum probably best to switch this to `MnObject`

PR #3535, opened 2017-11-10
refactor and reorganize view trigger and entity events unit tests

wesmangum:
### Proposed changes

Link to the issue: #3248


test/unit/view.triggers.spec.js
```diff
@@ -1,109 +1,138 @@
+import Backbone from 'backbone';
+import Marionette from '../../src/backbone.marionette';
```
paulfalgout: Hopefully we can remove this import entirely

PR #3536, opened 2017-11-13
Propagate only the first argument (options) passed to destroy

blikblum:
### Proposed changes
 - Instead of propagating all arguments  passed to destroy to events and other methods, pass only the first one (options)
 - Remove tests that are checking return of 'before:destroy' event that has no effect. The tests were different from the enunciate anyway

This change simplify the code and align with other method signatures like `region.reset`, `region.empty`, `region.show`

### DIFF
--- a/src/mixins/behaviors.js
+++ b/src/mixins/behaviors.js
@@ -75,12 +75,12 @@ export default {
     _invoke(this._behaviors, 'undelegateEntityEvents');
   },
 
-  _destroyBehaviors(...args) {
+  _destroyBehaviors(options) {
     // Call destroy on each behavior after
     // destroying the view.
     // This unbinds event listeners
     // that behaviors have registered for.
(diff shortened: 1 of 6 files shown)


paulfalgout:
I like and agree with this change for a variety of reasons.   Anyone have an issue with removing it prior to any sort of deprecation.  I think it's a very easy breaking change to deal with

JSteunou:
Just a little concerned about the tests that were removed. Seems to be linked to a different feature...

paulfalgout:
those tests are old and should have been removed.. probably pre v2.. maybe pre v1... but were instead maintained after their related feature was removed.  There's a handful of tests around like this..  It doesn't matter what you return in `onBeforeDestroy`

JSteunou:
Yeah that's what I though... Or there was a new feature I did ignore about :D

PR #3543, opened 2017-11-15
Optimize _getUIBindings

blikblum:
### Proposed changes
 - Takes advantage of short circuit evaluation to only execute `_.result(this, 'ui')` when `uiBindings` is falsy


paulfalgout:
This should be entirely uncontroversial..

PR #3544, opened 2017-11-15
Create rest arguments in triggerMethod only when there's a method

blikblum:
### Proposed changes
 - Currently the rest arguments of triggerMethod are created regardless the existence of a matching method. Changes to only create the rest arguments when there's a method (the minority of cases)

### DIFF
--- a/src/common/trigger-method.js
+++ b/src/common/trigger-method.js
@@ -33,7 +33,7 @@ export default function triggerMethod(event) {
   // call the onMethodName if it exists
   if (_.isFunction(method)) {
     // pass all args, except the event name
-    result = method.apply(this, _.rest(arguments));
+    result = method.apply(this, _.drop(arguments));
   }
 
   // trigger the event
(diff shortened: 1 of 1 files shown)


paulfalgout:
Since this is such a hot method and `onRender` is used frequently we'll get much better perf if we use the babel way instead of `_.drop` just inside the `if (_.isFunction`
```js
  for (var _len = arguments.length, args = Array(_len > 1 ? _len - 1 : 0), _key = 1; _key < _len; _key++) {
    args[_key - 1] = arguments[_key];
  } 
```

paulfalgout:
Hmm I dunno.. did a quick jsperf and using the babel `for` along with the `trigger.apply` seems to de-opt the method.. so we get near 10x the speed on events without an onHandler using `_.drop` and about 60% the speed on methods that do have a `onHandler`

blikblum:
>  we'll get much better perf if we use the babel way instead of _.drop just inside the if (_.isFunction

The benefit is that the code will not be executed at all in most cases (method does not exists).

paulfalgout:
the babel for loop is much faster than `_.drop` and we could put the loop inside the `if` just like the `_.drop` for the exact same benefit but with faster results.  _However_ for some reason this de-opts the function when the method does not exist.

PR #3548, opened 2017-11-15
reorganize and refactor common buildRegion unit tests

wesmangum:
### Proposed changes
- Reorganize tests to match order of codebase

Link to the issue: #3248

### DIFF
--- a/test/unit/common/build-region.spec.js
+++ b/test/unit/common/build-region.spec.js
@@ -1,66 +1,76 @@
-import * as Marionette from '../../../src/backbone.marionette';
+import View from '../../../src/view';
+import Region from '../../../src/region';
 
 describe('Region', function() {
   describe('.buildRegion', function() {
+    let DefaultRegionClass;
+    let view;
+    let fooSelector;
+    let barSelector;
+    let BarRegion;
+    let BazRegion;
+
     beforeEach(function() {
 
-      this.DefaultRegionClass = Marionette.Region.extend();
+      DefaultRegionClass = Region.extend();
 
-      this.view = new Marionette.View({
+      view = new View({
         template: _.noop,
-        regionClass: this.DefaultRegionClass
+        regionClass: DefaultRegionClass
       });
 
-      this.fooSelector = '#foo-region';
-      this.fooRegion = new this.DefaultRegionClass({el: this.fooSelector});
+      fooSelector = '#foo-region';
 
-      this.barSelector = '#bar-region';
-      this.BarRegion = Marionette.Region.extend({el: this.barSelector});
-      this.barRegion = new this.BarRegion();
+      barSelector = '#bar-region';
+      BarRegion = Region.extend({el: barSelector});
 
-      this.BazRegion = Marionette.Region.extend();
+      BazRegion = Region.extend();
     });
 
     describe('with a selector string', function() {
+      let region;
+
       beforeEach(function() {
-        this.region = this.view.addRegion(_.uniqueId('region_'), this.fooSelector);
+        region = view.addRegion(_.uniqueId('region_'), fooSelector);
       });
 
       it('uses the default region class', function() {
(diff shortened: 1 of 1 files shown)


test/unit/common/build-region.spec.js
```diff
@@ -1,294 +1,355 @@
...
-
-            this.region1 = this.view.addRegion(_.uniqueId('region_'),this.region1Definition);
-            this.region2 = this.view.addRegion(_.uniqueId('region_'),this.region2Definition);
-            this.region3 = this.view.addRegion(_.uniqueId('region_'),this.region3Definition);
+            baseDefinition = {regionClass: BazRegion};
+            region1Definition = _.defaults({el: fooSelector}, baseDefinition);
+            region2Definition = _.defaults({el: el}, baseDefinition);
+            region3Definition = _.defaults({el: $el}, baseDefinition);
+
+            new BazRegion({el: fooSelector});
```
paulfalgout: these don't seem necessary

test/unit/common/build-region.spec.js
```diff
@@ -1,294 +1,355 @@
...
+        let el;
         beforeEach(function() {
-          this.$el = $('<div id="baz-region">');
-          this.el = this.$el[0];
+          $el = $('<div id="baz-region">');
+          el = $el[0];
         });
 
         describe('with `el` also defined', function() {
+          let baseDefinition;
```
paulfalgout: some of these only need to be `let`s if they're tested outside of the `beforeEach` if it's just a variable used within a single function let's keep it `const`

test/unit/common/build-region.spec.js
```diff
@@ -1,294 +1,355 @@
...
         });
 
         it('throws a `NoElError`', function() {
-          expect(this.buildRegion).to.throw('An "el" must be specified for a region.');
+          expect(buildRegion).to.throw('An "el" must be specified for a region.');
         });
       });
 
       describe('with `regionClass` defined', function() {
+        let $el;
```
paulfalgout: these `el` and `$el` are only used inside the next `describe` block.. let's move them there

test/unit/common/build-region.spec.js
```diff
@@ -1,294 +1,355 @@
...
-          this.el = $('i-am-not-real');
-          this.definition = {el: this.el};
+          el = $('i-am-not-real');
+          definition = {el: el};
 
-          this.buildRegion = function() {
-            this.view.addRegion(_.uniqueId('region_'),this.definition);
-          }.bind(this);
+          buildRegion = function() {
+            view.addRegion(_.uniqueId('region_'),definition);
```
paulfalgout: `, `

test/unit/common/build-region.spec.js
```diff
@@ -1,294 +1,355 @@
...
-            expect(this.region.el).to.equal(this.el[0]);
+            expect(region.el).to.equal(el[0]);
           });
         });
       });
 
       describe('when el is an empty jQuery object', function() {
+        let el;
+        let definition;
+        let buildRegion;
```
paulfalgout: as below I think `buildRegion` is the only necessary let, the others can be `const` inside the `beforeEach`s

test/unit/common/build-region.spec.js
```diff
@@ -1,294 +1,355 @@
...
+          expect(region.el).to.equal(barSelector);
         });
       });
 
       describe('without `el` defined', function() {
+        let buildRegion;
+
         beforeEach(function() {
-          this.BarRegion = Marionette.Region.extend();
+          BarRegion = Region.extend();
```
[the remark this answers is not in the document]
wesmangum: The above assignment of `BarRegion` has an el assignment, which would make these tests fail. Is there a better way to handle the reassignment?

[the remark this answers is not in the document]
paulfalgout: how about `BazRegion`

PR #3553, opened 2017-11-23
POC Make MnObject extend Object

paulfalgout:
Makes Object slightly more purposeful maybe and makes renaming `Object` to `MnObject` possible less important.

We _could_ even export under both names?

There could be some significant downsides to this, but what are they?

### DIFF
--- a/src/backbone.marionette.js
+++ b/src/backbone.marionette.js
@@ -61,6 +61,7 @@ export {
   View,
   CollectionView,
   MnObject,
+  MnObject as Object,
   Region,
   Behavior,
   Application,
--- a/src/object.js
+++ b/src/object.js
@@ -15,28 +15,24 @@ const ClassOptions = [
   'radioRequests'
 ];
 
-// Object borrows many conventions and utilities from Backbone.
-const MarionetteObject = function(options) {
-  this._setOptions(options);
-  this.mergeOptions(options, ClassOptions);
-  this.cid = _.uniqueId(this.cidPrefix);
-  this._initRadio();
-  this.initialize.apply(this, arguments);
-};
+const MnObject = extend.call(Object, {
+  constructor(options) {
+    Object.prototype.constructor.apply(this, arguments);
+    this._setOptions(options);
+    this.mergeOptions(options, ClassOptions);
+    this.cid = _.uniqueId(this.cidPrefix);
+    this._initRadio();
+    this.initialize.apply(this, arguments);
+  },
 
-MarionetteObject.extend = extend;
-
-// Object Methods
-// --------------
-
-// Ensure it can trigger events with Backbone.Events
-_.extend(MarionetteObject.prototype, Backbone.Events, CommonMixin, DestroyMixin, RadioMixin, {
   cidPrefix: 'mno',
 
   // This is a noop method intended to be overridden
   initialize() {},
 
   triggerMethod
-});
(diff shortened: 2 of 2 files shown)


src/backbone.marionette.js
```diff
@@ -61,6 +61,7 @@ export {
   View,
   CollectionView,
   MnObject,
+  MnObject as Object,
```
paulfalgout: I think a big reason for adding this, regardless of extending `Object`, is that this works better for backwards compatibility.  But we can deprecate it and change all of the documentation to use `MnObject`  This allows any additional library to have a better change of working with v4 without changes and time to convert prior to v5 if it needs to be removed.  Any major objections?  @blikblum @bazineta @JSteunou @rafde ?

blikblum: The main issue i see is that the user won't know if is using system Object or Mn Object, just looking at the immediate code. It will depend if Object is being imported from Mn or not. 

Otherwise i don't see actual usage problems but this is unknown territory and may exist (or not) some issue that we are not foreseeing. 

So in the fence. 

Also, it works with ES module build where import/export is not transpiled?

blikblum: Found an actual usage of Object that will not work, when using as a wrapper:

form mdn:

"The Object constructor creates an object wrapper for the given value. If the value is null or undefined, it will create and return an empty object, otherwise, it will return an object of a Type that corresponds to the given value. If the value is an object already, it will return the value."

Usage:
```
function isObject(obj) {
  return obj === Object(obj);
}
```

paulfalgout: Ok so 2 issues.  1 is should we export `Object` in addition to `MnObject`?  I'm leaning towards yes for backwards compatibility of `Marionette.Object` and we can deprecate it if need be.  But we can export `MnObject` as a convienence for those who don't want to `import { Object as MnObject } from 'backbone.marionette'`  Though honestly I'm kind of leaning towards that as a pretty good solution anyhow..

Then 2.. regarding this PRs original purpose of extending `window.Object` I did get the wrapper working https://jsfiddle.net/d3a6hzw7/1/  seems doable whereas `Marionette.Object` would be 100% compatible with `window.Object` https://jsfiddle.net/d3a6hzw7/3/

blikblum: Object.keys is not working:

https://jsfiddle.net/d3a6hzw7/4/

Did i missed something?

paulfalgout: Ah sure enough it's nonenumerable.. fixed: https://jsfiddle.net/d3a6hzw7/5/

I'm buying this as an option less and less.. can't tell if it's valuable..  I do like possibly not having to use "MnObject" as a name and giving a slight validation to Marionette.Object having a reason to exist, but this is starting to feel a little gross..  maybe it's fine though?

blikblum: I think neither solution (MnObject or making MnObject acts as Object) are ideal, but IMO the usage of MnObject name have less drawbacks than trying to make MnObject act as Object.

In long term i see the option to migrate to ES6 class completely and use ES6 class mixins (http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/) like skatejs is doing.

So we could provide a few mixins (withCid, withEvents, withLifecycle etc) and let the user decide what features, name his/her base class will have.

PR #3555, opened 2017-11-23
Privatize and improve event / trigger pref 

paulfalgout:
`getEvents` and `getTriggers` have no public purpose.
I also added some guards if there are no events defined.
Additionally there is one breaking change.  handlers for `childViewEvents` must now be defined prior to `delegateEvents`  This will match every other events related API.  If someone needs to add a `childViewEvents` handle after the fact, they'll need to re-call `delegateEvents`.
This should significantly improve perf in some cases.

I had originally intended on simply reorganizing the behavior tests a bit, but one thing led to another.


paulfalgout:
I think I can improve on this further.  I'll try one other thing soon.

paulfalgout:
Ok after the 1st go I made an additional change to move the `childViewEventPrefix` to be determined prior to the event proxy.  It is currently calculated on every triggered event.  This is breaking for that reason, but now every event related API is determined around the same time.. either when `delegateEvents` is called, or for entity events, directly after `initialize`.  This makes a big difference if a view isn't going to listen to any child view.  It is worth noting that the child related events are a bit weird to be cached during `delegateEvents` which is primarily meant for ui..  However other than maybe `setElement` which doesn't make much more sense, there's not a better place to put it as it must be called during the backbone constructor.

This did end up being a bigger change than originally intended.

PR #3556, opened 2017-11-26
Implement setRenderer destructive option

blikblum:
### Proposed changes
 - when passing an options hash to `setRenderer` with destructive: false, do not reinit regions on view re-render
 
Link to the issue: #3427

### DIFF
--- a/src/config/renderer.js
+++ b/src/config/renderer.js
@@ -1,5 +1,8 @@
 // Static setter for the renderer
-export function setRenderer(renderer) {
+export function setRenderer(renderer, options) {
   this.prototype._renderHtml = renderer;
+  if (options && options.destructive === false) {
+    this.prototype._preventRegionReInit = true;
+  }
   return this;
 }
--- a/src/view.js
+++ b/src/view.js
@@ -85,7 +85,7 @@ const View = Backbone.View.extend({
 
     // If this is not the first render call, then we need to
     // re-initialize the `el` for each region
-    if (this._isRendered) {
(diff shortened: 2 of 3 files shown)


paulfalgout:
I'd support this for an experimental API.  Looking forward to v4.x and v5 for region improvements and I don't want to box ourselves in.

blikblum:
I'm ok with marking as an experimental API.
In fact, i think the renderer api should be extended to handle other things like dom:refresh and dom:remove events that does not make sense with non destructive renderers

src/config/renderer.js
```diff
@@ -1,5 +1,8 @@
 // Static setter for the renderer
-export function setRenderer(renderer) {
+export function setRenderer(renderer, options) {
   this.prototype._renderHtml = renderer;
+  if (options && options.destructive === false) {
```
scott-w: I think the feature could be good, though I don't like the name.

Maybe just call this `reInitializeRegion` or maybe `preserveRegion` so the flag check can be `options.preserveRegion`.

blikblum: I'm fine changing the name but should not be region specific. This flag may be needed for other things than reinitialize region like for example adding an onReady event which would have a different behavior from destructive and non destructive regions.

scott-w: What about `preventDestroy`? It's used in stuff like `region.show` and it feels like it does a similar job.

paulfalgout:
Something like this may make it to https://github.com/marionettejs/marionette

PR #3557, opened 2017-11-27
Refactor template-render mixin specs

paulfalgout:
Coverage remains the same.
This work narrows the test only to the unit being tested.


paulfalgout:
Merging with one thumb since there is no src changes

PR #3560, opened 2017-11-28
Make radio mixin spec more unit-specific

paulfalgout:
Tests only the unit in question.  the "as an option at instantation" are duplicating tests in the object.spec which are better defined there.

100% Coverage remains


paulfalgout:
I pushed a small change switching to `import Radio from 'backbone.radio';` rather than using `Backbone.Radio`

blikblum:
> I pushed a small change switching to import Radio from 'backbone.radio'; rather than using Backbone.Radio

Good call

PR #3566, opened 2017-12-08
Improve triggerMethod

paulfalgout:
I wasn't planning on making additional code changes apart from the spec work, but one rabbit trail led to another and I figured we may not want to introduce a new API that might be replaced soon.  

Partially resolves https://github.com/marionettejs/backbone.marionette/issues/2508

~~Renames `BackboneViewMixin` to `Marionette.Events`.  Still works for Bb Views but also sets up the potential for the rest of #2508 in v5.~~

Perhaps the biggest change is that this moves the behavior event proxy to the behavior.  This should be a perf improvement for `triggerMethod` in general.. particularly when no behaviors are applied.  I'm not sure the perf differences on the behaviors themselves.

### DIFF
--- a/src/behavior.js
+++ b/src/behavior.js
@@ -7,10 +7,8 @@
 // into portable logical chunks, keeping your views simple and your code DRY.
 
 import _ from 'underscore';
-import Backbone from 'backbone';
 import extend from './utils/extend';
 import getNamespacedEventName from './utils/get-namespaced-event-name';
-import triggerMethod from './common/trigger-method';
 import CommonMixin from './mixins/common';
 import DelegateEntityEventsMixin from './mixins/delegate-entity-events';
 import TriggersMixin from './mixins/triggers';
@@ -44,6 +42,9 @@ const Behavior = function(options, view) {
   // selector under an UI key.
   this.ui = _.extend({}, _.result(this, 'ui'), _.result(view, 'ui'));
 
+  // Proxy view triggers
+  this.listenTo(view, 'all', this.triggerMethod);
+
   this.initialize.apply(this, arguments);
 };
 
@@ -52,8 +53,7 @@ Behavior.extend = extend;
 // Behavior Methods
 // --------------
 
-// Ensure it can trigger events with Backbone.Events
-_.extend(Behavior.prototype, Backbone.Events, CommonMixin, DelegateEntityEventsMixin, TriggersMixin, UIMixin, {
+_.extend(Behavior.prototype, CommonMixin, DelegateEntityEventsMixin, TriggersMixin, UIMixin, {
   cidPrefix: 'mnb',
 
   // This is a noop method intended to be overridden
@@ -141,9 +141,7 @@ _.extend(Behavior.prototype, Backbone.Events, CommonMixin, DelegateEntityEventsM
     const behaviorTriggers = this.normalizeUIKeys(_.result(this, 'triggers'));
 
     return this._getViewTriggers(this.view, behaviorTriggers);
-  },
-
-  triggerMethod
+  }
 });
 
 export default Behavior;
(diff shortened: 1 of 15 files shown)


blikblum:
My opinion:

 *  #2508 is bad idea, because introduce the triggerMethod behavior (build method name and search in instance for a method) to all events triggers leading to performance overhead. Also is a big breaking change with potential to affect code outside of Marionette e.g. user code and Backbone plugins

 * We may want to extend BackboneViewMixin with `destroy` method so is possible to remove `destroyView` methods used in Marionette. `destroy` does not fit in Marionette.Events

* Backbone.View already has Backbone.Events mixed. By mixing Marionette.Events to a Backbone.View this is redundant

* The behavior event proxy to the behavior move can be accomplished without renaming the mixin

paulfalgout:
1. I removed `Marionette.Events` for the moment to get the behavior event proxy in
2. I think `Marionette.Events` is a good thing even if we never decide to merge `trigger` and `triggerMethod` as it would be one of the key steps to removing `backbone` as a dependency.  (That, `extend` and the base view functionality is really it)
3. I don't see much value if any from moving destroy to the `BackboneViewMixin`  Trying that is what led me to this point and it doesn't seem like any sort of real improvement, particularly since the same cannot be done with the render.  
If 2 and/or 3 is true `Marionette.Events` is a better name and is useful for more than just views.  Want `triggerMethod` on your models?  no problem.

src/common/trigger-method.js
```diff
@@ -7,15 +7,22 @@ import getOption from './get-option';
...
 // and turn it in to uppercase name onSection1Section2Section3
 function getEventName(match, prefix, eventName) {
   return eventName.toUpperCase();
 }
 
-const getOnMethodName = _.memoize(function(event) {
-  return 'on' + event.replace(splitter, getEventName);
-});
+const getOnMethodName = function(event) {
+  if (!methodCache[event]) {
```
blikblum: This is a new feature. Committed by accident?

test/unit/behavior.spec.js
```diff
@@ -682,12 +667,6 @@ describe('Behavior', function() {
 
       expect(listenToChangeStub).not.to.have.been.calledOnce;
     });
-
-    it('should still be bound to "on" on destroy', function() {
```
blikblum: Not true anymore?

If so maybe test that is not bound on destro

src/mixins/view.js
```diff
@@ -142,6 +141,7 @@ const ViewMixin = {
     this._deleteEntityEventHandlers();
 
     this.triggerMethod('destroy', this, options);
+    this._triggerEventOnBehaviors('destroy', this, options);
```
blikblum: It will not duplicate destroy triggerMethod call in behaviors?

paulfalgout: the listeners on the behavior are cleaned up by this point, so the destroy needs to be handled manually.  Not true for `before:destroy` however

src/common/trigger-method.js
paulfalgout: it's the same functionality that was there just less complex.  `_.memoize` has a bunch of other stuff going on that isn't needed.

test/unit/behavior.spec.js
paulfalgout: yep not true anymore.  `view.destroy().triggerMethod('foo');` will call `onFoo` on the view, but no longer on the view's behaviors.  This test was essentially duplicating the test to make sure `onDestroy` was called after the `destroy`  But testing that events listened to on the behavior itself are cleaned up on destroy is already being tested here: https://github.com/marionettejs/backbone.marionette/blob/next/test/unit/behavior.spec.js#L861

src/behavior.js
```diff
@@ -44,6 +42,9 @@ const Behavior = function(options, view) {
   // selector under an UI key.
   this.ui = _.extend({}, _.result(this, 'ui'), _.result(view, 'ui'));
 
+  // Proxy view triggers
+  this.listenTo(view, 'all', this.triggerMethod);
```
JSteunou: Hooo nice did know about this trick!

PR #3568, opened 2017-12-09
Add common view specs

paulfalgout:
This work adds a unit spec file for the common view methods.

I believe to have fixed a bug with rendering backbone views not setting the `_isRendered` flag unless the view had `supportsRenderLifecycle`

I also reduced the need for `_disableDetachEvents` to be attached to the view and brought it much closer to it's need in `view.destroy` which also made it easier to test.

What I haven't done is look for if/where this functionality is being tested in other files.  Hopefully that will come out in some sequent unit clean up

### DIFF
--- a/src/common/view.js
+++ b/src/common/view.js
@@ -8,24 +8,27 @@ export function renderView(view) {
   }
 
   view.render();
+  view._isRendered = true;
 
   if (!view.supportsRenderLifecycle) {
-    view._isRendered = true;
     view.triggerMethod('render', view);
   }
 }
 
-export function destroyView(view) {
+export function destroyView(view, disableDetachEvents) {
   if (view.destroy) {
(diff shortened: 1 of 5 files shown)


src/common/view.js
```diff
@@ -8,24 +8,27 @@ export function renderView(view) {
...
+  view._isRendered = true;
 
   if (!view.supportsRenderLifecycle) {
-    view._isRendered = true;
     view.triggerMethod('render', view);
   }
 }
 
-export function destroyView(view) {
+export function destroyView(view, disableDetachEvents) {
```
scott-w: This looks new to me, do you see this being publicly documented or is it only used internally?

paulfalgout: This entire module is not publicly accessible

PR #3582, opened 2018-01-31
Add filteredChildren container to CollectionView

### LINKED ISSUE #3580: View event monitor will mark filtered out children as attached
In v4 (and NextCollectionView for v3.x) the view event monitor will get all children and trigger attach events along with marking views attached for each child as it does not understand whether a child has been view filtered.

https://codepen.io/dimatabu/pen/zpbovw

paulfalgout:
Fixes #3580

Besides this bug, the ability to understand what views are currently not filtered out has been a semi-regular request.

This does add an extra container, but I don't think it should cost that much.  I intend to run this against the js-framework-benchmark to see how it affects thing, but I don't think it'll be bad.

The big question I _think_ is in the naming..  I'm not sure `filteredChildren` makes sense as a public API?  In context it kinda does..  _however_ one alternate option is to name _this_ container to `children` and make the current `children` private.  I don't _really_ know a use-case for needing both attached and detached children.

I could see arguments either way.

@marionettejs/marionette-core ?

This needs more tests and docs

### DIFF
--- a/src/child-view-container.js
+++ b/src/child-view-container.js
@@ -43,20 +43,22 @@ _.extend(Container.prototype, {
   // cid (and model itself). Additionally it stores
   // the view by index in the _views array
   _add(view, index = this._views.length) {
-    const viewCid = view.cid;
-
-    // store the view
-    this._viewsByCid[viewCid] = view;
-
-    // index it by model
-    if (view.model) {
-      this._indexByModel[view.model.cid] = viewCid;
-    }
+    this._addViewIndexes(view);
 
     // add to end by default
     this._views.splice(index, 0, view);
 
     this._updateLength();
+  },
+
+  _addViewIndexes(view) {
+    // store the view
+    this._viewsByCid[view.cid] = view;
+
+    // index it by model
+    if (view.model) {
+      this._indexByModel[view.model.cid] = view;
+    }
   },
 
   // Sort (mutate) and return the array of the child views.
@@ -83,12 +85,20 @@ _.extend(Container.prototype, {
   },
 
   // Replace array contents without overwriting the reference.
-  _set(views) {
+  // Should not add/remove views
+  _set(views, shouldReset) {
     this._views.length = 0;
 
     this._views.push.apply(this._views, views.slice(0));
 
-    this._updateLength();
+    if (shouldReset) {
+      this._viewsByCid = {};
+      this._indexByModel = {};
+
+      _.each(views, this._addViewIndexes.bind(this));
+
+      this._updateLength();
+    }
   },
 
   // Swap views by index
@@ -112,11 +122,8 @@ _.extend(Container.prototype, {
   },
 
   // Find a view by the `cid` of the model that was attached to it.
-  // Uses the model's `cid` to find the view `cid` and
-  // retrieve the view using it.
   findByModelCid(modelCid) {
-    const viewCid = this._indexByModel[modelCid];
-    return this.findByCid(viewCid);
+    return this._indexByModel[modelCid];
   },
 
   // Find a view by index.
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -48,11 +48,6 @@ const CollectionView = Backbone.View.extend({
 
     monitorViewEvents(this);
 
-    this.once('render', this._initialEvents);
-
-    // This children container isn't really used by a render, but it provides
-    // the ability to check `this.children.length` prior to rendering
-    // It also allows for cases where only addChildView is used
     this._initChildViewStorage();
     this._initBehaviors();
 
@@ -67,7 +62,10 @@ const CollectionView = Backbone.View.extend({
   },
 
   // Internal method to set up the `children` object for storing all of the child views
+  // `_children` represents all child views
+  // `children` represents only views filtered to be shown
   _initChildViewStorage() {
+    this._children = new ChildViewContainer();
     this.children = new ChildViewContainer();
   },
 
@@ -86,6 +84,8 @@ const CollectionView = Backbone.View.extend({
 
   // Configured the initial events that the collection view binds to.
   _initialEvents() {
+    if (this._isRendered) { return; }
+
     this.listenTo(this.collection, {
       'sort': this._onCollectionSort,
       'reset': this._onCollectionReset,
@@ -112,11 +112,9 @@ const CollectionView = Backbone.View.extend({
   _onCollectionReset() {
     this._destroyChildren();
 
-    this.children._init();
-
     this._addChildModels(this.collection.models);
 
-    this._showChildren();
+    this.sort();
   },
 
   // Handle collection update model additions and  removals
@@ -130,7 +128,7 @@ const CollectionView = Backbone.View.extend({
 
     this._detachChildren(removedViews);
 
-    this._showChildren();
+    this.sort();
 
     // Destroy removed child views after all of the render is complete
     this._removeChildViews(removedViews);
@@ -147,7 +145,7 @@ const CollectionView = Backbone.View.extend({
   },
 
   _removeChildModel(model) {
-    const view = this.children.findByModel(model);
+    const view = this._children.findByModel(model);
 
     if (view) { this._removeChild(view); }
 
@@ -158,6 +156,7 @@ const CollectionView = Backbone.View.extend({
     this.triggerMethod('before:remove:child', this, view);
 
     this.children._remove(view);
+    this._children._remove(view);
 
     this.triggerMethod('remove:child', this, view);
   },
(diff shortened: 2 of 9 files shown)


bazineta:
It seems to me that the name matches the most common convention, i.e., while 'filtered' in general can mean 'excluded' or 'included', the convention arising from Array.prototype.filter() means 'included', so I think it's reasonable as-is.

paulfalgout:
benchmarks differences here were all within the margin of error.  This change actually benchmarked faster by a very tiny amount, but I don't think it is actually faster

rafde:
> In context it kinda does.. however one alternate option is to name this container to children and make the current children private. I don't really know a use-case for needing both attached and detached children.

I think making the filteredChildren `children` and making the detached and attached children `_children` makes sense since you usually loop through what's been rendered.
If you want a public api to access de/attached children, could have a function like `getChildren()` and `getRenderedChildren()` and deprecate direct access to `this.children` or something like that.

scott-w:
From my experience with Backbone Paginator, I'm used to `children` being the children to show and `fullCollection` holding all the models if you want to iterate over them.

That said, if we could find a way to handle that state in Marionette and remove the need for a separate collection handling library, I'd be all for that :)

paulfalgout:
So there's a bit more to be done here. It turns out that handling the separate collection is pretty cheap, but it also provides cleaner ways to manage the empty view and handle other things.  I'm working on a bigger revision, but assuming all goes well, it will keep perf (actually slightly faster) while decreasing complexity quite a bit.  For the moment the full children collection will be `_children` and `children` will be only the views filtered to be shown.

paulfalgout:
@marionettejs/marionette-core

Ok.. so it wasn't _quite_ as straight-forward. Only `_set`ing the views on the "filtered" `children` was not updating the two view hashes on the containers.  So this meant that after filtering `hasView` and the various `findBy__` could be wrong.  So I added a "reset" flag on the private `_set`

I was concerned about perf hits, but with all of the changes adding 10k rows in the js-framework-benchmark produced only about 20ms on multiple runs which is within the margin of error (though obviously this will be marginally slower)

What this buys us is a cleaner API.  `children` now only contains the filtered views and `isEmpty` is only called once and the check is not dependent on a all filtered flag.

Most of the changes in this PR now are moving the methods into calling order.
I'll make some additional comments on the commit.

But this'll need new eyes.

src/child-view-container.js
```diff
@@ -43,20 +43,22 @@ _.extend(Container.prototype, {
   // cid (and model itself). Additionally it stores
   // the view by index in the _views array
   _add(view, index = this._views.length) {
-    const viewCid = view.cid;
+    this._addViewIndexes(view);
```
paulfalgout: this was made into a method so the same indexing could be used by "reset"

src/child-view-container.js
```diff
@@ -83,12 +85,20 @@ _.extend(Container.prototype, {
...
     this._views.push.apply(this._views, views.slice(0));
 
-    this._updateLength();
+    if (shouldReset) {
+      this._viewsByCid = {};
+      this._indexByModel = {};
+
+      _.each(views, this._addViewIndexes.bind(this));
+
+      this._updateLength();
```
paulfalgout: `shouldReset` will only ever be called if the length changes

src/child-view-container.js
```diff
@@ -112,11 +122,8 @@ _.extend(Container.prototype, {
   },
 
   // Find a view by the `cid` of the model that was attached to it.
-  // Uses the model's `cid` to find the view `cid` and
-  // retrieve the view using it.
   findByModelCid(modelCid) {
-    const viewCid = this._indexByModel[modelCid];
-    return this.findByCid(viewCid);
+    return this._indexByModel[modelCid];
```
paulfalgout: somewhat unrelated, but there was no reason to store the view cid instead of the view.

src/collection-view.js
```diff
@@ -48,11 +48,6 @@ const CollectionView = Backbone.View.extend({
 
     monitorViewEvents(this);
 
-    this.once('render', this._initialEvents);
```
paulfalgout: fairly unrelated to this PR, but moving this to `render` with some of the other changes was a perf benefit

src/collection-view.js
```diff
@@ -48,11 +48,6 @@ const CollectionView = Backbone.View.extend({
...
-
-    // This children container isn't really used by a render, but it provides
```
paulfalgout: no longer true

src/collection-view.js
```diff
@@ -86,6 +84,8 @@ const CollectionView = Backbone.View.extend({
 
   // Configured the initial events that the collection view binds to.
   _initialEvents() {
+    if (this._isRendered) { return; }
```
paulfalgout: this check prevents the need for `this.once` in the constructor

src/collection-view.js
```diff
@@ -446,45 +381,48 @@ const CollectionView = Backbone.View.extend({
...
+    if (!this._children.length) { return; }
 
-    delete this._addedViews;
+    const viewFilter = this._getFilter();
 
     if (!viewFilter) {
-      if (addedViews) { return addedViews; }
+      const shouldReset = this.children.length !== this._children.length;
+
+      this.children._set(this._children._views, shouldReset);
```
paulfalgout: This will "reset" the children if the current `this.children` length doesn't match all of the views.  But if they do match, then we can assume the indexes are correct and just `_set` for a possible change to the array sort

src/collection-view.js
```diff
@@ -567,21 +505,36 @@ const CollectionView = Backbone.View.extend({
...
   },
 
-  _renderChildren(views) {
-    if (this.isEmpty(!views.length)) {
+  _renderChildren() {
+    const views = this._addedViews || this.children._views;
+
+    this.triggerMethod('before:render:children', this, views);
+
+    if (this.isEmpty()) {
```
paulfalgout: one major change here is that the `render:children` events will now fire even if the empty view is shown.  Previously with the two `isEmpty` checks this was fairly difficult without artificially triggering those events in odd cases.  Now we can throw the event when 0 children are rendered.  This will simplify having an event for when the children are rendered even if there is no empty view.

src/collection-view.js
```diff
@@ -603,39 +556,77 @@ const CollectionView = Backbone.View.extend({
...
-    return elBuffer;
-  },
-
   // Override this method to do something other than `.append`.
   // You can attach any HTML at this point including the els.
   attachHtml(els) {
     this.Dom.appendContents(this.$container[0], els, {_$el: this.$container});
   },
 
+  isEmpty() {
```
paulfalgout: here's the major change to the empty view.. the rest of the methods below were just moved.

src/collection-view.js
```diff
@@ -603,39 +556,77 @@ const CollectionView = Backbone.View.extend({
...
       });
     }
 
-    this.children._swap(view1, view2);
+    this._children._swap(view1, view2);
     this.Dom.swapEl(view1.el, view2.el);
 
     // If the views are not filtered the same, refilter
-    if (this.Dom.hasEl(this.el, view1.el) !== this.Dom.hasEl(this.el, view2.el)) {
+    if (this.children.hasView(view1) !== this.children.hasView(view2)) {
```
paulfalgout: this was a nice change.

src/collection-view.js
```diff
@@ -723,15 +714,20 @@ const CollectionView = Backbone.View.extend({
 
   // Destroy the child views that this collection view is holding on to, if any
   _destroyChildren() {
-    if (!this.children || !this.children.length) {
+    if (!this._children.length) {
```
paulfalgout: it is impossible to not have `this._children` defined by this point

test/unit/collection-view/collection-view-viewmixin.spec.js
```diff
@@ -163,11 +163,8 @@ describe('CollectionView - ViewMixin', function() {
       collectionView._removeChildren();
     });
 
-    // Since the collectionView is destroyed we
```
paulfalgout: in this case this test was added early on when at one point all views were removed in an `_.each`.  This test was added when it was determined that it wasn't necessary.  Now the `_init` in the `_destroyView` does set the length to 0, but does it cheaply

JSteunou:
this is ok to me, even if I do not have the bandwith to review this more in depth :/ I very like the idea of `children` representing the attached / filtered children and `_children` being the full list / cache of children. Again :clap:  to @paulfalgout

paulfalgout:
I'm wrapping up my documentation work.  This is I believe the last of the alpha v4 work and needs to get in.  Unless there are objections I'm going to merge this soon. @marionettejs/marionette-core

PR #3590, opened 2018-03-08
Fix a broken link in the documentation

gpgekko:


docs/marionette.approuter.md
```diff
@@ -52,7 +52,7 @@ will be covered in more detail below.
 The AppRouter uses the `appRoutes` option to define how we respond to routes
 being accessed. To define routes, set the route as your key and the method to
 call as a string referencing a method on your controller. For more information
-on route definitions, see the [Backbone documentation][#backbone-routes].
+on route definitions, see the [Backbone documentation][backbone-routes].
```
paulfalgout: I don't think this is correct I think it needs to be `[Backbone documentation](http://backbonejs.org/#Router-routes)`

gpgekko: I thought the same at first, but these are some kind of named references to[ the links defined at the bottom of the document](https://github.com/marionettejs/backbone.marionette/pull/3590/commits/27eb08d9ebd40b29707fa9758f29819747e56aee#diff-9175f0f44dd039f2958d365391ec492eR243). It's some kind of Markdown magic that is used throughout this document.

paulfalgout: Well links are `[]()` with the link in the `()` not square brackets, but even so the `#` would be correct _if_ it were a link somewhere in the doc.. but I don't see it.  If it were correct this would go somewhere https://github.com/gpgekko/backbone.marionette/blob/27eb08d9ebd40b29707fa9758f29819747e56aee/docs/marionette.approuter.md#backbone-routes

gpgekko: As far as I can see in the preview it looks and functions like it should. It is also written in this style at [line 94](https://github.com/marionettejs/backbone.marionette/pull/3590/files#diff-9175f0f44dd039f2958d365391ec492eR94), which I didn't touch.
However, if you want me to do it different, I will.

paulfalgout: Ah yes a registered link name. You are correct!

PR #3597, opened 2018-05-07
typo fix in marionette functions doc

birbilis:
f is mentioned in the text, plus using foo would confuse the user anyway

### Proposed changes
 -
 -
 -

Link to the issue:


paulfalgout:
This is getting addressed as part of https://github.com/marionettejs/backbone.marionette/pull/3611

PR #3603, opened 2018-05-10
Rename BackboneViewMixin to Events

### LINKED ISSUE #3567: Rename BackboneViewMixin to Events
https://github.com/marionettejs/backbone.marionette/pull/3566#issuecomment-350265001
tl;dr
It is a mixin for `triggerMethod`.  I doubt that will change. `Marionette.Events` is one of the APIs that will be necessary to reduce or remove dependency on backbone.

blikblum:
Fixes #3567


paulfalgout:
This is good and simple.  I'm gonna merge it now and any complaints we can revert if necessary.

Issue #3594, opened 2018-05-06
Views addChildView'd at index 0 may get added at end


paulfalgout:
https://github.com/marionettejs/backbone.marionette/blob/next/src/collection-view.js#L655

The addedViews perf is incorrectly checking for `!index` which would include `0`, but in the case of a `0` index it should _not_ create the `_addedViews` array

Issue #3593, opened 2018-05-06
addChildView should support an array of views


paulfalgout:
Currently `addChildView` only handles a single view.  We could modify this to add multiple views which could be done more performant internally and with an easier API.

taburetkin:
```
addChildViews(children = []){

	if (!children.length) { return; }

	let awaitingRender = false;

	while(items.length){

		let item = items.shift();

		if (!_.isArray(args)) {
			args = [args, { index: 0 }];
		}

		//supporting old style addChildView
		let [ view, index, options = {} ] = args;

		if (_.isObject(index)) {
			options = index;			
		} else if(index != null && !('index' in options)) {
			options.index = index;
		}

		options.preventRender = !!items.length;

		if (!isView(view)) { continue; }

		this.addChildView(view, options);
		awaitingRender = options.preventRender;

	}

	if (awaitingRender) {
		this.sort();
	}

}
```

with this implementation i think that
this one
```
this.addChildView( view1, 0 )
this.addChildView( view2, 0 )
this.addChildView( view3, 0 )
```
will be equal to this
```
this.addChildViews([ 
 [view1 , 0],
 [view2 , 0],
 [view3 , 0],
]);
```

paulfalgout:
this.addChildViews([ 
 [view1 , 3],
 [view2 , 2],
 [view3 , 1],
]);

taburetkin:
will be equal to this
```
this.addChildView(view1, 3)
this.addChildView(view2, 2)
this.addChildView(view3, 1)
```
main thing that they both will behave the same

taburetkin:
and in my opinion its bad when you add views by one million years
and now, when the future is almost here and you can add them in one go
you suddenly found that behavior is different, not as usual...
its just a thoughts

paulfalgout:
In that case
```
this.addChildView(view1, 3, { preventRender: true });
this.addChildView(view2, 2, { preventRender: true });
this.addChildView(view3, 1);
```
is pretty good without needing to overcomplicate the library.

taburetkin:
and the one more thing
if there will be own indexer implemented as i offer
then individual indexes became more usefull because with own indexer you can order your views as you wish

taburetkin:
if this is pretty good
```
this.addChildView(view1, 3, { preventRender: true });
this.addChildView(view2, 2, { preventRender: true });
this.addChildView(view3, 1);
```
why you start this issue at all?

[...]

PR #3605, opened 2018-05-14
Prevent sort/filter when adding with an index

paulfalgout:
Fixes https://github.com/marionettejs/backbone.marionette/issues/3594

This work also lessens the need for https://github.com/marionettejs/backbone.marionette/issues/3593

Essentially if you're adding a view at a defined index, it skips sort and filter.  This is a small but important change.  It allows the following:
```js
// will re-sort and re-filter
myCollectionView.addChildView(new MyView());

// won't re-sort or re-filter
myCollectionView.addChildView(new MyView(), 0); 
myCollectionView.addChildView(new MyView(), 5); 
myCollectionView.addChildView(new MyView(), myCollectionView.children.length); 
```
The user will still need to handle the added views for any subsequent sort or filter, but before this change if a user wanted to insert a view at a particular location they'd either have to temporarily disable sorting and filtering or adding at the index was not useful since sorting and filtering were going to occur anyways.  It's worth noting that by default the CollectionView sorts by the `collection` so the `addChildView` index currently appears to not work by default until the sorting was fully understood.

This also makes something like `_.each(viewsArray, view => { myCv.addChildView(view, this.children.length); });` more reasonable as it won't re-sort / re-filter for each view.  Supporting an array could still be moderately more performant, by potentially utilizing a single append, but seemingly less important, particularly if children are added on the collectionview render prior to the collectionview's insertion in the DOM.

There's also one trivial change removing the `_.each` duplication by reusing `_removeChildViews`

I think this should be included with v4 as it could be considered breaking.  It doesn't need to make the alpha release necessarily.  @marionettejs/marionette-core

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -642,12 +642,19 @@ const CollectionView = Backbone.View.extend({
       this.render();
     }
 
+    const hasIndex = (typeof index !== 'undefined');
+
     // Only cache views if added to the end
-    if (!index || index >= this._children.length) {
+    if (!hasIndex || index >= this._children.length) {
       this._addedViews = [view];
     }
     this._addChild(view, index);
-    this.sort();
+
+    if (hasIndex) {
+      this._renderChildren();
+    } else {
+      this.sort();
+    }
 
     return view;
   },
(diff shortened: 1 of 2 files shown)


paulfalgout:
Two thoughts about 4.x  1. I'm anticipating some additions during the beta of v4, so there will be some changes between now and production version.  2. If we don't add this, should we document the caveats of it not being in there in v4.0 or can update the docs without having to add all of the info of caveats.

bazineta:
@paulfalgout I think in this case it's much easier to include the fix than to document why it's absent, really.

paulfalgout:
I'm very close to have the docs PR.  It's... a doozy.. but I'm feeling good about it.  Having this lull in activity has made space to read every line against the code.  I'm sure there'll be issues, but I think it'll be more cohesive.. expect it shortly.  I doubt I'll run into anything else like this, but maybe.

Issue #3608, opened 2018-05-31
PeerDependencies in packageJSON causing npm install errors


tirthaguha:
### PeerDependencies in packageJSON causing npm install errors

 1. Step to reproduce
     ```npm install underscore//installs v1.9.0``` 
     ```npm install backbone //installs v1.3.3```
     ```npm install backbone.marionette //installs 3.5.1```
     
>Throws annoying error ```-- UNMET PEER DEPENDENCY underscore@1.9.0```
     
 
 2. Proposed Solution
    in package json, change the peer dependency section
    **FROM**
    ```
    "peerDependencies": {
      "backbone": "~1.3.3",
      "underscore": "~1.8.3"
    },
    ```
    **TO**
    ```
    "peerDependencies": {
      "backbone": "^1.3.3",
      "underscore": "^1.8.3"
    },
    ```
### Expected behavior

There should not be any npm install error

### Actual behavior

```-- UNMET PEER DEPENDENCY underscore@1.9.0```

### Environment

1. Marionette version: 3.5.1
2. Backbone version: 1.3.3
3. Underscore version: 1.9.0
4. NodeJS 6.x, NPM 3.x

gam6itko:
It also can be
"peerDependencies": {
  "backbone": "^1.3",
  "underscore": "^1.8"
},

tirthaguha:
suites me. :+1:

paulfalgout:
Previously `Mn` had locked the versions because Backbone does not follow semver, but Marionette does.. so allowing for a breaking `1.9.x` could break an install that uses `^1.8.3`  This is still possible, but I feel less concerned about it if they're defined as peerDependencies.  That said.. unmet peer dependency is a warning is it not?

tirthaguha:
@paulfalgout that was the first thought. however, by not allowing ^ we are kind of forcing the newer applications to run on old underscoreJS version. 
Agree that peerDependency is an warning, but it is annoying enough, when you have a huge ecosystem running on marionette, and the build machines puke an warning with every build or every component.

paulfalgout:
So if a newer version of underscore is breaking and requires a fix in Marionette, what would we do?

Let's layout a scenario
Mn at v5.5.1 supports underscore ^2.1.1
Underscore 2.2.0 is released and unfortunately includes a breaking change such that v5.5.1 doesn't work without updates.  Underscore does not consider this to be a breaking semver change so it is up to Marionette to fix it.
[...]

PR #3609, opened 2018-05-31
Next

tirthaguha:
### Proposed changes
 - changes in peerDependency in packageJSON to allow higher versions of underscore.
 -
 -

Link to the issue:
https://github.com/marionettejs/backbone.marionette/issues/3608


JSteunou:
I'm sorry @tirthaguha but I'm against this change for all reasons @paulfalgout perfectly exposed already. We cannot put users at risk by allowing underscore minor updates that could be breaking. That would force Marionette to rush publish a fix on peerDependencies with a fixed range but users build could break until the new version is published.

Moreover if there was more than one Marionette major versions until this happens, that would force us to publish a fix on all Marionette major versions or leave users that are not on the latest major version in trouble.

PR #3614, opened 2018-07-03
Add yarn.lock file to .npmignore

blikblum:
### Proposed changes
 - Removes yarn.lock file from distributed package


paulfalgout:
why would we do this?  The lock file allows for managing updates to build tools

blikblum:
The lock file has no utility in the package uploaded to npm while it increases its size considerably

paulfalgout:
ah yep.. misread as `.gitignore` at first

PR #3615, opened 2018-07-04
Allow for use of Underscore 1.9.x.

adriantrunzo:
### Proposed changes
 - Use underscore >= 1.83 < 2 as a peer dependency.

Underscore.js 1.9.1 has been released, but currently the peer dependency set in Marionette limits Underscore.js to 1.8.3. The new version only contains additions and improvements, with seemingly no breaking changes: https://underscorejs.org/#changelog. Backbone lists its underscore dependency as >= 1.8.3: https://github.com/jashkenas/backbone/blob/master/package.json#L16, so it seems like Marionette should handle the update with no issues. All of the tests seem to pass with 1.9.1.

This update would allow for projects using the updated version of Underscore.js without having Marionette require a separate, older version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "peerDependencies": {
     "backbone": "~1.3.3",
-    "underscore": ">= 1.8.3 < 2"
+    "underscore": ">= 1.8.3 <= 1.9.x"
   },
   "devDependencies": {
     "babel-core": "6.25.0",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "peerDependencies": {
     "backbone": "~1.3.3",
-    "underscore": "~1.8.3"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -50,7 +50,7 @@
   },
   "peerDependencies": {
     "backbone": "~1.3.3",
-    "underscore": "~1.8.3"
+    "underscore": ">= 1.8.3 < 2"
```
paulfalgout: I think this should be `< 1.9.x` underscore does not follow semver so `1.10` can be breaking with no warning.

adriantrunzo: Ok, I will make that change.

adriantrunzo: Maybe `<= 1.9.x` will have clearer intentions yet?

paulfalgout: Oh yep

PR #3624, opened 2018-08-10
Remove stray backtick in upgrade guide

gpgekko:
Remove a stray backtick that was putting a random sentence in a code block.


gpgekko:
There is also a problem in the Recommended changes section (around line 230), where it breaks the website but does not break in the GitHub preview. I suspect it might be the unescaped asterix, but I don't have a Linux machine available at the moment to experiment / fix it.

PR #3630, opened 2018-08-29
Added addChildView preventRender logic

taburetkin:
### Proposed changes
 - preventRender logic for multiple adding child views
 - options argument `(view, index, options)` or `(view, options)`
 - possible options: index, preventRender

Link to the issue:

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -652,7 +652,7 @@ const CollectionView = Backbone.View.extend({
       this.render();
     }
 
-    let hasIndex = (typeof index !== 'undefined');
+    const hasIndex = (typeof index !== 'undefined');
 
     // Only cache views if added to the end and there is no awaiting views
     if (!preventRender && !this._addingMultipleViews && (!hasIndex || index >= this._children.length)) {
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -660,7 +660,6 @@ const CollectionView = Backbone.View.extend({
 
     this._addChild(view, index);
 
-    // `preventRender` passed, so we are waiting another `addChildView`
     if (options.preventRender) {
       this._hasUnrenderedViews = true;
       return view;
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -634,7 +634,17 @@ const CollectionView = Backbone.View.extend({
   },
 
   // Render the child's view and add it to the HTML for the collection view at a given index, based on the current sort
-  addChildView(view, index) {
+  addChildView(view, index, options = {}) {
+    if (_.isObject(index)) {
+      options = index;
+    }
+
+    let preventRender = options.preventRender;
+    //if options has defined index we should use it
+    if (options.index != null) {
+      index = options.index;
+    }
+
     if (!view || view._isDestroyed) {
       return view;
     }
@@ -643,13 +653,40 @@ const CollectionView = Backbone.View.extend({
       this.render();
     }
 
-    const hasIndex = (typeof index !== 'undefined');
+    //const -> let, because value may be changed later
+    let hasIndex = (typeof index !== 'undefined');
+
+    if (preventRender) {
+      //setting flag that we are waiting addChildView without `preventRender`
+      this._addingMultipleViews = true;
+
+      //if we added atleast one indexed view _addedViews perf should be skiped
+      if (!this._addingMultipleIndexedViews && hasIndex) {
+        this._addingMultipleIndexedViews = true;
+      }
+    }
 
     // Only cache views if added to the end
     if (!hasIndex || index >= this._children.length) {
       this._addedViews = [view];
     }
     this._addChild(view, index);
+
+    //if there was atleast one indexed view we should act as there an index exist
+    hasIndex = this._addingMultipleIndexedViews || hasIndex;
+
+    if (preventRender) {
+      //preventRender passed, so we are waiting another addChildView
+      return view;
+    } else {
+      //if we added atleast one indexed view _addedViews perf should be skiped
+      if (this._addingMultipleIndexedViews) {
+        delete this._addedViews;
+      }
+      //clearing after
+      delete this._addingMultipleViews;
+      delete this._addingMultipleIndexedViews;
+    }
 
     if (hasIndex) {
       this._renderChildren();
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -507,6 +507,12 @@ const CollectionView = Backbone.View.extend({
   },
 
   _renderChildren() {
+    // If there are unrendered views prevent add to end perf
+    if (this._hasUnrenderedViews) {
+      delete this._addedViews;
+      delete this._hasUnrenderedViews;
+    }
+
     const views = this._addedViews || this.children._views;
 
     this.triggerMethod('before:render:children', this, views);
@@ -635,36 +641,37 @@ const CollectionView = Backbone.View.extend({
 
   // Render the child's view and add it to the HTML for the collection view at a given index, based on the current sort
   addChildView(view, index, options = {}) {
+    if (!view || view._isDestroyed) {
+      return view;
+    }
+
     if (_.isObject(index)) {
       options = index;
     }
 
-    const preventRender = options.preventRender;
-
-    //if options has defined index we should use it
-    index = options.index || index;
-
-    if (!view || view._isDestroyed) {
-      return view;
+    // If options has defined index we should use it
+    if (options.index != null) {
+      index = options.index;
     }
 
     if (!this._isRendered) {
       this.render();
     }
 
-    const hasIndex = (typeof index !== 'undefined');
-
-    // Only cache views if added to the end and there is no awaiting views
-    if (!preventRender && !this._addingMultipleViews && (!hasIndex || index >= this._children.length)) {
-      this._addedViews = [view];
-    }
     this._addChild(view, index);
 
-
-    //preventRender passed, so we are waiting another addChildView
-    if (preventRender) {
-      this._addingMultipleViews = true;
+    // `preventRender` passed, so we are waiting another `addChildView`
+    if (options.preventRender) {
+      this._hasUnrenderedViews = true;
       return view;
+    }
+
+    const hasIndex = (typeof index !== 'undefined');
+    const isAddedToEnd = !hasIndex || index >= this._children.length;
+
+    // Only cache views if added to the end and there is no unrendered views
+    if (isAddedToEnd && !this._hasUnrenderedViews) {
+      this._addedViews = [view];
     }
 
     if (hasIndex) {
@@ -672,7 +679,7 @@ const CollectionView = Backbone.View.extend({
     } else {
       this.sort();
     }
-    delete this._addingMultipleViews;
+
     return view;
   },
 
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -656,14 +656,9 @@ const CollectionView = Backbone.View.extend({
     //const -> let, because value may be changed later
     let hasIndex = (typeof index !== 'undefined');
 
-    if (preventRender) {
-      //setting flag that we are waiting addChildView without `preventRender`
-      this._addingMultipleViews = true;
-
-      //if we added atleast one indexed view _addedViews perf should be skiped
-      if (!this._addingMultipleIndexedViews && hasIndex) {
-        this._addingMultipleIndexedViews = true;
-      }
+    //if we added atleast one indexed view _addedViews perf should be skiped
+    if (preventRender && hasIndex) {
+      this._addingMultipleIndexedViews = true;
     }
 
     // Only cache views if added to the end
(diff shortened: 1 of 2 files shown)


src/collection-view.js
```diff
@@ -634,7 +634,17 @@ const CollectionView = Backbone.View.extend({
   },
 
   // Render the child's view and add it to the HTML for the collection view at a given index, based on the current sort
-  addChildView(view, index) {
+  addChildView(view, index, options = {}) {
+    if (_.isObject(index)) {
+      options = index;
+    }
+
+    let preventRender = options.preventRender;
```
paulfalgout: all variables in `Mn` should be `const` unless otherwise necessary

src/collection-view.js
```diff
@@ -634,7 +634,17 @@ const CollectionView = Backbone.View.extend({
...
+    //if options has defined index we should use it
+    if (options.index != null) {
```
paulfalgout: following conventions elsewhere this should be `index = options.index || index;`

src/collection-view.js
```diff
@@ -643,14 +653,37 @@ const CollectionView = Backbone.View.extend({
       this.render();
     }
 
-    const hasIndex = (typeof index !== 'undefined');
+    //const -> let, because value may be changed later
```
paulfalgout: no need to comment here as `let` should indicate this

src/collection-view.js
```diff
@@ -643,14 +653,37 @@ const CollectionView = Backbone.View.extend({
...
+    let hasIndex = (typeof index !== 'undefined');
+
+    //if we added atleast one indexed view _addedViews perf should be skiped
```
paulfalgout: I don't believe that this is working. if 2 views are added `preventRender` without an index, only the first view will be attached. 
Additionally if someone adds a view with` preventRender` without an index and then a collection updates, the views added by the collection will not be attached and only the `_addedViews` will be attached.

I think the best approach is that `preventRender` should also prevent the `_addedViews` perf as the next render is not known.

test/unit/collection-view/collection-view-children.spec.js
```diff
@@ -204,6 +204,93 @@ describe('CollectionView Children', function() {
       this.sinon.spy(myCollectionView, 'sort');
     });
 
+    describe('when called with preventRender option', function() {
+
+      beforeEach(function() {
+        this.sinon.spy(myCollectionView, '_renderChildren');
```
paulfalgout: Mn does not test private functions.  So we need to test public API / hooks _or_ in some cases public to the module, but not necessarily exposed.  ie: `children._add` is considered public to the module, but not exposed for end users.  `_renderChildren` and flags like `_addingMultipleIndexedViews` should not be tested.  We should be free to modify implementation details without breaking tests. Tests should only break if a major version of Marionette is required.

taburetkin: yes, i was indoubt about this.
but can not realize yet should i just delete this two test or replace it with some else
for renderChildren maybe its possibe just to check onRenderChildren
but what about _addingMultipleIndexedViews 
does this mean that implementation is not good enough?

paulfalgout: correct I think testing for render events is fine.  Regarding the flag, it's important we test what the flag is doing and not how the flag is implemented.  Whether the flag is set or deleted is unimportant.  We need to test the scenarios when using `addChildView` that setting or deleting the flag cover.  However as mentioned elsewhere I think we should remove the `_addedViews` perf when using `preventRender` as if a render is not immediately called, subsequent other renders may break.  We _can_ however after this is implemented probably account for the perf bonus if we end up supporting an array of views.

taburetkin: i think that this can solve this issue
```
    // Only cache views if added to the end
    if (!hasIndex || index >= this._children.length) {
      this._addedViews || (this._addedViews = []);
      this._addedViews.push(view);
    }
```
last time you tell me that _addedViews not intended for continuos operations
but maybe in this scenario its possible to do it like this?

paulfalgout: I think attempting to maintain `_addedViews` is a very very bad idea with many caveats.
```js
myCollectionView.addChildView(new View(), { preventRender: true }); // this._addedViews exists with new view
myCollectionView.collection.add(fooModel); // this._addedViews exists so new view gets attached, fooModel's view does not
myCollectionView.addChildView(new View()); // new view gets added immediately.. still no fooModel view..
```

test/unit/collection-view/collection-view-children.spec.js
```diff
@@ -204,6 +204,78 @@ describe('CollectionView Children', function() {
...
+          .to.be.calledOnce.and.calledWith(myCollectionView, addView);
+      });
+
+    });
+
+    describe('when called with an index in options', function() {
+      const addIndex = 1;
+      beforeEach(function() {
+        myCollectionView.viewComparator = false;
+        myCollectionView.addChildView(addView, 0, { preventRender: true, index: 1 });
```
paulfalgout: probably should use `addIndex` here too

test/unit/collection-view/collection-view-children.spec.js
```diff
@@ -204,6 +204,78 @@ describe('CollectionView Children', function() {
...
+        expect(myCollectionView.onBeforeAddChild)
+          .to.be.calledOnce.and.calledWith(myCollectionView, addView);
+      });
+
+    });
+
+    describe('when called with an index in options', function() {
+      const addIndex = 1;
+      beforeEach(function() {
+        myCollectionView.viewComparator = false;
```
paulfalgout: why does the `viewComparator` need to be set here

test/unit/collection-view/collection-view-children.spec.js
```diff
@@ -204,6 +204,78 @@ describe('CollectionView Children', function() {
...
+          .to.have.been.calledOnce.and.calledWith(addView, addIndex);
+      });
+
+    });
+
+    describe('when called without preventRender after preventReder calls', function() {
+      const addViewIndex = 1;
+      beforeEach(function() {
+        const addView2 = new View({ template: _.noop });
+        myCollectionView.viewComparator = false;
```
paulfalgout: why is `viewComparator` set here?

src/collection-view.js
```diff
@@ -645,18 +654,25 @@ const CollectionView = Backbone.View.extend({
 
     const hasIndex = (typeof index !== 'undefined');
 
-    // Only cache views if added to the end
-    if (!hasIndex || index >= this._children.length) {
+    // Only cache views if added to the end and there is no awaiting views
+    if (!preventRender && !this._addingMultipleViews && (!hasIndex || index >= this._children.length)) {
```
paulfalgout: this still will not work because renders can occur separate from calling `addChildView`  any render that occurs apart from `addChildView` if `this._addedViews` exists will be broken.

taburetkin: i think i not follow
_addedViews will be used only if this is call with `preventRender: false` and there is no awaiting views. in other cases _addedViews will not be used at all
or i do not see something?

test/unit/collection-view/collection-view-children.spec.js
taburetkin: i've done requested test changes (not commited yet)
should i add additional tests or its enough?

src/collection-view.js
blikblum: If options.index === 0 will fail

src/collection-view.js
paulfalgout: ah I see.  yes I think you're mostly correct.

There is an unexpected side-effect though.  If `preventRender` was called and `_addingMultipleViews` is set, and then the collection adds views to the end it will trigger the `_addedViews` perf and only render those views without the unrendered view, which is not good.  I think maybe this flag should change to `_hasUnrenderedViews`.  Additionally the flag needs to be unset if the collection is re-rendered by some other way as well.

So I think perhaps the first line of `_renderChildren` should be:
```
if (this._hasUnrenderedViews) {
  delete this._addedViews;
  delete this._hasUnrenderedViews;
}
```

Had an idea though. on collection update `_addChild` happens first, which makes sense here as well.
This gets simpler if it is:
- `_addChild`
- `if preventRender return`
- `_addedViews =`

src/collection-view.js
paulfalgout: ah yes good catch.  My bad.  It must be as it was.

src/collection-view.js
taburetkin: did you mean this case?
```
const test = new ExtendedAddChildView({
	collection,
	childView: ModelView,
	viewComparator: false,
	onRender(){
		let preventRender = true;
		this.addChildView(new CustomView({ text:'c1' }), { preventRender });
    this.collection.add({id:4});
		this.addChildView(new CustomView({ text:'c2' }), { preventRender });
		this.addChildView(new CustomView({ text:'c3' }), 3, { preventRender });
		this.addChildView(new CustomView({ text:'c4' }), undefined, { preventRender });
		this.addChildView(new CustomView({ text:'c5' }));
	},
	
});
```
if so, then all views will be added
i think i am not understand the case
here is a codepen for this pull request: https://codepen.io/dimatabu/pen/jvMmxp

paulfalgout: Results show `2, 1, 4`
```js
	onRender(){
		this.addChildView(new CustomView({ text:'c1' }), { preventRender: true });
    this.collection.set([{id:1},{id:2},{id:4}]);
}
```
Results show `c1, 2, 1`
```js
	onRender(){
		this.addChildView(new CustomView({ text:'c1' }), { preventRender: true });
    this.collection.set([{id:1},{id:2}]);
}
```

paulfalgout: that's actually a _different_ issue than earlier mentioned.. still looking for other potential issues

taburetkin: yes, but there should call with preventRender: false
i mean, if you do:
addChildView (prevent:true)
addChildView (prevent:true)
... here something happens, say collection receive new models, and renderchildrens occurs
addChildView(prevent:true)
addChildView()
done. here renderChildren occurs and all views in own places
am i right?

taburetkin: i mean, if you add a view with preventRender it means that you must add another view, the final one without `preventRender: true` for putting views in a place

taburetkin: but maybe you want to tell that collection update event should be treaten as finalize in adding multiple views?

paulfalgout: I'm completely against adding this if it requires that you call the function a particular way at the end.  That makes no sense to me.
What should happen is that no matter what else causes a render the behavior should be the same.
Consider:  (also note the order of things)
```js
const ExtendedAddChildView = Mn.CollectionView.extend({
  addChildView(view, index, options = {}) {
    if (!view || view._isDestroyed) {
      return view;
    }
    if (_.isObject(index)) {
      options = index;
    }
    // If options has defined index we should use it
    if (options.index != null) {
      index = options.index;
    }

    this._addChild(view, index);

    // `preventRender` passed, so we are waiting another `addChildView`
    if (options.preventRender) {
      this._hasUnrenderedViews = true;
      return view;
    }
    
    if (!this._isRendered) {
      this.render();
    }
    
    const hasIndex = (typeof index !== 'undefined');
    const isAddedToEnd = !hasIndex || index >= this._children.length;

    // Only cache views if added to the end and there is no unrendered views
    if (isAddedToEnd && !this._hasUnrenderedViews) {
      this._addedViews = [view];
    }
    
    if (hasIndex) {
      this._renderChildren();
    } else {
      this.sort();
    }
    
    return view;
  },
  _renderChildren() {
    // If there are unrendered views prevent add to end perf
    if (this._hasUnrenderedViews) {
      delete this._addedViews;
      delete this._hasUnrenderedViews;
    }
    Mn.CollectionView.prototype._renderChildren.call(this);
  }
});
```

taburetkin: i've tried that
but now two test fails, trying to resolve this
*when the collectionView is not attached*
```
      it('should trigger "before:render" event on the childView', function() {
        expect(childView.onBeforeRender).to.have.been.calledOnce.and.calledWith(childView);
      });

      it('should trigger "render" event on the childView', function() {
        expect(childView.onRender).to.have.been.calledOnce.and.calledWith(childView);
      });
```

taburetkin:
```
    if (!this._isRendered) {
      this.render();
    }
    this._addChild(view, index);

    // `preventRender` passed, so we are waiting another `addChildView`
    if (options.preventRender) {
      this._hasUnrenderedViews = true;
      return view;
    }
```
otherwise we get two filed test in render child lifecocle mentioned above

paulfalgout: ah yep.  that makes sense

src/collection-view.js
```diff
@@ -634,22 +640,39 @@ const CollectionView = Backbone.View.extend({
...
+      index = options.index;
+    }
+
     if (!this._isRendered) {
       this.render();
     }
 
+    this._addChild(view, index);
+
+    // `preventRender` passed, so we are waiting another `addChildView`
```
paulfalgout: I don't think we should add opinion here in this comment.  There may be reasonable use cases we're not seeing apart from this. It simply means we're preventing `_renderChildren` as it does with `setComparator` and `setFilter`

taburetkin: so, removing comment?
setComparator and setFilter has no comments

paulfalgout: either way for my tastes.  The var used is descriptive.

taburetkin: no problem :)

paulfalgout:
2nd 👍 ?

paulfalgout:
Note: Needs to be "Squash and merge"d

Issue #3610, opened 2018-06-06
CollectionView does not consider the childViewContainer when rendering empty


paulfalgout:
For v4 `CollectionView` can render children inside of a selector on a template.

However the `emptyRegion` still uses the `el` of the view and not the container.

This is complicated as the empty region is created once at instantiation and there's not a reasonable way to currently update the selector on render.

adriantrunzo:
We just ran into this issue while rendering filterable tables:

```
const MyTableView = CollectionView.extend({
  tagName: 'table',

  childView: MyRowView,
  childViewOptions: { ... },
  childViewContainer: 'tbody',

  emptyView: MyEmptyView,
  ...
})
```

If a particular filter hides all children, the emptyView will show at the `el` level and wipe out the `childViewContainer`. Thus, after another call to `filter()` that results in views to be shown, the collection view cannot actually attach the children and the view is just blank.

Can the empty region just use the `childViewContainer` element if it is specified? That seems the most logical behavior. Further flexibility could be achieved by allowing for an `emptyViewContainer` property that would be the empty region `el`. Similar to `childViewOptions` and `emptyViewOptions`, if `emptyViewContainer` is not defined, the empty region would fallback to `childViewContainer` and then to the view `el`.

I am happy to help out with a pull request if we can agree on a good solution.

adriantrunzo:
Could CollectionView just throw an error in that case? When would it be useful to nest those containers?

paulfalgout:
nested containers will increase the complexity and performance of the collectionview.  I can't see a good use case for it that couldn't be handled better by a regular view/region and collectionview.  So I don't see the point of an error either if `emptyViewContainer` !== `childViewContainer` there are significant problems as we have currently.

avipinto:
Till 4.0.1 is out, "fixed" this bug the following way:
```   emptyView: function() {
    if (!this.fixedEmptyOnChildViewContainer) {
            this.fixedEmptyOnChildViewContainer = true;
            this._emptyRegion = new Mn.Region({ el: this.$(this.childViewContainer).get(0), replaceElement: false });
            this._emptyRegion._parentView = this;
        }
        return EmptyView;//my regular emptyView
    },
```

PR #3641, opened 2018-11-15
Introduce setElement to Region

paulfalgout:
I believe it is a first non-breaking step towards https://github.com/marionettejs/backbone.marionette/pull/3620 and it provides a solution path for https://github.com/marionettejs/backbone.marionette/issues/3610 where the `CollectionView` can update the `emptyRegion`'s `el` to match the `childViewContainer` at `render` without compromising the region instance the user may have already `getEmptyRegion`'d.

### DIFF
--- a/src/region.js
+++ b/src/region.js
@@ -30,15 +30,7 @@ const Region = function(options) {
   // Handle when this.el is passed in as a $ wrapped element.
   this.el = this.el instanceof Backbone.$ ? this.el[0] : this.el;
 
-  if (!this.el) {
-    throw new MarionetteError({
-      name: classErrorName,
-      message: 'An "el" must be specified for a region.',
-      url: 'marionette.region.html#additional-options'
-    });
-  }
-
-  this.$el = this.getEl(this.el);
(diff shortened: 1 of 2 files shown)


src/region.js
```diff
@@ -95,6 +87,49 @@ _.extend(Region.prototype, CommonMixin, {
...
+        url: 'marionette.region.html#additional-options'
+      });
+    }
+
+    // Check if `$el` is already correct
+    if (this.$el && this.$el[0] === this.el) { return; }
+
+    this.$el = this.getEl(this.el);
+
+    if(this.$el.length) {
```
paulfalgout: I believe ideally in a future release of Mn, Regions should no longer query the DOM and we should require that region instantiation and setElement include existing DOM elements.  querying the DOM requires awkward `parentEl` context or it can produce unwanted results that are difficult to track down.  View's should query the DOM and pass the element instead of the selector string.  This would also allow us to instantiate regions at the moment we need them instead of prior to anyone accessing or using them.

PR #3650, opened 2019-01-24
Preventing showing of shown view in a region

taburetkin:
### Proposed changes
 - Its possible now to show one view in multiple regions and this can lead to unexpected behavior. 
 - This fix checks if a view prerendered on the backend or somewhere else and if the view is Attached and not prerendered then error occurs

Link to the issue:
https://codepen.io/dimatabu/pen/bzVbvY?editors=1010

note:
throwed error contains unexistent url to the documentation and should be fixed

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -499,6 +499,8 @@ const CollectionView = Backbone.View.extend({
       view._isAttached = false;
       view.triggerMethod('detach', view);
     }
+
+    view._isShown = false;
   },
 
   // Override this method to change how the collectionView detaches a child view
@@ -538,8 +540,8 @@ const CollectionView = Backbone.View.extend({
 
     _.each(views, view => {
       renderView(view);
-      // corresponds that this view is not a pre-rendered one
-      view._attachedTo = true;
+      // corresponds that view is shown in a Region or CollectionView
+      view._isShown = true;
       this.Dom.appendContents(elBuffer, view.el, {_$contents: view.$el});
     });
 
@@ -645,6 +647,14 @@ const CollectionView = Backbone.View.extend({
   addChildView(view, index, options = {}) {
     if (!view || view._isDestroyed) {
       return view;
+    }
+
+    if (view._isShown) {
+      throw new MarionetteError({
+        name: classErrorName,
+        message: 'View is already shown in a Region or CollectionView',
+        url: 'marionette.collectionview.html#region-viewAlreadyShown'
+      });
     }
 
     if (_.isObject(index)) {
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -538,6 +538,8 @@ const CollectionView = Backbone.View.extend({
 
     _.each(views, view => {
       renderView(view);
+      // corresponds that this view is not a pre-rendered one
+      view._attachedTo = true;
       this.Dom.appendContents(elBuffer, view.el, {_$contents: view.$el});
     });
 
--- a/src/region.js
+++ b/src/region.js
@@ -300,13 +300,12 @@ _.extend(Region.prototype, CommonMixin, {
     delete this.currentView;
 
     if (!view._isDestroyed) {
-      view._isShown = false;
-
       if (shouldDestroy) {
         this.removeView(view);
       } else {
         this._detachView(view);
       }
+      view._isShown = false;
       this._stopChildViewEvents(view);
     }
 
--- a/src/region.js
+++ b/src/region.js
@@ -71,11 +71,11 @@ _.extend(Region.prototype, CommonMixin, {
 
     if (view === this.currentView) { return this; }
 
-    if (view._isAttached && view._attachedTo) {
+    if (view._isShown) {
       throw new MarionetteError({
         name: classErrorName,
-        message: 'A "view" can be attached to the DOM only if it prerendered on backend',
-        url: 'marionette.collectionview.html#region-showAttachedView'
+        message: 'View is already shown in a Region or CollectionView',
+        url: 'marionette.collectionview.html#region-viewAlreadyShown'
       });
     }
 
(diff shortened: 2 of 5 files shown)


src/collection-view.js
```diff
@@ -538,6 +538,8 @@ const CollectionView = Backbone.View.extend({
 
     _.each(views, view => {
       renderView(view);
+      // corresponds that this view is not a pre-rendered one
```
paulfalgout: I don't think that the comment is necessarily true.  This flag indicates that the view is attached to something else in marionette.  where as `_isAttached` only indicates that the view is attached to the DOM.  All prerendered views shown in a region or collection view will also have `_attachedTo = true`

src/region.js
```diff
@@ -71,6 +71,14 @@ _.extend(Region.prototype, CommonMixin, {
 
     if (view === this.currentView) { return this; }
 
+    if (view._isAttached && view._attachedTo) {
+      throw new MarionetteError({
+        name: classErrorName,
+        message: 'A "view" can be attached to the DOM only if it prerendered on backend',
```
paulfalgout: This message is incorrect. Pre-rendered DOM can be handed to the view by the frontend as well.  Also whether or not this view `_isAttached` is irrelevant here.  If you are showing the same view in two regions and the view with the regions isn't attached you're going to have the same leaky problems.  the only check needs to be `view._attachedTo`

test/unit/region.spec.js
```diff
@@ -484,6 +487,38 @@ describe('region', function() {
...
+    let region;
+    let anotherRegion;
+    let collectionView;
+
+    beforeEach(function() {
+      this.setFixtures('<div id="reg1"></div><div id="reg2"></div><div id="cv"></div><div id="view"></div>')
+      region = new Region({ el: '#reg1' });
+      anotherRegion = new Region({ el: '#reg2' });
+      collectionView = new CollectionView({ el: '#cv' });
+      testView = new View({ el: '#view', template: _.noop });
```
paulfalgout: no need to set the template here, let's leave it off.

src/region.js
```diff
@@ -71,6 +71,14 @@ _.extend(Region.prototype, CommonMixin, {
 
     if (view === this.currentView) { return this; }
 
+    if (view._isAttached && view._attachedTo) {
+      throw new MarionetteError({
```
paulfalgout: this error should also happen in `addChildView` for `CollectionView`

test/unit/region.spec.js
```diff
@@ -3,6 +3,7 @@ import Backbone from 'backbone';
 import Events from '../../src/mixins/events';
 import Region from '../../src/region';
 import View from '../../src/view';
+import CollectionView from '../../src/collection-view';
```
paulfalgout: We should also test for collectionview functionality in the relevant collectionview spec file which is probably with `addChildView` and `detachChildView`

test/unit/region.spec.js
```diff
@@ -340,6 +341,8 @@ describe('region', function() {
         $parentEl = region.$el.parent();
         regionHtml = $parentEl.html();
         region.replaceElement = true;
+        //have to redefine view, because its already in a region and can not be tested
+        view = new MyView();
```
paulfalgout: I wonder if this is needed because `_attachedTo` is never set to false

test/unit/region.spec.js
```diff
@@ -484,6 +487,38 @@ describe('region', function() {
...
+
+    beforeEach(function() {
+      this.setFixtures('<div id="reg1"></div><div id="reg2"></div><div id="cv"></div><div id="view"></div>')
+      region = new Region({ el: '#reg1' });
+      anotherRegion = new Region({ el: '#reg2' });
+      collectionView = new CollectionView({ el: '#cv' });
+      testView = new View({ el: '#view', template: _.noop });
+    });
+
+    it('should not throw an error if view is a prerendered one', function() {
```
paulfalgout: along with the changes above, I don't think pre-rendered will matter, so this test can be removed

src/region.js
taburetkin: what about this?
    if (view._isShown) {
      throw new MarionetteError({
        name: classErrorName,
        message: 'Given view is already shown in some marionette Region or CollectionView',
        url: 'marionette.collectionview.html#region-shownView'
      });
    }

taburetkin: > This message is incorrect. Pre-rendered DOM can be handed to the view by the frontend as well. Also whether or not this view `_isAttached` is irrelevant here. If you are showing the same view in two regions and the view with the regions isn't attached you're going to have the same leaky problems. the only check needs to be `view._attachedTo`

https://github.com/marionettejs/backbone.marionette/blob/37bb058441b6d045d93e7051f453cce8b4895a5c/test/unit/view.child-views.spec.js#L433

its not clear for me what exactly happens in this beforeEach hook, but it definitely reshowing already showed view and if i will check only for `_isShown` then i get 3 failed tests included this one

could you shed some light what here happens and how i have to change that

taburetkin: i moved _isShown clear to _empty it solves this issue,

taburetkin: btw, there is another place
```
          beforeEach(function() {
            ...
            anotherView = new AnotherView();
            ...
            myCollectionView.addChildView(anotherView);
          });
          it('should only append the added child', function() {
            this.sinon.stub(myCollectionView, 'attachHtml');

            // Only true if not maintaining collection sort
            myCollectionView.sortWithCollection = false;
            myCollectionView.addChildView(anotherView);
            ...
          });
```
in terms of this PR's change this can not be done any more
and i can't resolve what exactly this test is testing :-/

src/region.js
paulfalgout: I don't think it should swallow it

taburetkin:
added changes, not sure about testing detachChildView

src/region.js
```diff
@@ -288,6 +300,8 @@ _.extend(Region.prototype, CommonMixin, {
     delete this.currentView;
 
     if (!view._isDestroyed) {
+      view._isShown = false;
```
paulfalgout: this is a very small thing, but for consistency... in collectionview this flag is set after the detach takes place.  Can we move the to right before the `_stopChildViewEvents`?

Issue #3640, opened 2018-11-12
Use Marionette in child windows


sebamarynissen:
I'm using Marionette in an Electron app, and I am using it to render modals. The  approach I currently have is something like this:

```js
let view = new ModalView();
let child = window.open('', 'name');
let region = new Region({
  el: child.document.body
});
region.show(view);
```

This seems to work correctly, however the "onAttach" method does not get triggered. This is because Marionette checks if the region is present in the DOM using this fragment:
```js
// In region.js
const shouldTriggerAttach = !view._isAttached && this.Dom.hasEl(document.documentElement, this.el) && !this._shouldDisableMonitoring();
```
shouldTriggerAttach evaluates to false because document.documentElement indeed does no contain the regions element because it resides in my child window's document!

As a potential workaround, I propose to solve this by replacing this line by
```js
let shouldTriggerAttach = !view._isAttached;
if (shouldTriggerAttach) {
  const root = this.el.getRootNode();
  shouldTriggerAttach = root instanceof Document && !this._shouldDisableMonitoring();
}
```
Perhaps the most elegant solution is to re-introduce the isNodeAttached method from Marionette v3.x and implement it like
```js
function isNodeAttached(el) {
  return el.getRootNode() instanceof Document;
}
```

If this would be considered a good solution, I'd be happy to file a pull request for this. In any case I guess it's useful to be able to use Marionette in a child window just like we would in the main window.

**EDIT**: I've done some tests and while ```let root = el.getRootNode()``` returns a document element, ```root instanceof Document``` evaluates to ```false``` in a child window, whereas it evaluates to ```true``` in the main window, so the approach above wouldn't work. Don't really understand why though.

paulfalgout:
So we've seen this issue with iframes as well and I think a solution is warranted.  _however_ I don't think it's good to go around `hasEl`  (it would also be breaking to do so)

In fact it might make sense to add the `documentElement` to the DOM API so that the various lines end up being `this.Dom.hasEl(this.Dom.el, this.el)`.

Then it'd be `MyRegionOrView.setDomApi({ el: child.document.documentElement });` and all of the DOM related interactions are stored on the DOM API.  I think this is a good improvement..  it at least removes `document.documentElement` from being hardcoded in Mn code.

_However_ it's not a great solution, because then contextually you'll have to add the special dom api to every region and view used in the alternative documents... however since attachment is determined at instantiation for views, I don't really see a way around it.

[...]

Issue #3652, opened 2019-02-18
Fix coverage issue


paulfalgout:
We're missing coverage now here https://github.com/marionettejs/backbone.marionette/blob/master/src/common/monitor-view-events.js#L30

I believe the coverage loss likely occurred here #3636

Because there are no more destroys during destroys it may have rendered this check unnecessary.  But we should be absolutely sure there is not a corner case before removing this line.

cc: @taburetkin

PR #3655, opened 2019-02-19
Use ownerdocument for document element

paulfalgout:
Resolves https://github.com/marionettejs/backbone.marionette/issues/3640

~~While working on this I found a bug where showing a view in a region with `replaceElement: true` that already had a view will not get the `attach` events because the event checks the region's `el`.  So this also checks against the replaced view's el as a fix, but I need to add a test case to cover this logic branch.~~

[ ] Tests for replaceElement attached issue

### DIFF
--- a/src/collection-view.js
+++ b/src/collection-view.js
@@ -269,7 +269,7 @@ const CollectionView = Backbone.View.extend({
   setElement() {
     Backbone.View.prototype.setElement.apply(this, arguments);
 
-    this._isAttached = this.Dom.hasEl(document.documentElement, this.el);
+    this._isAttached = this._isElAttached();
 
     return this;
   },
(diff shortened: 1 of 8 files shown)


paulfalgout:
This PR is passing, but the `gulp-coveralls` is failing.  https://github.com/marionettejs/backbone.marionette/issues/3656

paulfalgout:
The bug was a false alarm.  Couldn't actually get to that state because `empty` always resolved the `el` first.  However while chasing that issue down I did look into the branch coverage issues mentioned here: https://github.com/marionettejs/backbone.marionette/issues/3652  I added them as a 2nd commit to this PR, but they can go in a separate PR if necessary.

I think after this v4.1 is ready to :shipit:

PR #3659, opened 2019-04-15
Update Tooling

paulfalgout:
This PR modernizes the tooling with npm 10, removing gulp and using scripts with rollup plugins primarily for all actions.


.eslintrc
```diff
@@ -1,4 +1,5 @@
 {
+  "extends": "eslint:recommended",
```
paulfalgout: I removed any rules defined in `eslint:recommended`

test/rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+
+function runSauce() {
+  easySauce({
+    name: 'Marionette.js',
+    username: process.env.SAUCE_USERNAME,
+    key: process.env.SAUCE_ACCESS_KEY,
+    port: '8080',
+    testPath: '/test/sauce.html',
+    framework: 'mocha',
+    platforms: [
```
paulfalgout: safari was removed from the matrix as, though the tests passed, the runner never registered with mocha, so something is up somewhere with safari and the sauce chain.  Not sure how to fix it, but I don't think safari should be a problem anyhow.

paulfalgout:
@marionettejs/marionette-core This is a fairly big change, but as there's no real code changes, and we  haven't been too active, I'll probably move forward with this sooner than later to get out v4.1   The only real code change is that rollup now adds `noConflict` back to the UMD build only which is a kind of nice addition.  I'll need to document v4.1 and actually run the build, but this new tooling is a substantial improvement to what was there.

PR #3664, opened 2019-04-28
WIP Next Region

paulfalgout:
An iteration and replacement for https://github.com/marionettejs/backbone.marionette/pull/3620

This is starting to feel really good.
This is a WIP and I'll add more notes here shortly, but I wanted to get it out for 👀 
I do think there's a handful of things we can introduce in a v4.x as well as some small deprecations that can be made for the breaking changes.
There are actually few breaking changes I believe the breaking changes that are there are minor and like inconsequential for most users.
I haven't done perf tests yet, but I feel moderately confident this will be more performant, and again I haven't tested it, but I believe this makes rendering views non-destructive to children.

I will work on documenting changes as well as breaking up this PR into sane commits as well as making v4.x PRs for features and deprecations we can do now.

But this I think is a good start.  @marionettejs/marionette-core 

**Breaking Changes**
- Region no longer accepts a jquery instances as `el` `el: $el[0]` should be used instead of `el: $el`  This removes region's dependencies on jquery.
- As previously discussed, the not documented `allowMissingEl` is removed, however the reasoning for adding this flag in the first place is not an issue with this revision
- Both `add:region` and `remove:region` events were removed.  They're even less useful now than they were previously, and there's a new hook for when regions are bound to the view `bind:regions` that should cover any use case this previously had.  This should make rendering cheaper overall as well.
- Regions no longer empty on `render`.  This may be the biggest breaking change to watch out for.  However for the general case where the user is setting the children in `onRender` the 2nd show will handle the empty effectively in the same way as before.  But there may be some cases where conditional child renders will no longer empty out the previous view.  Users will need to handle these cases now, but this also should allow children to be created once in an `onReady` and child views may not need to be re-rendered at all when the parent is rendered.
- Not documented, but Region's `@ui` allowed the user to build complex selectors such as:
`regionName: '@ui.foo @ui.bar'` producing `regionName: '#foo .bar'`.  This change prevents that and requires a single ui only.  Not only is the complex selector not recommended anyways, this prevents needing to requery the selector and the already queried ui is used instead.
- Not documented, but defining a Region on a view's region hash allowed the user to pass either and instance of a region or a region class. 
  ```js
   regions : {
     foo: myRegion
     bar: MyRegionClass
   }
  ```
  Both of these options are no longer supported as an `el` selector is required to be managed by the view, and both of these options prevent that.  However allowing the view to maintain regions outside of the view is problematic and certainly makes `showChildView` difficult to understand.  While I doubt many people are using this feature anyways, the way to support these options is:
  ```js
  const MyRegionClass = Region.extend({ el: '#foo' }); // some predefined selector
  const MyView = View.extend({
    initialize({ myRegion }) {
      this.myRegion = myRegion;
    },
    regions: {
      bar: {
        el: '#foo',
        regionClass: MyRegionClass
      }
    }
  ```
- Though largely considered a bug https://github.com/marionettejs/backbone.marionette/issues/3320 if a user was setting up regions to be within a child view, this will no longer work.  It should be documented however that using `addRegion` or `addRegions` _after_ the view is rendered will query all of the children until the render in which it will only query the rendered template.  It might be best to suggest simply adding a definition to `myView.regions` and then re-rendering.
- Regions are no longer instantiated prior to `View#initialize`.  This shouldn't really be an issue as getting a region will render and instantiation the region needed... but if someone was doing something strange and expecting the region init to be prior to the view's init, it would technically be breaking.
- The `el` on a region prior to showing a view was possibly the `selector` (if instantiated with a selector)  Now it will always be a DOM element or `undefined` if nothing was found by the query.  It might be an issue for users who instantiate a region that does a global look up for DOM that didn't exist at the time of instantiation.
- `Region#reset` was removed.  In this context it is no longer needed.

**New Features**
`View` `ready` event indicating the best spot to attach children
`Application` `region` can now be defined as a function returning a region or region definition
`Region#setElement` was made public and properly handles the various attached states of the elements


src/common/normalize-region.js
```diff
@@ -0,0 +1,19 @@
...
+import MarionetteError from '../utils/error';
+
+// return the region instance from the definition
+export default function(definition, regionClass) {
+  if (_.isString(definition)) {
+    definition = { el: definition };
+  }
+
+  if (!_.isString(definition.el)) {
+    console.log(definition);
```
paulfalgout: need to remove this

test/unit/view.dynamic-regions.spec.js
```diff
@@ -14,22 +14,12 @@ describe('itemView - dynamic regions', function() {
 
   describe('when adding a region to a layoutView, after it has been rendered', function() {
     beforeEach(function() {
-      this.MyView = Marionette.View.extend({
-        onAddRegion: function() {},
-        onBeforeAddRegion: function() {}
-      });
+      this.MyView = Marionette.View.extend();
```
paulfalgout: all of the test changes in this file relate to removing `add:region` / `remove:region`

test/unit/view.child-views.spec.js
```diff
@@ -637,53 +638,4 @@ describe('layoutView', function() {
       });
     });
   });
-
-  describe('manipulating regions', function() {
```
paulfalgout: // removing `add:region` / `remove:region`

test/unit/view.child-views.spec.js
```diff
@@ -550,6 +550,7 @@ describe('layoutView', function() {
     });
 
     it('should set custom region classes', function() {
+      this.layoutView.render();
```
paulfalgout: regions aren't initialized until the view is rendered

test/unit/view.child-views.spec.js
```diff
@@ -522,7 +522,7 @@ describe('layoutView', function() {
         war: '.craft',
         is: {
           regionClass: this.CustomRegion,
-          selector: '#a-fun-game'
+          el: '#a-fun-game'
```
paulfalgout: these tests failed silently before 😿

test/unit/view.child-views.spec.js
```diff
@@ -33,7 +33,7 @@ describe('layoutView', function() {
     this.ViewNoDefaultRegion = this.View.extend({
       regions: {
         regionOne: {
-          selector: '#regionOne',
+          el: '#regionOne',
```
paulfalgout: these tests failed silently before 😿

test/unit/view.child-views.spec.js
```diff
@@ -470,8 +470,8 @@ describe('layoutView', function() {
       this.layoutView.render();
     });
 
-    it('should call empty twice', function() {
-      expect(this.region.empty).to.have.been.calledThrice;
+    it('should call empty once', function() {
```
paulfalgout: this is a poorly written test, but childviews are no longer emptied on the parent's render

test/unit/region.spec.js
```diff
@@ -1317,30 +1226,6 @@ describe('region', function() {
     });
   });
 
-  describe('when calling "_ensureElement"', function() {
```
paulfalgout: removed `allowMissingEl`

test/unit/region.spec.js
```diff
@@ -1124,37 +1064,6 @@ describe('region', function() {
     });
   });
 
-  describe('when resetting a region', function() {
```
paulfalgout: remove `reset`

test/unit/region.spec.js
```diff
@@ -1104,12 +1042,14 @@ describe('region', function() {
     let region;
 
     beforeEach(function() {
-      itemView = new View();
-      itemView.render = this.sinon.stub();
-      itemView.addRegions({
-        MyRegion: '#region',
-        anotherRegion: '#region2'
+      itemView = new View({
```
paulfalgout: this test broke because it was `noop`ing render which is now what builds the regions it was testing

test/unit/region.spec.js
```diff
@@ -1031,22 +985,6 @@ describe('region', function() {
     });
   });
 
-  describe('when initializing a region and passing an "el" option', function() {
```
paulfalgout: `el` is no longer the string after initialization

test/unit/region.spec.js
```diff
@@ -212,7 +166,7 @@ describe('region', function() {
     beforeEach(function() {
       this.setFixtures('<div id="region"></div>');
       myRegion = new Region({
-        el: '#region'
+        el: $('#region')[0]
```
paulfalgout: I can revert this, but it works either way

test/unit/region.spec.js
```diff
@@ -156,38 +112,36 @@ describe('region', function() {
     });
 
     it('should return the region', function() {
-      expect(region._setElement(twoEl)).to.equal(region);
+      expect(region.setElement(twoEl)).to.equal(region);
```
paulfalgout: setElement was made public

test/unit/region.spec.js
```diff
@@ -89,52 +72,25 @@ describe('region', function() {
           $(this.el).html('some content');
         }
       });
-      myView = new MyView();
 
       this.setFixtures('<div id="region"></div>');
     });
 
     describe('when showing a view', function() {
```
paulfalgout: removed `allowMissingEl`

test/unit/region.spec.js
```diff
@@ -8,19 +8,10 @@ import CollectionView from '../../src/collection-view';
...
-        return new Region();
-      }).to.throw('An "el" must be specified for a region.');
-    });
-  });
-
   describe('when passing an el DOM reference in directly', function() {
     let el;
     let customRegion;
     let optionRegion;
-    let optionRegionJquery;
```
paulfalgout: removed the option to pass in `el` as a jquery object

test/unit/region.spec.js
```diff
@@ -8,19 +8,10 @@ import CollectionView from '../../src/collection-view';
 describe('region', function() {
   'use strict';
 
-  describe('when creating a new region and no configuration has been provided', function() {
```
paulfalgout: `el` can be undefined now during initialization

paulfalgout:
Note: nondestructive regions will not work without changing the dom api away from jquery
https://jsfiddle.net/qjmtpg9y/7/
when the view is rendered with `$el.html` jquery removes any previous listeners which breaks the existing child views.

paulfalgout:
This will eventually move to https://github.com/marionettejs/marionette

PR #3676, opened 2020-04-06
docs: Fix simple typo, overriden -> overridden

timgates42:
There is a small typo in lib/backbone.marionette.esm.js, lib/backbone.marionette.js, src/collection-view.js.

Should read `overridden` rather than `overriden`.


timgates42:
Thanks @paulfalgout - cleaned that up!

